PR #84, opened 2019-11-08
New badges that point to current readme for

lupine86:
These new badges render green when a VE version is fully deployed and yellow when there is a partial deployment (along with percent deployed).  Clicking on the badge navigates to the current Readme.md file for that image version.


lupine86:
This was completed with #106 so closing this one out.

PR #127, opened 2019-12-04
Integrate clang-tidy and scan-build into Linux.

pspoerri:
Tested by hand.
[ubuntu-log.txt](https://github.com/actions/virtual-environments/files/3922885/ubuntu-log.txt)


kaylangan:
@pspoerri thanks for the contribution!

Can you give a little more background on these tools and the benefit for having them pre-installed? If a tool can be installed during the build, we like to evaluate how much time is saved vs. how much space is used by having the tool pre-installed.

pspoerri:
> Can you give a little more background on these tools and the benefit for having them pre-installed?
clang-tidy and scan-build are extra tools belonging to clang.
- *[clang-tidy](https://clang.llvm.org/extra/clang-tidy/index.html)* is a linter tool which integrates a wide array of checks. Primary focus are code style checks.
- *[scan-build](https://clang-analyzer.llvm.org/) is the wrapper tool for the clang static analyzer.

[Here's a demo project](https://github.com/pspoerri/ClangTidy-Analyzer) that shows how clang-tidy and the scan-build are typically integrated with a CMake project.

Both the clang-tidy and the static analyzer are tools you would want to run on every pull-request.

> If a tool can be installed during the build, we like to evaluate how much time is saved vs. how much space is used by having the tool pre-installed.

I believe that both tools are useful for the developer community relying on C and C++ code-bases. Installing the tool with the image ensures that clang-tidy and scan-build are installed with the same version as the default clang installation.

images/linux/scripts/installers/clang.sh
```diff
@@ -1,31 +1,33 @@
...
+################################################################################
+
+# Source the helpers for use with the script
+source $HELPER_SCRIPTS/document.sh
+source $HELPER_SCRIPTS/apt.sh
+
+wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add -
+apt-add-repository "deb http://apt.llvm.org/$(lsb_release -cs)/ llvm-toolchain-$(lsb_release -cs)-6.0 main"
+apt-get update
+apt-get install -y clang-6.0 clang-tidy-6.0 clang-tools-6.0 lldb-6.0 lld-6.0 libfindbin-libs-perl
```
e-kwsm: * Is libfindbin-libs-perl required?
* I think clang-format is also useful

alepauly:
@pspoerri thank you so much for the contribution. However, after looking into it, I don't see enough benefit to preinstall this to make it worth the risk and ongoing maintenance effort. An alternative you might want to consider to simplify it's use on Actions is to write a setup-clang-tidy Action perhaps?

PR #136, opened 2019-12-09
Windows2019: Add Ninja

chuckatkins:
This adds the Ninja build tool (https://ninja-build.org/) for efficient scaling of parallel builds.


kaylangan:
@chuckatkins thanks for the contribution!

Do you know how much time it takes to install this during the build vs. having it pre-installed? Also, do you know the size? If a tool can be installed during the build, we like to evaluate how much time is saved vs. how much space is used by having the tool pre-installed.

chuckatkins:
It's a small tool (a few mb) so you can certainly manually install it as part of the build in a setup step (it's available as a pre-built binary).  The same can be said though for a significant number of the utilities that come pre-installed in the image (jq for instance).  It's more about having a baseline of common tools already available.  Especially since cmake is baked and in, using the Ninja generator can yield a significant improvement in build times for c++ code even in the small resource constrained environment the actions run in.

chuckatkins:
@kaylangan ping?

stevencpp:
By the way, Visual Studio 2019 already ships a copy of Ninja at `Common7\IDE\CommonExtensions\Microsoft\CMake\Ninja\ninja.exe` . For C++ projects you generally run vcvarsall.bat and that even adds ninja to the PATH. It's 1.8.2, whereas the latest is 1.9.0, but it might still be good enough to just always add it to PATH by default for projects that don't need all the other stuff vcvarsall.bat brings in ?
For linux and mac though it'd be really helpful to install it. It's small but can still take a long time to install with a package manager, and hardcoding links to specific versions to download it while running the jobs isn't very user friendly for such a basic tool.

alepauly:
> Visual Studio 2019 already ships a copy of Ninja

Thanks for the info @stevencpp. 

@chuckatkins I confirmed ninja is already in the machines through VS 2019 so I'll go ahead and close this. If you think there's still value please let me know so we can reconsider.

Issue #56, opened 2019-10-23
Update CMake on Linux


wipe2238:
**Tool information**
- Tool name: cmake
- Add or update? Update
- Desired version: 3.15.4 (as on windows/macOs)
- Approximate size:

**Virtual environments affected**
- [ ] macOS 10.14
- [x] Ubuntu 16.04 LTS
- [x] Ubuntu 18.04 LTS
- [ ] Windows Server 2016 R2
- [ ] Windows Server 2019

**Can this tool be installed during the build?**
Yes (download .tar.gz from cmake.org, extract, add to PATH).

It would be nice if we had same version everywhere in future; there won't be need to check versions on each OS before tweaking project with newer features.

**Are you willing to submit a PR?**
???

al-cheb:
Hello, @wipe2238 

Thank you for taking the time to report an issue. I have created PR to fix this issue(PR https://github.com/actions/virtual-environments/pull/274)

maxim-lobanov:
This issue will also resolve https://github.com/actions/virtual-environments/issues/110

zbeekman:
#274 was merged and upgraded to 3.16.2. I think this issue can be closed now

PR #141, opened 2019-12-12
Update cmake.sh

merlinLia:
cmake 3.15.5 has been released in 2019.10.30. Update CMake to 3.15 stable could let us to make it possible to apply the latest CMake in our build. Link to the previous PR in original repos: #https://github.com/microsoft/azure-pipelines-image-generation/pull/1424

### DIFF
--- a/images/linux/scripts/installers/cmake.sh
+++ b/images/linux/scripts/installers/cmake.sh
@@ -12,7 +12,7 @@ echo "Checking to see if the installer script has already been run"
 if command -v cmake; then
     echo "Example variable already set to $EXAMPLE_VAR"
 else
-	curl -sL https://cmake.org/files/v3.12/cmake-3.12.4-Linux-x86_64.sh -o cmakeinstall.sh \
+	curl -sL https://cmake.org/files/v3.15/cmake-3.15.5-Linux-x86_64.sh -o cmakeinstall.sh \
 	&& chmod +x cmakeinstall.sh \
 	&& ./cmakeinstall.sh --prefix=/usr/local --exclude-subdir \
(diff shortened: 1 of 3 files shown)


kaylangan:
@azuredaveops can you or someone on your team take a look at this?

stevencpp:
The windows-latest image already has CMake 3.16, so could we get the same on all platforms ? Related issue: #56

compnerd:
Is there a reason to install this this way rather than use the KitWare APT repo?

alepauly:
> KitWare

@dakale do you know why we might have originally opted to install through the sh scripts instead of apt? I wonder if maybe the apt source wasn't available then. If so, using apt might be fine as long as we pin the version.

alepauly:
@merlinLia sorry for the delay here. I'll try to get someone to look at this. In the meantime, can you remove the Readme changes? those should get generated automatically next time we build the images.

zbeekman:
This looks like it's stale/outdated to me. Latest CMake on linux is now 3.16.x. But the install is still via the self extracting script. As a CMake user and contributor, I don't have a problem with installing via the self extracting script, however, adding the kitware APT repo, say in [`repo.sh`](https://github.com/actions/virtual-environments/blob/master/images/linux/scripts/base/repos.sh) would allow the cmake install script to be a simple `apt-get install cmake==<version>`.

If this is of interest I could work up a PR.

Otherwise, I think this PR is outdated and can be closed as CMake 3.16.x is now installed on linux.

al-cheb:
Closed in https://github.com/actions/virtual-environments/pull/274 PR

Issue #146, opened 2019-12-12
Unable to build latest Ubuntu 18.04 image - "Could not open input file: composer-setup.php"


ropnop:
**Describe the bug**
I am trying to use Packer to build the latest Ubuntu 1804 image from `/images/linux/ubuntu1804.json`, but the installation is failing and Packer is aborting with this error:

```
azure-arm: The following NEW packages will be installed:
    azure-arm:   snmp
    azure-arm: 0 upgraded, 1 newly installed, 0 to remove and 30 not upgraded.
    azure-arm: Need to get 0 B/158 kB of archives.
    azure-arm: After this operation, 563 kB of additional disk space will be used.
    azure-arm: Selecting previously unselected package snmp.
    azure-arm: (Reading database ... 241772 files and directories currently installed.)
    azure-arm: Preparing to unpack .../snmp_5.7.3+dfsg-1.8ubuntu3.3_amd64.deb ...
    azure-arm: Unpacking snmp (5.7.3+dfsg-1.8ubuntu3.3) ...
    azure-arm: Setting up snmp (5.7.3+dfsg-1.8ubuntu3.3) ...
    azure-arm: Processing triggers for man-db (2.8.3-2ubuntu0.1) ...
    azure-arm: Installer corrupt
    azure-arm: Could not open input file: composer-setup.php
```

**Virtual environments affected**
- [ ] macOS 10.15
- [ ] Ubuntu 16.04 LTS
- [x ] Ubuntu 18.04 LTS
- [ ] Windows Server 2016 R2
- [ ] Windows Server 2019

**Expected behavior**
A successful `packer build` to create the Ubuntu 1804 image

**Actual behavior**
An installation step fails and Packer aborts the build with this error:
```
azure-arm: The following NEW packages will be installed:
    azure-arm:   snmp
    azure-arm: 0 upgraded, 1 newly installed, 0 to remove and 30 not upgraded.
    azure-arm: Need to get 0 B/158 kB of archives.
    azure-arm: After this operation, 563 kB of additional disk space will be used.
    azure-arm: Selecting previously unselected package snmp.
    azure-arm: (Reading database ... 241772 files and directories currently installed.)
    azure-arm: Preparing to unpack .../snmp_5.7.3+dfsg-1.8ubuntu3.3_amd64.deb ...
    azure-arm: Unpacking snmp (5.7.3+dfsg-1.8ubuntu3.3) ...
[...]

PR #152, opened 2019-12-13
compare php composer against latest signature

ropnop:
This addresses #146 

The SHA384 hash of the `composer-setup.php` script is hardcoded in the build script for the Ubuntu images. This means that whenever the install script is updated, the entire packer build fails and errors out on that step.

According to [this page from Composer](https://getcomposer.org/doc/faqs/how-to-install-composer-programmatically.md), the better way is to fetch the latest signature dynamically and compare it against the downloaded installer file.

I modified the php steps slightly to compare the hash against the latest official one from https://composer.github.io/installer.sig

Let me know if you think this is the best approach. The other approach would be to pin the installer to a specific hash and then downloading it by reference like `https://raw.githubusercontent.com/composer/getcomposer.org/76a7060ccb93902cd7576b67264ad91c8a2700e2/web/installer`


majorvin:
hey @ropnop  can you rebase your branch? 
@kaylangan i've tested the fix above, can we get someone to approve it after the rebase?

ropnop:
Merged with upstream master

ropnop:
I think the history is good now? Let me know, it got a little messy

alepauly:
> 
> 
> I think the history is good now? Let me know, it got a little messy

@ropnop, something went wrong on that merge so I just went ahead and cherry-picked your commit unto a new clean branch and resubmitted as #194. I'll close this one and merge through that other one. Thanks again and sorry for the pain getting it merged :-)

Issue #166, opened 2019-12-16
Unable to build Ubuntu16.04  cannot find zipfile directory in one of maven.zip


majorvin:
**Describe the bug**
After manually fixing the bug listed in the PR (https://github.com/actions/virtual-environments/issues/146) getting another error installing java tools

**Virtual environments affected**
- [ ] macOS 10.15
- [x ] Ubuntu 16.04 LTS
- [ ] Ubuntu 18.04 LTS
- [ ] Windows Server 2016 R2
- [ ] Windows Server 2019

**Expected behavior**
Should continue and complete the image generation

**Actual behavior**
```
done.
    azure-arm: done.
    azure-arm: ANT_HOME=/usr/share/ant
    azure-arm: Archive:  maven.zip
==> azure-arm:   End-of-central-directory signature not found.  Either this file is not
==> azure-arm:   a zipfile, or it constitutes one disk of a multi-part archive.  In the
==> azure-arm:   latter case the central directory and zipfile comment will be found on
==> azure-arm:   the last disk(s) of this archive.
==> azure-arm: unzip:  cannot find zipfile directory in one of maven.zip or
==> azure-arm:         maven.zip.zip, and cannot find maven.zip.ZIP, period.
==> azure-arm: Provisioning step had errors: Running the cleanup provisioner, if present...
==> azure-arm:
==> azure-arm: The resource group was not created by Packer, deleting individual resources ...
```

mrpullen:
Same issue - @majorvin I've modified the java-tools.sh file per the PR above - This also impacts ubuntu 18.04 LTS as well.

kaylangan:
I believe that this was fixed in #167 / #172. If that's incorrect, let me know and I'll re-open this.

PR #167, opened 2019-12-16
Upgrade maven to 3.6.3

majorvin:
For some reason, 3.6.2. is deleted in the repository bumping to 3.6.3 instead..

Fix for: https://github.com/actions/virtual-environments/issues/166


kaylangan:
Fixed with #172

PR #168, opened 2019-12-16
Use toolcache for PyPy on Ubuntu 1604 and 1804

konradpabjan:
Automation was recently done to a publish PyPy straight to the tools cache (same thing we do for Python and Ruby): https://dev.azure.com/mseng/AzureDevOps/_build?definitionId=9703&_a=summary

Currently we would manually download the latest version of PyPy and add it to the toolscache (Windows and Mac). For Ubuntu 1604 and 1804, PyPy would be downloaded here and whenever there was a new version, we would have to change the download URLs (not fun, extra maintenance and not very consistent).

This PR removes the manual downloading of PyPy that was going on here and switches over Ubuntu 1604 and 1804 to use PyPy that is in the toolscache.

Corresponding PR for adding PyPy 2.7.13 to Ubuntu 1604 and 1804 tools cache: https://mseng.visualstudio.com/AzureDevOps/_git/CIPlat.Externals/pullrequest/524840?_a=overview

Corresponding PR for adding PyPy 3.6.9 to Ubuntu 1604 and 1804 tools cache: https://mseng.visualstudio.com/AzureDevOps/_git/CIPlat.Externals/pullrequest/524846


konradpabjan:
@brcrista @alepauly @alepauly Could you take a look

images/linux/scripts/installers/python.sh
```diff
@@ -10,47 +10,9 @@ source $HELPER_SCRIPTS/document.sh
 # Install Python, Python 3, pip, pip3
 apt-get install -y --no-install-recommends python python-dev python-pip python3 python3-dev python3-pip
```
konradpabjan: Was planning on removing all these extra downloaded versions of Python but I'll leave this for another PR (want to make sure PyPy + tools cache is working correctly for Ubuntu)

images/linux/scripts/installers/python.sh
```diff
@@ -10,47 +10,9 @@ source $HELPER_SCRIPTS/document.sh
...
 
-# Install PyPy 2.7 to $AGENT_TOOLSDIRECTORY
-wget -q -P /tmp https://bitbucket.org/pypy/pypy/downloads/pypy2.7-v7.1.0-linux64.tar.bz2
-tar -x -C /tmp -f /tmp/pypy2.7-v7.1.0-linux64.tar.bz2
-rm /tmp/pypy2.7-v7.1.0-linux64.tar.bz2
-mkdir -p $AGENT_TOOLSDIRECTORY/PyPy/2.7.13
-mv /tmp/pypy2.7-v7.1.0-linux64 $AGENT_TOOLSDIRECTORY/PyPy/2.7.13/x64
-touch $AGENT_TOOLSDIRECTORY/PyPy/2.7.13/x64.complete
-
-# add pypy to PATH by default
```
brcrista: We'll want to keep this and the one for PyPy3 below.  Unfortunately that adds something to keep in sync when we update PyPy -- not sure if there's a better way.  Python will have the same problem.

brcrista: I think a good place for it would be `hosted-tool-cache.sh` where you are documenting it now.

images/linux/scripts/installers/python.sh
```diff
@@ -10,47 +10,9 @@ source $HELPER_SCRIPTS/document.sh
...
-ln -s $AGENT_TOOLSDIRECTORY/PyPy/3.6.9/x64/bin/python3 $AGENT_TOOLSDIRECTORY/PyPy/3.6.9/x64/bin/python
-
-# Install latest Pip for PyPy3
-$AGENT_TOOLSDIRECTORY/PyPy/3.6.9/x64/bin/pypy3 -m ensurepip
-$AGENT_TOOLSDIRECTORY/PyPy/3.6.9/x64/bin/pypy3 -m pip install --ignore-installed pip
-
 # Run tests to determine that the software installed as expected
 echo "Testing to make sure that script performed as expected, and basic scenarios work"
-for cmd in python pip pypy python3 pip3 pypy3; do
+for cmd in python python3; do
```
brcrista: We'll want to keep the checks for `pip` and `pip3` here.

images/linux/scripts/installers/python.sh
konradpabjan: In the `setup.sh` file that gets downloaded and run, there are symlinks that get created.  
```
# PEP 394-like symlinks (but don't include minor version)
ln -s pypy python2
ln -s python2 python
```
Couldn't we add an extra symlink to `setup.sh`? We need this so that `pypy` can be used without the `setup-python` action?
```
ln -s pypy /usr/local/bin/pypy
```

konradpabjan: As long as there are only 2 versions of PyPy in the tools cache then there shouldn't be any conflicts. For Python this will have to be addressed differently

brcrista: > The idea of having to make a change here every time there is a version update in the tools cache just doesn't seem right

That's a fair point.  See [how we do it for Windows](https://github.com/actions/virtual-environments/blob/07b9e016112891bbec8108d7dc941a84aeb8989b/images/win/scripts/Installers/Windows2019/Install-Python.ps1#L27) currently.  So whichever pattern you use, I think it should be consistent.

I think that should work -- you might have to `sudo` to write to `/usr/local/bin`.  I don't see what the problem with CPython would be, could you explain?

konradpabjan: Based off of the options, I decided to add `pypy` and `pypy3` to PATH in `hosted-tool-cache.sh`. This was done largely to be consistent with how Windows does it. Also contemplated doing something like  `export PATH`, but decided against that. In the future we will be able to completely remove `python.sh` 🔥 when the extra versions of python that get downloaded get removed.

The change I made runs on the assumption that there will every only be 2 versions of PyPy in the tools cache (which i think is safe to have). 

Regarding CPython I was originally thinking of modifying the `setup.sh` file that gets downloaded and having the link get created there as one of the steps. Modifying `setup.sh` and adding the symlink there would work for PyPy because there would only be 2 versions downloaded. With Python there are a lot more than 2 so this would cause a problem with what becomes the default?. For CPython, if the symlink gets created just like PyPy now (some if statement), then we can choose which version of Python is the default one that gets added to PATH.

brcrista: > The change I made runs on the assumption that there will every only be 2 versions of PyPy in the tools cache

There aren't any plans right now to have more versions, and if there are we can always change this.  Thanks for calling out the assumption, though.

konradpabjan:
@alepauly I merged in all the changes for the tools cache so we're good to go! 🎉

PR #184, opened 2019-12-19
Install Google Chrome, Chromedriver and Selenium

AndreyMaslennikov:
In this PR we've extended `images/linux/scripts/installers/google-chrome.sh` script in order to install Chromedriver compatible with installed Google Chrome version and latest stable selenium-server-standalone version

### DIFF
--- a/images/linux/scripts/installers/google-chrome.sh
+++ b/images/linux/scripts/installers/google-chrome.sh
@@ -58,6 +58,14 @@ SELENIUM_VERSION_MAJOR_MINOR=$(echo $SELENIUM_VERSION | cut -d '.' -f 1,2)
 
 # Download selenium standalone server (hardcoded version 3.141.59)
 echo "Downloading selenium-server-standalone v$SELENIUM_VERSION..."
-wget https://selenium-release.storage.googleapis.com/$SELENIUM_VERSION_MAJOR_MINOR/selenium-server-standalone-$SELENIUM_VERSION.jar
+SELENIUM_JAR_NAME=selenium-server-standalone-$SELENIUM_VERSION.jar
+wget https://selenium-release.storage.googleapis.com/$SELENIUM_VERSION_MAJOR_MINOR/$SELENIUM_JAR_NAME
+
+echo "Testing to make sure that script performed as expected, and basic scenarios work"
+if [ ! -f "$SELENIUM_JAR_NAME" ]; then
+    echo "Selenium server was not installed"
+    exit 1
+fi
+
 mv selenium-server-standalone-$SELENIUM_VERSION.jar /usr/share/java/selenium-server-standalone.jar
 echo "CLASSPATH=/usr/share/java/selenium-server-standalone.jar:." | tee -a /etc/environment
--- a/images/linux/scripts/installers/google-chrome.sh
+++ b/images/linux/scripts/installers/google-chrome.sh
@@ -51,9 +51,13 @@ fi
 echo "Lastly, documenting what we added to the metadata file"
 DocumentInstalledItem "Chromedriver ($(chromedriver --version))"
 
+# Determine latest selenium standalone server version
+SELENIUM_LATEST_VERSION_URL=https://api.github.com/repos/SeleniumHQ/selenium/releases/latest
+SELENIUM_VERSION=$(curl $SELENIUM_LATEST_VERSION_URL | jq '.name' | tr -d '"' | cut -d ' ' -f 2)
+SELENIUM_VERSION_MAJOR_MINOR=$(echo $SELENIUM_VERSION | cut -d '.' -f 1,2)
+
 # Download selenium standalone server (hardcoded version 3.141.59)
-SELENIUM_VERSION=3.141.59
 echo "Downloading selenium-server-standalone v$SELENIUM_VERSION..."
-wget https://selenium-release.storage.googleapis.com/3.141/selenium-server-standalone-$SELENIUM_VERSION.jar
+wget https://selenium-release.storage.googleapis.com/$SELENIUM_VERSION_MAJOR_MINOR/selenium-server-standalone-$SELENIUM_VERSION.jar
 mv selenium-server-standalone-$SELENIUM_VERSION.jar /usr/share/java/selenium-server-standalone.jar
 echo "CLASSPATH=/usr/share/java/selenium-server-standalone.jar:." | tee -a /etc/environment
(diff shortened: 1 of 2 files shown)


images/linux/scripts/installers/google-chrome.sh
```diff
@@ -25,3 +25,51 @@ fi
...
+LATEST_CHROMEDRIVER_VERSION=$(curl "https://chromedriver.storage.googleapis.com/LATEST_RELEASE_$CHROME_VERSION")
+
+# Download and unpack latest release of chromedriver
+echo "Downloading chromedriver v$LATEST_CHROMEDRIVER_VERSION..."
+wget "https://chromedriver.storage.googleapis.com/$LATEST_CHROMEDRIVER_VERSION/chromedriver_linux64.zip"
+unzip chromedriver_linux64.zip
+rm chromedriver_linux64.zip
+mv "chromedriver" "/usr/bin/chromedriver"
+chown root:root /usr/bin/chromedriver
+chmod +x /usr/bin/chromedriver
```
AlenaSviridenko: maybe move `/usr/bin/chromedriver` to variable like `CHROMEDRIVER_BIN` and use it instead of 3 hard-coded path?

images/linux/scripts/installers/google-chrome.sh
```diff
@@ -25,3 +25,50 @@ fi
...
+SELENIUM_LATEST_VERSION_URL=https://api.github.com/repos/SeleniumHQ/selenium/releases/latest
+SELENIUM_VERSION=$(curl $SELENIUM_LATEST_VERSION_URL | jq '.name' | tr -d '"' | cut -d ' ' -f 2)
+SELENIUM_VERSION_MAJOR_MINOR=$(echo $SELENIUM_VERSION | cut -d '.' -f 1,2)
+
+# Download selenium standalone server (hardcoded version 3.141.59)
+echo "Downloading selenium-server-standalone v$SELENIUM_VERSION..."
+SELENIUM_JAR_NAME=selenium-server-standalone-$SELENIUM_VERSION.jar
+wget https://selenium-release.storage.googleapis.com/$SELENIUM_VERSION_MAJOR_MINOR/$SELENIUM_JAR_NAME
+
+echo "Testing to make sure that script performed as expected, and basic scenarios work"
```
AlenaSviridenko: does it makes sense to add note to these `echo`s (also on the line 45 for chromedriver) about what we are testing just to make it a bit clearer?
something like `Testing Selenium JAR downaloaded correctly`
Also it is not clear what are the basic scenarios we are testing here?

images/linux/scripts/installers/google-chrome.sh
AndreyMaslennikov: Makes sense for me, thanks

images/linux/scripts/installers/google-chrome.sh
AndreyMaslennikov: This `echo` is used for each tool we install. I believe it would be inconsistent with the other ones

AndreyMaslennikov: Also, as you can see, there are logs, which tool we're installing on the lines 36 and 62. If something goes wrong, we'll be able to determine what tool caused the failure with these logs

AlenaSviridenko: okay, thanks for the clarification.

images/linux/scripts/installers/google-chrome.sh
```diff
@@ -25,3 +25,52 @@ fi
 # Document what was added to the image
 echo "Lastly, documenting what we added to the metadata file"
 DocumentInstalledItem "Google Chrome ($(google-chrome --version))"
+
+CHROME_VERSION=$(google-chrome --version | grep -Eo "([0-9]+\.?){4}" | cut -d "." -f 1,2,3)
```
Lctrs: Can be simplified to :

```suggestion
CHROME_VERSION=$(google-chrome --product-version)
CHROME_VERSION=${CHROME_VERSION%.*}
```

AndreyMaslennikov: Looks good. Thanks for your suggestion

AndreyMaslennikov:
> thanks @AndreyMaslennikov! - looks like you're running CI builds? please let me know when they succeed and I'll merge.

Hello @alepauly! Yeah, I've ran image generation for both Ubuntu 16.04 and 18.04. Google Chrome, Chromedriver and Selenium installation looks good, but both pipelines fell on unzipping `maven.zip`. I've tried to reproduce this issue locally, and it looks like archive `maven.zip` v3.6.2 is broken. I've updated maven to 3.6.3 and ran CI again. I'll ping you when image generation is succeed. Please, let me know if you have any concerns about it.

alepauly:
@andreymaslennikov - yeah that change for maven is already in master, so it looks like you just have to rebase to the latest master. The CI build doesnt actually integrate right now, just builds off your fork if you specify it.

AndreyMaslennikov:
Hello @alepauly!
We have tested these changes on Ubuntu 16 and 18 images and they look good. I believe we can merge this PR.

images/linux/scripts/installers/google-chrome.sh
```diff
@@ -25,3 +25,53 @@ fi
...
+wget https://selenium-release.storage.googleapis.com/$SELENIUM_VERSION_MAJOR_MINOR/$SELENIUM_JAR_NAME
+
+echo "Testing to make sure that script performed as expected, and basic scenarios work"
+if [ ! -f "$SELENIUM_JAR_NAME" ]; then
+    echo "Selenium server was not installed"
+    exit 1
+fi
+
+mv "selenium-server-standalone-$SELENIUM_VERSION.jar" "/usr/share/java/selenium-server-standalone.jar"
+echo "CLASSPATH=/usr/share/java/selenium-server-standalone.jar:.:$CLASSPATH" | tee -a /etc/environment
```
headius: Forcing Selenium into the default CLASSPATH is not a good idea. This change has broken JRuby on Github Actions, because that Selenium jar includes an incompatible version of SnakeYAML.

See https://github.com/actions/virtual-environments/issues/242

AndreyMaslennikov: Hello @headius!

As a probable solution, we can remove selenium jar from the `$CLASSPATH` environment variable, declare the `$SELENIUM_JAR_PATH` environment variable, in which the path to the selenium jar will be stored, and specify parameter `-classpath` when running selenium tests.

For example:
`java -classpath $SELENIUM_JAR_PATH:$CLASSPATH executable.java`

Let me know if it looks good for you.

pkuczynski: That sounds like a reasonable idea for me.

headius: +1 seems good as long as CLASSPATH goes back to being empty by default.

PR #187, opened 2019-12-19
Add PHP 7.4

ADmad:


alepauly:
@ADmad we can abandon this now. The same change had been merged a few days ago but we had a migration snafu that stomped on top of it, should be fixed now in #189 - sorry about that!

ADmad:
@alepauly No worries, have the changes in #189 been deployed?

alepauly:
@admad, it hasn't deployed and unfortunately we are in a deployment freeze due to the holidays so this won't make it out until the first week of January.

PR #210, opened 2019-12-26
Default to tool cache Python for Ubuntu 1604/1804

konradpabjan:
Removes some extra versions of Python that get installed on Ubuntu 1604/1804. These were added before the tools cache were setup (clean up as part of tech-debt). 2 versions of Python get added to PATH so that the `python`, `python2` and `python3` commands can be used without `setup-python`. The version that get added to path are done so to align with the current default behavior.

Current Default Python on Ubuntu 1604:
image

Current Default Python on Ubuntu 1804:
image


images/linux/scripts/installers/1804/hosted-tool-cache.sh
```diff
@@ -31,14 +31,23 @@ DocumentInstalledItem "Python:"
...
+	fi
 done;
 
 # PyPy is also configured using the setup-python action
 pypys=$(ls $AGENT_TOOLSDIRECTORY/PyPy)
 for pypy in $pypys; do
 	DocumentInstalledItemIndent "PyPy $pypy"
 	# Add symlinks for pypy2 and pypy3 to usr/local/bin, there should only be 2 versions of PyPy in the tools cache that is downloaded
-	if [ ${pypy:0:1} -eq "3" ] ; then
+	if [ ${pypy:0:1} == "3" ] ; then
```
konradpabjan: Updated just for consistency with what we do for Python. Should be the same behavior. In bash if there is a string like "3.5" being compared using `-eq` then it will complain with something like  "Integer expected".

konradpabjan:
@brcrista could you take a look. This should be the last of the Python/PyPy tech debt 😎

images/linux/scripts/installers/1604/hosted-tool-cache.sh
```diff
@@ -31,14 +31,23 @@ DocumentInstalledItem "Python:"
 pythons=$(ls $AGENT_TOOLSDIRECTORY/Python)
 for python in $pythons; do
 	DocumentInstalledItemIndent "Python $python"
+	# Add symlinks for python, python2 and python3 so that Python can be used out of the box without setup-python
+	if [ ${python:0:3} == "2.7" ] ; then
+		# add python 2.7 for python and python2
+		ln -s $AGENT_TOOLSDIRECTORY/Python/$python/x64/bin/python /usr/local/bin/python
+		ln -s $AGENT_TOOLSDIRECTORY/Python/$python/x64/bin/python /usr/local/bin/python2
```
brcrista: We'll definitely want `pip` and `pip3` in `PATH`, along with the various command-line tools that come with Python.  I think the best thing to do here would just be to add the tool cache directories to `PATH` rather than create symlinks.

images/linux/scripts/installers/1604/hosted-tool-cache.sh
```diff
@@ -31,14 +31,23 @@ DocumentInstalledItem "Python:"
...
+	elif [ ${python:0:3} == "3.5" ] ; then
+		# add python 3.5 for python3 which is the default for 1604
```
brcrista: For both this and Ubuntu 18, we can take the opportunity to update to the latest Python.  The only reason we were using an older version was because of what was available through APT.

images/linux/scripts/installers/python.sh
```diff
@@ -1,27 +0,0 @@
...
-################################################################################
-##  File:  python.sh
-##  Desc:  Installs Python 2/3
-################################################################################
-
-# Source the helpers for use with the script
-source $HELPER_SCRIPTS/document.sh
-
-# Install Python, Python 3, pip, pip3
-apt-get install -y --no-install-recommends python python-dev python-pip python3 python3-dev python3-pip
```
brcrista: `python-dev` installs the .so and header files for developing C extensions for Python.  The tool cache versions have these installed as well.  I think `python setup.py` will handle setting the include file and library paths when building an extension, so hopefully this just works with the change.

I had [added a canary build](https://github.com/microsoft/azure-pipelines-canary/pull/114) for building C extensions with the versions in tool cache.  Could you make sure this works without running `Use Python Version` as well?

images/linux/scripts/installers/1604/hosted-tool-cache.sh
brcrista: You could also just add every tool cache directory to `PATH` -- `ls` should guarantee us a sorted order here.  That would make commans like `tox -e py37` work on Actions, which I think is something we should enable.

konradpabjan:
Closing for now. Waiting on the following PR to go in first since there will be some significant changes going in: https://github.com/actions/virtual-environments/pull/216

After the tools cache is switched over to use the new packages I'll go back and address this issue.

Issue #83, opened 2019-11-08
BOOST_ROOT directory is missing


rkennedy:
[Documentation says](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/software-installed-on-github-hosted-runners) `BOOST_ROOT` refers to the directory where the Boost library is installed. Although the environment variable is set to a plausible value, I find the directory doesn't exist.

**Virtual environments affected**
- [ ] macOS 10.15
- [ ] Ubuntu 16.04 LTS
- [x] Ubuntu 18.04 LTS
- [ ] Windows Server 2016 R2
- [ ] Windows Server 2019

**Expected behavior**
I expected to see a list of hundreds of .hpp files.

**Actual behavior**
I used the following workflow:

```yaml
name: Build
on:
- push
jobs:
  build:
    runs-on: ubuntu-latest
    steps:
    - run: |
        ls -lR ${BOOST_ROOT}
```

This is the output I see:

```
ls: cannot access '/usr/local/share/boost/1.69.0': No such file or directory
```

Have I misunderstood how to use Boost in this environment?

colin-daniels:
I also had some issues with this and just ended up installing boost via `apt-get` to get things working, but you're right, to me at least it doesn't seem like boost 1.69 is actually installed. Listing packages via `apt list --installed` you only get:
```
libboost-context1.65.1/bionic,now 1.65.1+dfsg-0ubuntu5 amd64 [installed,automatic]
libboost-filesystem1.65.1/bionic,now 1.65.1+dfsg-0ubuntu5 amd64 [installed,automatic]
libboost-program-options1.65.1/bionic,now 1.65.1+dfsg-0ubuntu5 amd64 [installed,automatic]
libboost-system1.65.1/bionic,now 1.65.1+dfsg-0ubuntu5 amd64 [installed,automatic]
libboost-thread1.65.1/bionic,now 1.65.1+dfsg-0ubuntu5 amd64 [installed,automatic]
```

And actually searching the filesystem for anything with boost in the name via `sudo find / -iname '*boost*'` results in:
```
/proc/sys/vm/watermark_boost_factor
/var/lib/dpkg/info/libboost-system1.65.1:amd64.triggers
/var/lib/dpkg/info/libboost-context1.65.1:amd64.triggers
/var/lib/dpkg/info/libboost-context1.65.1:amd64.md5sums
/var/lib/dpkg/info/libboost-system1.65.1:amd64.list
[...]

PR #240, opened 2020-01-08
Build boost from a tarball and bump version to 1.72.0

e-kwsm:
#83 indicates that installation of Boost on Ubuntu 18.04 have a trouble.
I'm not sure if zip archives including 1.72.0, the latest release, are always available, and just 3f4654678a4ddcdb46c1f2d7c6194be7d835415e works.
So a1ac83b64573104105973e476f6e2c185a049445 was added, where Boost is built from a tarball, but it will take up time.

Anyway, I would appreciate if the newer versions are available.

### DIFF
--- a/images/linux/scripts/installers/boost.sh
+++ b/images/linux/scripts/installers/boost.sh
@@ -15,6 +15,12 @@ do
     BOOST_SYMLINK_VER=`echo "${BOOST_VERSION//[.]/_}"`
     BOOST_ROOT="BOOST_ROOT_$BOOST_SYMLINK_VER"
 
+    wget -q https://dl.bintray.com/boostorg/release/${BOOST_VERSION}/source/boost_${BOOST_SYMLINK_VER}.tar.bz2
+    tar -xf boost_${BOOST_SYMLINK_VER}.tar.bz2
+    cd boost_${BOOST_SYMLINK_VER}
+    ./bootstrap.sh --prefix=$BOOST_LIB/$BOOST_VERSION
+    ./b2 install
+
     echo "$BOOST_ROOT=$BOOST_LIB/$BOOST_VERSION" | tee -a /etc/environment
     if [[ $BOOST_VERSION == $BOOST_DEFAULT ]]; then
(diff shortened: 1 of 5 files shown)


images/linux/scripts/installers/boost.sh
```diff
@@ -7,17 +7,19 @@
...
     BOOST_SYMLINK_VER=`echo "${BOOST_VERSION//[.]/_}"`
     BOOST_ROOT="BOOST_ROOT_$BOOST_SYMLINK_VER"
-    BOOST_ZIP="boost_`echo $BOOST_VERSION`_gcc.zip"
 
-    unzip $BOOST_ZIP_PATH/$BOOST_ZIP -d $BOOST_LIB
+    wget -q https://dl.bintray.com/boostorg/release/${BOOST_VERSION}/source/boost_${BOOST_SYMLINK_VER}.tar.bz2
+    tar -xf boost_${BOOST_SYMLINK_VER}.tar.bz2
+    cd boost_${BOOST_SYMLINK_VER}
+    ./bootstrap.sh --prefix=$BOOST_LIB/$BOOST_VERSION
+    ./b2 install
```
maxim-lobanov: Previously, version 1.69 was built with the following options:
```
sudo ./b2 install --prefix=$BOOST_DIR --build-dir=$BOOST_TEMP_DIR --layout="tagged" variant="release" link="static,shared" runtime-link="static,shared"
sudo ./b2 install --prefix=$BOOST_DIR --build-dir=$BOOST_TEMP_DIR --layout="tagged" variant="debug" link="static,shared" runtime-link="static,shared"
```

But if we do that during image-generation, it will add about 30-50 minutes for each Boost version.

kaylangan:
Can we update 1.69 to 1.72? Or do we need 1.69 also?

e-kwsm:
Personally I am using [Boost.Histogram](https://www.boost.org/doc/libs/1_70_0/libs/histogram/doc/html/index.html), which was newly added in 1.70.0.

al-cheb:
Closed in https://github.com/actions/virtual-environments/pull/316 PR

PR #258, opened 2020-01-13
Move installation of curl in ubuntu18 to last

alepauly:
In #204, a duplication curl install call was removed. Since then, the builds are failing unable to find curl during verification. Switching to the end to make sure it's there.


alepauly:
Force merge since it's low risk and to attempt to unblock 1/13 rollouts.

PR #266, opened 2020-01-13
Chrome update to 79.0.3945

MarcusFelling:
windows-2019 image was updated with Chrome 79.


miketimofeev:
@MarcusFelling we don't need to update these files manually. Docs are updated automatically after the new image rollout.

MarcusFelling:
@miketimofeev - The automated process must be broken then. The Docs do not have the correct version of Chrome or ChromeDriver.

Issue #248, opened 2020-01-10
Add CHROMEDRIVER_BIN variable to environment


cdunklau:
As of the 20200102.1 update, chromedriver is installed, but it doesn't look like the CHROMEDRIVER_BIN variable gets added to the environment. This would be nice to have, in the same vein as CHROME_BIN.

**Tool information**
- Tool name: chromedriver
- Add or update? Tweak env var for existing tool
- Desired version: N/A
- Approximate size: N/A

**Area for Triage**:
Testing and code coverage

**Question, Bug, or Feature?**:
Feature

**Virtual environments affected**
I need it for Linux, but the change is probably useful for users of the other images
- [ ] macOS 10.15
- [x] Ubuntu 16.04 LTS
- [x] Ubuntu 18.04 LTS
- [ ] Windows Server 2016 R2
- [ ] Windows Server 2019

**Can this tool be installed during the build?**
N/A

**Are you willing to submit a PR?**
Sure

kaylangan:
@cdunklau thanks for the feedback. It'd be great if you submitted a PR for this change!

cdunklau:
Looks like #262 duplicates this, but I didn't realize the env var was documented on the other VMs. It's been resolved in #269, but N.B. the env var name is `CHROMEWEBDRIVER`, not `CHROMEDRIVER_BIN` as I initially suggested.

Issue #262, opened 2020-01-13
ChromeWebDriver env variable missing on Ubuntu


nicojs:
**Describe the bug**
The [ChromeWebDriver env variable](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/software-installed-on-github-hosted-runners#chrome-driver) seems to be missing on ubunty

**Area for Triage**: 
Bash

**Question, Bug, or Feature?**: 
Bug

**Virtual environments affected**
- [ ] macOS 10.15
- [x] Ubuntu 16.04 LTS
- [x] Ubuntu 18.04 LTS
- [ ] Windows Server 2016 R2
- [ ] Windows Server 2019

**Expected behavior**
```yml
    steps:
    - run: env
```

Should also list: CHROMEWEBDRIVER

**Actual behavior**

CHROMEWEBDRIVER is missing

maxim-lobanov:
Hi @nicojs,
This section describes Windows image and this variable is available there. Ubuntu docs doesn't contain this section and variable.
On Ubuntu images, Chrome Driver is available on `/usr/bin/chromedriver`
Probably, it makes sense to introduce this variable on Ubuntu images for consistency.

kaylangan:
Related to #184

PR #269, opened 2020-01-14
Add ChromeWebDriver env variable on Ubuntu

MaksimZhukov:
In scope of this Pull Request we've fixed [this bug](https://github.com/actions/virtual-environments/issues/262)


al-cheb:
By convention in Linux environment variables  are uppercase. I think we should follow this approach.

cdunklau:
This resolves #248 too, albeit with a different env var name. Thanks!

PR #274, opened 2020-01-15
Update CMake on Linux

al-cheb:
Issue:
https://github.com/actions/virtual-environments/issues/56

Fix:
Update Cmake version to  Cmake 3.15.4

### DIFF
--- a/images/linux/scripts/installers/cmake.sh
+++ b/images/linux/scripts/installers/cmake.sh
@@ -12,7 +12,7 @@ echo "Checking to see if the installer script has already been run"
 if command -v cmake; then
     echo "Example variable already set to $EXAMPLE_VAR"
 else
-	curl -sL https://cmake.org/files/v3.12/cmake-3.12.4-Linux-x86_64.sh -o cmakeinstall.sh \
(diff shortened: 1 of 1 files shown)


alepauly:
@al-cheb #141 attempts to update to 3.15.5, any reason we can't do that here? (that other one is blocked on conflicts right now)

al-cheb:
@alepauly I have checked installed CMake version on MacOS, Windows hosted agents. These images contain CMake 3.16.2. Would it make sense to update Cmake version to 3.16.2 or 3.15.6 on Ubuntu images?

alepauly:
> 
> 
> @alepauly I have checked installed CMake version on MacOS, Windows hosted agents. These images contain CMake 3.16.2. Would it make sense to update Cmake version to 3.16.2 or 3.15.6 on Ubuntu images? The second choice we can rework script to install Cmake using apt tool - https://apt.kitware.com/ .

If it's easy to install 3.16.2 for consistency I think it makes sense. We can defer changing install mechanism for later as needed. I dont think we need to do it/risk it now.

zbeekman:
FWIW adding the Kitware APT repo, say to [`repos.sh`](https://github.com/actions/virtual-environments/blob/master/images/linux/scripts/base/repos.sh) will make upgrading and pinning CMake a little easier. I'm happy to submit a PR if you like.

PR #276, opened 2020-01-16
Add -Install to Get-WUInstall call to actually install Windows Updates.

alepauly:
Added missing param to actually install Windows updates.


alepauly:
Low risk, forcing merge to get new builds going.

PR #296, opened 2020-01-20
Update gem to latest version

dmitry-shibanov:
- Update gem to the latest version (3.1.2) on Linux and Windows
- Include gem version to docs.


images/win/scripts/Installers/Install-Ruby.ps1
```diff
@@ -17,4 +17,8 @@ $latestRubyBinPath2_5 = Get-ChildItem -Path $toolcacheRubyPath | Sort-Object {[S
 
 Add-MachinePathItem $latestRubyBinPath2_5
 $env:Path = Get-MachinePath
+
+# Update ruby gem to latest version
+gem update --system
```
sergey-akhalkov: I believe we should report to docs the latest version of gem.

dmitry-shibanov: Yes, sure. We have added docs generation.

mjroghelia:
Looks good to me.

Issue #263, opened 2020-01-13
Windows x64 Image broken when attempting to use commands provided by git bash


jtpetty:
**Describe the bug**
If an agent is attempting to run a bash script, and that script tries to run 'find', it attempts to run the Windows version of 'find' instead of that bash version of 'find'. This fails since those two commands take very different parameters. It results in the following output:

```
FIND: Parameter format not correct
```

I believe this a result of this PR #211 

This started happening on Jan 10, 2020 after 5:45p EST

I was able to add a step to add those paths back to the PATH and my pipeline works again.

ethomson:
More reports of this issue: https://github.community/t5/GitHub-Actions/Windows-tests-worked-yesterday-broken-today/td-p/43574

DaanDeMeyer:
As could be expected, there's already lots of scripts that depend on the unix tools being available from the PATH without running in the git bash shell so #211 is a breaking change (I wasn't really expecting it to get merged without some discussion to be honest especially since I explicitly asked for pointers on how to actually test the changes in the pull request).

The tools really should really never have been added to the PATH when running cmd or powershell in the first place but they were so now we're stuck in a mess where having them in the PATH breaks some workflows (My use case works perfectly since #211 was deployed) but removing them breaks other existing workflows.

If #211 is reverted since it's a breaking change, the documentation should be updated to clearly mention that MinGW won't work by default since it's own libstdc++ will be overriden by the one put in the PATH by the Git installation and manual intervention is required to make sure the correct libstdc++ is found first.

It might also be possible that chocolatey is simply not adding the git tools to the PATH correctly. However, I'm not sure if it's possible to fix that without adding libstdc++ to the PATH again which breaks chocolatey MinGW for everyone else.

alepauly:
> so #211 is a breaking change (I wasn't really expecting it to get merged without some discussion to be honest especially since I explicitly asked for pointers on how to actually test the changes in the pull request).

@DaanDeMeyer yeah, that's on me for trying to multi-task and not looking carefully enough. Removing Mingw paths seemed clear enough but I didn't think it through as I should have. Lesson learned :-)

There's currently no simple way to test other than building the machine on your own but that's quite a bit of effort. We're working on adding some CI and checks within the PRs in the coming months.

bryanmacfarlane:
This also broke our runner builds.  

I think the fix is as simple as adding a path to PATH

alepauly:
[...]

PR #304, opened 2020-01-22
Put necessary directories to the PATH

ivan-nosar:
In #211  the `C:\Program Files\Git\bin` was removed from the `PATH`. This folder contains a lot of bash tools such as `find` and its removing makes all these tools unreachable from `bash`(#263 ). 

Despite the assumption that `choco` must add this folder t the `PATH` while installing `git` - it was not done automatically so we have to add this folder to `PATH` manually.

### DIFF
--- a/images/win/scripts/Installers/Install-Git.ps1
+++ b/images/win/scripts/Installers/Install-Git.ps1
@@ -12,4 +12,6 @@ choco install git -y --package-parameters="/GitAndUnixToolsOnPath /WindowsTermin
 # Disable GCM machine-wide
 [Environment]::SetEnvironmentVariable("GCM_INTERACTIVE", "Never", [System.EnvironmentVariableTarget]::Machine)
 
+Add-MachinePathItem "C:\Program Files\Git\bin"
+
 exit 0


ivan-nosar:
Waiting for validation: https://dev.azure.com/mseng/AzureDevOps/_build/results?buildId=11354172
We are going to merge this PR once we sure that everything is good

al-cheb:
@alepauly

There are two bash.exe at git folder:
- "C:\Program Files\Git\bin\bash"
```
$ where FIND
C:\Program Files\Git\usr\bin\find.exe - Correct version in the first order
C:\Windows\System32\find.exe
```
- "C:\Program Files\Git\usr\bin\bash"
```
$ where FIND
C:\Windows\System32\find.exe     - First order Windows version of find
C:\Program Files\Git\usr\bin\find.exe
```
Correct bash version should be point at  "C:\Program Files\Git\bin\bash" ( but we remove it in #211), but  now we are using bash version point at  "C:\Program Files\Git\usr\bin\bash" and the script was failing with a `FIND: Parameter format not correct` exception as it wasn't finding the correct version of find.

PR #311, opened 2020-01-23
Set a default value for BOOST_ROOT = C:\hostedtoolcache\windows

al-cheb:
Issue:
b2.exe tool cannot properly handle with paths containing spaces, by default BOOST_ROOT = C:\Program Files\Boost\

Fix:
Set BOOST_ROOT=$Env:AGENT_TOOLSDIRECTORY(C:\hostedtoolcache\windows\Boost by default)


maxim-lobanov:
Just to confirm, we should merge this PR only after PR with switching to the GitHub package registry

PR #312, opened 2020-01-23
Add available ghc-8.8.2 to installed versions

jneira:


jneira:
Already requested in #312, sorry for the noise

PR #314, opened 2020-01-23
Update docs and pass GitHub token

dmitry-shibanov:
- add docs page for image generation
- help page was moved from [here](https://github.com/microsoft/azure-pipelines-image-generation/blob/master/help/CreateImageAndAzureResources.md) (I think that the file was lost after migration to virtual environments, that is why I returned it back).
- pass GitHub access token to script `helpers/GenerateResourcesAndImage.ps1`

### DIFF
--- a/helpers/GenerateResourcesAndImage.ps1
+++ b/helpers/GenerateResourcesAndImage.ps1
@@ -86,10 +86,10 @@ Function GenerateResourcesAndImage {
         [Switch] $Force
     )
 
-    if (([string]::IsNullOrEmpty($version)))
+    if (([string]::IsNullOrEmpty($GithubFeedToken)))
     {
-        Write-Error "You have to specify valid GitHub PAT to dowload tool packages from GitHub Package Registry"
-        exit 0
+        Write-Error "You have to specify valid GitHub PAT to download tool packages from GitHub Package Registry"
+        exit 1
     }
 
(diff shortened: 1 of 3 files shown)


helpers/GenerateResourcesAndImage.ps1
```diff
@@ -77,10 +80,18 @@ Function GenerateResourcesAndImage {
...
         [int] $SecondsToWaitForServicePrincipalSetup = 30,
+        [Parameter(Mandatory = $True)]
+        [string] $GithubFeedToken,
         [Parameter(Mandatory = $False)]
         [Switch] $Force
     )
 
+    if (([string]::IsNullOrEmpty($version)))
+    {
+        Write-Error "You have to specify valid GitHub PAT to dowload tool packages from GitHub Package Registry"
```
maxim-lobanov: typo `download`

maxim-lobanov:
@alepauly , Is there any other docs about image-generation that should be updated to notify customers about `github_feed_token` variable?

helpers/GenerateResourcesAndImage.ps1
dmitry-shibanov: Done

alepauly:
@maxim-lobanov, this is good. Maybe we'll need to link it from somewhere in the main readme eventually. @dmitriy-shibanov, thanks for reviving those instructions!

PR #333, opened 2020-01-28
[DefaultLimitNOFILE=50000] setting modification in Ubuntu16/18 images

andy-mishechkin:
[DefaultLimitNOFILE=50000] setting has been added to /etc/systemd/system.d for increasing the max number of open files in non-interactive sessions.


images/linux/ubuntu1604.json
```diff
@@ -77,7 +77,8 @@
                 "systemctl disable apt-daily-upgrade.service",
                 "echo '* soft nofile 50000 \n* hard nofile 50000' >> /etc/security/limits.conf",
                 "echo 'session required pam_limits.so' >> /etc/pam.d/common-session",
-                "echo 'session required pam_limits.so' >> /etc/pam.d/common-session-noninteractive"
+                "echo 'session required pam_limits.so' >> /etc/pam.d/common-session-noninteractive",
+                "echo 'DefaultLimitNOFILE=50000' >> /etc/systemd/system.conf"
```
sergey-akhalkov: Just wondering, why the limit is specified to 50000 (no more no less) ?

miketimofeev: @sergey-akhalkov we don't know either looks like some random value that is definitely enough(max value is 65536)

andy-mishechkin: I just repeat the previous limit value (in ["echo '* soft nofile 50000 \n* hard nofile 50000' >> /etc/security/limits.conf",])

al-cheb: Seems like the PR contains a lot of non relative files

Issue #268, opened 2020-01-13
Increase max number of open files. 


MmtBkn:
**Describe the bug**
When I run my javascript builds on Ubuntu images, I get the following error message; 
`EMFILE: too many open files`

Related SO Question;
https://stackoverflow.com/questions/8965606/node-and-error-emfile-too-many-open-files

**Area for Triage**: 
Ubuntu image

**Question, Bug, or Feature?**: 
Bug

**Virtual environments affected**
- [ ] macOS 10.15
- [x] Ubuntu 16.04 LTS
- [x] Ubuntu 18.04 LTS
- [ ] Windows Server 2016 R2
- [ ] Windows Server 2019

**Expected behavior**
Build shouldn't throw `EMFILE: too many open files` error. 

**Actual behavior**
Build throws `EMFILE: too many open files` error. 

See https://github.com/veritone/veritone-sdk/commit/48fe61ccd51f5358bed65a6a765f0aeeb5cef1a8/checks?check_suite_id=398732036

Increasing the maximum number of open files on OS level should resolve the issue. When I try to add this into the workflow file, I get permission denied. 
`ulimit -Sn 65536`

MikailBag:
Did you try this?
```bash
sudo ulimit -Sn 65536
```

miketimofeev:
@MmtBkn we have implemented a fix and it should be rolled out in 1-2 weeks. Feel free to open a new issue if it still not working.

miketimofeev:
@MmtBkn the image has been deployed. Could you check that everything is fine now, please?

MmtBkn:
Everything works fine, thank you.

RainbowMango:
@miketimofeev and @andy-mishechkin I met this issue on ubuntu-20.04 when I tried to [promote from ubuntu-18.04 to ubuntu-20.04](https://github.com/karmada-io/karmada/pull/2404). Yeah, sure it works on ubuntu-18.04.

You can get the full logs from :
https://github.com/karmada-io/karmada/runs/7952115182?check_suite_focus=true

I tried to find what's the default node number on ubuntu-20.04 as per #340, but seems the configuration changed a lot, I can't figure out it. Can you help to point that?

RainbowMango:
Yeah, I find it:
https://github.com/actions/runner-images/blob/c08d3333864a501c7e0a703c655e391d360da917/images/linux/scripts/base/limits.sh#L9-L10

[...]

PR #340, opened 2020-01-30
Increase max number of open files

andy-mishechkin:
According to https://github.com/actions/virtual-environments/issues/268, we have modified Ubuntu 16/18 images for adding the [DefaultLimitNOFILE=50000] setting to /etc/systemd/system.conf.
This setting will be used in non-interactive sessions for increasing the max open files limit

### DIFF
--- a/images/linux/ubuntu1604.json
+++ b/images/linux/ubuntu1604.json
@@ -77,10 +77,10 @@
                 "systemctl disable apt-daily.timer",
                 "systemctl disable apt-daily-upgrade.timer",
                 "systemctl disable apt-daily-upgrade.service",
-                "echo '* soft nofile 50000 \n* hard nofile 50000' >> /etc/security/limits.conf",
+                "echo '* soft nofile 65536 \n* hard nofile 65536' >> /etc/security/limits.conf",
                 "echo 'session required pam_limits.so' >> /etc/pam.d/common-session",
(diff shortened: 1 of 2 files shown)


images/linux/ubuntu1604.json
```diff
@@ -79,7 +79,8 @@
                 "systemctl disable apt-daily-upgrade.service",
                 "echo '* soft nofile 50000 \n* hard nofile 50000' >> /etc/security/limits.conf",
                 "echo 'session required pam_limits.so' >> /etc/pam.d/common-session",
-                "echo 'session required pam_limits.so' >> /etc/pam.d/common-session-noninteractive"
+                "echo 'session required pam_limits.so' >> /etc/pam.d/common-session-noninteractive",
+                "echo 'DefaultLimitNOFILE=50000' >> /etc/systemd/system.conf"
```
sergey-akhalkov: Just wondering why we should not use Max possible number, what the motivation to set 50000?

andy-mishechkin: Previous max file limit just has been repeated in this setting

sergey-akhalkov: Let's change it to the max possible value if you see not issues with it.

andy-mishechkin: Max files limit has been changed to 65536

Issue #337, opened 2020-01-29
[Ubuntu] Failed to apt-get libmysqlclient-dev


ivailop:
**Describe the bug**
On Ubuntu 16.04, since today (1/29/2020), installation of `libmysqlclient-dev` errors out:
```
sudo apt-get libmysqlclient-dev
...
Err:87 http://security.ubuntu.com/ubuntu xenial-security/main amd64 libmysqlclient-dev amd64 5.7.28-0ubuntu0.16.04.2
  404  Not Found
Fetched 30.0 MB in 0s (30.3 MB/s)
E: Failed to fetch http://security.ubuntu.com/ubuntu/pool/main/m/mysql-5.7/libmysqlclient-dev_5.7.28-0ubuntu0.16.04.2_amd64.deb  404  Not Found

E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?
```

**Area for Triage**: 

**Question, Bug, or Feature?**: 
Bug

**Virtual environments affected**
- [ ] macOS 10.15
- [x] Ubuntu 16.04 LTS
- [ ] Ubuntu 18.04 LTS
- [ ] Windows Server 2016 R2
- [ ] Windows Server 2019

**Expected behavior**
Install successfully

**Actual behavior**
Errors out.

```
E: Failed to fetch http://security.ubuntu.com/ubuntu/pool/main/m/mysql-5.7/libmysqlclient-dev_5.7.28-0ubuntu0.16.04.2_amd64.deb  404  Not Found
```

ivailop:
FYI, similar issue (but with `5.7.27`) happened couple of months ago - https://github.com/microsoft/azure-pipelines-image-generation/issues/1410

To prevent this environment volatility wouldn't it be better to have a more stable version of MySQL (e.g. `5.5`, instead of `5.7`) installed on the hosted images?

miketimofeev:
@ivailop thanks for reporting this issue. MySQL will be updated to 5.7.29 on the next rollout, and the problem should disappear. 
We will also consider adding libmysqlclient-dev during the image generation step to avoid such inconvenience in the future.

al-cheb:
Temporary workaround:
```
jobs:
  build:
    runs-on: [ubuntu-16.04]
    steps:
    - name: libmysql
      run: |
        sudo apt-get update && sudo apt-get install libmysqlclient-dev
```

[...]

PR #341, opened 2020-01-30
Add libmysqlclient-dev installation on Ubuntu

miketimofeev:
We have several complaints about libmysqlclient-dev installation on Ubuntu images. 
https://github.com/actions/virtual-environments/issues/337
https://github.com/microsoft/azure-pipelines-image-generation/issues/1410
This happens when MySQL version is superseded in ubuntu repositories, but we still have an older one on the deployed VM and it requires the same outdated libmysqlclient-dev version, which is no longer available via apt-get.
The solution is to install libmysqlclient-dev during image generation to have the same versions of packages at the end of the build. 

Changes:
- Add libmysqlclient-dev installation to MySQL Client installation script

### DIFF
--- a/images/linux/scripts/installers/mysql.sh
+++ b/images/linux/scripts/installers/mysql.sh
@@ -13,7 +13,7 @@ export ACCEPT_EULA=Y
 apt-get install mysql-client -y
 
 # InstallMySQL database development files
-apt-get install libmysqlclient-dev
+apt-get install libmysqlclient-dev -y
 
 # Install MySQL Server
 MYSQL_ROOT_PASSWORD=root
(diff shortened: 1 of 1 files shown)


images/linux/scripts/installers/mysql.sh
```diff
@@ -12,6 +12,9 @@ export ACCEPT_EULA=Y
 # Install MySQL Client
 apt-get install mysql-client -y
 
+# InstallMySQL database development files
+apt-get install libmysqlclient-dev
```
alepauly: should `-y` be specified? I'm guessing yes...

miketimofeev: yeah, let's add it just to be on the safe side

Issue #50, opened 2019-10-18
Add Windows Subsystem for Linux to Windows


dcermak:
**Tool information**
- Tool name: Windows Subsystem for Linux
- Add or update? Add
- Desired version: latest
- Approximate size:
- If this is an add request:
  - Brief description of tool: Virtual Linux environment for Windows 10
  - URL for tool's homepage: https://docs.microsoft.com/en-us/windows/wsl/about

**Virtual environments affected**
- [ ] macOS 10.14
- [ ] Ubuntu 16.04 LTS
- [ ] Ubuntu 18.04 LTS
- [x] Windows Server 2016 R2 (not sure if available here)
- [x] Windows Server 2019

**Can this tool be installed during the build?**
No, it requires a reboot of the Windows machine.

**Are you willing to submit a PR?**
Willing yes, able: no (I have literally zero Windows experience). In theory you just have to run:
``` Powershell
Enable-WindowsOptionalFeature -Online -FeatureName Microsoft-Windows-Subsystem-Linux
```
as the administrator and reboot.

TomasHubelbauer:
> Are you willing to submit a PR?

Where exactly would one send a PR for this anyway, there's no agent provisioning code in this repo, no?
In any case, I'd like to vouch for this feature to be added!

dcermak:
> > Are you willing to submit a PR?
> 
> Where exactly would one send a PR for this anyway, there's no agent provisioning code in this repo, no?

I don't think you can, at least I haven't found a way to do so.

eine:
@dcermak, is there any specific advantage in installing WSL instead of using a proper MSYS2 installation (#30) and/or docker containers?

TomasHubelbauer:
I feel like "proper" is a loaded word. In one way, WSL is a part of Windows and WSL2 is an actual Linux kernel, so one could argue that makes it more proper than MSYS2 which you need to install yourself and is a recompiled binary for Windows, not the actual Linux binary. I personally think WSL is more important to add, because you can (as I understand it) install MSYS2 yourself in the workflow run, not so much for WSL which needs to be enabled as a Windows component, which requires restart and at the current, the agents are not set up so that they could pick up a workflow run after their restart.

dcermak:
eine <notifications@github.com> writes:

> @dcermak, is there any specific advantage in installing WSL instead of
> using a proper MSYS2 installation (#30) and/or docker containers?

Yes, if you want to test a WSL image, you actually do need WSL ;-)

[...]

Issue #30, opened 2019-10-11
Add MSYS2 to Windows


kaylangan:
**Tool information**
- Tool name: MSYS2
- Add or update? Add
- Desired version: Latest
- Approximate size: 
- If this is an add request:
  - Brief description of tool: Providing MSYS2 instead of [MinGW](https://help.github.com/en/articles/software-in-virtual-environments-for-github-actions#mingw) would enable a way easier package management for MinGW (via Pacman) and also faster builds.
  - URL for tool's homepage: https://www.msys2.org/

**Virtual environments affected**
- [ ] macOS 10.14
- [ ] Ubuntu 16.04 LTS
- [ ] Ubuntu 18.04 LTS
- [x] Windows Server 2016 R2
- [x] Windows Server 2019

**Can this tool be installed during the build?**
Yes, via `choco install msys2`.

**Are you willing to submit a PR?**

1138-4EB:
Ref actions/starter-workflows#95 (/cc @krcroft @andymckay)

Ref [numworks/setup-msys2](https://github.com/numworks/setup-msys2) [1138-4EB/setup-msys2](https://github.com/1138-4EB/setup-msys2) (/cc @Ecco)

eine:
I suggest to download and extract the available MSYS2 tarball, instead of using chocolatey. This is because chocolatey requires ~5min but the tarball is extracted in 30s. Please, see 
https://github.com/actions/toolkit/pull/165#issuecomment-566247350 for other references related to current workarounds for this issue.

rgl:
@eine that time difference is probably because one is updating the base tarball binaries and the other is not.

eine:
@rgl, although it might be the case, I don't think it is the only reason. Precisely, an option is available in setup-msys2 which allows to not only extract the tarball, but also execute `pacman -Syu`. Extraction time is ~40s (https://github.com/numworks/setup-msys2/runs/348237664) and extraction plus update requires ~70s (https://github.com/numworks/setup-msys2/runs/348238118). Conversely, `choco install msys2` takes 270s (https://github.com/eine/ghdl-packaging/commit/069ae680c38a1832950a377618dc4a1798d6ccf3/checks?check_suite_id=231637124) at least.

[...]

PR #342, opened 2020-01-31
Install WSL and MSYS2 on Windows Images

AndreyMaslennikov:
In this PR we enabled WSL Windows feature, requested in #50, and installed Debian Image to use it in the WSL.
We also installed MSYS2 tool


images/win/scripts/Installers/Install-DebianImage.ps1
```diff
@@ -0,0 +1,29 @@
+################################################################################
+##  File:  Install-DebianImage.ps1
+##  Desc:  Install Debian Image for WSL
+################################################################################
+
+Import-Module -Name ImageHelpers
+
+debianTempDir = "C:\Temp\debian"
```
sergey-akhalkov: Is it possible to do not hardcode path to the temp folder?

images/win/scripts/Installers/Install-MSYS2.ps1
```diff
@@ -0,0 +1,11 @@
+################################################################################
+##  File:  Install-MSYS2.ps1
+##  Desc:  Install MSYS2 via chocolaty
+################################################################################
+
+Import-Module -Name ImageHelpers
+
+$msys2Path = "C:\tools\msys64"
+choco install msys2 -y --params "/InstallDir:$msys2Path"
```
alepauly: I'm rather sure that the mingw toolchains can be installed through msys2 so ideally we do that here and remove the separate installation of mingw. Also, I wonder if it's possible to tell GitForWindows to just use the installed msys2/mingw rather than bring in it's own. Any thoughts?

images/win/scripts/Installers/Install-DebianImage.ps1
```diff
@@ -0,0 +1,29 @@
...
+
+if (-not (Test-Path $debianPath)) {
+    New-Item -Path debianTempDir -ItemType Directory
+}
+
+Push-Location debianTempDir
+
+$debianZipName = "debian.zip"
+Invoke-WebRequest -Uri "https://aka.ms/wsl-debian-gnulinux" -OutFile $debianZipName
```
alepauly: curious why did you choose Debian?

images/win/Windows2016-Azure.json
```diff
@@ -217,10 +217,28 @@
                 "{{ template_dir }}/scripts/Installers/Install-ServiceFabricSDK.ps1"
             ]
         },
+        {
+            "type": "powershell",
+            "scripts":[
+                "{{ template_dir }}/scripts/Installers/Install-WSL.ps1"
```
alepauly: I don't think we should make these changes to the Windows2016 - I don't see any requests asking for them and it just adds one more element to destabilize an older setup. @thejoebourneidentity thoughts?

images/win/scripts/Installers/Install-MSYS2.ps1
alepauly: Also, how does this affect existing builds that depend on the currently installed mingw?

alepauly: another comment. if avoiding chocolatey is not substantially difficult, I think we should avoid it since waiting for choco packages adds another layer of delay and potential problems that we can't control. [This comment](https://github.com/actions/virtual-environments/issues/30#issuecomment-566275476) by @eine suggests it should be easy enough to just untar.

eine:
As commented by @alepauly, I believe that it'd be desirable to get the tarball from MSYS2's site (as done in setup-msys2), instead of using chocolatey. Not only it avoids an unnecessary layer of delay/complexity, but it is also faster.

https://github.com/numworks/setup-msys2/blob/4895c26cd1bbdf4052003c6d771c3c713932bf36/index.js#L24-L26

> I'm rather sure that the mingw toolchains can be installed through msys2 so ideally we do that here and remove the separate installation of mingw. Also, I wonder if it's possible to tell GitForWindows to just use the installed msys2/mingw rather than bring in it's own. Any thoughts?

> Also, how does this affect existing builds that depend on the currently installed mingw?

Actually, if MSYS2 is installed, GitForWindows should not be required at all. MSYS2 provides three flavours: MSYS, MINGW64 or MINGW32. Hence, starting an MINGW64 shell from MSYS2 provides the same features as GitForWindows, plus `pacman` is available.

Precisely, issue #30 (which should probably be closed by this PR) is about *replacing* the existing `mingw` package with MSYS2, not about adding MSYS2 only. Hence, the default `bash` shell would be MSYS2's MINGW64.

Furthermore, I think that the following points should also be considered:

- If the tarball is used, an MSYS2 shell should be executed once, for the setup to be completed. Any dummy command will work. The result will be an installation with versions dating back to when the tarball was published. In setup-msys2, it is supported to optionally update the installation too (https://github.com/numworks/setup-msys2/blob/4895c26cd1bbdf4052003c6d771c3c713932bf36/index.js#L39-L42). However, I believe that the default environment should NOT be updated.

- If both, `mingw` and MSYS2, are installed, I think that `$msys2Path = "C:\tools\msys64"` might be a conflict.

- If both, `mingw` and MSYS2, are installed, I suggest adding a shell type (`msys`), which is equivalent to `bash`, but which uses the MSYS2 installation instead of `mingw`. Otherwise, custom shell entrypoints will still be required for multi-line YAML fields: https://github.com/actions/toolkit/pull/165#issuecomment-566247350

Ref actions/starter-workflows#95

TomasHubelbauer:
This looks great! I don't mind the choice of Debian as long as I can install Ubuntu instead, but perhaps we could also install both at the same time? Would that be too wasteful? Ubuntu is much more popular than Debian in WSL so I think many people would be switching to it anyway, so it makes sense to have it out of the box. Thoughts?

images/win/scripts/Installers/Install-MSYS2.ps1
MSP-Greg: Re using the tar, I created an [`Install-MSYS2.ps1`](https://github.com/MSP-Greg/msys2-action/blob/install-msys2/Install-MSYS2.ps1) script mentioned in [Issue #30 comment](https://github.com/actions/virtual-environments/issues/30#issuecomment-575688053). Also installs typical MSYS2 shell tools and basic 32 & 64 bit compilers.

Having used MSYS2 for a while, the main tar is not revised often, as the current release is dated 2019-May-24.  Hence, the more important code is the update code, which hopefully can be run any time images are updated...

MSP-Greg:
[Above](https://github.com/actions/virtual-environments/pull/342#discussion_r373746137), I mentioned a ps1 script written to install MSYS2.  Please see the following, which is a link to MSYS2's script that they use for CI on Azure Pipelines:

https://github.com/msys2/MINGW-packages/blob/master/azure-pipelines.yml

Also note that they have a repo with a base install of MSYS2, which is used in their script:
https://github.com/msys2/msys2-ci-base

images/win/scripts/Installers/Install-MSYS2.ps1
eine: > Hence, the more important code is the update code, which hopefully can be run any time images are updated...

I beg to differ. Updating by default prevents users from testing their tools on a stable environment.

MSP-Greg:
@eine 

> I beg to differ. Updating by default prevents users from testing their tools on a stable environment.

No.  From my perspective, a stable environment is the equivalent of a fresh MSYS2 install, since pacman does not normally allow older packages to be installed.  If you want to download and install older packages, have at it.  Most users will not want that.

Given your statement, what then determines when MSYS2 (or anything for that matter) gets updated?

eine:
> From my perspective, a stable environment is the equivalent of a fresh MSYS2 install, since pacman does not normally allow older packages to be installed.

That's exactly what I said. The tar should be uncompressed and any dummy command should executed (say `uname -a`), so that the installation is completed. But that command should NOT be `pacman -Syu`.

Conversely, you [said](https://github.com/actions/virtual-environments/pull/342#discussion_r373746137) that "the more important code is the update code, which hopefully can be run any time images are updated", which I understand as updating MSYS (executing `pacman -Syu`) each time that environment image (`windows-latest`) are updated. This is what I differ with.

> Given your statement, what then determines when MSYS2 (or anything for that matter) gets updated?

1. Maintainers of MSYS2, when they release a new version to the upstream location: http://repo.msys2.org/distrib/x86_64/
2. Any user, by executing `pacman -Syu` explicitly, or by using action setup-msys2, which will hopefully be updated according to this PR.

MSP-Greg:
@eine 

Ever used MSYS?  One of the main changes in MSYS2 was the use of pacman.

> Maintainers of MSYS2, when they release a new version to the upstream location

Since pacman is able to update all of one's MSYS2 install, releases are not done as frequently as other software.

Since you mentioned them, why does all their CI for packages run with a fully updated version?

Why do all their docs mention performing the equivalent of `pacman -Syyuu` immediately after install?

Actions should try to update/refresh MSYS2 anytime a new image is released.  People will prefer it, and the extra resources (time, downloads) needed to update it wouldn't be needed.

eine:
@MSP-Greg

> Ever used MSYS? One of the main changes in MSYS2 was the use of pacman.

> Since pacman is able to update all of one's MSYS2 install, releases are not done as frequently as other software.

These are two facts. What do you mean by stating them?

> Since you mentioned them, why does all their CI for packages run with a fully updated version?

I believe that's an standard procedure to detect regression bugs in the development branch of a tool, innit? I.e. this question/discussion would better fit in their repo/forum, rather than here.

> Why do all their docs mention performing the equivalent of `pacman -Syyuu` immediately after install?

I'm afraid that's not true. It is not the first time you word your very personal point of view as if it was some objetive fact. I'd like to please ask you to do an effort to make it clear what it is, and what you'd like to be.

On the one hand, I didn't find `pacman -Syyuu` or any equivalent command to be mentioned in the docs. I found `-Syu` and `-Su`, which might be considered as `-Syuu`. Still, it is explicitly suggested to do it in separate steps.

On the other hand, they suggest it to make it immediately after install, *as long as* you want to update the packages. It makes sense to suggest using updating commands in order to have packages updated.

I acknowledge that the frontpage (https://www.msys2.org/) might be misleading. But I believe that the docs are clear enough:

- https://github.com/msys2/msys2/wiki/MSYS2-installation#ii-installation: no reference to any `pacman` command at all. It is said that "*you should start MSYS2*", "*after this initial run you MUST restart MSYS2*" and "*now you can update the base MSYS2 packages*". Note that *should*, *MUST*, and *can* are used. Not *MUST*, *MUST* and *MUST*.
- https://github.com/msys2/msys2/wiki/MSYS2-installation#iii-updating-packages: the procedure to update the packages "*in the right way*" is explained. It is explicitly said that executing `pacman -Syuu` is likely not to work if done just once: "*Follow the instructions. Repeat this step until it says there are no packages to update*". Moreover, "*there's an update-core script that helps you to update MSYS2 in the right way*". Note, that there might be further complexities to account for: "*updating bash, MSYS2 or Pacman itself can cause subsequent package updates to fail. For this reason, the safest procedure for updating MSYS2 is to do it in two stages*".
- https://github.com/msys2/msys2/wiki/MSYS2-installation#v-issues-and-workarounds: "***Please read "III." above, carefully :-) We will continue working towards trouble-free updates***".

> Actions should try to update/refresh MSYS2 anytime a new image is released. People will prefer it, and the extra resources (time, downloads) needed to update it wouldn't be needed.

I think you forgot to add *some* before *people*. Having it updated by default involves a maintenance burden, due to the references above, and it will break multiple use cases. Those use cases are maintained by people, which will be forced to keep installating a separate "clean" MSYS2. [Once again](https://github.com/actions/virtual-environments/issues/30#issuecomment-575937387), I feel that you are ok with having your particular preferences placed, but seem to show little regard to other users' needs.

Honestly, I suggest you to get involved in the development of setup-msys2. I believe that many of your ideas are useful for many users indeed, and I'd be really happy to have those supported by a single action, instead of fragmenting the ecosystem. Nevertheless, I believe that most of them need to be optional; thus, should not be applied to this PR.

Please, note that, regarding setup time (which you seem to be more concerned about), setup-msys2 can be combined with a caching approach, so that you have your custom MSYS2 install ready-to-use.

MSP-Greg:
@eine

`pacman -Syyuu` is the equivalent of much of what you mentioned.  From my perspective, using an MSYS2 install from nothing but the current tar is equivalent to using Windows 10 with no updates applied.

It happened a while ago, so I can't remember all the details, but sometimes package updates cannot be applied piecemeal.

There was an issue with the AppVeyor MSYS2 install, and I needed one package updated.  Jumped thru all sorts of hoops trying to figure out how to do it, and couldn't force it with script.  I was writing code and running it locally to show dependency chains, etc.

The main reason for trying to do so was that the MSYS2 install at the time was outdated, and, given all the packages that were installed, running `pacman -Syyuu` took a long time, probably 8 to 10 minutes.

I asked about it in MSYS2, and the response was equivalent to "what you're trying to do should work, but there will be some rare package updates that can only be performed by `pacman -Syyuu`".  So, if Actions doesn't update their MSYS2 install, the same thing may happen...

Anyway, maybe we'll just agree that we disagree.

> Honestly, I suggest you to get involved in the development of setup-msys2

For now, I'm just waiting for an independent, normal install in the image(s).  I have an action (https://github.com/MSP-Greg/msys2-action), which uses the embedded MSYS2, and is directed towards gcc use.

It's also somewhat Ruby focused, as older Ruby versions (which I encourage people not to use) are compiled with OpenSSL 1.0.2.  There's logic to detect the Ruby version and switch OpenSSL packages accordingly.

Anyway, I'll have a look at `setup-msys2`...

eine:
> From my perspective, using an MSYS2 install from nothing but the current tar is equivalent to using Windows 10 with no updates applied.

I mostly agree, and I'd expect any average user to update MSYS2 periodically. However, this is not always true. Some users (which are not familiar with the environment) will just install it and use it straightaway. Moreover, in some "computer rooms" installations are preset, reset after each restart and do not provide admin permissions to the user. These are the use cases that (as maintainer of PKGBUILD recipes for MSYS2), I'd like to be able to test. That allows me to know whether it should work, or if users need to talk to their admins.

> It happened a while ago, so I can't remember all the details, but sometimes package updates cannot be applied piecemeal.

Just for the sake of completeness, should you remember or find a link, please do add it here or in #30. I believe it might be useful for someone in the future.

> The main reason for trying to do so was that the MSYS2 install at the time was outdated, and, given all the packages that were installed, running `pacman -Syyuu` took a long time, probably 8 to 10 minutes.

I think that situations such as this one you describe should be handled by requesting/requiring MSYS2 maintainers to release a new package. As you said, MSYS2 tarballs are updated once or twice a year. Overall, I'm ok with that. Nevertheless, if the latest is so outdated that 10min are required, I believe that's a fair point to nudge them. There should not be 10+ breaking changes to the core in a year anyway. Moreover, as you pointed, I believe they don't want thousands of potential CI jobs hitting their servers so hard.

> For now, I'm just waiting for an independent, normal install in the image(s). I have an action ([MSP-Greg/msys2-action](https://github.com/MSP-Greg/msys2-action)), which uses the embedded MSYS2, and is directed towards gcc use.

I understand that Ruby-specific details might not fit in setup-msys2, so it would still make sense for you to keep your own action. Nonetheless, features such as having gcc and other build-essentials set up can be useful for many users. Hence, setup-msys2 might support some "preset" groups which are optionally cached separatedly. Of course, this is just me thinking loud. Regardless of our discussion, GitHub employees/maintainers might take any approach.

MSP-Greg:
@eine

> Nevertheless, if the latest is so outdated that 10min are required

Sorry, the main reason that `pacman -Syyuu` took so long is that they (AppVeyor) have a lot packages installed,  see:
https://ci.appveyor.com/project/MSP-Greg/appveyor-ruby

But, the fact that they have all of those packages installed means something.  As stated elsewhere, I feel that Actions should install standard compiler tools and leave individual packages up to the people using them.

> as maintainer of PKGBUILD recipes for MSYS2

What packages do you maintain?  Just wondering what you normally work with...

> I understand that Ruby-specific details

It's not really Ruby specific, more about allowing installation of OpenSSL 1.0.2 when needed.  I just provided an easy 'translation' for it into Ruby terms...

eine:
> What packages do you maintain? Just wondering what you normally work with...

GHDL, a FOSS analyser, compiler, simulator and synthesiser for VHDL: https://github.com/ghdl/ghdl/tree/master/dist/msys2-mingw It supports three different backends, so up to 6 different PKGBUILD files might exist. Currently, only two of them are maintained. Moreover, there are also variants to test 'stable' releases: https://github.com/eine/ghdl-packaging/tree/master/msys2-mingw. Hence, the full matrix might have up to 18 items.

images/win/scripts/Installers/Install-DebianImage.ps1
AndreyMaslennikov: The only reason to use Debian is because this image takes up less space than others.
I don't see any objections to using Ubuntu instead, so if that makes sense, let's use it

images/win/Windows2016-Azure.json
AndreyMaslennikov: As I can see there is a request for the both Windows2016 and 2019 images in the issue #50.
We also planned to use bash from the WSL as default . Wouldn't it be confused if we'll use different `bash.exe` for these images?

images/win/scripts/Installers/Install-DebianImage.ps1
AndreyMaslennikov: Sure, I changed the path to `"$env:AGENT_TEMPDIRECTORY\debian"`

MSP-Greg:
See draft PR https://github.com/actions/virtual-environments/pull/355 for a ps1 script installing MSYS2 via download from repo.msys2.org.

images/win/scripts/Installers/Install-MSYS2.ps1
AndreyMaslennikov: @alepauly, MSYS2 installation is already implemented in the #355 PR. Could you please also take a look at this PR? If it looks good for you, we probably should merge #355 (after adding some tests), and remove MSYS2 installation from this PR. What do you think about it?

eine:
@AndreyMaslennikov, any updates on this?

maxim-lobanov:
@eine , we have reworked our approach a bit and still working on implementation / validation.
This PR can be closed but @m-minasyan will publish the new one soon

eine:
Ref #928 #916 #906 #899 #342

joelvaneenwyk:
Awesome work on MSYS2 and so glad to see that merged in. The reworked pull requests seem to no longer include WSL support though. Any chance that part can come back? Help needed on that? Seems fairly trivial to enable and would make me (and likely others) insanely happy. If there are concerns / risks around enabling it please let us know. Having WSL enabled seems to also be the only legitimate path for Linux containers on Docker for Windows now that LCOW seems deprecated.

PR #354, opened 2020-02-03
Add net core path to docs

dmitry-shibanov:
- add net core path to docs
- add net core path to doc generation

### DIFF
--- a/images/win/Windows2019-Readme.md
+++ b/images/win/Windows2019-Readme.md
@@ -126,6 +126,7 @@ _Environment:_
 ## .NET 4.8
 
 _Version:_ 4.8.03761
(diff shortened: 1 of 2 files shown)


images/win/Windows2019-Readme.md
```diff
@@ -125,7 +125,8 @@ _Environment:_
 
 ## .NET 4.8
 
-_Version:_ 4.8.03761
+_Version:_ 4.8.03761
+_Path:_ C:\Program Files (x86)\Microsoft SDKs\Windows\v10.0A\bin\NETFX 4.8 Tools
```
alepauly: probably fine here but in the future we dont really want to modify the readme manually since it won't represent what the code would have generated.

dmitry-shibanov: Yes, thank you for the response and advice. I just wanted to add it manually to prevent questions about it until next image rollouts.

Issue #353, opened 2020-02-03
Ubuntu16 build failed on PowerShell Core installation


andy-mishechkin:
**Describe the bug**
Ubunut16 build faults with PowerShell installation error. It looks like PowerShell installation in Ubuntu 16 has been broken

**Area for Triage**: 
Packages

**Question, Bug, or Feature?**: 
Bug

**Virtual environments affected**
- [ ] macOS 10.15
- [x] Ubuntu 16.04 LTS
- [ ] Ubuntu 18.04 LTS
- [ ] Windows Server 2016 R2
- [ ] Windows Server 2019

**Expected behavior**
The build should pass

**Actual behavior**
Build falls on `powershellcore.sh` script execution with error:
`powershell : Depends: libicu60 but it is not installable`

andy-mishechkin:
The PR has been created:
https://github.com/actions/virtual-environments/pull/361

andy-mishechkin:
The next PR has been created: https://github.com/actions/virtual-environments/pull/379

andy-mishechkin:
Build has been successful

PR #361, opened 2020-02-04
Fix PowerShell Core installation in ubuntu16

andy-mishechkin:
Issue: https://github.com/actions/virtual-environments/issues/353
"packages-microsoft-prod.deb" package contains link to PowerShell Core for Ubuntu 18.04.
This release of PowerShell Core doesn't work without libicu60 library, which is not present in Ubuntu 16.04, therefore we had a Ubuntu 16 build faults.
For prevention of this issue, libicu60 installation has been added to powershellcore.sh script

### DIFF
--- a/images/linux/scripts/installers/1604/powershellcore.sh
+++ b/images/linux/scripts/installers/1604/powershellcore.sh
@@ -8,9 +8,9 @@
 source $HELPER_SCRIPTS/document.sh
 
 LSB_RELEASE=$(lsb_release -rs)
-echo "Installing the libicu60 library"
-wget http://security.ubuntu.com/ubuntu/pool/main/i/icu/libicu60_60.2-3ubuntu3_amd64.deb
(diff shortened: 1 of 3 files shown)


images/linux/scripts/installers/1604/powershellcore.sh
```diff
@@ -8,8 +8,11 @@
 source $HELPER_SCRIPTS/document.sh
 
 LSB_RELEASE=$(lsb_release -rs)
+#echo "Installing the libicu60 library"
+#wget http://security.ubuntu.com/ubuntu/pool/main/i/icu/libicu60_60.2-3ubuntu3_amd64.deb
+#dpkg -i libicu60_60.2-3ubuntu3_amd64.deb
```
sergey-akhalkov: Does it really make sense to leave commented code here? Let's just remove it if we cannot see any good reasons.

PR #374, opened 2020-02-06
Add lerna to windows images

dsame:
Install lerna npm globally
Test lerna is in path

### DIFF
--- a/images/win/scripts/Installers/Install-NodeLts.ps1
+++ b/images/win/scripts/Installers/Install-NodeLts.ps1
@@ -31,4 +31,4 @@ npm install -g gulp-cli
 npm install -g parcel-bundler
 npm install -g --save-dev webpack webpack-cli
 npm install -g yarn
-npm install --global lerna
+npm install -g lerna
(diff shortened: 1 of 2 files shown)


images/win/scripts/Installers/Install-NodeLts.ps1
```diff
@@ -31,3 +31,4 @@ npm install -g gulp-cli
 npm install -g parcel-bundler
 npm install -g --save-dev webpack webpack-cli
 npm install -g yarn
+npm install --global lerna
```
maxim-lobanov: please use `-g` instead of `--global` just to be consistent with previous lines

images/win/scripts/Installers/Validate-NodeLts.ps1
```diff
@@ -26,6 +26,16 @@ else
     exit 1
 }
 
+if((Get-Command -Name 'lerna'))
```
AlenaSviridenko: I think we don't need double brackets in this condition, `if(Get-Command -Name 'lerna')` should be enough.

PR #378, opened 2020-02-06
Removed duplicate from Ubuntu1804-README.md

kmadof:
I'm not 100% sure whether it is duplicate or there is multiple docker client installed side by side and it it wrong version.

### DIFF
--- a/images/linux/Ubuntu1804-README.md
+++ b/images/linux/Ubuntu1804-README.md
@@ -35,7 +35,6 @@ The following software is installed on machines with the 20200119.1 update.
 - CMake (cmake version 3.16.2)
 - Docker Compose (docker-compose version 1.22.0, build f46880fe)
 - Docker (Docker version 3.0.8, build 2355349d)
-- Docker (Docker version 3.0.8, build 2355349d)
 - .NET Core SDK:
   - 3.0.102
   - 3.0.101


miketimofeev:
@kmadof thanks for reporting this. Readme is being created automatically during image generation, so we should find the root cause there, no need to fix the readme file itself.

miketimofeev:
Looks like these 2 scripts provide the same output
https://github.com/actions/virtual-environments/blob/master/images/linux/ubuntu1604.json#L130#L131

kmadof:
@miketimofeev I see your point. I see two possible causes:

- one script is not needed
- one script should not edit docker image

Am I right?

kmadof:
@miketimofeev I see your point. I see two possible causes:

- one script is not needed
- one script should not change README.md

Am I right?

miketimofeev:
@kmadof looks so.
@alepauly it seems that docker.sh script extends docker-moby.sh and I'm a bit confused why do we need both of them?

alepauly:
> @alepauly it seems that docker.sh script extends docker-moby.sh and I'm a bit confused why do we need both of them?

@ds-ms made those changes originally- and he probably knows best. Deepak, can you helps here? thanks.

ds-ms:
Earlier when I made that change only one of those scripts was used in the packer template. 
But, from the looks of it, it seems to be a duplicate entry.

We can merge the scripts into one, `docker.sh` is being used for caching some images during the vm image generation, and `docker-moby.sh` does only `moby` install.

kmadof:
Do you think that this PR is good place to solve that issue or maybe sth more accurate is needed?

ds-ms:
This seems like the right place to me, given that the root cause of the problem is duplicate scripts.
Thoughts?

al-cheb:
Hello, @kmadof 
Please don't update autogenerated documentation file. Duplicated and will be fixed in https://github.com/actions/virtual-environments/pull/470 PR.

PR #379, opened 2020-02-06
Ubuntu16 build has been fixed

andy-mishechkin:
According to https://github.com/actions/virtual-environments/issues/353, dotnetcore-sdk.sh has been modified.
Addition of extra repository 'microsoft-ubuntu-bionic-prod' has been removed from this script (right microsoft package already is added through repos.sh script), because it impacted to PowerShell Core installation and build has been broken
After this modification Ubuntu16 build is successful.

### DIFF
--- a/images/linux/scripts/installers/1604/dotnetcore-sdk.sh
+++ b/images/linux/scripts/installers/1604/dotnetcore-sdk.sh
@@ -34,11 +34,6 @@ for latest_package in ${LATEST_DOTNET_PACKAGES[@]}; do
     echo "Determing if .NET Core ($latest_package) is installed"
     if ! IsInstalled $latest_package; then
         echo "Could not find .NET Core ($latest_package), installing..."
-        #curl https://packages.microsoft.com/keys/microsoft.asc | gpg --dearmor > microsoft.gpg
-        #mv microsoft.gpg /etc/apt/trusted.gpg.d/microsoft.gpg
-        #sh -c 'echo "deb [arch=amd64] https://packages.microsoft.com/repos/microsoft-ubuntu-bionic-prod bionic main" > /etc/apt/sources.list.d/dotnetdev.list'
-        #apt-get install apt-transport-https
-        #apt-get update
         apt-get install $latest_package -y
     else
         echo ".NET Core ($latest_package) is already installed"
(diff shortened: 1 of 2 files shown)


miketimofeev:
@andy-mishechkin do we need the same for ubuntu18 dotnet installation script just to be consistent?

andy-mishechkin:
I didn't investigate Ubuntu 18 deployment yet, and I need some time for this

images/linux/scripts/installers/1604/dotnetcore-sdk.sh
```diff
@@ -34,11 +34,6 @@ for latest_package in ${LATEST_DOTNET_PACKAGES[@]}; do
     echo "Determing if .NET Core ($latest_package) is installed"
     if ! IsInstalled $latest_package; then
         echo "Could not find .NET Core ($latest_package), installing..."
-        curl https://packages.microsoft.com/keys/microsoft.asc | gpg --dearmor > microsoft.gpg
-        mv microsoft.gpg /etc/apt/trusted.gpg.d/microsoft.gpg
-        sh -c 'echo "deb [arch=amd64] https://packages.microsoft.com/repos/microsoft-ubuntu-bionic-prod bionic main" > /etc/apt/sources.list.d/dotnetdev.list'
-        apt-get install apt-transport-https
```
sergey-akhalkov: Could you please confirm we don't really need `apt-transport-https` package?

images/linux/scripts/installers/1604/dotnetcore-sdk.sh
```diff
@@ -34,11 +34,6 @@ for latest_package in ${LATEST_DOTNET_PACKAGES[@]}; do
...
-        apt-get update
         apt-get install $latest_package -y
```
sergey-akhalkov: Could you please also confirm that all packages in the LATEST_DOTNET_PACKAGES are publicly available (no need to and any new packages sources / etc.) and can be installed without any issues?

al-cheb:
> @andy-mishechkin do we need the same for ubuntu18 dotnet installation script just to be consistent?

You should change docker-moby.sh script for Ubuntu 18.04 to preserve the consistence.

images/linux/scripts/installers/1604/dotnetcore-sdk.sh
andy-mishechkin: This package also will be installed in kubernetes-tools.sh
I think we don't need install it twice in Ubuntu16 build

images/linux/scripts/installers/1604/dotnetcore-sdk.sh
andy-mishechkin: Confirmed
The last version of .NET Core is 3.1 (https://dotnet.microsoft.com/download)
It presents in LATEST_DOTNET_PACKAGES

images/linux/scripts/installers/1604/dotnetcore-sdk.sh
sergey-akhalkov: Does it matter when this is installed - earlier or later? Are we using it somehow during the image generation?

PR #360, opened 2020-02-04
Add Chrome driver directory to path

dmitry-shibanov:
- create directory for chromedriver in usr/local/share/chrome_driver
- add symlink to usr/local/share/chrome_driver/chromedriver

PR #380, opened 2020-02-07
Install Firefox(Gecko) webdriver to linux images

dsame:
Add Gecko webdriver to linux images
Move selenium installation to its own script

### DIFF
--- a/images/linux/scripts/installers/firefox.sh
+++ b/images/linux/scripts/installers/firefox.sh
@@ -49,4 +49,5 @@ if ! command -v geckodriver; then
 fi
 
 echo "Lastly, documenting what we added to the metadata file"
-DocumentInstalledItem "Geckodriver ($(geckodriver --version)); Gecko Driver is available via GECKOWEBDRIVER environment variable"
+ver=`./geckodriver --version|head -1|awk '{print $2}'`
+DocumentInstalledItem "Geckodriver (${ver}); Gecko Driver is available via GECKOWEBDRIVER environment variable"
--- a/images/linux/scripts/installers/firefox.sh
+++ b/images/linux/scripts/installers/firefox.sh
@@ -23,3 +23,26 @@ echo "Lastly, documenting what we added to the metadata file"
 #           ($HOME is /home/packer which is owned by packer.)
 HOME=/root
 DocumentInstalledItem "Firefox ($(firefox --version))"
+
+# Download and unpack latest release of geckodriver
+URL=https:`curl -s https://api.github.com/repos/mozilla/geckodriver/releases/latest|grep 'browser_download_url.*linux64.tar.gz'|sed -e 's/^.*"https://'|sed -e 's/"//g'`
+echo "Downloading geckodriver $URL..."
+wget "$URL" -O geckodriver.tar.gz
+tar -xzf geckodriver.tar.gz
+rm geckodriver.tar.gz
+
+GECKODRIVER_BIN="/usr/bin/geckodriver"
+mv "geckodriver" $GECKODRIVER_BIN
+chown root:root $GECKODRIVER_BIN
+chmod +x $GECKODRIVER_BIN
+echo "GECKOWEBDRIVER=$GECKODRIVER_BIN" | tee -a /etc/environment
+
+# Run tests to determine that the geckodriver installed as expected
+echo "Testing to make sure that script performed as expected, and basic scenarios work"
+if ! command -v geckodriver; then
+    echo "geckodriver was not installed"
+    exit 1
+fi
+
+echo "Lastly, documenting what we added to the metadata file"
+DocumentInstalledItem "Geckodriver ($(geckodriver --version)); Gecko Driver is available via GECKOWEBDRIVER environment variable"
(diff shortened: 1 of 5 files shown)


images/linux/scripts/installers/firefox.sh
```diff
@@ -23,3 +23,26 @@ echo "Lastly, documenting what we added to the metadata file"
...
+echo "Downloading geckodriver $URL..."
+wget "$URL" -O geckodriver.tar.gz
+tar -xzf geckodriver.tar.gz
+rm geckodriver.tar.gz
+
+GECKODRIVER_BIN="/usr/bin/geckodriver"
+mv "geckodriver" $GECKODRIVER_BIN
+chown root:root $GECKODRIVER_BIN
+chmod +x $GECKODRIVER_BIN
+echo "GECKOWEBDRIVER=$GECKODRIVER_BIN" | tee -a /etc/environment
```
maxim-lobanov: Could you please take a look at this PR? Could you please install geckodriver using consistent way.
https://github.com/actions/virtual-environments/pull/360/files

I mean that we should create the separate folder for driver, something like `/usr/local/share/firefox_driver` or `/usr/local/share/gecko_driver`, put `geckodriver` there. and set `GeckoWebDriver` variable to point to the folder instead fo binary file. Also we need to make symlink to "/usr/bin/geckodriver".

We do that to be consistent with guide: https://docs.microsoft.com/en-us/azure/devops/pipelines/test/continuous-test-selenium?view=azure-devops#decide-how-you-will-deploy-and-test-your-app

images/linux/scripts/installers/firefox.sh
```diff
@@ -23,3 +23,26 @@ echo "Lastly, documenting what we added to the metadata file"
 #           ($HOME is /home/packer which is owned by packer.)
 HOME=/root
 DocumentInstalledItem "Firefox ($(firefox --version))"
+
+# Download and unpack latest release of geckodriver
+URL=https:`curl -s https://api.github.com/repos/mozilla/geckodriver/releases/latest|grep 'browser_download_url.*linux64.tar.gz'|sed -e 's/^.*"https://'|sed -e 's/"//g'`
```
maxim-lobanov: I would prefer to use `jq` to parse JSON instead of `grep` because it will be more readable. Just nitpick, feel free to ignore if you prefer your way.
Probably, something like below will work for you:
`curl -s https://api.github.com/repos/mozilla/geckodriver/releases/latest | jq '.assets[] | select(.name | endswith("linux64.tar.gz")) | .browser_download_url '`

images/linux/scripts/installers/firefox.sh
```diff
@@ -23,3 +23,26 @@ echo "Lastly, documenting what we added to the metadata file"
...
+echo "Downloading geckodriver $URL..."
+wget "$URL" -O geckodriver.tar.gz
+tar -xzf geckodriver.tar.gz
+rm geckodriver.tar.gz
+
+GECKODRIVER_BIN="/usr/bin/geckodriver"
+mv "geckodriver" $GECKODRIVER_BIN
+chown root:root $GECKODRIVER_BIN
```
maxim-lobanov: I think this line is not needed because we run this script with sudo

images/linux/scripts/installers/firefox.sh
```diff
@@ -23,3 +23,26 @@ echo "Lastly, documenting what we added to the metadata file"
...
+
+# Run tests to determine that the geckodriver installed as expected
+echo "Testing to make sure that script performed as expected, and basic scenarios work"
+if ! command -v geckodriver; then
+    echo "geckodriver was not installed"
+    exit 1
+fi
+
+echo "Lastly, documenting what we added to the metadata file"
+DocumentInstalledItem "Geckodriver ($(geckodriver --version)); Gecko Driver is available via GECKOWEBDRIVER environment variable"
```
maxim-lobanov: `geckodriver --version` produce the following output: `geckodriver 0.26.0 (e9783a644016 2019-10-10 13:38 +0000)`
Can we keep only `geckodriver 0.26.0` for readability?

images/linux/scripts/installers/firefox.sh
dsame: I am not sure `jq` is always available this why i prefer  to use grep on linux platform

images/linux/scripts/installers/firefox.sh
```diff
@@ -23,3 +23,31 @@ echo "Lastly, documenting what we added to the metadata file"
 #           ($HOME is /home/packer which is owned by packer.)
 HOME=/root
 DocumentInstalledItem "Firefox ($(firefox --version))"
+
+# Download and unpack latest release of geckodriver
+URL=https:`curl -s https://api.github.com/repos/mozilla/geckodriver/releases/latest|grep 'browser_download_url.*linux64.tar.gz'|sed -e 's/^.*"https://'|sed -e 's/"//g'`
```
al-cheb: nitpick: `URL=$(curl -s https://api.github.com/repos/mozilla/geckodriver/releases/latest|grep 'browser_download_url.*linux64.tar.gz'|sed -e 's/^.* "//'|sed -e 's/"//g')`

PR #386, opened 2020-02-09
Add Julia 1.3.1

### LINKED ISSUE #271: Add Julia
**Tool information**
- Tool name: Julia
- Add or update? Add
- Desired version: Latest stable, currently that's v1.3.1
- Approximate size: ~350 MB installed on Linux (70-90 MB download size)
- If this is an add request:
  - Brief description of tool: The Julia programming language. There are many action-like tools written in Julia that run frequently and would benefit a lot from a default installation of Julia, such as [CompatHelper.jl](https://github.com/bcbi/CompatHelper.jl), [Registry Auto Merging](https://github.com/JuliaRegistries/RegistryCI.jl) or [VersionVigilante.jl](https://github.com/bcbi/VersionVigilante.jl)
  - URL for tool's homepage: https://julialang.org/

**Area for Triage**:

**Question, Bug, or Feature?**:
Feature

**Virtual environments affected**
- [X] macOS 10.15
- [ ] Ubuntu 16.04 LTS
- [X] Ubuntu 18.04 LTS
- [ ] Windows Server 2016 R2
- [X] Windows Server 2019

**Can this tool be installed during the build?**
Yes, it can be installed using the [julia-actions/setup-julia](https://github.com/julia-actions/setup-julia) action. It takes approximately 10s per build. Many of the tools mentioned above run many times an hour. Overall this causes multiple TB traffic per day to download and install the tool.

**Are you willing to submit a PR?**
Yes

SaschaMann:
**Why is this tool needed?**

(see #271)

There are many action-like tools written in Julia that run frequently and would benefit a lot from a default installation of Julia, such as [CompatHelper.jl](https://github.com/bcbi/CompatHelper.jl), [Registry Auto Merging](https://github.com/JuliaRegistries/RegistryCI.jl) or [VersionVigilante.jl](https://github.com/bcbi/VersionVigilante.jl).

While it can be installed using the [julia-actions/setup-julia](https://github.com/julia-actions/setup-julia) action, it takes approximately 10s per build and many of the tools mentioned above run many times an hour. Overall this causes multiple TB traffic per day to download and install the tool. The Julia ecosystem is growing, and there are more tools like this in development, so this will likely become an even bigger issue over time.

---

A few notes on the PR:

- The scripts are based on the Go installer scripts, apart from the installation logic itself.
- In the future, it may be possible to determine the latest stable Julia version programmatically instead of relying on hardcoded versions. This requires https://github.com/JuliaLang/julia/issues/33817 to be fully fixed first, so it's not possible yet.
- I don't have any experience with Azure VMs. I'm not sure how to do `Step 4. Test your changes by creating VHD and deploying a VM` / `1. Generating required Azure resources and running a packer build for a targeted image is automated here`. **Therefore, this is completely untested right now.** Is there a more in-depth guide on how to do this?
- Where are the macOS scripts added?
- Should the GPG signature of the linux archive be verified?

---

(fixes #271)


al-cheb:
Hello, @SaschaMann 

1. It's very difficult to support hard-coded software version.  Please add logic to parse the latest release version and install it for Ubuntu.
get_latest_release - https://gist.github.com/lukechilds/a83e1d7127b78fef38c2914c4ececc3c

```
$ get_latest_release JuliaLang/julia
v1.3.1
```

2. Please use choco to install Julia. It's very simple:
```
PS >  choco search -r Julia
Julia|1.3.1
```

SaschaMann:
Thanks for the help!

> It's very difficult to support hard-coded software version. Please add logic to parse the latest release version and pass it for Ubuntu.

AFAIK The release is tagged before the official binaries are available. This could potentially cause an issue if the environment is built during that brief period. It's highly unlikely but waiting for https://github.com/JuliaLang/julia/issues/33817 (there will be a JSON file with all available Julia binaries to download) would be an alternative that doesn't have this problem.

johnnychen94:
> AFAIK The release is tagged before the official binaries are available.

I made a [jill.py](https://github.com/johnnychen94/jill.py) that can install the latest stable Julia version:

```bash
jill install --confirm --update --upstream Official
```

Currently, it's only supported for Linux and macOS, but it is possible to support Windows (I haven't done it because I don't have access to a Windows machine)

images/win/scripts/Installers/Validate-Julia.ps1
```diff
@@ -0,0 +1,25 @@
...
+}
+else
+{
+    Write-Host "Julia is not on the path."
+    exit 1
+}
+
+# Add description of the software to Markdown
+$SoftwareName = "Julia (x64)"
+$juliaVersion = $(julia --version) -match "\d+\.\d+\.\d+"
```
al-cheb:
```
PS > $(julia --version) -match "\d+\.\d+\.\d+"
True
```
I think it should be:
`$(julia --version).split() -match "\d+.\d+\.\d+"`

SaschaMann: Fixed in https://github.com/actions/virtual-environments/pull/386/commits/68fb8ac37ca6a4778ce938bf5d87fd1fbe7b084e

images/win/scripts/Installers/Install-Julia.ps1
```diff
@@ -0,0 +1,6 @@
+################################################################################
+##  File:  Install-Julia.ps1
+##  Desc:  Install Julia
+################################################################################
+
+choco install julia -y
```
al-cheb: Julia by default uses `%USERPROFILE%\AppData\Local` for installation. %USERPROFILE% will be deleted after image generation will be finished.
We should choose another folder,  as you mentioned previously: `C:\Julia`.
`choco install julia --ia "/D=C:\Julia"`

SaschaMann: Fixed in: https://github.com/actions/virtual-environments/pull/386/commits/036ff7c43c3c2e54050d040d3577b479a899b5bc

images/linux/scripts/installers/julia.sh
```diff
@@ -0,0 +1,62 @@
...
+##  Desc:  Installs Julia, and adds Julia to the path
+################################################################################
+
+# Source the helpers for use with the script
+source $HELPER_SCRIPTS/document.sh
+
+# This function fetches the latest Julia release from the GitHub API
+# Based on https://gist.github.com/lukechilds/a83e1d7127b78fef38c2914c4ececc3c
+function GetLatestJuliaRelease () {
+    curl --silent "https://api.github.com/repos/julialang/julia/releases/latest" |
```
fredrikekre: I don't think this gets you the release with highest version number (I assume that's what you are after), IIRC it just returns the latest (as in time) release.

DilumAluthge: @fredrikekre is correct. So if we release 1.6.0, and the next day release 1.5.1, the GitHub API (if I understand correctly) will return 1.5.1 as the "latest release."

I think the correct thing to do is for the JuliaLang community to host a URL that simply redirects to the GitHub release of the **stable** Julia release with the highest version number. So e.g. right now it would redirect to: https://github.com/JuliaLang/julia/releases/tag/v1.3.1

Maybe @staticfloat or @maleadt could help with setting up this URL? I think it should be relatively straightforward:
1. Use the GitHub API to get the list of all releases.
2. Filter out the pre-releases, release candidates, etc.
3. Sort the remaining releases by SemVer version number in decreasing order.
4. Pick the release with highest version number.

We would run this computation once per day as a cron job, and save the result somewhere, and then that result is hosted at a JuliaLang.org URL.

SaschaMann: @DilumAluthge see https://github.com/JuliaLang/julia/issues/33817

DilumAluthge: It looks like that (if it is ever implemented) would provide all of the versions? So we’d still have to go through and figure out what the latest version is?

SaschaMann: Yes.

Alternatively, the steps you outlined could be added to this build script.

@al-cheb Do you know if `jq` is available in the build environment?

Issue #275, opened 2020-01-15
Add Tkinter


RealOrangeOne:
**Tool information**
- Tool name: Tkinter
- Add or update? Add, ish
- Desired version: latest
- Approximate size: 
- If this is an add request:
  - Brief description of tool: Enables the use (or at least import of) the `tkinter` Python module in the standard library
  - URL for tool's homepage: https://wiki.python.org/moin/TkInter

**Area for Triage**:
Python
Packages

**Question, Bug, or Feature?**:
Feature, I guess?

**Virtual environments affected**
- [x] macOS 10.15
- [x] Ubuntu 16.04 LTS
- [x] Ubuntu 18.04 LTS
- [x] Windows Server 2016 R2
- [x] Windows Server 2019

**Can this tool be installed during the build?**
Not to my knowledge. Attempting to install `python3-tk` and `tk-dev` before `@actions/setup-python`, but it doesn't seem to work. Likely requires a rebuild of Python.

**Are you willing to submit a PR?**
If I knew what to do, sure.

thejoebourneidentity:
Thanks for the suggestion @RealOrangeOne , we're accepting this issue and will begin working on it shortly!

andy-mishechkin:
Tkniter already is present in Windows and MacOS images, but because it's a GUI interface, you can't check it from serial console, where you'll lost control, when you run tkniter application.
Additional libraries have been added to Ubuntu images for Tkniter support

andy-mishechkin:
We have tested all Ubuntu version and have got the correct results from test python script - closing the ticket.
Please feel free to reopen this ticket, if you'll find some issues.

miketrethewey:
To be fair, the Windows part of the issue wasn’t really addressed in a descriptive way about how to get around it as opposed to skipping tests or whatever regarding TKinter and Windows.

dmitry-shibanov:
Hello @miketrethewey, thank you for your response. We use PyInstaller to install Python versions on  Windows image. It includes Tkinter by default so everything should work.
Example for Python 3.7:
```
import sys
print(sys.version)
import tkinter as tk
print(tk.TkVersion)
```
Output:
![image](https://user-images.githubusercontent.com/28229110/75750266-e632e880-5d34-11ea-9e16-6c64246ab565.png)
If it doesn’t work for you, could you please share minimal repro steps?

miketrethewey:
[...]

PR #404, opened 2020-02-13
tk installation has been added to Ubuntu 16/18 basic.sh

andy-mishechkin:
According to https://github.com/actions/virtual-environments/issues/275, tk library installation has been added to Ubuntu 16/18.
This library is used for correct tkniter work in Ubuntu.

### DIFF
--- a/images/linux/scripts/installers/1804/basic.sh
+++ b/images/linux/scripts/installers/1804/basic.sh
@@ -97,6 +97,9 @@ apt-get install -y --no-install-recommends xvfb
 echo "Install libgtk"
 apt-get install -y --no-install-recommends libgtk-3-0
 
+echo "Install tk"
+apt install -y tk
+
 echo "Install fakeroot"
 apt-get install -y --no-install-recommends fakeroot
 
--- a/images/linux/scripts/installers/1604/basic.sh
+++ b/images/linux/scripts/installers/1604/basic.sh
@@ -47,6 +47,7 @@ apt-fast install -y --no-install-recommends \
     dbus \
     xvfb \
     libgtk-3-0 \
+    tk \
     fakeroot \
     dpkg \
     rpm \


images/linux/scripts/installers/1804/basic.sh
```diff
@@ -97,6 +97,9 @@ apt-get install -y --no-install-recommends xvfb
 echo "Install libgtk"
 apt-get install -y --no-install-recommends libgtk-3-0
 
+echo "Install tk"
+apt install -y tk
```
AlenaSviridenko: do we also need `--no-install-recommends` here as we do for other tools? also, I see that `--no-install-recommends` is used for 1604 as well

andy-mishechkin: I didn't found the option  '--no-install-recommends' in apt man page (as far a I know apt it's a different tool then apt-get).
Also we used the same command in building Ubuntu HostedToolCache.Python, and this build was was without any errors

AlenaSviridenko: oh, it's `apt`, not `apt-get`, didn't notice that. Thanks for clarification.

alepauly: you mention that this is used when building python in hosted toolcache? why is it not available here already?

Also, can we add some validation to make sure this is in fact functional?

maxim-lobanov: @alepauly , It is not enough to add this package to HostedToolCache. This package is needed in runtime to use Tkinter packages so we have to add it to the image.
Also we will build canary test for validation

andy-mishechkin: @sergey-akhalkov . we have a successful builds and testing of all Python versions on Ubuntu 16/18.
Please merge this PR

andy-mishechkin:
@sergey-akhalkov . we have a successful builds and testing of all Python versions on Ubuntu 16/18.
Please merge this PR

PR #410, opened 2020-02-14
action: QA for executable statements

Kreyren:
These statements are printing in terminal which i believe is unexpected.

```console
gitpod /workspace/virtual-environments $ if ! command -v cat; then echo test; fi ; echo $?
/usr/bin/cat
0
gitpod /workspace/virtual-environments $ if command -v cat; then echo test; fi ; echo $?
/usr/bin/cat
test
0
```


miketimofeev:
These outputs caused by GitPod, we don't experience any issues when building an image.

Issue #282, opened 2020-01-16
Jobs stuck when trying to untar an archive


eine:
Since today, trying to untar a `tar.xz` archive on `windows-latest` seems not to be possible:

- [MWE1](https://github.com/eine/setup-msys2/blob/07a6f112cd2ac8bef7ab984f2a2a5f9213196b11/main.js): `tc.extractTar` [stuck](https://github.com/eine/setup-msys2/commit/07a6f112cd2ac8bef7ab984f2a2a5f9213196b11/checks?check_suite_id=404893652)
- [MWE2](https://github.com/eine/setup-msys2/blob/c0a28de30a068489254ed39d4c29a57ffdc9d769/main.js): `tar -x -J --force-local -f` [failure](https://github.com/eine/setup-msys2/commit/c0a28de30a068489254ed39d4c29a57ffdc9d769/checks?check_suite_id=404905818)
- [MWE3](https://github.com/eine/setup-msys2/blob/59a2bc91b344be13c09ced79e4f70de97bf1a833/main.js): `tar -xf distrib` [stuck](https://github.com/eine/setup-msys2/commit/59a2bc91b344be13c09ced79e4f70de97bf1a833/checks?check_suite_id=404908739)
- [MWE4](https://github.com/eine/setup-msys2/blob/a4db4e2813844e3ab9027ceb45c8604e5bd40a6d/main.js): `tar -xJf distrib` [stuck](https://github.com/eine/setup-msys2/commit/a4db4e2813844e3ab9027ceb45c8604e5bd40a6d/checks?check_suite_id=404909806)

/cc @ericsciple

thboop:
Related to https://github.com/actions/virtual-environments/issues/263

thejoebourneidentity:
@alepauly can you verify that this is fixed? ty!

eine:
@thechrisjohnson, the MWEs above are still failing/stuck: https://github.com/eine/setup-msys2/runs/417809983

miketimofeev:
@thejoebourneidentity @eine the issue is already fixed, but the VM image is not deployed yet. It should be available in a week or two.

miketimofeev:
@eine the image has been deployed. Could you check the behavior, please?

eine:
@miketimofeev I restarted an old job, and I also pushed some new. In all cases, runs are still stuck:

- https://github.com/eine/setup-msys2/actions/runs/38608071
- https://github.com/eine/setup-msys2/actions/runs/38609226

[...]

PR #415, opened 2020-02-17
Restore Git bash utilities to the PATH

miketimofeev:
This is a quick fix for issue https://github.com/actions/virtual-environments/issues/282
In [#211](https://github.com/actions/virtual-environments/pull/211) C:\Program Files\Git\usr\bin was removed from the PATH caused to using [Windows tar](https://devblogs.microsoft.com/commandline/tar-and-curl-come-to-windows/)(C:\Windows\System32\tar.exe) instead of the tar comes along with Git. Windows doesn't have xz support by default, tries to use mingw xz utility and hangs.

Changes:
- Restore `"C:\Program Files\Git\usr\bin"` in the PATH, until MSYS2 is added(https://github.com/actions/virtual-environments/issues/30)

### DIFF
--- a/images/win/scripts/Installers/Install-Git.ps1
+++ b/images/win/scripts/Installers/Install-Git.ps1
@@ -12,6 +12,7 @@ choco install git -y --package-parameters="/GitAndUnixToolsOnPath /WindowsTermin
 # Disable GCM machine-wide
 [Environment]::SetEnvironmentVariable("GCM_INTERACTIVE", "Never", [System.EnvironmentVariableTarget]::Machine)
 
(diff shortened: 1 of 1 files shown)


maxim-lobanov:
Is any regression expected after these changes?
In scope of #211 , this line was removed so, probably, we can bring some issues back again

miketimofeev:
> Is any regression expected after these changes?
> In scope of #211 , this line was removed so, probably, we can bring some issues back again

@maxim-lobanov complains were mostly about mingw64 directory in the PATH. And after #211  changes some of the tools are broken as explained in this comment
https://github.com/actions/virtual-environments/pull/211#pullrequestreview-346292514

Issue #363, opened 2020-02-04
Update/Add [net472]


zfengms:
**Tool information**
- Tool name: .net framework 4.7.2
- Add or update? Add
- Desired version: 4.7.2
- Approximate size: 
- If this is an add request:
  - Brief description of tool: Our application needs .net framework 4.7.2 but could not find it on hosted agent
  - URL for tool's homepage: https://dotnet.microsoft.com/download/dotnet-framework/thank-you/net472-developer-pack-offline-installer

**Area for Triage**:
.NET Framework

**Question, Bug, or Feature?**:
Feature

**Virtual environments affected**
- [ ] macOS 10.15
- [ ] Ubuntu 16.04 LTS
- [ ] Ubuntu 18.04 LTS
- [x] Windows Server 2016 R2
- [x] Windows Server 2019

**Can this tool be installed during the build?**
Do not know

**Are you willing to submit a PR?**
No

vmapetr:
@zfengms .net framework 4.7.2 developer tools were added to Windows images. They will be available after next image rollout (1-2 weeks approximately)

miketimofeev:
@zfengms the image has been rolled out. Could you check everything is fine now?

maxim-lobanov:
I am closing the issue since image was deployed. Feel free to reopen if you see any issues

PR #419, opened 2020-02-18
Add dotnet 4.7.2 developer tools

vmapetr:
This PR fixes issue https://github.com/actions/virtual-environments/issues/363 and it adds .NET framework 4.7.2 SDK, Targeting Pack and Developer tools as visual studio components to VS17 and VS19

### DIFF
--- a/images/win/scripts/Installers/Windows2019/Install-VS2019.ps1
+++ b/images/win/scripts/Installers/Windows2019/Install-VS2019.ps1
@@ -73,6 +73,9 @@ $WorkLoads = '--allWorkloads --includeRecommended ' + `
               '--add Microsoft.Component.VC.Runtime.UCRTSDK ' + `
               '--add Microsoft.Net.ComponentGroup.4.6.2.DeveloperTools ' + `
               '--add Microsoft.Net.ComponentGroup.4.7.1.DeveloperTools ' + `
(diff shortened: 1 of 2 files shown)


images/win/scripts/Installers/Windows2019/Install-VS2019.ps1
```diff
@@ -73,6 +73,9 @@ $WorkLoads = '--allWorkloads --includeRecommended ' + `
               '--add Microsoft.Component.VC.Runtime.UCRTSDK ' + `
               '--add Microsoft.Net.ComponentGroup.4.6.2.DeveloperTools ' + `
               '--add Microsoft.Net.ComponentGroup.4.7.1.DeveloperTools ' + `
+              '--add Microsoft.Net.Component.4.7.2.SDK ' + `
```
maxim-lobanov: Do we really need `.SDK` and `. TargetingPack` here?
I wonder because for all other frameworks only `DeveloperTools` is installed

vmapetr: SDK is required to develop applications for the .NET Framework and
Targeting pack is required for targeting a particular .NET framework which is different from currently installed .NET framework.
Considering that we add additional .NET framework, they are vital.

Issue #428, opened 2020-02-19
WDK should be installed after VS 2019 and not before


jimlamb:
**Describe the bug**
The WDK needs to be installed after VS 2019, not before.

**Area for Triage**: 

**Question, Bug, or Feature?**: 
Bug

**Virtual environments affected**
- [ ] macOS 10.15
- [ ] Ubuntu 16.04 LTS
- [ ] Ubuntu 18.04 LTS
- [ ] Windows Server 2016 R2
- [x] Windows Server 2019

**Expected behavior**
If you take the output of the VS 2019 KMDF project template and try to build it, it should build successfully.

**Actual behavior**
If the WDK is installed before VS 2019, you'll get the following error when you try to build that basic KMDF project:

    Error MSB8020: The build tools for WindowsKernelModeDriver10.0 (Platform Toolset = 'WindowsKernelModeDriver10.0') cannot be found.

vmapetr:
Hi @jimlamb! This issue has been discussed in https://github.com/actions/virtual-environments/issues/259 previously. We working on the fix now.

jimlamb:
Thanks, @vmapetr. I'm the one who helped nibanks diagnose the issue. Let me know if I can help verify the fix.

Issue #259, opened 2020-01-13
Update [Windows Driver Kit] to latest version (10.0.18362.1) 


heavenwing:
**Tool information**
- Tool name: Windows Driver Kit in Windows Server 2019
- Add or update? : Update
- Desired version: 10.0.18362.1
- Approximate size: not sure
- If this is an add request:
  - Brief description of tool: to build windows driver project
  - URL for tool's homepage: https://docs.microsoft.com/en-us/windows-hardware/drivers/download-the-wdk

**Area for Triage**:

**Question, Bug, or Feature?**: Feature

##[error]C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\MSBuild\Microsoft\VC\v160\Microsoft.CppBuild.targets(382,5): Error MSB8020: The build tools for WindowsKernelModeDriver10.0 (Platform Toolset = 'WindowsKernelModeDriver10.0') cannot be found. To build using the WindowsKernelModeDriver10.0 build tools, please install WindowsKernelModeDriver10.0 build tools.  Alternatively, you may upgrade to the current Visual Studio tools by selecting the Project menu or right-click the solution, and then selecting "Retarget solution".

**Virtual environments affected**
- [ ] macOS 10.15
- [ ] Ubuntu 16.04 LTS
- [ ] Ubuntu 18.04 LTS
- [ ] Windows Server 2016 R2
- [ * ] Windows Server 2019

**Can this tool be installed during the build?**
No

**Are you willing to submit a PR?**
No

nibanks:
From what I can tell from [Install-WDK.ps1](https://github.com/actions/virtual-environments/blob/master/images/win/scripts/Installers/Windows2019/Install-WDK.ps1) it should be installing `Version: 10.0.18362.0`:
```PowerShell
# Version: 10.0.18362.0
# Update Validate-WDK.ps1 if the version changes!
# There doesn't seem to be any way to check the version programmatically
```
I'm trying to onboard building a universal Windows driver and I've explicitly targeted this version:
```xml
  <PropertyGroup Label="Globals">
    <ProjectGuid>{C31B028C-E91C-4CF7-A8E7-F385B2AF5F85}</ProjectGuid>
    <TemplateGuid>{1bc93793-694f-48fe-9372-81e2b05556fd}</TemplateGuid>
    <TargetFrameworkVersion>v4.5</TargetFrameworkVersion>
    <MinimumVisualStudioVersion>12.0</MinimumVisualStudioVersion>
[...]

PR #431, opened 2020-02-20
WDK installation fix for win19

miketimofeev:
We have several issues related to WDK:
https://github.com/actions/virtual-environments/issues/259
https://github.com/actions/virtual-environments/issues/428
The root cause is an incorrect installation order - WDK is supposed to be installed after VS19, not before so that the extension can be added to VS.

Changes:
- Installation order changed
- Try-catch on package installation added in order to break the build if the package is not installed
- Validate-WDK script works properly now, previously it's just put hardcoded value to the softwarelist(the same can be done for windows16 if needed)
- Get-VS19ExtensionVersion function created to use it in both Validate-WDK and Validate-Wix scripts

### DIFF
--- a/images/win/scripts/ImageHelpers/InstallHelpers.ps1
+++ b/images/win/scripts/ImageHelpers/InstallHelpers.ps1
@@ -83,8 +83,8 @@ function Install-EXE
 }
 
 function Stop-SvcWithErrHandling
-<# 
-.DESCRIPTION 
+<#
+.DESCRIPTION
 Function for stopping the Windows Service with error handling
 
 .AUTHOR
@@ -98,7 +98,7 @@ Switch for stopping the script and exit from PowerShell if one service is absent
 #>
 {
     param (
-        [Parameter(Mandatory, ValueFromPipeLine = $true)] [string] $ServiceName, 
+        [Parameter(Mandatory, ValueFromPipeLine = $true)] [string] $ServiceName,
         [Parameter()] [switch] $StopOnError
     )
 
@@ -126,8 +126,8 @@ Switch for stopping the script and exit from PowerShell if one service is absent
 }
 
 function Set-SvcWithErrHandling
-<# 
-.DESCRIPTION 
+<#
+.DESCRIPTION
 Function for setting the Windows Service parameter with error handling
 
 .AUTHOR
@@ -140,7 +140,7 @@ The name of stopping service
 Hashtable for service arguments
 #>
 {
-    
+
     param (
         [Parameter(Mandatory, ValueFromPipeLine = $true)] [string] $ServiceName,
         [Parameter(Mandatory)] [hashtable] $Arguments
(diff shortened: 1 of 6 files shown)


images/win/scripts/ImageHelpers/InstallHelpers.ps1
```diff
@@ -160,3 +160,31 @@ Hashtable for service arguments
...
+function Get-VS19ExtensionVersion
+{
+    param (
+        [string] [Parameter(Mandatory=$true)] $packageName
+    )
+
+    $vsProgramData = Get-Item -Path "C:\ProgramData\Microsoft\VisualStudio\Packages\_Instances"
+    $instanceFolders = Get-ChildItem -Path $vsProgramData.FullName
+
+    if($instanceFolders -is [array])
```
vmapetr: nitpick: missing space `if ($instanceFolders -is [array])`

miketimofeev: thanks! Fixed

images/win/scripts/ImageHelpers/InstallHelpers.ps1
```diff
@@ -160,3 +160,31 @@ Hashtable for service arguments
...
     }
 }
+
+function Get-VS19ExtensionVersion
+{
+    param (
+        [string] [Parameter(Mandatory=$true)] $packageName
+    )
+
+    $vsProgramData = Get-Item -Path "C:\ProgramData\Microsoft\VisualStudio\Packages\_Instances"
```
al-cheb: What is the purpose Get-Item -Path ? We already have fullName.

images/win/scripts/ImageHelpers/InstallHelpers.ps1
```diff
@@ -160,3 +160,31 @@ Hashtable for service arguments
...
+    $instanceFolders = Get-ChildItem -Path $vsProgramData.FullName
```
al-cheb: .FullName is not necessary

images/win/scripts/ImageHelpers/InstallHelpers.ps1
```diff
@@ -160,3 +160,31 @@ Hashtable for service arguments
...
+
+    if ($instanceFolders -is [array])
+    {
+        Write-Host "More than one instance installed"
+        exit 1
+    }
+
+    $stateContent = Get-Content -Path ($instanceFolders.FullName + '\state.packages.json')
```
al-cheb: Use Join-Path

images/win/scripts/ImageHelpers/InstallHelpers.ps1
miketimofeev: Thanks, removed

images/win/scripts/ImageHelpers/InstallHelpers.ps1
miketimofeev: Thanks, removed

images/win/scripts/ImageHelpers/InstallHelpers.ps1
miketimofeev: Switched, thanks

Issue #286, opened 2020-01-17
add lerna


awentzel:
**Tool information**
- Tool name: Lerna
- Add or update? Add
- Desired version: Latest
- Approximate size: 
- If this is an add request:
  - Brief description of tool: Lerna is a common package manager for monorepositories
  - URL for tool's homepage: https://lerna.js.org/

**Area for Triage**:

**Question, Bug, or Feature?**:

**Virtual environments affected**
- [ ] macOS 10.15
- [ ] Ubuntu 16.04 LTS
- [ ] Ubuntu 18.04 LTS
- [] Windows Server 2016 R2
- [X] Windows Server 2019

**Can this tool be installed during the build?**

**Are you willing to submit a PR?**

alepauly:
@awentzel - can this be installed during a build?

awentzel:
@alepauly we have been installing during the build for a few years.  However, it takes twice as long to install on Windows as Linux or Mac.  Lerna is very common for enterprise projects.  Oddly, building on Windows, in general, takes 36% longer with twice the memory and same cores as Circle CI.  Trying to find a way to squeeze faster builds out of Windows environments or we'll have to keep using Circle CI.  We can Linux, but, it would be nice to run on Windows to ensure reliability for Windows users.

miketimofeev:
PR has been merged https://github.com/actions/virtual-environments/pull/374

miketimofeev:
@awentzel Win16 and 19 images have been deployed. Could you check that lerna is presented, please?

miketimofeev:
@awentzel I'm going to close the issue. Feel free to reopen if you have any concerns, thanks!

PR #435, opened 2020-02-20
Remove unnecessary Node.js related dependencies

### LINKED ISSUE #427: Cache is large on macos-latest
👋 When setting up caching via `actions/cache`, it's common and recommended to cache `~/.npm` after running npm install/ci.

On macOS, this directory is pre-populated with npm packages that are potentially not related to the user's build and is over 1 GB in size. This limits the ability to cache build dependencies due to size constraints and slows down caching due to the larger cache size.

See https://github.com/actions/cache/issues/187

This applies to other ecosystems like Gradle as well

imbsky:
Closes #427

### DIFF
--- a/images/linux/scripts/installers/nodejs.sh
+++ b/images/linux/scripts/installers/nodejs.sh
@@ -7,13 +7,9 @@
 # Source the helpers for use with the script
 source $HELPER_SCRIPTS/document.sh
 
-# Install LTS Node.js and related build tools
-curl -sL https://raw.githubusercontent.com/mklement0/n-install/stable/bin/n-install | bash -s -- -ny -
-~/n/bin/n lts
-npm install -g grunt gulp n parcel-bundler typescript
-npm install -g --save-dev webpack webpack-cli
-npm install -g npm
-rm -rf ~/n
+# Install LTS Node.js
+curl -sL https://deb.nodesource.com/setup_12.x | bash -
+apt-get install -y nodejs
 
 # Install Yarn repository and key
 curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add -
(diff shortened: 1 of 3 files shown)


imbsky:
CC: @maxim-lobanov @joshmgross

maxim-lobanov:
@imbsky , I think the initial issue was about macOS (I have put some info there).
As for Linux / Windows, yep, probably some preinstalled packages are unnecessary but, unfortunately, we can't predict if it breaks someone.
Is it real issue on Windows / Linux?

imbsky:
No, this should happen on all operating systems. (Because you used npm to install some packages. and this is why I didn't install yarn using npm in this PR.) Also, for general use, a global installation of these tools is more likely to destroy the user's environment.

imbsky:
By the way, I can't update macOS code because it's not open source, so could you do that? Note that you need to install yarn via brew. (It should not be installed via npm for the above reasons.)

imbsky:
As a less clean fix, removing the contents of the npm cache directory after a global install is not a bad option. But, it's better to remove the unnecessary tools from code and install yarn via brew.

imbsky:
This should not happen if you are using yarn. At least in my environment. So I strongly suspect that the problem is caused by the installation of unnecessary tools via npm.

imbsky:
I just resolved the conflict. Please let me know if you have any additional questions or concerns.

maxim-lobanov:
I am still concerned about removing those packages because customers' pipelines can depend on them and will start fail after these changes.
For example, Lerna. It was added recently, in scope of this issue #286 so your changes remove this package and will cause regression.
The same for other packages, they were added by customers' requests

imbsky:
Well, I don't think it will take that long to install lerna. Also, installing the npm libraries globally is disruptive and shouldn't have been done casually in CI's VM image in the first place. 99% of people install dependencies from a package.json. The problems that this causes are primarily due to actions that deviate from the recommended workflow.

alepauly:
Thanks for the PR @imbsky but we can't stop installing those packages without some preparation. Your comments are definitely heard and we can continue to discuss the approach of [not] installing packages by default but I'll close this PR since this would definitely break existing users.

imbsky:
I see. For example, what about running the cache clean command after installation? If you okay with that, I will open a PR to do it.

imbsky:
I think this is the simplest solution that doesn't break the existing environment.

maxim-lobanov:
@imbsky , 
I think `running the cache clean command after installation` should work for us.

imbsky:
Okay

imbsky:
There is a lot of code to install npm packages, so I think it's inefficient to purge cache everywhere every time. I would like to run it once when the entire installation process is complete. Is there a way to do that?

maxim-lobanov:
@imbsky , We can add separate script and put it to the packer template as a last step

imbsky:
Oh, I see.

Issue #104, opened 2019-11-20
Add C++ support (GCC and Clang) on all platforms


kcgen:
**Tool information**
- **Tool names**: defacto C++ compilers: GCC and Clang
- **Add or update?** Add
- **Desired versions**: Latest version available in standard package repositories. Currently these are:

   | Compiler | Version | OS | Package Manager(s) |
   |---|---|---|---|
   | GCC  | 9  | all  | apt, Brew, MacPorts, and MSYS2 |
   | Clang | 8 | Ubuntu-19.04 | apt |
   | Clang | 9 | macOS and Windows | Brew and Chocolatey |

- **Approximate size**: Clang depends on GCC (that's why it often makes sense to simply install both; in addition to their technical differences). Typically the combo is a couple GB on Windows and smaller on macOS and Linux.

- **Brief description of packages**: These  C++ compilers are used to build literally every C++ autotools/cmake project out there (tens of thousands of open-source applications and tools; perhaps more on GitHub).

   The main goal of this request is to allow the vast majority of open source C/C++ packages to be ./configure'd and built with either compiler (assuming the user pulls in any package-specific dependencies) "out of the box".

- **Homepages**: GCC - https://gcc.gnu.org/, Clang - https://clang.llvm.org/

**Virtual environments affected**
- :ballot_box_with_check: macOS 10.15
- :ballot_box_with_check:Ubuntu 16.04 LTS
- :ballot_box_with_check: Ubuntu 18.04 LTS
- :ballot_box_with_check: Windows Server 2016 R2
- :ballot_box_with_check: Windows Server 2019

**Can they be installed during the build?**
Yes. I'm currently installing the following in my workflow:

| Compilers | OS | Package Manager | Time |
|---|---|---|---|
| GCC and Clang |Ubuntu   | apt  | < 60s |
| GCC |  macOS | Brew and MacPorts | between 3 min and 5 min |
| GCC and Clang | Windows  | MSYS2's pacman | between 8 min and 12 min  |

Compiling my application takes less time than installing these.

Because of this excessive time and resource usage, I made use of GitHub's new [cache](https://github.com/actions/cache) feature available in CI workflows, however, the cache limit is (currently) too small to save most of these development stacks.

Because it seems absurd that GitHub's cache size cannot accommodate C++ "hello world", I flagged 
[...]

PR #447, opened 2020-02-22
Install clang-format

NilsIrl:
https://github.com/actions/virtual-environments/issues/104#issuecomment-589962007

### DIFF
--- a/images/linux/scripts/installers/clang.sh
+++ b/images/linux/scripts/installers/clang.sh
@@ -18,6 +18,7 @@ function InstallClang {
         apt-get install -y "clang-$version" "lldb-$version" "lld-$version"
     else
         ./llvm.sh $version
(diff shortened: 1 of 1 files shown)


EFanZh:
How about making `clang-format-9` as the default `clang-format` using `update-alternatives`? Since `clang-9` and `clang++-9` has already been set as default.

NilsIrl:
> How about making `clang-format-9` as the default `clang-format` using `update-alternatives`

Great idea! Pushed a commit for this.

miketimofeev:
@NilsIrl I'm afraid you've missed the name for the clang-format
`update-alternatives: --install needs <link> <name> <path> <priority>`

NilsIrl:
> I'm afraid you've missed the name for the clang-format

Thank you, fixed it.

PR #462, opened 2020-02-25
Add Images.CI for Windows and Ubuntu images (Azure DevOps)

maxim-lobanov:
This PR adds `images.CI` folder in the root of repository. This folder contains set of scripts and yml files to run image generation builds in Azure DevOps CI.

Builds will be run on `https://dev.azure.com/github/virtual-environments`

### DIFF
--- a/images.CI/azure-pipelines/image-generation.yml
+++ b/images.CI/azure-pipelines/image-generation.yml
@@ -36,6 +36,18 @@ jobs:
                         -GitHubFeedToken $(GITHUB_TOKEN)
 
   - task: PowerShell@2
+    displayName: 'Create release for VM deployment'
+    inputs:
+      targetType: filePath
+      filePath: ./images.CI/create-release.ps1
+      arguments: -BuildId $(Build.BuildNumber) `
+                        -Organization $(RELEASE_TARGET_ORGANIZATION) `
+                        -DefinitionId $(RELEASE_TARGET_DEFINITION_ID) `
+                        -Project $(RELEASE_TARGET_PROJECT) `
+                        -ImageName ${{ parameters.image_type }} `
+                        -AccessToken $(RELEASE_TARGET_TOKEN)
+
+  - task: PowerShell@2
     displayName: 'Clean up resources'
     condition: always()
     inputs:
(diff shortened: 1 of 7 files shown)


images.CI/build-image.ps1
```diff
@@ -0,0 +1,60 @@
...
+    'StorageAccountLocation',
+    'OSDiskUri',
+    'OSDiskUriReadOnlySas',
+    'TemplateUri',
+    'TemplateUriReadOnlySas',
+    ':  ->'
+)
+
+Write-Host "Build $Image VM"
+packer build    -var "capture_name_prefix=$ResourcesNamePrefix" `
```
alepauly: can you please double check that the variables expected as secrets are actually marked as such in the packer json files

maxim-lobanov: Actually, only `client_secret` and `github_token` are marked as `sensitive-variables` in packer templates.
Other variables, like `client_id`, `capture_name_prefix`, `resource_group`, `storage_account` and etc should not be marked as sensitive because it can  complicate life of our customers who will build image locally.

These variables won't be shown in our CI because they are marked as secret on AzDo level:
image

As for the autogenerate values like temporary resource group and machine id (values that we can't know in advance), we use filter to exclude them from logs (see 56-59 lines)

images.CI/azure-pipelines/image-generation.yml
```diff
@@ -0,0 +1,69 @@
...
+  steps:
+  - task: PowerShell@2
+    displayName: 'Download custom repository'
+    condition: and(ne(variables['CUSTOM_REPOSITORY_URL'], ''), ne(variables['CUSTOM_REPOSITORY_BRANCH'], ''))
+    inputs:
+      targetType: 'inline'
+      script: |
+        Remove-Item -path './*' -Recurse -Force
+        Write-Host "Download $(CUSTOM_REPOSITORY_BRANCH) branch from $(CUSTOM_REPOSITORY_URL)"
+        # git on self-hosted agent produces some output to stderr even during successful cloning
```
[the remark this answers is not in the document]
maxim-lobanov: Good catch, thank you!

PR #466, opened 2020-02-26
Add Bazel support to Linux and Windows

nikita-bykov:
PR for Issue: #221 

This adds installation Bazel for Windows using `choco` and for Linux using `apt-get`


images/linux/scripts/installers/bazel.sh
```diff
@@ -0,0 +1,22 @@
...
+
+# Run tests to determine that the software installed as expected
+echo "Testing to make sure that script performed as expected, and basic scenarios work"
+if ! command -v bazel; then
+    echo "Bazel was not installed"
+    exit 1
+fi
+
+# Document what was added to the image
+echo "Lastly, documenting what we added to the metadata file"
```
sergey-akhalkov: nitpick: not sure we need echo for docs

maxim-lobanov: We should, at least to be consistent with other tools.
For example, AzCopy: https://github.com/actions/virtual-environments/blob/master/images/linux/scripts/installers/azcopy.sh

sergey-akhalkov: Yeah it totally makes sense to be consistent, just wondering why we need echo for docs.

Issue #290, opened 2020-01-19
Update Docker to 3.2.0 or higher for BuildKit 


jjangga0214:
**Tool information**
- Tool name: Docker
- Add or update? update
- Desired version: 3.2.0 or higher (See [release-notes](https://docs.docker.com/ee/ucp/release-notes/))
- Approximate size: 

As of 3.2.0, a feature called `BuildKit` can be used. This significantly boosts docker build time, which makes sense to use it in CI. 

**Area for Triage**: Docker

**Question, Bug, or Feature?**: Question

**Virtual environments affected**
- [ ] macOS 10.15
- [x] Ubuntu 16.04 LTS
- [x] Ubuntu 18.04 LTS
- [ ] Windows Server 2016 R2
- [ ] Windows Server 2019

**Can this tool be installed during the build?**

It's update of enterprise version of docker, so I don't know if people want to do it manually. 
So I think the best is for GitHub to update it by itself.  

**Are you willing to submit a PR?**

If applicable.

alepauly:
Our installed version of moby might already support this, we just need to make sure the version maps to Docker 3.2.0 or later.

m-minasyan:
Hi @jjangga0214. I made several fixes and we now correctly display the version as docker-moby, since we have docker-moby installed instead of docker. docker-moby has been updated to the latest version.

jjangga0214:
@m-minasyan Thanks. Good to hear I can use `buildkit` in GitHub Action.
> we now correctly display the version as docker-moby
But what do you mean by this? `images/linux/Ubuntu1804-README.md` still shows the old version `3.0.10`.

I see `DocumentInstalledItem "Docker-Moby ($docker_version)"` in 
https://github.com/m-minasyan/virtual-environments/blob/f0e4d780060478aa6aa178ba7dc7335eb7ec0f1a/images/linux/scripts/installers/docker-moby.sh#L51, and guess the markdown file is auto-generated (on production VM creation phase?).

So does it mean docker has not updated yet, though your PR(#470) is merged?

m-minasyan:
@jjangga0214 
Version `3.0.10` is the latest version of Docker-Moby. 
Please check if `buildkit` is working properly at the moment, because at the moment I think there should be no problems with it.

m-minasyan:
@jjangga0214 
Documentation has been updated and Docker Moby has the latest version installed. If you have questions, you can re-open issue. Thank you.

PR #470, opened 2020-02-26
Change documenting of Docker version

m-minasyan:
1. Change documenting of `Docker` version to `Docker-Moby` version in `docker-moby.sh` for Ubuntu 16 and Ubuntu 18.
2. Delete duplicate of documenting `Docker` version.


miketimofeev:
Do we really need both docker.sh and docker-moby.sh?

images/linux/scripts/installers/docker.sh
```diff
@@ -26,9 +26,4 @@ docker pull debian:9
 docker pull alpine:3.7
 docker pull alpine:3.8
 docker pull alpine:3.9
-docker pull alpine:3.10
-
-## Add version information to the metadata file
-echo "Documenting Docker version"
-DOCKER_VERSION=`docker -v`
```
maxim-lobanov: Just wonder, why we delete these lines.
Previously, `Docker` doc entry was produced in 2 places:
1) docker-moby (that you fixed)
2) docker (that you have removed)
So where the original `Docker` is installed? Or we don't install original docker at all?

m-minasyan: Its shows same thing.
https://github.com/actions/virtual-environments/releases/tag/ubuntu18%2F20200217.1

m-minasyan:
> Do we really need both docker.sh and docker-moby.sh?

I think we need only one of it, because its do same things.

al-cheb:
@m-minasyan Please merge latest changes to your branch from the master branch.

images/linux/scripts/installers/docker.sh
```diff
@@ -1,32 +0,0 @@
...
-##  File:  docker.sh
-##  Desc:  Installs the correct version of docker onto the image, and pulls
-##         down the default docker image used for building on ubuntu
-################################################################################
-
-source $HELPER_SCRIPTS/apt.sh
-source $HELPER_SCRIPTS/document.sh
-
-DOCKER_PACKAGE=moby
-
```
maxim-lobanov: Should we also remove `docker.sh` from packer templates?

maxim-lobanov:
Fix for the issue #290

Issue #213, opened 2019-12-29
add .dotnet/tools to path


Jetersen:
**Describe the bug**
DotNet global tool path should be added to path.

I tried to fix it in https://github.com/actions/setup-dotnet however I was suggested to submit a patch for the virtual environments.

See https://github.com/actions/setup-dotnet/pull/70

**Area for Triage**: .NET Core

**Feature**:

**Virtual environments affected**
- [x] macOS 10.15
- [x] Ubuntu 16.04 LTS
- [x] Ubuntu 18.04 LTS
- [N/A] Windows Server 2016 R2
- [WORKS] Windows Server 2019

**Expected behavior**
after installing a dotnet tool it should be accessible.

```bash
dotnet tool install -g gitversion
dotnet gitversion -versiom
5.1.3+Branch.master.Sha.bef8ebc0b62b3ddd0cdafe09b66d68bbfcaf90d5
```

**Actual behavior**
See my repro:
https://github.com/casz/setup-dotnet-global-tool-repro
See my fix working for setup-dotnet action:
https://github.com/casz/setup-dotnet-global-tool-repro/blob/testingFix/.github/workflows/ci.yml

```bash
dotnet tool install -g gitversion
dotnet gitversion -version
Could not execute because the specified command or file was not found.
Possible reasons for this include:
  * You misspelled a built-in dotnet command.
  * You intended to execute a .NET Core program, but dotnet-gitversion does not exist.
  * You intended to run a global tool, but a dotnet-prefixed executable with this name could not be found on the PATH.
```

casz:
This does not behave as intended:
https://github.com/actions/virtual-environments/blob/95954b3f022a1c4626d112bd4fe82d0812fa6f4b/images/linux/scripts/installers/1804/dotnetcore-sdk.sh#L94

casz:
Added tests for all OSes listed here: https://help.github.com/en/actions/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions#jobsjob_idruns-on

See https://github.com/casz/setup-dotnet-global-tool-repro/runs/366837035

casz:
Most likely a .NET Core area rather than Git area 😅 
cc @bishal-pdmsft @pulkitaggarwl

casz:
I can confirm this also affects Linux and Mac environments on Azure Pipeline.

casz:
for anyone using GitHub actions and would like a workaround

Use my branch for actions/setup-dotnet
```yaml
[...]

Issue #118, opened 2019-11-28
The composer global bin path isn't added to `$PATH`


georgeboot:
**Describe the bug**
The composer global bin path isn't added to `$PATH`

**Virtual environments affected**
- [ ] macOS 10.15
- [ ] Ubuntu 16.04 LTS
- [x] Ubuntu 18.04 LTS
- [ ] Windows Server 2016 R2
- [ ] Windows Server 2019

**Expected behavior**
The composer global path (`/home/worker/.composer/vendor/bin`) is added to the `$PATH`.

**Actual behavior**
It isn't in the path. and thus you have to call globally installed executables with their full path. As this might change some day, this isn't ideal.

Note that I have only tested this on 18.04, but it's likely an issue on the other environments too.

vsafonkin:
PR: [#391](https://github.com/actions/virtual-environments/pull/391)

vsafonkin:
Hello @georgeboot , unfortunately, adding "export PATH" command to user's .bashrc file didn't resolve problem, we will continue to work on it.

georgeboot:
Hmm that is strange. Sure it's added to the correct rc file?

vsafonkin:
Yes, user's bashrc contains export instruction, but it works only with terminal session, if we use image for ci it's not working ((

vsafonkin:
PR: [https://github.com/actions/virtual-environments/pull/477](https://github.com/actions/virtual-environments/pull/477)

vsafonkin:
The path `/home/runner/.config/composer/vendor/bin` was added into $PATH. It can use for Github Actions pipelines (user `runner`).

@georgeboot , I close this issue, please reopen it if you have any concerns, thanks!

PR #477, opened 2020-02-28
Update PATH environment variable to add .dotnet and php composer default folder

al-cheb:
Customers requested to add  .dotnet and php composer default folders to the PATH to eliminate manual update `PATH` environment variable to the workflows.

Fix add updatepath.sh  script to update all paths at the single place:
1. `/home/runner/.dotnet/tools` - customer request https://github.com/actions/virtual-environments/issues/213
2. `/home/runner/.config/composer/vendor/bin` -  customer request https://github.com/actions/virtual-environments/issues/118
3. move ${CARGO_HOME} from `rust.sh` to `updatepath.sh`

PS. `/home/runner` - is the $HOME folder for `runner` user under that user all builds run.

### DIFF
--- a/images/linux/scripts/installers/rust.sh
+++ b/images/linux/scripts/installers/rust.sh
@@ -15,7 +15,6 @@ export CARGO_HOME=/usr/share/rust/.cargo
 curl https://sh.rustup.rs -sSf | sh -s -- -y
 
 # Add Cargo and Rust binaries to the machine path
-echo "PATH=${CARGO_HOME}/bin:$PATH" | tee -a /etc/environment
 source $CARGO_HOME/env
 
 # Install common tools
--- a/images/linux/scripts/installers/rust.sh
+++ b/images/linux/scripts/installers/rust.sh
@@ -14,7 +14,7 @@ export CARGO_HOME=/usr/share/rust/.cargo
 
 curl https://sh.rustup.rs -sSf | sh -s -- -y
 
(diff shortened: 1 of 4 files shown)


images/linux/scripts/installers/rust.sh
```diff
@@ -15,7 +15,6 @@ export CARGO_HOME=/usr/share/rust/.cargo
 curl https://sh.rustup.rs -sSf | sh -s -- -y
 
 # Add Cargo and Rust binaries to the machine path
```
miketimofeev: Do we need to change the comment?

images/linux/scripts/installers/rust.sh
```diff
@@ -14,8 +14,7 @@ export CARGO_HOME=/usr/share/rust/.cargo
 
 curl https://sh.rustup.rs -sSf | sh -s -- -y
 
-# Add Cargo and Rust binaries to the machine path
-echo "PATH=${CARGO_HOME}/bin:$PATH" | tee -a /etc/environment
```
alepauly: I think the verification at line 32 below is going to fail now since the PATH isn't updated, is that true?

al-cheb: @alepauly,
This `source $CARGO_HOME/env` command imports PATH variable from the `$CARGO_HOME/env` , file content:
`export PATH="/usr/share/rust/.cargo/bin:$PATH"`

PR #504, opened 2020-03-05
Unite Ubuntu powershell installation scripts

miketimofeev:
The only difference between both installation scripts was a workaround for Ubuntu 16.04. It's not needed anymore since Powershell core 7.


images/linux/scripts/installers/powershellcore.sh
```diff
@@ -7,10 +7,14 @@
...
 source $HELPER_SCRIPTS/document.sh
 
-# libicu64, which comes with php-intl module, has powershell breaking issue https://github.com/PowerShell/PowerShell/issues/9746
+LSB_RELEASE=$(lsb_release -rs)
+
+# libicu64, which comes with php-intl module, has powershell breaking issue on Ubuntu 16.04 https://github.com/PowerShell/PowerShell/issues/9746
 # Fix - install additional libicu65 where the issue is fixed
-echo "install libicu65"
-apt get install libicu65
+if [ $LSB_RELEASE == "16.04" ]; then
```
al-cheb: Must be 18.04

miketimofeev: Why? This workaround is for 16.04 only.

miketimofeev: That's my fault, thanks. But it seems we don't need it anymore - I'll check

miketimofeev: Yeah, it's not needed anymore with Powershell core 7.

PR #535, opened 2020-03-11
Removed installing 2.2 dotnetcore version in linux and windows images

hiyadav:
This change is because 2.2 version of .NET Core has been deprecated in January, 2020. As we try only to keep the versions which are not End Of Line (2.1, 3.0 and 3.1)

We have also added a message in .NET Core CLI task in Azure DevOps, which states that 2.2 is EOL and will be removed by end of March 2020. The users can however still choose to use this version by using UseDotNet V2 task to download the required version by them.

In .NET Core, its recommended to always build using the latest SDK, if however a specific version of the SDK is required, than use of UseDotNet task is recommended.

### DIFF
--- a/images/linux/scripts/installers/dotnetcore-sdk.sh
+++ b/images/linux/scripts/installers/dotnetcore-sdk.sh
@@ -48,7 +48,7 @@ for latest_package in ${LATEST_DOTNET_PACKAGES[@]}; do
 done
 
 # Get list of all released SDKs from channels which are not end-of-life or preview
-release_urls=("https://dotnetcli.blob.core.windows.net/dotnet/release-metadata/2.1/releases.json" "https://dotnetcli.blob.core.windows.net/dotnet/release-metadata/2.2/releases.json" "https://dotnetcli.blob.core.windows.net/dotnet/release-metadata/3.0/releases.json" "https://dotnetcli.blob.core.windows.net/dotnet/release-metadata/3.1/releases.json")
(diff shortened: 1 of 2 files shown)


hiyadav:
@alepauly @bishal-pdMSFT can you help in getting this PR reviewed and merged ?

maxim-lobanov:
@hiyadav, thank you for contribution!
Could you please update PR title and provide the reason to remove those versions.
Also I wonder if it can be potential regression for some customers?

hiyadav:
Hey @maxim-lobanov 
This change is because 2.2 version of .NET Core has been deprecated in January, 2020. As we try only to keep the versions which are not End Of Line (2.1, 3.0 and 3.1), I made the change to remove 2.2 version, as we did earlier for 1.0, 1.1 and 2.0. 

For informing customers, we have also added a message in .NET Core CLI task in Azure DevOps, which states that 2.2 is EOL and will be removed by end of March 2020. The users can however still choose  to use this version by using UseDotNet V2 task to download the required version by them.

In .NET Core, its recommended to always build using the latest SDK, if however a specific version of the SDK is required, than use of UseDotNet task is recommended.

maxim-lobanov:
@hiyadav , thank you for explanation, It sounds good to me!
I have run CI builds for validation.

hiyadav:
@alepauly 
That is true, however I'm not aware of the mechanism to notify users in GitHub Actions ? 

One way could be to add a message in `setup-` action, but I'm not sure who manages the `setup-` action.

We should however, take action on this soon, as .NET Core 2.2 will no longer get any security updates as well and thus might be risking the users and our environment

maxim-lobanov:
@hiyadav, we have decided to keep dotnetcore 2.2 on our images for 3 more months. Please see #567

maxim-lobanov:
Hello!
We have remove dotnetcore 2.2 from all images: #1016 , #1017 and macOS.
Closing this PR

PR #536, opened 2020-03-11
Add readme output step to Image-generation CI

nikita-bykov:
We added a new step to Image.CI, which writes a README file containing an array of installed software on the image.

Issue: https://github.com/actions/virtual-environments-internal/issues/234

Windows2016: https://github.visualstudio.com/virtual-environments/_build/results?buildId=62058&view=results
Ubuntu1604: https://github.visualstudio.com/virtual-environments/_build/results?buildId=62059&view=results
Ubuntu1804: https://github.visualstudio.com/virtual-environments/_build/results?buildId=62027&view=results


images.CI/azure-pipelines/image-generation.yml
```diff
@@ -55,6 +55,13 @@ jobs:
                         -ImageName ${{ parameters.image_type }} `
                         -AccessToken $(RELEASE_TARGET_TOKEN)
 
+  - task: PowerShell@2
+    displayName: 'Output Readme file content'
+    inputs:
+      targetType: 'inline'
+      script: |
+        Get-Content ((Get-ChildItem -Path "images" -Include ${{ parameters.image_readme_name }} -Recurse -Depth 1).FullName)
```
maxim-lobanov: Suggest to split this command to two lines:
```
$docsPath = (Get-ChildItem -Path "images" -Include ${{ parameters.image_readme_name }} -Recurse -Depth 1 | Select-Object -First 1
$docsPath | Get-Content
```

miketimofeev:
@nikita-bykov could you provide build examples, please?

nikita-bykov:
@miketimofeev Added build examples to description.

images.CI/azure-pipelines/image-generation.yml
```diff
@@ -55,6 +55,14 @@ jobs:
                         -ImageName ${{ parameters.image_type }} `
                         -AccessToken $(RELEASE_TARGET_TOKEN)
 
+  - task: PowerShell@2
+    displayName: 'Output Readme file content'
```
vmapetr: I believe it's called InstalledSoftware.md

images.CI/azure-pipelines/image-generation.yml
```diff
@@ -55,6 +55,14 @@ jobs:
...
+    inputs:
+      targetType: 'inline'
+      script: |
+        $docsPath = Get-ChildItem -Path "images" -Include ${{ parameters.image_readme_name }} -Recurse -Depth 1 | Select-Object -First 1
```
vmapetr: I'm not getting it - why we showing output for Readme file from repo, thats updated after image deployment, instead of showing content for InstalledSoftware.md thats generated with Image?

images.CI/azure-pipelines/image-generation.yml
```diff
@@ -55,6 +55,14 @@ jobs:
...
+        Get-Content -Path $docsPath
```
vmapetr: Also, I believe, such things should be downloaded into build artifacts.

images.CI/azure-pipelines/image-generation.yml
maxim-lobanov: There is no any `InstalledSoftware.md`.
All packer templates just replace existing file in the repository. Please see packer templates for details

images.CI/azure-pipelines/image-generation.yml
maxim-lobanov: If we upload it to build artifacts, we have to download it locally to read and validate. Step logs are available in web UI. If you see that it makes sense, we can publish build artifact in additional to this step. But artifact is not convenient way for initial purpose of PR

images.CI/azure-pipelines/image-generation.yml
vmapetr: There is always `InstalledSoftware.md` generated in the produced image root folder. Can't we download it into build agent, right after image generation?

vmapetr: And I'm actually don't understand, why we need to show the output for Readme, if we can't get InstalledSoftware, because Readme content always will be outdated, if you compare it with the generated image.

maxim-lobanov: We have already do that.
image
but they are downloaded to templates directory and this step searches them in this directory

vmapetr: Missed that, thanks for the explanation!

PR #537, opened 2020-03-11
Add retry logic for VS19 extensions installation

miketimofeev:
The last 2 days VS extension marketplace returns Error 429, which breaks image building.
```
==> vhd: Provisioning with powershell script: C:\agent\_work\2\s\images\win/scripts/Installers/Windows2019/Install-AnalysisExtenstion.ps1
    vhd: Downloading Microsoft.DataTools.AnalysisServices.vsix extension
    vhd: Installing Microsoft.DataTools.AnalysisServices.vsix extension
    vhd: Exception calling "DownloadFile" with "2" argument(s): "The remote server returned an error: (429)."
```
This PR adds retry logic for extension download process.

Changes:
- Add Install-VsixExtension function
- Change Install-Wix and Install-AnalysisExtenstion scripts to use the helper function

### DIFF
--- a/images/win/scripts/ImageHelpers/InstallHelpers.ps1
+++ b/images/win/scripts/ImageHelpers/InstallHelpers.ps1
@@ -170,7 +170,6 @@ function Install-VsixExtension
     )
 
     $FilePath = "${env:Temp}\$Name"
-    $exitCode = -1
     $retries = 20
 
     while($retries -gt 0)
--- a/images/win/scripts/ImageHelpers/InstallHelpers.ps1
+++ b/images/win/scripts/ImageHelpers/InstallHelpers.ps1
@@ -161,6 +161,74 @@ Hashtable for service arguments
     }
 }
 
+function Install-VsixExtension
+{
+    Param
+    (
+        [String]$Url,
+        [String]$Name
+    )
+
+    $FilePath = "${env:Temp}\$Name"
+    $ReleaseInPath = 'Enterprise'
+    $exitCode = -1
+    $retries = 20
+
+    while($true)
+    {
+        try
+        {
+            Write-Host "Downloading $Name..."
+            (New-Object System.Net.WebClient).DownloadFile($Url, $FilePath)
+            break
+        }
+        catch
+        {
+            Write-Host "There is an error during $Name downloading"
+            $_
+            if ($retries -gt 0)
+            {
+                $retries--
+                Write-Host "Waiting 30 seconds before retrying. Retries left: $retries"
+                Start-Sleep -Seconds 30
+            }
+            else {
+                Write-Host "File can't be downloaded"
+                $_
+                exit 1
+            }
+        }
+    }
+
+    $ArgumentList = ('/quiet', $FilePath)
+
+    Write-Host "Starting Install $Name..."
+    try
+    {
+        $process = Start-Process -FilePath "C:\Program Files (x86)\Microsoft Visual Studio\2019\$ReleaseInPath\Common7\IDE\VSIXInstaller.exe" -ArgumentList $ArgumentList -Wait -PassThru
+    }
+    catch
+    {
+        Write-Host "There is an error during $Name installation"
+        $_
+        exit 1
+    }
+
+    $exitCode = $process.ExitCode
+
+    if ($exitCode -eq 0 -or $exitCode -eq 1001) # 1001 means the extension is already installed
+    {
+        Write-Host "$Name installed successfully"
+    }
+    else
+    {
+        Write-Host "Unsuccessful exit code returned by the installation process: $exitCode."
+        exit 1
+    }
+
+    #Cleanup installation files
+    Remove-Item -Force -Confirm:$false $FilePath
+}
 function Get-VS19ExtensionVersion
 {
     param (
--- a/images/win/scripts/ImageHelpers/InstallHelpers.ps1
+++ b/images/win/scripts/ImageHelpers/InstallHelpers.ps1
@@ -170,11 +170,10 @@ function Install-VsixExtension
     )
 
     $FilePath = "${env:Temp}\$Name"
-    $ReleaseInPath = 'Enterprise'
     $exitCode = -1
     $retries = 20
 
-    while($true)
+    while($retries -gt 0)
     {
         try
         {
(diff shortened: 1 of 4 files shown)


images/win/scripts/ImageHelpers/InstallHelpers.ps1
```diff
@@ -161,6 +161,75 @@ Hashtable for service arguments
...
+                Write-Host "Waiting 30 seconds before retrying. Retries left: $retries"
+                Start-Sleep -Seconds 30
+            }
+            else {
+                Write-Host "File can't be downloaded"
+                $_
+                exit 1
+            }
+        }
+    }
```
AlenaSviridenko: loop looks a bit extra complex with `while($true)`, may be something like this will also work?
```
while($retries -gt 0)
{
    try
    {
        Write-Host "Downloading $Name..."
        (New-Object System.Net.WebClient).DownloadFile($Url, $FilePath)
        break
    }
    catch
    {
        Write-Host "There is an error during $Name downloading"
        $_

        $retries--

        if ($retries -eq 0) {
            Write-Host "File can't be downloaded"
            $_
            exit 1
        }

        Write-Host "Waiting 30 seconds before retrying. Retries left: $retries"
        Start-Sleep -Seconds 30
    }
}
```

al-cheb:
Maybe it makes sense to combine Install-AnalysisExtenstion.ps1 and Install-Wix.ps1 into one script something like - Install-VSExtension.ps1

images/win/scripts/ImageHelpers/InstallHelpers.ps1
```diff
@@ -161,6 +161,75 @@ Hashtable for service arguments
...
+function Install-VsixExtension
+{
+    Param
+    (
+        [String]$Url,
+        [String]$Name
+    )
+
+    $FilePath = "${env:Temp}\$Name"
+    $ReleaseInPath = 'Enterprise'
```
al-cheb: Do we really need a hardcoded value?

miketimofeev: I'm not sure, that part was taken from Install-Wix.ps1, probably we can get rid of it

miketimofeev:
> Maybe it makes sense to combine Install-AnalysisExtenstion.ps1 and Install-Wix.ps1 into one script something like - Install-VSExtension.ps1

I'd prefer to leave Wix-extension in the same script where wixtoolset is installing.

images/win/scripts/ImageHelpers/InstallHelpers.ps1
miketimofeev: That makes sense! Changed, thank you.

al-cheb: @AlenaSviridenko Maybe we should remove duplicate error message?
```
 Write-Host "File can't be downloaded"
            $_
```

images/win/scripts/ImageHelpers/InstallHelpers.ps1
```diff
@@ -161,6 +161,75 @@ Hashtable for service arguments
...
+function Install-VsixExtension
+{
+    Param
+    (
+        [String]$Url,
+        [String]$Name
+    )
+
+    $FilePath = "${env:Temp}\$Name"
+    $exitCode = -1
```
al-cheb: This variable is always replaced on 217 line and we don't need it. Or we should change logic and put the variable in try-catch block 208-217,

images/win/scripts/ImageHelpers/InstallHelpers.ps1
AlenaSviridenko: @al-cheb I think `$_` contains information about why it can't be dowloaded, like response error code or something else, so it doesn't looks like a duplication between `$_` and `File can't be downloaded`
please, correct me if I am missing something

images/win/scripts/ImageHelpers/InstallHelpers.ps1
```diff
@@ -161,6 +161,75 @@ Hashtable for service arguments
     }
 }
 
+function Install-VsixExtension
+{
+    Param
+    (
+        [String]$Url,
```
vmapetr: nitpick: uri instead of string

images/win/scripts/Installers/Windows2019/Install-Wix.ps1
```diff
@@ -2,52 +2,13 @@
...
-        return -1
-    }
-}
 
+Import-Module -Name ImageHelpers -Force;
 
 choco install wixtoolset -y --force
 
+$extensionUrl = "https://wixtoolset.gallerycdn.vsassets.io/extensions/wixtoolset/wixtoolsetvisualstudio2019extension/1.0.0.4/1563296438961/Votive2019.vsix"
+$extensionName = "Votive2019.vsix"
```
vmapetr: Url is always contains extension? Maybe we don't need additional argument for it?

images/win/scripts/ImageHelpers/InstallHelpers.ps1
al-cheb: @AlenaSviridenko:

First occurence:
```
Write-Host "There is an error during $Name downloading"
        $
```

Second occurence:
```
Write-Host "File can't be downloaded"
            $_
```

AlenaSviridenko: @al-cheb oh, you mean that duplication. Then yes, I agree, we can remove it

images/win/scripts/Installers/Windows2019/Install-Wix.ps1
miketimofeev: Not always - this particular extension is not from VS marketplace, that's why it has the name in the uri.

images/win/scripts/ImageHelpers/InstallHelpers.ps1
miketimofeev: It's also a part of the old function. I guess they didn't know that empty != 0 😃
Removed.

images/win/scripts/ImageHelpers/InstallHelpers.ps1
miketimofeev: Let's leave Url for consistency – all the other helpers function use Url, not Uri

vmapetr: Actually, I'm talking about argument type. [uri] instead of [string]

Issue #491, opened 2020-03-03
No current user owned `${XDG_CONFIG_HOME}` (aka `${HOME}/.config`) directory 


alepauly:
As originally reported by @ivailop in https://github.com/microsoft/azure-pipelines-image-generation/issues/1271:

**Describe the bug**
The hosted "Ubuntu 16.04" image has no current user owned `${XDG_CONFIG_HOME}` (aka `${HOME}/.config`) directory.

**Expected behavior**
Have existing `${XDG_CONFIG_HOME}` (aka `${HOME}/.config`) pre-created with current user (i.e. `vsts`) being the owner of it.

**Actual behavior**
`${XDG_CONFIG_HOME}` (aka `${HOME}/.config`) directory does not exist. As being a de-facto location for storing user-specific configuration, installing dependencies using `sudo` would automatically create it with `root` being the owner of it. Subsequent non-sudo access to it would fail with `Permission denied`.

To reproduce:
```
- script: |
    ls -lsa ~/.config # -> ls: cannot access '/home/vsts/.config': No such file or directory
    sudo npm install -g eslint
    ls -lsa ~/.config # -> ls: cannot open directory '/home/vsts/.config': Permission denied
```

**Image impacted**
Ubuntu 16.04

miketimofeev:
@ivailop could you please clarify why this directory should be owned for the current user for your builds? 
Is it possible to add `mkdir ~/.config` as the first step of your build? I'm asking because the default user is created not during the image generation step but on some later VM provisioning step and adding this directory globally might be quite complicated.

ivailop:
> could you please clarify why this directory should be owned for the current user for your builds?

It must be owned by the current user as being in its `HOME` directory. 
Normally, system dependencies (using `sudo`) are installed in the **very beginning** of a workflow. And as shown above that could lead to `~/.config` being stolen (`root` owned, instead of user).

Later on, any tool being run (non-sudo) that needs to write access to `~/.config` is being denied such!

> Is it possible to add `mkdir ~/.config` as the first step of your build? 

Yes, this is what I end up doing as a workaround. But it would be nice if someone does not have to discover a non-obvious problem like that and search for workarounds to fix it.

miketimofeev:
@ivailop thanks for the answers. Does it make sense to pre-create configstore directory(under ~/.config) with the same permissions as well?

ivailop:
[...]

PR #544, opened 2020-03-12
Add XDG_CONFIG_HOME directory to Ubuntu images

miketimofeev:
The hosted Ubuntu images have no current user-owned ${XDG_CONFIG_HOME} (aka ${HOME}/.config) directory, which causes issues for some tasks(installing dependencies using sudo would automatically create it with the root being the owner of it)
https://github.com/actions/virtual-environments/issues/491

**Changes:**
- Add $HOME/.config/configstore directory owned by user
- Add XDG_CONFIG_HOME environmental variable


images/linux/ubuntu1604.json
```diff
@@ -129,6 +129,14 @@
             ],
             "execute_command": "sudo sh -c '{{ .Vars }} {{ .Path }}'"
         },
+        {
+            "type": "shell",
+            "inline": [
+                "mkdir -p /etc/skel/.config/configstore",
+                "echo 'export XDG_CONFIG_HOME=$HOME/.config' | tee -a /etc/skel/.bashrc"
```
al-cheb: This line doesn't work without using external provisionator, the same issue was at https://github.com/actions/virtual-environments/blob/master/images/linux/scripts/installers/dotnetcore-sdk.sh.

miketimofeev: I've tested it on deployed VM and it works fine. Are there any changes made by the provisioner that breaks it?

images/linux/ubuntu1804.json
```diff
@@ -132,6 +132,14 @@
...
             "execute_command": "sudo sh -c '{{ .Vars }} {{ .Path }}'"
         },
+        {
+            "type": "shell",
+            "inline": [
+                "mkdir -p /etc/skel/.config/configstore",
+                "echo 'export XDG_CONFIG_HOME=$HOME/.config' | tee -a /etc/skel/.bashrc"
+            ],
+            "execute_command": "sudo sh -c '{{ .Vars }} {{ .Path }}'"
+        },
```
maxim-lobanov: It would be great to add any comment here, why we add this logic. Link to the issue or something similar

images/linux/ubuntu1604.json
miketimofeev: I've got your idea, but let's leave it for self-hosted agents.

images/linux/ubuntu1804.json
miketimofeev: Unfortunately, JSON doesn't support comments

maxim-lobanov: does it make sense to move these lines to very small shell script? It sounds overkilled but just want to make sure that we will remember the reason.
May be something like:
```
"inline": [
                "# It is my comment",
                "mkdir -p /etc/skel/.config/configstore",
                "echo 'export XDG_CONFIG_HOME=$HOME/.config' | tee -a /etc/skel/.bashrc"
            ],
```

miketimofeev: Good point, thanks! I've added a comment

images/linux/ubuntu1604.json
```diff
@@ -129,6 +129,15 @@
             ],
             "execute_command": "sudo sh -c '{{ .Vars }} {{ .Path }}'"
         },
+        {
+            "type": "shell",
+            "inline": [
```
alepauly: Can we move this to a script that's shared for both images? I would also include the inline calls above to set ImageVersion and ImageOS. I rather we minimize the logic directly in the json file since it's harder to test locally.

miketimofeev: Sure thing! I'll make the changes.

images/linux/scripts/installers/envvariables.sh
```diff
@@ -0,0 +1,7 @@
+#Set ImageVersion and ImageOS env variables
+"echo ImageVersion=$IMAGE_VERSION | tee -a /etc/environment",
```
al-cheb: 1. Remove ,
2. Remove  double quotes

miketimofeev: my fault, thanks!

images/linux/ubuntu1804.json
```diff
@@ -126,9 +126,12 @@
         },
         {
             "type": "shell",
-            "inline": [
-                "echo ImageVersion={{user `image_version`}} | tee -a /etc/environment",
-                "echo ImageOS={{user `image_os`}} | tee -a /etc/environment"
+            "scripts": [
+                "{{template_dir}}/scripts/installers/envvariables.sh"
```
maxim-lobanov: nitpick: may be `configure-environment.sh` so we will be able to reuse this file in future for some minor system settings

miketimofeev: Yeah, that was the initial idea. I like your naming better - renamed

images/linux/ubuntu1604.json
miketimofeev: Changes have been made

images/linux/ubuntu1804.json
al-cheb: In that case it makes sense to combine `/updatepath.sh` and `configure-environment.sh` into one script.

PR #549, opened 2020-03-13
Change the listed Helm version

cannibal-kush:
Because in fact version 3 is installed

https://github.com/actions/virtual-environments/blob/0c2643cb8d11706a08c366b3929e34f34b95197f/images/linux/scripts/installers/1604/kubernetes-tools.sh#L20


al-cheb:
Hello, @cannibal-kush 
Please don't update auto-generated documentation. It will be fixed soon in - https://github.com/actions/virtual-environments/releases/tag/ubuntu16%2F20200308.0

cannibal-kush:
This is an issue as i have helm 3 installed on my build agent but the documentation leads me to believe it is version 2. I suggest you look at that process

maxim-lobanov:
@cannibal-kush , Usually, deployment takes 3-4 days (during this period your build can use updated or non-updated machines) and we update docs only when the all machines use new version of image.

gregdasney:
@maxim-lobanov  Helm 3 introduced breaking changes.  If agents take 3-4 days to be updated how do I ensure that my builds will work correctly during the interim since helm2 vs helm3 require some changes to the scripts?

eg:
helm2 requires `helm init --client-only` to work, but this command fails with helm3
helm3 requires `/home/<USER>/.config/helm` to exist

There are other breaking changes but these are just two examples.

PR #555, opened 2020-03-13
Fix errors in Initialize-VM.ps1

olandese:
With this pull request the following errors will be solved when running the Initialize-VM.ps1:

Error 1: 

``` 
Stop-Process : Cannot find a process with the name "Explorer". Verify the process name and call the cmdlet again.
Stop-Process -Name Explorer -Force -ErrorAction Continue
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : ObjectNotFound: (Explorer:String) [Stop-Process], ProcessCommandException
    + FullyQualifiedErrorId : NoProcessFoundForGivenName,Microsoft.PowerShell.Commands.StopProcessCommand

```
Error 2: 

```

+ Set-ExecutionPolicy Unrestricted
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : PermissionDenied: (:) [Set-ExecutionPolicy], SecurityException
    + FullyQualifiedErrorId : ExecutionPolicyOverride,Microsoft.PowerShell.Commands.SetExecutionPolicyCommand
```

### DIFF
--- a/images.CI/build-image.ps1
+++ b/images.CI/build-image.ps1
@@ -36,6 +36,9 @@ $SensitiveData = @(
     ':  ->'
 )
 
+Write-Host "Show Packer Version"
+packer --version
+
 Write-Host "Build $Image VM"
 packer build    -var "capture_name_prefix=$ResourcesNamePrefix" `
                 -var "client_id=$ClientId" `
--- a/images/win/scripts/Installers/Windows2016/Initialize-VM.ps1
+++ b/images/win/scripts/Installers/Windows2016/Initialize-VM.ps1
@@ -8,7 +8,13 @@ function Disable-InternetExplorerESC {
     $UserKey = "HKLM:\SOFTWARE\Microsoft\Active Setup\Installed Components\{A509B1A8-37EF-4b3f-8CFC-4F3A74704073}"
     Set-ItemProperty -Path $AdminKey -Name "IsInstalled" -Value 0 -Force
     Set-ItemProperty -Path $UserKey -Name "IsInstalled" -Value 0 -Force
-    Stop-Process -Name Explorer -Force -ErrorAction Continue
+
+    $ieProcess = Get-Process -Name Explorer -ErrorAction SilentlyContinue
+
+    if ($ieProcess){
+        Stop-Process -Name Explorer -Force -ErrorAction Continue
+    }
+
     Write-Host "IE Enhanced Security Configuration (ESC) has been disabled."
 }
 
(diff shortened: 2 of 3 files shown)


miketimofeev:
@olandese `"execution_policy": "unrestricted"` can't be used that way. There is an error:
`unknown configuration key: "execution_policy"`

olandese:
@miketimofeev this is strange. 
I am running the packer build locally on my pc (packer version 1.5.4) and it runs well.
The option I applied is described here:

https://packer.io/docs/provisioners/powershell.html#execution_policy

miketimofeev:
@olandese it might be related to outdated packer version 1.3.4, which used on the image generation agents. We will consider a possible update.

olandese:
@miketimofeev it's probably related to that. It would be a good choice to update the packer version. My changes will prevent the ugly errors to appear :)

al-cheb:
Hello, @olandese
1. I don't think we should change executionpolicy to `Unresticted`, because by default value, that is very good settings, is bypass -  ` To run ps scripts on windows packer defaults this to "bypass" and wraps the command to run`. I prefer to remove your changes from packer config.
2. We have not got any issues with executionpolicy and my concern is to remove statement at all:
`Set-ExecutionPolicy -ExecutionPolicy Unrestricted -Scope LocalMachine  -ErrorAction Continue | Out-Null`

olandese:
Hi @al-cheb,

1. The execution policy I added in packer was to solve Error 2 (see my opening post).
2. If the Set-ExecutionPolicy is not need, then indeed, better remove that

I will update my PR to solve only the problem when trying to stop the Explorer Process

al-cheb:
@olandese, Thank you for your contribution and good idea about updating packer. I agree we should update packer maybe not to the latest one, but at least 1.4.5.

olandese:
> @olandese, Thank you for your contribution and good idea about updating packer. I agree we should update packer maybe not to the latest one, but at least 1.4.5.

Thank you guys! 
Can you please run a build to check if everything works fine?

alepauly:
@olandese before I merge, can you please cleanup the branch to include only the commits we want? There's a few unnecessary commits and I rather not squash/merge to mantain your authorship. Thanks!

olandese:
> @olandese before I merge, can you please cleanup the branch to include only the commits we want? There's a few unnecessary commits and I rather not squash/merge to mantain your authorship. Thanks!

Hi @alepauly, sure, which commits do you want to keep/remove?

alepauly:
> Hi @alepauly, sure, which commits do you want to keep/remove?

@olandese: If you can rebase/squash locally and repush, seems like a simple well-messaged commit is all that's needed.

olandese:
> > Hi @alepauly, sure, which commits do you want to keep/remove?
> 
> @olandese: If you can rebase/squash locally and repush, seems like a simple well-messaged commit is all that's needed.

Done!

olandese:
Thank you guys for the merge!
Btw, is packer been updated on the build agents?

maxim-lobanov:
@olandese , yep, we have updated build agents to use Packer 1.5.4 👍

Issue #4, opened 2019-10-01
Update/Add nvm


scottohara:
**Tool information**
- Tool name: nvm
- Add or update? Add
- Desired version: latest
- Approximate size: 
- If this is an add request:
  - Brief description of tool: install/switch between Node.js versions, auto-installs & uses the node version as specified in a project’s `.nvmrc` file
  - URL for tool's homepage: https://github.com/nvm-sh/nvm

**Virtual environments affected**
- [ ] macOS 10.14
- [x] Ubuntu 16.04 LTS
- [x] Ubuntu 18.04 LTS
- [ ] Windows Server 2016 R2
- [ ] Windows Server 2019

(Already installed one macOS virtual env)

**Can this tool be installed during the build?**

Yes, but not easily. See https://github.community/t5/GitHub-Actions/How-to-share-shell-profile-between-steps-or-how-to-use-nvm-rvm/td-p/33185

**Are you willing to submit a PR?**

scottohara:
`setup-node` requires an explicit version of nodejs needed for the job to be specified in the `with:` property.

Many nodejs projects already have a file in their repo that specifies a desired node version (e.g.`.nvmrc`), so having to repeat the desired node version in `.github/workflows/*.yml` would seem unnecessary duplication.

If setup-node could detect and use the version specified in an existing .nvmrc then arguably installing nvm would not be necessary.

Additional points:

1. The macOS virtual env already has nvm preinstalled (but the Linux ones don’t)
2. CI services such as TravisCI have nvm preinstalled for this very purpose

thejoebourneidentity:
I'm working with the `setup-node` owners to determine if https://github.com/actions/setup-node/issues/32 is something they plan on fixing. Leaving this issue open for now to track.

thejoebourneidentity:
Thanks for the suggestion @scottohara, we're going to add nvm as requested.

@maxim-lobanov , let's do it, let me know if you need any additional assistance.

maxim-lobanov:
@scottohara , before we start adding it to the Linux, could you please confirm that it works as expected on macOS images? I am asking because as I know, nvm doesn't work properly on macOS due to NVM initialization way and you have to write initialization trick in each pipeline:
```
# Delete prefix from system node config
npm config delete prefix
# Need for using `nvm` inside script
. "$HOME/.nvm/nvm.sh"
nvm use $version
nvm alias default $(nvm current)

echo "Node version: $(node -v)"
echo "npm version: $(npm -v)"

[...]

PR #582, opened 2020-03-19
Add nvm for linux

dmitry-shibanov:
- Add nvm to ubuntu

Initial issue: https://github.com/actions/virtual-environments/issues/4


maxim-lobanov:
Could you please update title to reflect OS where you install NVM?
Could you please update description with the list of changes and mention initial issue?

images/linux/scripts/installers/nvm.sh
```diff
@@ -0,0 +1,24 @@
...
+# Source the helpers for use with the script
+source $HELPER_SCRIPTS/document.sh
+
+export NVM_DIR="/opt/.nvm"
+mkdir -m 777 $NVM_DIR
+curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.35.3/install.sh | bash
+chmod -R 777 $NVM_DIR
+echo "NVM_DIR=$NVM_DIR" | tee -a /etc/environment
+echo '[ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"  # This loads nvm' >> /etc/profile.d/nvm.sh
+echo '[ -s "$NVM_DIR/bash_completion" ] && \. "$NVM_DIR/bash_completion"  # This loads nvm bash_completio' >> /etc/profile.d/nvm.sh
```
maxim-lobanov: typo: `bash_completio`

images/linux/scripts/installers/nvm.sh
```diff
@@ -0,0 +1,24 @@
...
+################################################################################
+##  File:  nvm.sh
+##  Desc:  Installs Nvm
+################################################################################
+
+# Source the helpers for use with the script
+source $HELPER_SCRIPTS/document.sh
+
+export NVM_DIR="/opt/.nvm"
+mkdir -m 777 $NVM_DIR
```
maxim-lobanov: Why we need this magic with permissions in two places? 11 and 13 lines.
If they are necessary, please put small comment

images/linux/scripts/installers/nvm.sh
dmitry-shibanov: Done.

images/linux/scripts/installers/nvm.sh
dmitry-shibanov: Done.

images/linux/scripts/installers/nvm.sh
```diff
@@ -0,0 +1,25 @@
...
+################################################################################
+##  File:  nvm.sh
+##  Desc:  Installs Nvm
+################################################################################
+
+# Source the helpers for use with the script
+source $HELPER_SCRIPTS/document.sh
+
+export NVM_DIR="/etc/skel/.nvm"
+# should create directory with full permissions.
```
sergey-akhalkov, line 11: It is still not clear why we need full permissions here and below, could you please clarify and add a comment?

dmitry-shibanov: Yes, It was previously added when nvm was installed to /opt/cause it's system dir. Nvm installs new node versions to its dir. But now we install It to skel and nvm dir copies to $HOME dir of user, so I removed this action. I'll rerun my checks if everything is ok I'll leave a comment and close conversation.

dmitry-shibanov: @sergey-akhalkov , removed full permissions and work as expected

boredland:
Hi! How is this supposed to work? I tried this with this miniature workflow and it didn't:
```
name: nvm
on:
  push:
jobs:
  nvm:
    runs-on: ubuntu-latest
    steps:
      - name: try out nvm
        run: nvm install
```

al-cheb:
@boredland , please don't forget shell params:
`shell: bash -l {0}`

 -l        Make bash act as if it had been invoked as a login shell (see INVOCATION below).
       --login
              files ~/.bash_profile, ~/.bash_login, or ~/.profile.  By default, bash reads these  files  when  it  is
              invoked as a login shell (see INVOCATION below).

PR #585, opened 2020-03-20
Install MSYS2 to Windows.

vsafonkin:
Issue: [#30](https://github.com/actions/virtual-environments/issues/30)

Added scripts for installation and validation [MSYS2](https://www.msys2.org/). This PR is based on [PR #355 ](https://github.com/actions/virtual-environments/pull/355) from @MSP-Greg

Installation time: 6 - 7 min
Size: ~4.2 Gb


images/win/scripts/Installers/Validate-Msys2.ps1
```diff
@@ -0,0 +1,57 @@
...
+# Adding description of the software to Markdown
+
+# `gcc --version` gives output like:
+# gcc.exe (x86_64-posix-seh-rev0, Built by MinGW-W64 project) 5.3.0
+# Copyright (C) 2015 Free Software Foundation, Inc.
+# This is free software; see the source for copying conditions.  There is NO
+# warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+
+$SoftwareName = "MinGW"
+$(gcc --version).Split([System.Environment]::NewLine)[0] -match "\d\.\d\.\d$"
```
al-cheb: 10.0.0 will be parsed incorrect

images/win/scripts/Installers/Install-Msys2.ps1
```diff
@@ -0,0 +1,80 @@
...
+
+# References
+# https://github.com/msys2/MINGW-packages/blob/master/azure-pipelines.yml
+# https://packages.msys2.org/group/
+
+$env:orig_path = $env:PATH
+$env:git_path  = "C:\Program Files\Git"
+
+# get info from https://sourceforge.net/projects/msys2/files/Base/x86_64/
+$msy2_uri  = "http://repo.msys2.org/distrib/x86_64/msys2-base-x86_64-20190524.tar.xz"
```
AlenaSviridenko: nitpick: $msys2_uri. Also, could ypu please adjust variables naming to our common style that is used in other installers files and get rid of underlines?

images/win/scripts/Installers/Validate-Msys2.ps1
vsafonkin: Fixed

images/win/scripts/Installers/Install-Msys2.ps1
vsafonkin: Done

images/win/scripts/Installers/Install-Msys2.ps1
```diff
@@ -0,0 +1,80 @@
...
+# References
+# https://github.com/msys2/MINGW-packages/blob/master/azure-pipelines.yml
+# https://packages.msys2.org/group/
+
+$env:origPath = $env:PATH
+$env:gitPath  = "C:\Program Files\Git"
+
+# get info from https://sourceforge.net/projects/msys2/files/Base/x86_64/
+$msys2Uri  = "http://repo.msys2.org/distrib/x86_64/msys2-base-x86_64-20190524.tar.xz"
+$msys2File = "C:\Windows\Temp\msys2.tar.xz"
```
maxim-lobanov: `$msys2File = $env:TEMP "msys2.tar.xz"`

images/win/scripts/Installers/Install-Msys2.ps1
```diff
@@ -0,0 +1,80 @@
...
+
+# Download the latest msys2 x86_64
+Write-Host "Starting msys2 download"
+(New-Object System.Net.WebClient).DownloadFile($msys2Uri, $msys2File)
+Write-Host "Finished download"
+
+$msys2FileU = "/$msys2File".replace(':', '')
```
maxim-lobanov: Why we need this line?
`$msys2File = "C:\Windows\Temp\msys2.tar.xz"` I don't see any `:`

images/win/scripts/Installers/Install-Msys2.ps1
```diff
@@ -0,0 +1,80 @@
...
+
+# extract tar.xz to C:\
+Write-Host "Starting msys2 extraction"
+&$tar -Jxf $msys2FileU -C /c/
+Remove-Item $msys2File
+Write-Host "Finished extraction"
+
+$env:PATH = "C:\msys64\mingw64\bin;C:\msys64\usr\bin;$env:origPath"
+
+$ErrorActionPreference = "Continue"
```
maxim-lobanov: As I know, it is bad practice to change ErrorAction for the whole script. Why we need it?

images/win/scripts/Installers/Install-Msys2.ps1
```diff
@@ -0,0 +1,80 @@
...
+Invoke-Expression "bash -c `"pacman-key --init 2>&1`""
+
+Write-Host "bash -c pacman-key --populate msys2"
+Invoke-Expression "bash -c `"pacman-key --populate msys2 2>&1`""
+
+Write-Host "pacman --noconfirm -Syyuu"
+pacman.exe -Syyuu --noconfirm 2>$null
+pacman.exe -Syuu --noconfirm 2>$null
+
+# install msys2 packages
```
maxim-lobanov: Why we need the comment if we have write-host below

images/win/scripts/Installers/Install-Msys2.ps1
```diff
@@ -0,0 +1,80 @@
...
+$env:PATH = "$env:gitPath\mingw64\bin;$env:origPath"
+$tar      = "$env:gitPath\usr\bin\tar.exe"
+
+# extract tar.xz to C:\
+Write-Host "Starting msys2 extraction"
+&$tar -Jxf $msys2FileU -C /c/
+Remove-Item $msys2File
+Write-Host "Finished extraction"
+
+$env:PATH = "C:\msys64\mingw64\bin;C:\msys64\usr\bin;$env:origPath"
```
maxim-lobanov: I see that we modify PATH a lot in this script. Could you please add a few comments to explain why we need it?

images/win/scripts/Installers/Validate-Msys2.ps1
```diff
@@ -0,0 +1,57 @@
...
+    Write-Host "g++ is successfully installed:"
+    g++ --version | Write-Host
+}
+else
+{
+    Write-Host "g++ is not on PATH"
+    exit 1
+}
+
+if (Get-Command -Name 'make')
```
maxim-lobanov: Will this PR affect default tools somehow, like `tar` and etc. 
As I remember, @miketimofeev had some problems with it previously.

images/win/scripts/Installers/Install-Msys2.ps1
vsafonkin: C:\ => C\

images/win/scripts/Installers/Install-Msys2.ps1
```diff
@@ -0,0 +1,80 @@
+################################################################################
+##  File:  Install-Msys2.ps1
+##  Desc:  Install Msys2 and 64-bit gcc, cmake, & llvm (32-bit commented out)
+################################################################################
+
+# References
+# https://github.com/msys2/MINGW-packages/blob/master/azure-pipelines.yml
+# https://packages.msys2.org/group/
+
+$env:origPath = $env:PATH
```
al-cheb: I think it's not necessary to use $env at all: 
$env:origPath => $origPath 
$env:gitPath = $gitPath

images/win/scripts/Installers/Install-Msys2.ps1
al-cheb: nitpick: Join-Path $env:Temp "msys2.tar.xz"

images/win/scripts/Installers/Install-Msys2.ps1
```diff
@@ -0,0 +1,80 @@
...
+&$tar -Jxf $msys2FileU -C /c/
+Remove-Item $msys2File
+Write-Host "Finished extraction"
+
+$env:PATH = "C:\msys64\mingw64\bin;C:\msys64\usr\bin;$env:origPath"
+
+$ErrorActionPreference = "Continue"
+
+Write-Host "bash -c pacman-key --init"
+Invoke-Expression "bash -c `"pacman-key --init 2>&1`""
```
al-cheb: Could we remove `Invoke-Expression "` or script failed without it?

images/win/scripts/Installers/Validate-Msys2.ps1
miketimofeev: We should definitely check all the utilities like tar, perl, and openssl. All of them are supposed to be not from msys or mingw.

images/win/scripts/Installers/Install-Msys2.ps1
vsafonkin: Done

images/win/scripts/Installers/Install-Msys2.ps1
vsafonkin: Removed it.

MSP-Greg:
Re gcc

MSYS2/MinGW gcc output is similar to:
```
gcc.exe (Rev1, Built by MSYS2 project) 9.3.0
Copyright (C) 2019 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

As to Path, as mentioned elsewhere, the Windows Image's Path has enough conflicts already.  No need to add the MSYS2/MinGW folders permanently to Path, but they should be in place for installation and checks/validation.

images/win/scripts/Installers/Validate-Msys2.ps1
MSP-Greg: Re path issues/conflicts, see:
https://github.com/MSP-Greg/actions-image-testing/runs/524240890?check_suite_focus=true#step:11:4

vsafonkin:
Hi @MSP-Greg , we will not add MSYS2 folders to PATH permanently, but we need to validate MSYS2 installation and check some installed packages. Which packages are most important for you?

MSP-Greg:
@vsafonkin

Thanks to all for the work on this.  As to validation, make, gcc, and tar would check both usr and mingw64 folders.  Some of the cmake exe's might be good to add, but I generally use make...

MSP-Greg:
Reminder - once MSYS2 is added, there is no longer a need for the embedded MSYS2 installs in the four Ruby builds in C:\hostedtoolcache\windows\Ruby.  I haven't checked for a while, but each one had an embedded MSYS2.

If you check a release of Windows Ruby (https://github.com/oneclick/rubyinstaller2/releases/tag/RubyInstaller-2.7.0-1), the base releases are ~ 10 MB, whereas the release with an included DevKit/MSYS2 install is ~ 130 MB...

If one is using the normal method of 'connecting' Ruby with MSYS2 (by requiring `devkit`), there will be no change to CI...

vsafonkin:
@MSP-Greg , thanks for your point about embedded msys! We will investigate the possibility of getting rid of it because we're building those Ruby versions on the same windows images so it looks like we need msys2 preinstalled firstly.

MSP-Greg:
@vsafonkin

**Off-topic, kind of Ruby specific**

I don't know, as I think the Ruby build/install code is in the super top secret vault area.  JFYI, the current script for installing MSYS2 does not contain all the dependencies needed to build Ruby.

I thought keeping it a small as possible was a concern.  I also thought that once one starts including packages beyond build tools, it's difficult to decide what should be included.  AppVeyor did, and they've got a lot of packages (see https://ci.appveyor.com/project/MSP-Greg/appveyor-ruby).

As to the pre-installed Rubies, they are available in GitHub releases as exe's or 7z files (without devkit/MSYS2); it might be easier to use those.  After all, that is what normal users would install...

One thing I did see, below (line 75) is updating RubyGems for the system default Ruby (currently 2.5.x).  You might consider doing so for all installed Rubies, or none.  To update for just one may be confusing?

https://github.com/actions/virtual-environments/blob/cd130bd7047b24f05cf09002908ed20e852ec087/images/win/scripts/Installers/Download-ToolCache.ps1#L62-L76

vsafonkin:
@MSP-Greg , we are going to move ruby build/install code to opensource soon.

> As to the pre-installed Rubies, they are available in GitHub releases as exe's or 7z files (without devkit/MSYS2); it might be easier to use those. After all, that is what normal users would install...

Could you please clarify what GitHub releases do you mean?

MSP-Greg:
@vsafonkin 

> Could you please clarify what GitHub releases do you mean?

Sorry. See the releases/tags listed at https://github.com/oneclick/rubyinstaller2/tags.  There are multiple files per release, both 64 and 32 bit, and with and without devkit/msys2.  For example, the 2.7.0 release without devkit/msys2 is:

https://github.com/oneclick/rubyinstaller2/releases/download/RubyInstaller-2.7.0-1/rubyinstaller-2.7.0-1-x64.7z

Other than the root folder, it's everything needed.

images/win/scripts/Installers/Install-Msys2.ps1
evandrocoan: I would say it is because there are about 3 different `bash`es on the system path. So, putting ours bash on top of the system path assure us running our bash binary, not someone else's. Running the wrong bash binary will probably install things on the wrong Unix Environment and not in ours.

images/win/scripts/Installers/Install-Msys2.ps1
evandrocoan: Good point. Some times I lost myself doing the same. But when I realize it, all comments become `write-host`, so the `write-host` itself becomes the "comment".

images/win/scripts/Installers/Install-Msys2.ps1
```diff
@@ -0,0 +1,76 @@
...
+##  Desc:  Install Msys2 and 64-bit gcc, cmake, & llvm (32-bit commented out)
+################################################################################
+
+# References
+# https://github.com/msys2/MINGW-packages/blob/master/azure-pipelines.yml
+# https://packages.msys2.org/group/
+
+$origPath = $env:PATH
+$gitPath  = "C:\Program Files\Git"
+
```
evandrocoan: Perhaps use `$env:ProgramFiles` instead of `C:\Program Files`?

vsafonkin: Done, thank you!

vsafonkin:
@MSP-Greg , I've got it, thank you!

Issue #587, opened 2020-03-20
Update/Add Composer to Windows-latest (2019)


MIchaelMainer:
**Tool information**
- Tool name: Composer
- Add or update? Add
- Desired version: Latest
- Approximate size: 
- If this is an add request:
  - Brief description of tool: Composer is the dependency manager for PHP
  - URL for tool's homepage: https://getcomposer.org/doc/00-intro.md#installation-windows

**Area for Triage**:
PHP

**Question, Bug, or Feature?**:
Adding Composer is a feature, fixing the configuration is a buug. PHP is not so useful without Composer so let's consider it a feature.

**Virtual environments affected**
- [ ] macOS 10.15
- [ ] Ubuntu 16.04 LTS
- [ ] Ubuntu 18.04 LTS
- [ ] Windows Server 2016 R2
- [x] Windows Server 2019

**Can this tool be installed during the build?**
Yes, but there is a configuration that prevents it as the openssl extension is not enabled. [See this post for more information](https://superuser.com/questions/1010080/the-openssl-extension-is-missing-on-windows-while-installing-composer). When running the commands below, 

      curl -s https://getcomposer.org/installer -o "$env:PHPROOT/composer-setup.php"
      php "$env:PHPROOT/composer-setup.php" --install-dir=$env:PHPROOT --filename=composer
      php -r "unlink('composer-setup.php');"

I get the following message:
*The openssl extension is missing, which means that secure HTTPS transfers are impossible.*
*If possible you should enable it or recompile php with --with-openssl*

**Update** Yes, it can be installed during the build. But really, Composer should be in the image.
      $installDir = "c:\tools\php"
      choco install composer --ia "/DEV=$installDir /PHP=$installDir"

**Are you willing to submit a PR?**
You betcha!

maxim-lobanov:
Thank you for contribution! I have run CI builds in your PR for validation

maxim-lobanov:
Changes were merged #588 
Waiting for VM Rollouts

maxim-lobanov:
Changes were delivered and available on Windows 2019 image. Please let me know if you have any issues with it

PR #588, opened 2020-03-20
Update Install-PHP.ps1 to include Composer

MIchaelMainer:
1. Install Composer from Chocolatey. PHP needs its dependency manager.
2. Enable php_openssl.dll - believe this is necessary in the case someone wants to update Composer. 

**Reference**
https://chocolatey.org/packages/composer
https://github.com/actions/virtual-environments/issues/587
https://superuser.com/questions/1010080/the-openssl-extension-is-missing-on-windows-while-installing-composer


maxim-lobanov:
@MIchaelMainer , thank you for contribution! Changes looks good, but I think we should update `Validate-PHP.ps1` script too

PR #589, opened 2020-03-20
add Homebrew on Linux

dsame:
Install Homebrew package manager 
New Tool.

* Create installer script to install Homebrew according to https://docs.brew.sh/Homebrew-on-Linux
* Call the installer script from packer.io templates 
* Add `etc-environemnt.sh` with helper functions to manipulate /etc/environment
* Reboot VM
* Validate Homebrew is available after a reboot

**Total size (including related brew tap) 430MB. Installation time:  5 minutes**


images/linux/scripts/installers/homebrew.sh
```diff
@@ -0,0 +1,22 @@
...
+
+# Source the helpers
+source $HELPER_SCRIPTS/document.sh
+
+#  Install the Homebrew on Linux
+/bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install.sh)"
+
+# Validate the installation
+echo "Validate the installation"
+if ! command -v aws; then
```
al-cheb: I suppose should be brew instead of aws

images/linux/scripts/installers/homebrew.sh
```diff
@@ -0,0 +1,22 @@
...
+    echo "aws was not installed"
```
al-cheb: The same here

images/linux/ubuntu1604.json
```diff
@@ -186,7 +186,8 @@
                 "{{template_dir}}/scripts/installers/vcpkg.sh",
                 "{{template_dir}}/scripts/installers/zeit-now.sh",
                 "{{template_dir}}/scripts/installers/updatepath.sh",
-                "{{template_dir}}/scripts/installers/dpkg-config.sh"
+                "{{template_dir}}/scripts/installers/dpkg-config.sh",
+                "{{template_dir}}/scripts/installers/homebrew.sh"
```
maxim-lobanov: I think we should add homebrew installation at the begin of this list because potentially it can be used to install other software

images/linux/scripts/installers/homebrew-doc.sh
```diff
@@ -0,0 +1,12 @@
...
+##  File:  homebrew-doc.sh
+##  Desc:  Document the installation of Homebrew on Linux
+################################################################################
+
+# Source the helpers
+source $HELPER_SCRIPTS/document.sh
+
+# Document the installed version
+echo "Document the installed version"
+DocumentInstalledItem "$(cat /tmp/homebrew.txt)"
```
maxim-lobanov: Why we need a separate script for documentation?

dsame:
16.04 build https://github.visualstudio.com/virtual-environments/_build/results?buildId=63089&view=results
18.04 build https://github.visualstudio.com/virtual-environments/_build/results?buildId=63090&view=results

images/linux/scripts/installers/homebrew-doc.sh
dsame: Because homebrew must be installed and run as `runner` but the file with meta information used for documentation is owned by `root` user.

maxim-lobanov: Is it possible to use the single script and just run a few command using the other user? Probably, it will be more convenient than split installation to a few parts because we don't use this approach for other tools

images/linux/scripts/installers/homebrew.sh
```diff
@@ -0,0 +1,24 @@
...
+##  File:  homebrew.sh
+##  Desc:  Installs the Homebrew on Linux
+################################################################################
+
+# Source the helpers
+source $HELPER_SCRIPTS/document.sh
+
+#  Install the Homebrew on Linux
+/bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install.sh)"
+echo 'eval $(/home/linuxbrew/.linuxbrew/bin/brew shellenv)' >> ~/.profile
```
maxim-lobanov: `~/.profile` will be removed after image generation because user will be recreated. Will it work as you expected?

images/linux/scripts/installers/homebrew-doc.sh
dsame: https://github.com/actions/virtual-environments/pull/589/commits/cdb7f4b0cf28281c6253fe5e1d48876a3c451ac0

images/linux/scripts/installers/homebrew.sh
dsame: moved to /etc/profile.d https://github.com/actions/virtual-environments/pull/589/commits/358c29ef00d560b29d44adeff502bf6c56b52da3

images/linux/scripts/helpers/document.sh
```diff
@@ -9,7 +9,7 @@ function WriteItem {
         echo "METADATA_FILE environment variable must be set to output to Metadata Document!"
         return 1;
     else
-        echo -e "$1" >> "$METADATA_FILE"
+        sudo /bin/sh -c "echo -e \"$1\" >> \"$METADATA_FILE\""
```
maxim-lobanov: Will `echo -e "$1" | sudo tee -a $METADATA_FILE` is more readable?

images/linux/scripts/installers/homebrew.sh
```diff
@@ -0,0 +1,23 @@
...
+##  File:  homebrew.sh
+##  Desc:  Installs the Homebrew on Linux
+################################################################################
+
+# Source the helpers
+source $HELPER_SCRIPTS/document.sh
+
+#  Install the Homebrew on Linux
+/bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install.sh)"
+echo 'eval $(/home/linuxbrew/.linuxbrew/bin/brew shellenv)' | sudo tee /etc/profile.d/brew.sh
```
al-cheb: This approach will not work for us. You should get environment variables from command `brew shellenv` and put them to the `/etc/environment` file.

```
HOMEBREW_PREFIX="/home/linuxbrew/.linuxbrew"
HOMEBREW_CELLAR="/home/linuxbrew/.linuxbrew/Cellar"
HOMEBREW_REPOSITORY="/home/linuxbrew/.linuxbrew/Homebrew"
```
To append to the PATH env variable, please edit `https://github.com/actions/virtual-environments/blob/master/images/linux/scripts/installers/updatepath.sh` script.

```PATH="/home/linuxbrew/.linuxbrew/bin:/home/linuxbrew/.linuxbrew/sbin${PATH+:$PATH}"```

Comment (maxim-lobanov) [reply images/linux/scripts/installers/homebrew.sh]: Agree, in general.
But I think we should not use `updatepath.sh` file and should update /etc/environments in homebrew.sh. I think everything related to brew should be in the single place

Comment (dsame) [reply images/linux/scripts/installers/homebrew.sh]: https://github.com/actions/virtual-environments/pull/589/commits/37f9820527b9c0f3de0e40d875b3c2a3580f8ac2

Comment (dsame) [reply images/linux/scripts/installers/homebrew.sh]: 16.04 build https://github.visualstudio.com/virtual-environments/_build/results?buildId=63719&view=logs&j=011e1ec8-6569-5e69-4f06-baf193d1351e&t=5431112d-2b61-5a5f-7042-ef698f761043
tested on VM https://ms.portal.azure.com/#@microsoft.onmicrosoft.com/resource/subscriptions/417cd125-8a26-41ac-8dcf-30b6d4dbbfed/resourceGroups/github-imagegeneration/providers/Microsoft.Compute/virtualMachines/vsedoli2/overview
for the user `testAdm2` and  another user created from scratch without any additional permissions `u1`

**Comment by AlenaSviridenko** [comment]
@dsame could you please also provide tool size and installation time?

File: images/linux/scripts/helpers/etc-environment.sh [hunk 8]
```diff
@@ -0,0 +1,65 @@
...
+    else
+        addEtcEnvironmentVar $var_name $var_value
+    fi
+}
+
+function addEtcEnvironmentPathElement {
+    element="$1"
+    etc_path=$(getEtcEnvironmentVar PATH)
+    setEtcEnvironmentVar PATH "${element}:${etc_path}"
+}
```
maxim-lobanov, line 47: I would suggest to build function `PrependEtcEnvironmentVariable` and `AppendEtcEnvironmentVariable`
```
etc_value=$(getEtcEnvironmentVar $1)
setEtcEnvironmentVar $1 "${2}:${etc_value}"
```

It is okay to have one more function `prependEtcEnvironmentPath` if you prefer

images/linux/scripts/helpers/etc-environment.sh
```diff
@@ -0,0 +1,65 @@
...
+
+function replaceEtcEnvironmentVar {
+    var_name="$1"
+    var_value="$2"
+
+    # modify /etc/environemnt in place by replacing a string that begins with var_name
+    sudo sed -ie "s%^${var_name}=.*$%${var_name}=\"${var_value}\"%" /etc/environment
+}
+
+function setEtcEnvironmentVar {
```
maxim-lobanov, line 32: `setEtcEnvironmentVariable` - let's use the full `variable` word for readability
The same for other function

images/linux/ubuntu1604.json
```diff
@@ -132,6 +132,34 @@
...
+            "pause_before": "30s",
+            "timeout": "10m",
+            "start_retry_timeout": "10s",
+            "scripts": [
+                "{{template_dir}}/scripts/installers/homebrew-validate.sh"
+            ],
+            "environment_vars": [
+                "METADATA_FILE={{user `metadata_file`}}",
+                "HELPER_SCRIPTS={{user `helper_script_folder`}}",
+                "DEBIAN_FRONTEND=noninteractive"
```
maxim-lobanov, line 159: I think we don't need these variables for validation script

images/linux/scripts/installers/homebrew.sh
```diff
@@ -0,0 +1,40 @@
...
+    exit 1
+fi
+
+# Document the installed version
+echo "Document the installed version"
+DocumentInstalledItem "Homebrew on Linux ($(brew -v 2>&1))"
+
+# Reboot VM in order to validate brew can be run after it
+echo "Reboot VM"
+sudo reboot
```
maxim-lobanov, line 40: Do we wanto to move reboot step as a separate step to the templates? So it will be clearer when machine is rebooted

images/linux/scripts/helpers/etc-environment.sh
dsame: The changes (and related to it) are implemented in the another PR: 
https://github.com/actions/virtual-environments/pull/660/commits/ba017c929014c482545acb2eb5a3fdaa4a71ddce
https://github.com/actions/virtual-environments/pull/660/commits/144559b5f252ca47b23345fcef5c27eff1b39537
https://github.com/actions/virtual-environments/pull/660/commits/a2517ced678009aba36207d0c61d740f7f5c35f0
https://github.com/actions/virtual-environments/pull/660/commits/3ebe6154b26c8807ab82b109bd3a85e0dc407b6a

images/linux/ubuntu1604.json
```diff
@@ -132,6 +132,36 @@
...
+            ],
+            "execute_command": "/bin/sh -c '{{ .Vars }} {{ .Path }}'"
+        },
+        {
+            "type": "shell",
+            "expect_disconnect": true,
+            "scripts": [
+                "{{template_dir}}/scripts/base/reboot.sh"
```
alepauly, line 151: why do we need this reboot specifically? is it just to make sure? It would be nice to avoid it since that adds a considerable amount of time to the build.

maxim-lobanov: @alepauly , Unfortunately, reboot is necessary to apply brew installation :(

rbellamy:
I realize I'm bumping an already-merged discussion, but I have to ask - why the reboot?

I'm trying to track the work in this repository to build Docker-hosted github runners that have the same environment and tools as the images created in this repository, and a mid-provision reboot seems like a really drastic requirement for installing this tool.

images/linux/scripts/installers/homebrew-validate.sh
```diff
@@ -0,0 +1,14 @@
...
+################################################################################
+##  File:  homebrew-validate.sh
+##  Desc:  Validate the Homebrew can run after reboot without extra configuring
+################################################################################
+
+# Validate the installation
+echo "Validate the Homebrew can run after reboot"
+
+if ! command -v brew; then
+    echo "brew cat not run after reboot"
```
sjackman, line 11: Typo here. `cat not` should be `cannot`

sjackman:
A reboot isn't necessary to install Homebrew. I suspect it's to set up the environment, by
```sh
eval $(/home/linuxbrew/.linuxbrew/bin/brew shellenv)
```

rbellamy:
Yes, that seems likely. And if so, my question seems even more relevant - surely there's a more graceful way to address that stanza without having to result to the "nuclear" option of rebooting in the middle of provisioning?

dsame:
> surely there's a graceful way to address that stanza without having to result to the "nuclear" option of rebooting in the middle of provisioning?

The problem is to set up the environment in `/etc/environment` in order to let the brew running within shells started with `-norc` option which is the case for runners tasks. 

Also the brew is used during the provision of images (AWS SAM CLI for example) and issuing `eval $(/home/linuxbrew/.linuxbrew/bin/brew shellenv)` inside many different tasks becomes cumbersome and error prone. 

While the reboot takes ~20 secs it was decided this is the most optimal way to ensure `brew` is  available in the next provision tasks and uses `/etc/environment` exactly as it will do in runner tasks.

dsame:
> A reboot isn't necessary to install Homebrew. I suspect it's to set up the environment, by
> 
> ```shell
> eval $(/home/linuxbrew/.linuxbrew/bin/brew shellenv)
> ```

Yes, and it is used this way in the installation script
https://github.com/dsame/virtual-environments/blob/16c6c787c598e7df12ea12fdd7e79d4596ee9265/images/linux/scripts/installers/homebrew.sh#L13

But next it is necessary to make `brew` available for the runners task which are started within `--norc` environment and thus having no use of "profile" files.

This is why the `brew shellenv` output is duplicated to `/etc/environment` and in order to validate `brew` within no-profile environment (see https://github.com/dsame/virtual-environments/blob/16c6c787c598e7df12ea12fdd7e79d4596ee9265/images/linux/scripts/installers/homebrew-validate.sh) it is desirable to have reboot step between installation and validation. 

As a bonus it is possible to use brew in the next provision steps without issuing `eval ...`

rbellamy:
>The problem is to set up the environment in /etc/environment in order to let the brew running within shells started with -norc option which is the case for runners tasks.

I think that is what confuses me, as the runner is deliberately NOT sourcing `/etc/environment` - https://github.com/actions/runner/issues/455.

>Also the brew is used during the provision of images (AWS SAM CLI for example) and issuing eval $(/home/linuxbrew/.linuxbrew/bin/brew shellenv) inside many different tasks becomes cumbersome and error prone.

That makes sense. Thank you very much for the explanation!

rbellamy:
Gonna put this here just in case someone lands here looking to understand how to build a VM from this repo: both `git` and `ruby` are required. Neither of which are provisioned BEFORE this provisioning script is run. For Docker, this means you'll have to install both explicitly for this script to run.

dsame:
> I think that is what confuses me, as the runner is deliberately NOT sourcing `/etc/environment` - [actions/runner#455](https://github.com/actions/runner/issues/455).
> 

Hm, now it seems something confuses me :)

First, it is not possible to do not source /etc/environment as it is read by any PAM enabled (and configured to read it) process/library and there's definitely at least one of such run on linux startup/connection.

Also it must be said "--noprofile --norc" has no connection to `/etc/environment` at all. "--noprofile --norc " are the shell options while `/etc/environment` is the confing file for PAM system. 

Hm.. If the https://github.com/actions/runner/issues/455 is still actual, please reopen it and ping me, so i'll be able to take a close look. I am not sure `/etc/environment` is actually read at the moment of parsing YAML file, but i can hardly imagine the situation it is not.

rbellamy:
@dsame thanks for sticking with me on this. I think I misspoke - it's not that the runner doesn't load `/etc/environment` since, as you say, PAM is enabled and configured to read it. The runner's environment IS loading the `/etc/environment`. https://github.com/actions/runner/issues/455 was more about the fact that the `${{ env.* }}` defaults aren't pre-populated with the values from `/etc/environment`.

Issue #20, opened 2019-10-06
Add MongoDB Community Edition


Ignigena:
**Tool information**
- Tool name: mongo and mongod
- Add or update? Add
- Desired version: 3.x, 4.x
- Approximate size: 250-300MB
- If this is an add request:
  - Brief description of tool: NoSQL database server and CLI tool
  - URL for tool's homepage: https://github.com/mongodb/mongo

**Virtual environments affected**
- [x] macOS 10.14
- [x] Ubuntu 16.04 LTS
- [x] Ubuntu 18.04 LTS
- [x] Windows Server 2016 R2
- [x] Windows Server 2019

**Can this tool be installed during the build?**

**Are you willing to submit a PR?**

maxim-lobanov:
Tool can be installed in runtime using the following script
```
# Ubuntu
sudo apt install -y mongodb
# macOS
brew tap mongodb/brew
brew install mongodb-community
# Windows
choco install mongodb
```
Installation takes about 1 minute.

dsame:
PR https://github.com/actions/virtual-environments/pull/591

dsame:
Ubuntu PR https://github.com/actions/virtual-environments/pull/591

dsame:
The changes are merged and will be in the production  on the next deploy

PR #591, opened 2020-03-20
Add MongoDB 4.2 to linux images

dsame:
* Create installed scripts according to https://docs.mongodb.com/manual/tutorial/install-mongodb-on-ubuntu/
* Call the installed scripts from packer.io templates

The related issue https://github.com/actions/virtual-environments/issues/20

Time to install: <10 secs
Size on disk: nearly 380mb

The Ubuntu packages are not used because they install the obsolete 2.6 version of MongoDB


AlenaSviridenko:
Looks like mongoDB installation scripts are the same for both ubuntu versions. If that's correct, could you please move it to common installers folder if possible? 
Or please let me know if there are any reasons why we need 2 different files.

maxim-lobanov:
> Looks like mongoDB installation scripts are the same for both ubuntu versions. If that's correct, could you please move it to common installers folder if possible?
> Or please let me know if there are any reasons why we need 2 different files.

Looks like line 12 is different. I would suggest to add helper functions to determine OS ([like these ones](https://github.com/actions/virtual-environments/blob/master/images/win/scripts/ImageHelpers/InstallHelpers.ps1#L305)) and just add single script with branching.

AlenaSviridenko:
@maxim-lobanov oh, I see, `ubuntu xenial` vs `ubuntu bionic`. Agree with refactoring based on OS version

images/linux/scripts/installers/mongodb.sh
```diff
@@ -0,0 +1,26 @@
...
+##  File:  homebrew.sh
+##  Desc:  Installs Mongo DB
+################################################################################
+
+# Source the helpers
+source $HELPER_SCRIPTS/document.sh
+
+#  Install Mongo DB
+wget -qO - https://www.mongodb.org/static/pgp/server-4.2.asc | sudo apt-key add -
+version=$(lsb_release -d|grep 'Ubuntu 16' > /dev/null  && echo xenial || echo bionic)
```
maxim-lobanov: Can we add helper function like [this one](https://github.com/actions/virtual-environments/blob/master/images/win/scripts/ImageHelpers/InstallHelpers.ps1#L305) for Ubuntu

dsame:
build 1604 https://github.visualstudio.com/virtual-environments/_build/results?buildId=63294&view=results
build 1804 https://github.visualstudio.com/virtual-environments/_build/results?buildId=63412&view=results

images/linux/scripts/helpers/install.sh
```diff
@@ -0,0 +1,15 @@
...
+
+function TestIsXenial
+{
+    lsb_release -d|grep 'Ubuntu 16' > /dev/null
+}
+
+function TestIsBionic
+{
+    lsb_release -d|grep 'Ubuntu 18' > /dev/null
+}
```
maxim-lobanov: I think it is better to have the following functions:
1) `isUbuntu16` (rename from `TestIsXenial`)
2) `isUbuntu18` (rename from `TestIsBionic`)
3) `getOSVersionLabel` that will return `xenial` or `bionic`

images/linux/scripts/helpers/install.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+################################################################################
+##  File:  install-helpers.sh
```
sergey-akhalkov: nitpick: "install.sh" is not clear file name and I cannot understand what I can expect there, I would suggest to rename file to "os.sh" if it makes sense to you.

alepauly: @dsame I agree with @sergey-akhalkov, the script name is not clear.

Issue #94, opened 2019-11-15
Add AWS SAM CLI


MichalMiko:
**Tool information**
- Tool name: AWS SAM CLI
- Add or update? Add
- Desired version: latest
- Approximate size: 500kB
- If this is an add request:
  - Brief description of tool: AWS SAM provides you with a command line tool, the AWS SAM CLI, that makes it easy for you to create and manage serverless applications
  - URL for tool's homepage: https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/serverless-sam-cli-install.html

**Virtual environments affected**
- [x] macOS 10.15
- [x] Ubuntu 16.04 LTS
- [x] Ubuntu 18.04 LTS
- [ ] Windows Server 2016 R2
- [ ] Windows Server 2019

**Can this tool be installed during the build?**
Yes, using:
`$ pip install aws-sam-cli`

But it can take some time, it requires: Flask, boto3, aws-lambda-builders, chevron, aws-sam-translator, pywin32, docker, cookiecutter, PyYAML, click, serverlessrepo, jmespath, requests, python-dateutil, dateparser 

**Are you willing to submit a PR?**
Of course ;)

maxim-lobanov:
The fastest way to install AWS SAM CLI in runtime:
```
# macOS
brew tap aws/tap
brew install aws-sam-cli
# Ubuntu
pip install aws-sam-cli
```
Installation takes about 20 seconds

alepauly:
We should also install it on Windows images if it makes sense.

alnaranjo:
+1

dsame:
Linux PR https://github.com/actions/virtual-environments/pull/615

AlenaSviridenko:
Images with AWS SAM CLI were deployed, closing this issue. 
Please, feel free to contact up if needed.

PR #615, opened 2020-03-25
Install AWS SAM CLI

dsame:
# Description

Install AWS SAM CLI on Ubuntu images
New Tool.

Adds _sam_ utility according to the installation instructions https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/serverless-sam-cli-install-mac.html

This PR does not include adding the AWS SAM docker container.

**Total size (including related brew tap) 650Kb. Installation time: between 5 and 10 minutes**

#### Related issue:
https://github.com/actions/virtual-environments/issues/94

## Check list

### DIFF
--- a/images/linux/scripts/installers/aws-sam-cli.sh
+++ b/images/linux/scripts/installers/aws-sam-cli.sh
@@ -1,0 +1,24 @@
+#!/bin/bash
+################################################################################
+##  File:  aws-sam-cli.sh
+##  Desc:  Installs AWS SAM CLI
+##         Must be run as non-root user after homebrew and clang
+################################################################################
+
+# Source the helpers for use with the script
+source $HELPER_SCRIPTS/document.sh
+
+# Install aws sam cli
+brew tap aws/tap
+brew install aws-sam-cli
+
+# Run tests to determine that the software installed as expected
+echo "Testing to make sure that script performed as expected, and basic scenarios work"
+if ! sam --version; then
+    echo "AWS SAM CLI was not installed"
+    exit 1
+fi
+
+# Document what was added to the image
+echo "Lastly, documenting what we added to the metadata file"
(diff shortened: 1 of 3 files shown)


images/linux/scripts/installers/aws-sam-cli.sh
```diff
@@ -0,0 +1,23 @@
...
+##  File:  aws-sam-cli.sh
+##  Desc:  Installs AWS SAM CLI
+################################################################################
+
+# Source the helpers for use with the script
+source $HELPER_SCRIPTS/document.sh
+
+# Install aws sam cli
+brew tap aws/tap
+brew install aws-sam-cli
```
maxim-lobanov: Since installation of AWS-Sam-CLI is pretty small, I would suggest to add this logic to `aws.sh` file instead of creating the new one. So all tools related to AWS will be in the single place:
https://github.com/actions/virtual-environments/blob/master/images/linux/scripts/installers/aws.sh

6iix:
When is this going to be published? I know it was 4 days ago merged, just curious when I will be able to leverage this awesome work.

maxim-lobanov:
Hello @6iix , usually, it takes about 1 week to update all VM machines.
But, currently, we have an active incident related to disk space on Ubuntu images so it can take a bit more.

6iix:
@maxim-lobanov :) thanks for the update. I will be eagerly waiting. This is all awesome work, I am new to actions so I'm happy for the fast turnarounds.

chenrui333:
> Hello @6iix , usually, it takes about 1 week to update all VM machines.
> But, currently, we have an active incident related to disk space on Ubuntu images so it can take a bit more.

@maxim-lobanov thanks for the work, can we have a update once it got rolled out to all Ubuntu images? This is indeed an anticipated feature.

Issue #40, opened 2019-10-15
Update/Add Mysql 8


paulrrogers:
**Tool information**
- Tool name: Mysql
- Add or update? update
- Desired version: 8.0
- Approximate size: 
- If this is an add request:
  - Brief description of tool: RDBMS
  - URL for tool's homepage: https://dev.mysql.com/

**Virtual environments affected**
- [ ] macOS 10.14
- [x] Ubuntu 16.04 LTS
- [x] Ubuntu 18.04 LTS
- [ ] Windows Server 2016 R2
- [ ] Windows Server 2019

**Can this tool be installed during the build?**
yes, minutes

**Are you willing to submit a PR?**
maybe

maxim-lobanov:
@paulrrogers , currently, our images contain MySQL 5.7. Do you have acontext, will it be a breaking change for other customers who use MySQL 5.7 if we update it to MySQL 8?

paulrrogers:
@maxim-lobanov Mysql 8 has significant changes, though if BC is a concern it can be configured to behave similarly to the defaults in v5.7.

https://dev.mysql.com/doc/refman/8.0/en/upgrading-from-previous-series.html

Darleev:
Hello, @paulrrogers.
Mysql 8 update on Ubuntu 20 pull request has been merged. All changes will be available after the next image rollout.
I will let you know additionally when the change is applied.

Darleev:
Hello, @paulrrogers,
We have deployed MySQL 8 on Ubuntu 20 only due to potential impact and significant changes in latest MySQL version. So, MySQL 8 has already been delivered to Ubuntu 20, feel free to use it.

I'm going to close the issue.
Please let us know in case of any questions.

PR #621, opened 2020-03-25
mysql update to version 8

Darleev:
# Description
Update MySQL from version 5.7 to 8.
In scope of the [ticket](https://github.com/actions/virtual-environments/issues/40), it was requested to update mysql to version 8 in Ubuntu images

#### Related issue:
https://github.com/actions/virtual-environments/issues/40

## Check list
Test are in progress. Ubuntu images with MySQL 8 only were successfully deployed and tested.

### DIFF
--- a/images/linux/scripts/installers/mysql.sh
+++ b/images/linux/scripts/installers/mysql.sh
@@ -9,17 +9,14 @@ source $HELPER_SCRIPTS/document.sh
 
 export ACCEPT_EULA=Y
 
-# Install MySQL Client
-apt-get install mysql-client -y
-
-# InstallMySQL database development files
-apt-get install libmysqlclient-dev -y
-
 # Install MySQL Server
 MYSQL_ROOT_PASSWORD=root
 echo "mysql-server mysql-server/root_password password $MYSQL_ROOT_PASSWORD" | debconf-set-selections
 echo "mysql-server mysql-server/root_password_again password $MYSQL_ROOT_PASSWORD" | debconf-set-selections
-
+debconf-set-selections <<< 'mysql-apt-config mysql-apt-config/select-server select mysql-8.0'
+wget -c https://dev.mysql.com/get/mysql-apt-config_0.8.14-1_all.deb
+dpkg -i mysql-apt-config_0.8.14-1_all.deb
+apt update
 apt-get install -y mysql-server
 
 # Install MS SQL Server client tools (https://docs.microsoft.com/en-us/sql/linux/sql-server-linux-setup-tools?view=sql-server-2017)
@@ -46,5 +43,5 @@ DocumentInstalledItem "MySQL Server (user:root password:root)"
 DocumentInstalledItem "MS SQL Server Client Tools"
 
 # Disable mysql.service
-systemctl is-active --quiet mysql.service && systemctl stop mysql.service 
-systemctl disable mysql.service 
+systemctl is-active --quiet mysql.service && systemctl stop mysql.service
+systemctl disable mysql.service


images/linux/scripts/installers/mysql.sh
```diff
@@ -9,17 +9,14 @@ source $HELPER_SCRIPTS/document.sh
 
 export ACCEPT_EULA=Y
 
-# Install MySQL Client
-apt-get install mysql-client -y
-
-# InstallMySQL database development files
-apt-get install libmysqlclient-dev -y
```
al-cheb: Why do we delete libmysqlclient-dev?

images/linux/scripts/installers/mysql.sh
```diff
@@ -9,17 +9,14 @@ source $HELPER_SCRIPTS/document.sh
...
-
 # Install MySQL Server
 MYSQL_ROOT_PASSWORD=root
 echo "mysql-server mysql-server/root_password password $MYSQL_ROOT_PASSWORD" | debconf-set-selections
 echo "mysql-server mysql-server/root_password_again password $MYSQL_ROOT_PASSWORD" | debconf-set-selections
-
+debconf-set-selections <<< 'mysql-apt-config mysql-apt-config/select-server select mysql-8.0'
+wget -c https://dev.mysql.com/get/mysql-apt-config_0.8.14-1_all.deb
```
al-cheb: Why do we use hardcoded version? Please parse to the latest version- https://dev.mysql.com/downloads/repo/apt/

images/linux/scripts/installers/mysql.sh
Darleev: mysql-client will be installed in scope of mysql-server installation - it was verified. Thank you for the tip, libmysqlclient-dev definitely must be included into installation script. Fixed.

images/linux/scripts/installers/mysql.sh
Darleev: Thank you, you are absolutely right, hardcoded value is not good solution here. I have applied another one.

images/linux/scripts/installers/mysql.sh
```diff
@@ -9,19 +9,20 @@ source $HELPER_SCRIPTS/document.sh
...
-# InstallMySQL database development files
-apt-get install libmysqlclient-dev -y
-
 # Install MySQL Server
 MYSQL_ROOT_PASSWORD=root
 echo "mysql-server mysql-server/root_password password $MYSQL_ROOT_PASSWORD" | debconf-set-selections
 echo "mysql-server mysql-server/root_password_again password $MYSQL_ROOT_PASSWORD" | debconf-set-selections
-
+debconf-set-selections <<< 'mysql-apt-config mysql-apt-config/select-server select mysql-8.0'
+package_version=$(curl https://dev.mysql.com/downloads/repo/apt/ 2> /dev/null | grep ".deb" | awk -F"[()]" '{print $2}' | grep "\.deb")
```
al-cheb: Why do we need two greps?
`curl https://dev.mysql.com/downloads/repo/apt/ 2> /dev/null | grep "\.deb" | awk -F"[()]" '{print $2}'`

Darleev: thank you. resolved

images/linux/scripts/installers/mysql.sh
```diff
@@ -9,19 +9,30 @@ source $HELPER_SCRIPTS/document.sh
...
-# InstallMySQL database development files
-apt-get install libmysqlclient-dev -y
+if isUbuntu16 || isUbuntu18 ; then
+    apt-get install mysql-client -y
+fi
 
-# Install MySQL Server
-MYSQL_ROOT_PASSWORD=root
-echo "mysql-server mysql-server/root_password password $MYSQL_ROOT_PASSWORD" | debconf-set-selections
-echo "mysql-server mysql-server/root_password_again password $MYSQL_ROOT_PASSWORD" | debconf-set-selections
```
maxim-lobanov, line 21: Looks like before your changes these lines were invoked on Ubuntu 18 but after your changes that are not invoked. Is it expected?

Darleev: Thank you, you are absolutely right, I moved this part accidentally.

images/linux/scripts/installers/mysql.sh
```diff
@@ -9,19 +9,31 @@ source $HELPER_SCRIPTS/document.sh
...
+    apt-get install mysql-client -y
+fi
 
-# InstallMySQL database development files
-apt-get install libmysqlclient-dev -y
+if isUbuntu20 ; then
+    # Install mysql 8 for Ubuntu 20.
 
-# Install MySQL Server
+    debconf-set-selections <<< 'mysql-apt-config mysql-apt-config/select-server select mysql-8.0'
```
alepauly, line 19: is this code so the install call below will default to the 8 instead of 5?

images/linux/scripts/installers/mysql.sh
```diff
@@ -9,19 +9,31 @@ source $HELPER_SCRIPTS/document.sh
 
 export ACCEPT_EULA=Y
 
-# Install MySQL Client
-apt-get install mysql-client -y
+if isUbuntu16 || isUbuntu18 ; then
+    apt-get install mysql-client -y
```
alepauly, line 13: not available on 20.04 yet?

Darleev: I have tried to install mysql 8 on clean ubuntu VM and mysql-client was installed as a part of mysql-server package.

images/linux/scripts/installers/mysql.sh
Darleev: Yes, it is. By default package configuration contains mysql version 5, that is why we need to manually specify version 8 in the script for ubuntu 20.04.

PR #630, opened 2020-03-26
Revert "Install MSYS2 to Windows."

AlenaSviridenko:
Reverts actions/virtual-environments#585


evandrocoan:
Why reverting to it back on https://github.com/actions/virtual-environments/pull/632?

AlenaSviridenko:
Hi @evandrocoan !
Currently we are trying to minimize the possible impact on customers while pandemic is on, MSYS2 is  a pretty significant change and it will be deployed after the crysis is going down.
Thanks for your patience.

PR #631, opened 2020-03-26
Update terraform version to v0.12.24

chenrui333:
# Description

Just update the README to the latest v0.12.24.

Just ran into a issue as below.
image

## Check list


maxim-lobanov:
Hello, thank you for your contribution but we should not update documentation manually. it is updated automatically when the deployment has finished. Usually, deployment takes a few days so documentation can be a bit outdated.
I am closing this PR and the files will be updated soon

chenrui333:
Good to know, thanks!

PR #632, opened 2020-03-26
Install MSYS2 to Windows.

vsafonkin:
Issue: [#30](https://github.com/actions/virtual-environments/issues/30)

Added scripts for installation and validation MSYS2.

Installation time: 6 - 7 min
Size: ~4.2 Gb


images/win/scripts/Installers/Install-Msys2.ps1
```diff
@@ -0,0 +1,76 @@
...
+pacman.exe -Scc --noconfirm
+
+Write-Host "Installed mingw64 packages"
+pacman.exe -Qs --noconfirm mingw-w64-x86_64-
+
+Write-Host "Installed mingw32 packages"
+pacman.exe -Qs --noconfirm mingw-w64-i686-
+
+Write-Host "Installed msys2 packages"
+pacman.exe -Qs --noconfirm
```
MSP-Greg: Please consider changing lines 65 thru 72 to:
```ps1
Write-Host "Installed mingw64 packages"
pacman.exe -Q | grep ^mingw-w64-x86_64-

Write-Host "Installed mingw32 packages"
pacman.exe -Q | grep ^mingw-w64-i686-

Write-Host "Installed msys2 packages"
pacman.exe -Q | grep -v ^mingw-w64-
```

images/win/scripts/Installers/Install-Msys2.ps1
```diff
@@ -0,0 +1,76 @@
+################################################################################
+##  File:  Install-Msys2.ps1
+##  Desc:  Install Msys2 and 64-bit gcc, cmake, & llvm (32-bit commented out)
```
MSP-Greg: Please remove '(32-bit commented out)'

images/win/scripts/Installers/Install-Msys2.ps1
```diff
@@ -0,0 +1,76 @@
...
+# Add msys2 bin tools folders to PATH temporary
+$env:PATH = "C:\msys64\mingw64\bin;C:\msys64\usr\bin;$origPath"
+
+Write-Host "bash pacman-key --init"
+bash.exe -c "pacman-key --init 2>&1"
+
+Write-Host "bash pacman-key --populate msys2"
+bash.exe -c "pacman-key --populate msys2 2>&1"
+
+Write-Host "pacman --noconfirm -Syyuu"
```
eine, line 41: Ref https://github.com/actions/virtual-environments/pull/342#issuecomment-581086902

images/win/scripts/Installers/Install-Msys2.ps1
```diff
@@ -0,0 +1,76 @@
...
+pacman.exe -Syyuu --noconfirm
+pacman.exe -Syuu --noconfirm
```
eine, line 43: Please, see https://github.com/actions/virtual-environments/pull/342#issuecomment-580985905

images/win/scripts/Installers/Install-Msys2.ps1
miketimofeev: This PR is based on @MSP-Greg suggested one that's why it's here. I'd prefer to wait for image deployment and then we can collect some feedback about working scenarios and decide if this approach suits or not.

eine: Actually, I'm good with the merged solution, even though it constraints the possible use cases. As commented in previous discussions, users can keep using other custom MSYS2 installations through action setup-msys2. I'd just like it to be explicitly documented, so that users of GitHub Actions are aware that this is not a default MSYS2 install.

miketimofeev: Could you provide your vision on documentation, please? What points should be mentioned?

eine: Sure. For example:

> In order to reduce startup time for most common use cases, the MSYS2 installation provided in `windows` environments is installed from binaries, then updated, and some packages such as `base-devel`, `compression`, `clang` and `cmake` are preinstalled. Please open an issue in LINK for requesting additional packages to be preinstalled.
>
> As a result, users willing to use a stable environment should install MSYS2 themselves. See actions/virtual-environments#632 for a reference. Alternatively Action setup-msys2 can be used.

vsafonkin: Hi @eine, we will add information about non-default MSYS2 installation to docs, thank you for your participate!

images/win/scripts/Installers/Install-Msys2.ps1
```diff
@@ -0,0 +1,76 @@
...
+
+# References
+# https://github.com/msys2/MINGW-packages/blob/master/azure-pipelines.yml
+# https://packages.msys2.org/group/
+
+$origPath = $env:PATH
+$gitPath  = "$env:ProgramFiles\Git"
+
+# get info from https://sourceforge.net/projects/msys2/files/Base/x86_64/
+$msys2Uri  = "http://repo.msys2.org/distrib/x86_64/msys2-base-x86_64-20190524.tar.xz"
```
eine, line 14: A more recent release is available at https://github.com/msys2/msys2-installer/releases. Nightly releases are available too. It might be worth updating.

At the same time, since everything is updated anyway, it would be desirable to retrieve the non-base package.

eine:
Ref #928 #916 #906 #899 #632 #342

Issue #507, opened 2020-03-05
Reproducing the Windows 2019 image


MarcDenman:
**Describe the bug**
For various reasons we are generating our own copy of the `Windows Server 2019` image. We have found the process to be very flaky and our build of the image often fails at seemingly random points (20 minutes, 50 minutes, several hours in). A rerun of the build will go past the point of the previous failure with no changes needed.

Are you experiencing this behaviour as well or does it run well for you with no random failures? Our current best approach is running several isolated packer builds simultaneously and hoping one finishes with no errors! 

**Area for Triage**: 
Deployment/Release

**Question, Bug, or Feature?**: 
Question

**Virtual environments affected**
- [ ] macOS 10.15
- [ ] Ubuntu 16.04 LTS
- [ ] Ubuntu 18.04 LTS
- [ ] Windows Server 2016 R2
- [x] Windows Server 2019

**Expected behavior**
Windows 2019 packer build runs correctly on a repeatable basis

**Actual behavior**
Random failures whilst building the Windows 2019 image.

alepauly:
Hi @MarcDenman, thanks for reaching out! What's the rate of flakiness that you are seeing and is there any specific type of error you are seeing? We do see occasional (~1/6th of builds) failures mostly due to network issues or third-party package sources not responding. Also, are you running the builds off a cloud VM, a local box, a server? What type of network connection does it have?

jmos5156:
In short yes. I've experienced the same and more. There are several issues already discussed in this repo from WinRM timeout, download failures, validation checks not working, to installs not completing. I'd say it's around 1/4 of all builds I do.

There are a couple of suggestions of issues with Azure but even so, this repo offers you the full bells and whistles Windows 2019 agent build which just seems to flaky to run for several hours, at cost, for it to fail. That's too much investment IMHO.

I've run this in most European Azure locations with varying levels of success.
Firstly I have trimmed down the vast numbers of installs to only what I need. I don't need several python versions, for example. I carefully review the validation steps as it annoys me that the process can fail simply because of an error in the validation process. I don't like this as it has branched me away from this repo and would prefer to remain closer to the changes seen here.

Also, watch out for dependencies between packages. This has caught me out several times.
I have resorted to running several builds each tackling a single install just to test the code. If these work individually, I will attempt a full build, but there's no guarantee it will work.

alepauly:
> I carefully review the validation steps as it annoys me that the process can fail simply because of an error in the validation process.

[...]

PR #646, opened 2020-03-31
Chocolatey Helper

elduddz:
# Description
This is looking at the issues we have been having building a successful image in UK South. #507 

This is a mix of an improvement and a change to start looking at behaviours around protecting the process from intermittent network outages, now granted this process would need to be applied all over the place, but I think this will aid the success rate of completions in places other than CUS.

The issue I see is that when an error occurs no matter how far down the line, the packer build asks you what to do or gives up, if you have gone off to do something else, 5 hours later you find the build it stuck waiting for you to abort or retry, this isnt an option for a pipeline, you only get the option of ask, which pauses indefinately or cleanup, there isnt a retry.

We are trying to automate this process, so having the build ask for what you want to do isnt really an option, so we are defaulting to destroy everything. I guess if packer had an auto retry (or at least I havent found the option)

## caveats:
I am only reviewing the win19 build as that is the one I need.

This is not a complete coverage, I have only applied fixes on the steps where the build has dropped out, I want to know if I am on the right track and there are still elements outside of this project that are pulled in.

I also noticed that packer has a little message saying vhd will be deprecated, so I have updated the json file to use managed disks, as I am putting this in a Shared image gallery, this saves a step.

#### Related issue: 
#507

## Check list
   - this still builds in CUS, as I still cannot get a build in UKS, but this does get the process further down the line, but as the network download / outage is so intermittent, one package that might have been ok on one pass now fails.


elduddz:
an alternative is a request I have raised with hashicorp:
https://github.com/hashicorp/packer/issues/8972#issuecomment-606459263

maybe we can get some likes and pleases?

images/win/scripts/ImageHelpers/ImageHelpers.psm1
```diff
@@ -4,6 +4,7 @@ param()
 . $PSScriptRoot\PathHelpers.ps1
 . $PSScriptRoot\InstallHelpers.ps1
 . $PSScriptRoot\MarkdownHelpers.ps1
+. $PSScriptRoot\InstallChoco.ps1
```
maxim-lobanov, line 7: I suggest to rename it to `ChocoHelpers.ps1` so this file can be reused for more choco-related helpers in future

images/win/scripts/Installers/Install-PowershellCore.ps1
```diff
@@ -2,5 +2,4 @@
 ##  File:  Install-PowershellCore.ps1
 ##  Desc:  Install PowerShell Core
 ################################################################################
-
-Invoke-Expression "& { $(Invoke-RestMethod https://aka.ms/install-powershell.ps1) } -UseMSI -Quiet"
+Install-Choco "powershell-core"
```
maxim-lobanov, line 5: What is the issue with installation from `https://aka.ms/install-powershell.ps1`. I didn't notice any issues with it previously.

images/win/scripts/ImageHelpers/ImageHelpers.psm1
elduddz: ok I have updated this file

images/win/scripts/ImageHelpers/MarkdownHelpers.ps1
```diff
@@ -1,5 +1,5 @@
 function Add-ContentToMarkdown {
-    [CmdletBinding()]
+
```
elduddz, line 2: Is this supposed to be here? I just noticed I am deleting the binding? but I never touched this file myself?

images/win/scripts/Installers/Install-PowershellCore.ps1
maxim-lobanov: I have looked at it deeper, it looks like both [install-powershell.ps1](https://raw.githubusercontent.com/PowerShell/PowerShell/master/tools/install-powershell.ps1) and [choco install powershell-core](https://gitlab.com/DarwinJS/ChocoPackages/-/blob/d57372d292b4166313363fa262460bccb2ad29e9/pwshpackages/powershell-core/tools/chocolateyinstall.ps1) do the same things: download installer from `https://github.com/PowerShell/PowerShell/releases/download/v${release}/${packageName}` and apply it.
The single difference: `install-powershell.ps1` determines latest version in [runtime](https://raw.githubusercontent.com/PowerShell/PowerShell/master/tools/metadata.json) but in `choco install powershell-core` the version is [hardcoded](https://gitlab.com/DarwinJS/ChocoPackages/-/commit/d57372d292b4166313363fa262460bccb2ad29e9) and package owner has to update it manually every time when the new version is released.
Considering this point, I would prefer to continue using `https://aka.ms/install-powershell.ps1`.
@alepauly for any thoughts

maxim-lobanov:
@elduddz , just a gentle reminder :)
We would like to apply this contribution and extend this approach to more choco install invocations in future.
But there is the latest comment that prevent us to merge this pull-request (about PowerShell Core installation). Could you please revisit my comment when you have a chance?

Feel free to just revert `Install-PowershellCore.ps1` file and we will take care about retry logic there from our side.

elduddz:
> @elduddz , just a gentle reminder :)
> We would like to apply this contribution and extend this approach to more choco install invocations in future.
> But there is the latest comment that prevent us to merge this pull-request (about PowerShell Core installation). Could you please revisit my comment when you have a chance?
> 
> Feel free to just revert `Install-PowershellCore.ps1` file and we will take care about retry logic there from our side.

Ah sorry, slipped my mind. I'll sort it out tonight.

images/win/scripts/Installers/Install-PowershellCore.ps1
elduddz: reverted

maxim-lobanov:
Both checks are green. I am merging this one.
Thank you for contribution!

images/win/scripts/ImageHelpers/ChocoHelpers.ps1
```diff
@@ -0,0 +1,33 @@
+function Install-Choco {
+    [CmdletBinding()]
+    param(
+        [Parameter(Mandatory)]
+        [string]$install,
```
al-cheb, line 5: nitpick: Use upper-case for parameters

al-cheb, line 5: Could you please rename to be more clear? For example PackageName.

images/win/scripts/ImageHelpers/ChocoHelpers.ps1
```diff
@@ -0,0 +1,33 @@
...
+        [int]$retries = 5
+    )
+
+    begin { }
+    process {
+        $condition = $false
+        $count = 0
+        do {
+            Write-Output "running: powershell choco install $install -y"
+            powershell choco install $install -y
```
al-cheb, line 15: Why do we need to run another instance of powershell?

Issue #635, opened 2020-03-27
Windows environment path has an illegal character


jetersen:
**Describe the bug**
Seems like there is an illegal character coming from Microsoft SQL

```
Caused by: java.nio.file.InvalidPathException: Trailing char < > at index 36: C:\Program Files (x86)\Microsoft SQL 
```

https://github.com/jenkinsci/hashicorp-vault-plugin/runs/540184036#step:4:193

This used to work in the past so a recent change made it borked:

https://github.com/jenkinsci/hashicorp-vault-plugin/actions/runs/47619076

**Area for Triage**: 
Java

**Question, Bug, or Feature?**: 
Bug

**Virtual environments affected**
- [ ] macOS 10.15
- [ ] Ubuntu 16.04 LTS
- [ ] Ubuntu 18.04 LTS
- [ ] Windows Server 2016 R2
- [X] Windows Server 2019

**Expected behavior**
Environment variables should be possible to normalize.

**Actual behavior**
Code that tries to normalizes the windows path sees an illegal character

Jetersen:
https://github.com/jenkinsci/configuration-as-code-plugin/pull/1302/checks?check_run_id=541313850#step:5:599

al-cheb:
Hello, @Jetersen 
I can reproduce the bug. Thank you for mention about it.

```
 - name: Trail space
      run: |
        $env:Path.split(";") -match " $"
```

"C:\Program Files (x86)\Microsoft SQL " - contains space  as the last character

Jetersen:
Hopefully it would be fixed at the source? 🤔 I assume this is a bug in the Microsoft SQL installer?

al-cheb:
@Jetersen , I don't think it's a bug in the Microsoft SQL installer. Looks like we reach env var 2048 limit - https://devblogs.microsoft.com/oldnewthing/20100203-00/?p=15083

Jetersen:
Well that's somewhat disappointing 😞

I guess we need to start using more variables and using short paths.

Jetersen:
This stackoverflow goes over some of the steps to go through: https://stackoverflow.com/questions/34491244/environment-variable-is-too-large-on-windows-10

Some even suggest Microsoft SQL is partly responsible for the issue.

al-cheb:
I think the root cause: `setx`  Path is truncating the path to 1024 characters.

Jetersen:
That's even more concerning 😰

jetersen:
[...]

PR #647, opened 2020-03-31
Fix user environment  PATH variable

al-cheb:
# Description
Environment variable Path has a space as the last character in a folder name. Windows will not accept a  name with a trailing space character.

`Caused by: java.nio.file.InvalidPathException: Trailing char < > at index 36: C:\Program Files (x86)\Microsoft SQL `

The root cause of the issue is in the `setx` command limitation: `WARNING: The data being saved is truncated to 1024 characters.` In the script `Install-DotnetSDK.ps1` we are trying to update PATH variable using `setx` after truncation there is a chance to  have names with a trailing space at the end.

PowerShell  `[System.Environment]::SetEnvironmentVariable` static method doesn't have this limitation and support 32767 characters.
#### Related issue:
https://github.com/actions/virtual-environments/issues/635
## Check list

### DIFF
--- a/images/win/scripts/Installers/Install-DotnetSDK.ps1
+++ b/images/win/scripts/Installers/Install-DotnetSDK.ps1
@@ -100,10 +100,10 @@ function RunPostInstallationSteps()
 {
     Add-MachinePathItem "C:\Program Files\dotnet"
     # Run script at startup for all users
-    $cmdDotNetPath = @"
+    $cmdDotNetPath = @'
 @echo off
-SETX PATH "%USERPROFILE%\.dotnet\tools;%PATH%"
-"@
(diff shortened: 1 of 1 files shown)


images/win/scripts/Installers/Install-DotnetSDK.ps1
```diff
@@ -100,10 +100,10 @@ function RunPostInstallationSteps()
...
     # Run script at startup for all users
-    $cmdDotNetPath = @"
+    $cmdDotNetPath = @'
 @echo off
-SETX PATH "%USERPROFILE%\.dotnet\tools;%PATH%"
-"@
+powershell -executionpolicy bypass -noprofile -command "[System.Environment]::SetEnvironmentVariable('PATH',"""$env:USERPROFILE\.dotnet\tools;$env:PATH""", 'USER')"
+'@
 
     $cmdPath = "C:\Program Files\dotnet\userpath.bat"
```
maxim-lobanov: Can we change file format from bat to ps1? Or it won't work in autologon

Issue #649, opened 2020-03-31
Conflict of CPack in PATH


dragon-archer:
**Describe the bug**
There are two `cpack.exe` in PATH
1. The alias of `choco pack` for Chocolatey
2. The pack module of the CMake suite

The first one (placed in `%ChocolateyInstall%\bin\`) is in front of the second one (placed in CMake  installtion dir), so when use `cpack` in cmd, it will call the first one.
But actually, the first one is just an alias, while the second one is an important program, and the second one is more important the the first one. 

**Area for Triage**: 
Packages

**Question, Bug, or Feature?**: 
Bug

**Virtual environments affected**
- [ ] macOS 10.15
- [ ] Ubuntu 16.04 LTS
- [ ] Ubuntu 18.04 LTS
- [x] Windows Server 2016 R2
- [x] Windows Server 2019

**Expected behavior**
The cpack in CMake suite is in front of the  alias in PATH, and `cpack` command will call the second one above.

**Actual behavior**
The `cpack` command call the alias

**Solution**
There are two ways to solve this problem
1. Remove the alias (can be done with `del /F /Q %ChocolateyInstall%\bin\cpack.exe`)
2. Change the PATH order, put the Chocolatey behind CMake

xnorpx:
this one is really annoying please fix

al-cheb:
Hello, @dragon-archer 
The default cmake has been set to - `C:\Program Files\CMake\bin\cpack.exe`

![cpack](https://user-images.githubusercontent.com/47745270/80343430-86594980-886e-11ea-9569-edf2d17d40b7.JPG)

Feel free to open the thread if you have any concerns.

antonioborondo:
The same problem is happening again.

PR #651, opened 2020-04-01
Remove choco alias cpack.exe that conflicts with CMake cpack

al-cheb:
# Description

There are two cpack.exe in PATH
1. The alias of choco pack for Chocolatey
2. The pack module of the CMake suite

The first one (placed in `%ChocolateyInstall%\bin\`) is in front of the second one (`%ProgramFiles%\CMake\bin\`), so when use cpack, it will call the `%ChocolateyInstall%\bin\cpack.exe` . But actually, the `%ChocolateyInstall%\bin\cpack.exe` is just an alias, while `%ProgramFiles%\CMake\bin\cpack.exe` is an important program, and the second one is more important the the first one.

https://github.com/chocolatey/choco/commit/e9dcb0cc12d49791511e8456b6e94438b81c369c
```
NOTE: `cpack` has been deprecated as it has a name collision with CMake. Please 
 use `choco pack` instead. The shortcut will be removed in v1.
```

#### Related issue:
https://github.com/actions/virtual-environments/issues/649
https://github.com/chocolatey/choco/issues/89
## Check list

### DIFF
--- a/images/win/scripts/Installers/Windows2016/Initialize-VM.ps1
+++ b/images/win/scripts/Installers/Windows2016/Initialize-VM.ps1
@@ -104,6 +104,10 @@ Invoke-Expression ((new-object net.webclient).DownloadString('https://chocolatey
 # Turn off confirmation
 choco feature enable -n allowGlobalConfirmation
 
+# https://github.com/chocolatey/choco/issues/89
+# Remove some of the command aliases, like `cpack` #89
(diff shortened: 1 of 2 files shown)


gep13:
>@sergey-akhalkov said...
Should we worry about regression?

It is possible that some folks might be using `cpack` in some of their builds, so this might cause a problem.  Most of the documentation in our Chocolatey docs suggests that folks should use `choco pack` when building, so we are encouraging folks to do this, but there are likely some that still do it.

PR #660, opened 2020-04-02
Migrate /etc/environment modifications to helpers

dsame:
# Description
Improvement
Use functions from `helpers/etc-environment.sh` in order to

* modify `/etc/environment` from install scripts:
    - installers/rust.sh
    - installers/1604/php.sh
    - installers/1604/php.sh
    - installers/dotnetcore-sdk.sh
* remove legacy `updatepath.sh` used to modify `/etc/environment` 
* rename the helpers functions for better readability

#### Related issue:
https://github.com/actions/virtual-environments-internal/issues/334 

## Check list
- [x ] Related issue / work item is attached
- [x ] Changes are tested and related VM images are successfully generated

### DIFF
--- a/images/linux/scripts/helpers/etc-environment.sh
+++ b/images/linux/scripts/helpers/etc-environment.sh
@@ -9,34 +9,34 @@
 #     The values containing '%' will break the functions
 
 function getEtcEnvironmentVar {
-    var_name="$1"
-    # remove `var_name=` and possible quotes from the line
-    grep "^${var_name}=" /etc/environment |sed -E "s%^${var_name}=\"?([^\"]+)\"?.*$%\1%"
+    variable_name="$1"
+    # remove `variable_name=` and possible quotes from the line
+    grep "^${variable_name}=" /etc/environment |sed -E "s%^${variable_name}=\"?([^\"]+)\"?.*$%\1%"
 }
 
 function addEtcEnvironmentVar {
-    var_name="$1"
-    var_value="$2"
+    variable_name="$1"
+    variable_value="$2"
 
-    echo "$var_name=\"$var_value\"" | sudo tee -a /etc/environment
+    echo "$variable_name=\"$variable_value\"" | sudo tee -a /etc/environment
 }
 
 function replaceEtcEnvironmentVar {
-    var_name="$1"
-    var_value="$2"
+    variable_name="$1"
+    variable_value="$2"
 
-    # modify /etc/environemnt in place by replacing a string that begins with var_name
-    sudo sed -ie "s%^${var_name}=.*$%${var_name}=\"${var_value}\"%" /etc/environment
+    # modify /etc/environemnt in place by replacing a string that begins with variable_name
+    sudo sed -ie "s%^${variable_name}=.*$%${variable_name}=\"${variable_value}\"%" /etc/environment
 }
 
 function setEtcEnvironmentVar {
-    var_name="$1"
-    var_value="$2"
+    variable_name="$1"
+    variable_value="$2"
 
-    if grep "$var_name" /etc/environment > /dev/null; then
-        replaceEtcEnvironmentVar $var_name $var_value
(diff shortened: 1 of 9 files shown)


images/linux/scripts/helpers/etc-environment.sh
```diff
@@ -0,0 +1,64 @@
...
+# Process /etc/environment as if it were shell script with `export VAR=...` expressions
+#    The PATH variable is handled specially in order to do not override the existing PATH
+#    variable. The value of PATH variable read from /etc/environment is added to the end
+#    of value of the exiting PATH variable exactly as it would happen with real PAM app read
+#    /etc/environment
+#
+# TODO: there might be the others variables to be processed in the same way as "PATH" variable
+#       ie MANPATH, INFOPATH, LD_*, etc. In the current implementation the values from /etc/evironments
+#       replace the values of the current environment
+function  sourceEtcEnvironment {
```
maxim-lobanov, line 58: Maybe `reloadEtcEnvironment` will be clearer name?

images/linux/scripts/helpers/etc-environment.sh
```diff
@@ -0,0 +1,64 @@
...
+    echo "$var_name=\"$var_value\"" | sudo tee -a /etc/environment
+}
+
+function replaceEtcEnvironmentVar {
+    var_name="$1"
+    var_value="$2"
+
+    # modify /etc/environemnt in place by replacing a string that begins with var_name
+    sudo sed -ie "s%^${var_name}=.*$%${var_name}=\"${var_value}\"%" /etc/environment
+}
```
maxim-lobanov, line 30: It looks like these functions `getEtcEnvironmentVar`, `addEtcEnvironmentVar`, `replaceEtcEnvironmentVar` are internal and not intendant to use in other scripts.
As I know, bash doesn't support local function but may be we can use specific naming for them? like `_getEtcEnvironmentVar`, `_addEtcEnvironmentVar`, `_replaceEtcEnvironmentVar`? Or left comment in this file

images/linux/scripts/helpers/etc-environment.sh
```diff
@@ -0,0 +1,64 @@
...
+
+function setEtcEnvironmentVar {
```
maxim-lobanov, line 32: `setEtcEnvironmentVariable` - let's use the full `variable` word for readability

images/linux/scripts/helpers/etc-environment.sh
```diff
@@ -0,0 +1,64 @@
...
+    var_value="$2"
+
+    if grep "$var_name" /etc/environment > /dev/null; then
+        replaceEtcEnvironmentVar $var_name $var_value
+    else
+        addEtcEnvironmentVar $var_name $var_value
+    fi
+}
+
+function addEtcEnvironmentPathElement {
```
maxim-lobanov, line 43: I would suggest to build function `PrependEtcEnvironmentVariable` and `AppendEtcEnvironmentVariable`
```
etc_value=$(getEtcEnvironmentVar $1)
setEtcEnvironmentVar $1 "${2}:${etc_value}"
```

It is okay to have one more function `prependEtcEnvironmentPath` if you prefer

images/linux/scripts/helpers/etc-environment.sh
maxim-lobanov: The same for other functions please

images/linux/scripts/helpers/etc-environment.sh
dsame: Done

images/linux/scripts/helpers/etc-environment.sh
dsame: I do not think so. While currently these functions are not used outside but they are not intended to be private and some day the may be useful in outer code. Nothing should prevent developers from using them if they needed

Issue #672, opened 2020-04-03
Not enough entropy on Ubuntu VMs


dtitov:
**Describe the bug**
I have discovered that your Ubuntu VMs suffer from this issue: https://serverfault.com/questions/214605/gpg-does-not-have-enough-entropy

I have a [cryptographic library](https://github.com/uio-bmi/crypt4gh) and when I run GitHub Actions pipeline on Ubuntu, the build hangs forever, because library can’t generate encryption keys due to the issue above (FYI: MacOS VMs are not affected by this).

I have applied a workaround by installing and running `rng-tools` on the VM (as per suggestion from the link above). But it would be nice if you could fix this issue on your side. 

I also realize that security-wise, using `/dev/urandom` as a source of entropy is discouraged. However, I believe that it's important only for production use and it's completely fine for the testing environment, which GitHub Actions is. Also, maybe you would be able to come up with a better solution, because, again - this is just a workaround.

**Area for Triage**: 
- Java

**Question, Bug, or Feature?**: 
- Bug

**Virtual environments affected**
- [ ] macOS 10.15
- [x] Ubuntu 16.04 LTS
- [x] Ubuntu 18.04 LTS
- [ ] Windows Server 2016 R2
- [ ] Windows Server 2019

**Expected behavior**
The build of https://github.com/uio-bmi/crypt4gh runs normally without installing `rng-tools` or performing any other additional manipulation.

**Actual behavior**
The build of https://github.com/uio-bmi/crypt4gh hangs forever if `rng-tools` are not installed.

al-cheb:
Hello, @dtitov 
I am going to add `haveged` package on Ubuntu images.  Could you please validate your build with the package - `sudo apt-get install haveged` ?

dtitov:
Hi, @al-cheb,

I tried it and it worked:  https://github.com/uio-bmi/crypt4gh/runs/564321973?check_suite_focus=true 👍

al-cheb:
Hi, @dtitov 
We have pre-installed `haveged` on Ubuntu images.  Feel free to open the thread if you have any concerns.

al-cheb:
Hello, @dtitov
We have pre-installed `haveged` on Ubuntu images - https://github.com/actions/virtual-environments/blob/releases/ubuntu18/20200430/images/linux/Ubuntu1804-README.md
Feel free to open the thread if you have any concerns.

PR #679, opened 2020-04-06
Add haveged random generator tool on Ubuntu images

al-cheb:
# Description
The build hangs forever with a [cryptographic library](https://github.com/uio-bmi/crypt4gh) , because the library can’t generate encryption keys due to the issue [GPG does not have enough entropy](https://serverfault.com/questions/214605/gpg-does-not-have-enough-entropy)

> We need to generate a lot of random bytes. It is a good idea to perform
> some other action (type on the keyboard, move the mouse, utilize the
> disks) during the prime generation; this gives the random number
> generator a better chance to gain enough entropy.
> Not enough random bytes available.  Please do some other work to give
> the OS a chance to collect more entropy!

Haveged - 200kb (~10 seconds)

#### Related issue:
https://github.com/actions/virtual-environments/issues/672

## Check list

### DIFF
--- a/images/linux/scripts/installers/rndgenerator.sh
+++ b/images/linux/scripts/installers/rndgenerator.sh
@@ -1,0 +1,24 @@
+#!/bin/bash
+################################################################################
+##  File:  rndgenerator.sh
+##  Desc:  Install random number generator
+################################################################################
+
+# Source the helpers for use with the script
+source $HELPER_SCRIPTS/document.sh
+
+# Install haveged
+apt-get -y install haveged
+
+# Run tests to determine that the software installed as expected
+echo "Testing to make sure that script performed as expected, and basic scenarios work"
+for cmd in haveged; do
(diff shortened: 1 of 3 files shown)


images/linux/scripts/installers/rndgenerator.sh
```diff
@@ -0,0 +1,24 @@
...
+apt-get -y install haveged
+
+# Run tests to determine that the software installed as expected
+echo "Testing to make sure that script performed as expected, and basic scenarios work"
+for cmd in haveged; do
+    if ! command -v $cmd; then
+        echo "$cmd was not installed or not found on PATH"
+        exit 1
+    fi
+done
```
maxim-lobanov, line 20: Why do we need a loop here?

al-cheb: Just for feature tools if someone decides to add rng-tools.

PR #680, opened 2020-04-06
Install MongoDB on Windows images

dsame:
# Description
New tool
Install MongoDB

**total size:** 930MB
**installation time:** less than a minute

#### Related issue:
https://github.com/actions/virtual-environments/issues/20

## Check list

### DIFF
--- a/images/win/scripts/Installers/Install-MongoDB.ps1
+++ b/images/win/scripts/Installers/Install-MongoDB.ps1
@@ -6,4 +6,3 @@
 choco install mongodb
 
 Add-MachinePathItem "$($env:SystemDrive)\Program Files\MongoDB\Server\4.2\bin"
-$env:Path = Get-MachinePath
--- a/images/win/scripts/Installers/Install-MongoDB.ps1
+++ b/images/win/scripts/Installers/Install-MongoDB.ps1
@@ -1,0 +1,9 @@
+####################################################################################
+##  File:  Install-MongoDB.ps1
+##  Desc:  Install MongoDB
+####################################################################################
+
+choco install mongodb
+
+Add-MachinePathItem "$($env:SystemDrive)\Program Files\MongoDB\Server\4.2\bin"
+$env:Path = Get-MachinePath
--- a/images/win/Windows2016-Azure.json
+++ b/images/win/Windows2016-Azure.json
@@ -656,6 +656,12 @@
             ]
         },
         {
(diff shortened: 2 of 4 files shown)


images/win/scripts/Installers/Install-MongoDB.ps1
```diff
@@ -0,0 +1,8 @@
+####################################################################################
+##  File:  Install-MongoDB.ps1
+##  Desc:  Install MongoDB
+####################################################################################
+
+Import-Module -Name ImageHelpers -Force
+[Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
+Install-MSI -MsiUrl "https://fastdl.mongodb.org/win32/mongodb-win32-x86_64-2012plus-4.2.5-signed.msi" -MsiName "MongoDB.msi"
```
al-cheb, line 8: Why don't we use choco?

dsame: done https://github.com/actions/virtual-environments/pull/680/commits/e9543008c288328e85588cfd899bb435434f8a09

images/win/scripts/Installers/Install-MongoDB.ps1
```diff
@@ -0,0 +1,9 @@
+####################################################################################
+##  File:  Install-MongoDB.ps1
+##  Desc:  Install MongoDB
+####################################################################################
+
+choco install mongodb
+
+Add-MachinePathItem "$($env:SystemDrive)\Program Files\MongoDB\Server\4.2\bin"
```
al-cheb, line 8: Are you sure what 4.2 will not be changed?

dsame: There's no way to be sure mongodb/choco team won't change the path. The validation script is a waring tool the path has to be changed

images/win/scripts/Installers/Install-MongoDB.ps1
```diff
@@ -0,0 +1,9 @@
...
+$env:Path = Get-MachinePath
```
maxim-lobanov, line 9: We, actually, don't need this line because this line modifies Path only for current file and it doesn't make sense to do that in the last line

dsame: fixed

PR #695, opened 2020-04-08
Update CMake to 3.17.0 on all environments

rick-de-water:
# Description
The newest version of CMake adds some nice new features, but is missing on the linux environments. I've simply updated the version number in the install script.

## Check list
- [x] Documentation is updated

### DIFF
--- a/images/linux/scripts/installers/cmake.sh
+++ b/images/linux/scripts/installers/cmake.sh
@@ -12,7 +12,7 @@ echo "Checking to see if the installer script has already been run"
 if command -v cmake; then
     echo "Example variable already set to $EXAMPLE_VAR"
 else
-	curl -sL https://cmake.org/files/v3.16/cmake-3.16.2-Linux-x86_64.sh -o cmakeinstall.sh \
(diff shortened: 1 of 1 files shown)


maxim-lobanov:
@rick-de-water , your first commit was the best one :)
We should not update documentation manually. We need to change only `images/linux/scripts/installers/cmake.sh`

rick-de-water:
> @rick-de-water , your first commit was the best one :)
> We should not update documentation manually. We need to change only `images/linux/scripts/installers/cmake.sh`

Ah I see, I was wondering how you would keep that file up to date properly ^^

rick-de-water:
I did a rollback, so it should just be the first commit now.

maxim-lobanov:
> > @rick-de-water , your first commit was the best one :)
> > We should not update documentation manually. We need to change only `images/linux/scripts/installers/cmake.sh`
> 
> Ah I see, I was wondering how you would keep that file up to date properly ^^

It is updated automatically when the new image is deployed to all VMs.

PR #700, opened 2020-04-09
use the right software name

mprins:
# Description
seems to be a copy-paste error resulting in strange documentation

## Check list


maxim-lobanov:
@mprins , that is definitely typo from copy-paste :)
Thank you for finding and fixing.

PR #705, opened 2020-04-09
Add Python tools installation from Github releases for Windows

vmapetr:
This PR changes approach for installation of Python tools for Windows systems.

**Details:**
Main point of Python toolset provisioner is to download required Python versions from github releases and install them during the image generation. Tools and versions that will be installed can be found in `toolset-%image_version%.json` configuration file in `/images/win` folder. In case if tool declared in configuration contains wildcard in build number, latest available version will be installed.

**Additional:**
Python npm-packages installation logic was removed from toolcache provisioners.

**How it was tested:**
 [Windows2016](https://dev.azure.com/github/virtual-environments/_build/results?buildId=64917&view=results)
[Windows2019](https://dev.azure.com/github/virtual-environments/_build/results?buildId=64916&view=results)


images/win/scripts/Installers/Validate-Toolset.ps1
```diff
@@ -0,0 +1,101 @@
...
+
+    foreach ($version in $tool.versions) {
+        # Check if version folder exists
+        $foundVersionPath = Join-Path $toolPath $version | Join-Path -ChildPath $tool.arch | Get-Item
+        if ($foundVersionPath -eq $null) {
+            Write-Host "Expected $($tool.name) $version ($($tool.arch)) folder is not installed!"
+            exit 1
+        }
+
+        $foundVersion = $($foundVersionPath.parent.fullname.Split("\")[-1])
```
maxim-lobanov, line 85: What if `$foundVersionPath` contains array with a few versions?
I believe, it is possible if we specify in toolset:
`@("3.7.*", "3.7.2", "3.7.3")`

images/win/toolset-2016.json
```diff
@@ -0,0 +1,32 @@
...
+            "arch": "x86",
+            "platform" : "win32",
+            "versions": [
+                "2.7.*",
+                "3.5.*",
+                "3.6.*",
+                "3.7.*",
+                "3.8.*"
+            ],
+            "default": ""
```
maxim-lobanov, line 29: Let's just remove empty field

images/win/scripts/Installers/Validate-Toolset.ps1
vmapetr: Good point! I've made some fixes in installed versions selection, please take a look.

images/win/toolset-2016.json
vmapetr: Done.

Issue #692, opened 2020-04-08
vs2017-win2016 random dll library load errors


FrancescElies:
Summarizing my comments from #691 CC @miketimofeev

**Describe the bug**
Yesterday (7. April 2020) we started getting some random failures, the code did not change code on our side but something seems to have changed on the image, comparing logs we noticed the following.

Running same pipeline for same commit twice behaves differently, sometimes we cannot load dlls.

**Area for Triage**: 
C/C++, Python

**Virtual environments affected**
- [ ] macOS 10.15
- [ ] Ubuntu 16.04 LTS
- [ ] Ubuntu 18.04 LTS
- [x] Windows Server 2016 R2
- [ ] Windows Server 2019

**Expected behavior**
I expect two builds of the same commit running on vs2017-win2016  to work the same.

**Actual behavior**
When we call the `dir` command we get slightly different outputs
- Failing Build 🔴
```
2020-04-08T04:51:40.9092142Z  Volume in drive D is Temporary Storage
2020-04-08T04:51:40.9093101Z  Volume Serial Number is 4099-6268
...
2020-04-08T04:51:42.0603173Z rootdir: d:\a\1\s\src
```
- Passed  Build 🟢 (for the same commit)
```
2020-04-08T04:51:13.6285940Z  Volume in drive D is Temp
2020-04-08T04:51:13.6286720Z  Volume Serial Number is 7AAE-11A5
...
2020-04-08T04:51:14.7481347Z rootdir: D:\a\1\s\src
```

At least drive D seems to be slightly different, rootdir is now also lowercase, Windows path does not differentiate between lowercase and uppercase right?

We consistently **get dll loading problems** when `drive D is Temporary Storage`. 

Could someone shed a bit of light on this?

alepauly:
@FrancescElies I took the liberty to send you an email to the email address on your profile so we can get some more details and try to figure why your builds started failing. I hope that's ok. Thanks.

FrancescElies:
Hi thanks for your reply, I got the email :)

al-cheb:
Hello, @FrancescElies 
Just to double check. Could you please add the current path to the `PATH`. For PowerShell example:
`$env:PATH = "$(pwd);$env:PATH"`

FrancescElies:
@al-cheb i will give it a try on Tuesday (14. April). This will be a long weekend. İ hope it's fine.

FrancescElies:
[...]

Issue #691, opened 2020-04-08
Windows build changes without a release log


ThadHouse:
Did the windows build agents get updated on 4/6/20 without a changelog being posted? Our PR builds were working until about 1 AM on 4/7/20, and now they all fail with library load errors. Trying to find what caused this, as we have not committed anything new in days, and after 8 hours am still stuck. Would gladly appreciate some help in figuring out if anything changed on Azures end.

FrancescElies:
I believe we are seeing the same problem here.

We started getting some random failures, the code did not change on our side but something seems to have changed on the image, comparing logs we noticed the following.

When we call the `dir` command we get slightly different outputs
- Failing Build 🔴
```
2020-04-08T04:51:40.9092142Z  Volume in drive D is Temporary Storage
2020-04-08T04:51:40.9093101Z  Volume Serial Number is 4099-6268
...
2020-04-08T04:51:42.0603173Z rootdir: d:\a\1\s\src
```
- Passed  Build 🟢 (for the same commit)
```
2020-04-08T04:51:13.6285940Z  Volume in drive D is Temp
2020-04-08T04:51:13.6286720Z  Volume Serial Number is 7AAE-11A5
...
2020-04-08T04:51:14.7481347Z rootdir: D:\a\1\s\src
```


At least the the colume drive D seems to be slightly different, rootdir is now also lowercase, Windows path does not differentiate between lowercase and uppercase right?

We consistently get dll loading problems when `drive D is Temporary Storage`. 

Could someone shed a bit of light on this?

FrancescElies:
Just in case it helps the problems we are getting are when loading a dll via python.

```
OSError: cannot load library 'myfile.dll': error 0x7e.  Additionally, ctypes.util.find_library() did not manage to locate a library called 

```

ThadHouse:
I was able to get a full loader log of my issue. The library being loaded from Java is `wpiutiljni.dll`, which is being loaded via a full path. That library depends on `wpiutild.dll`, and when loading the loader seems to be entirely skipping the current working directory, which the top of the log says is the directory where both dlls are located at.

[FullLog.txt](https://github.com/actions/virtual-environments/files/4448740/FullLog.txt)

ThadHouse:
The most interesting load log I think is attempting to load from `C:\Program F\wpiutild.dll`. There are 65 paths it checks before that one. I think whats happening is that the search path is getting too long, and overfilling a buffer.

ThadHouse:
[...]

PR #707, opened 2020-04-10
Disable MS SQL Server 2019 Express  installation

al-cheb:
# Description
After installation MS SQL Server 2019 Express we have faced with  incorrect dynamic-link library search order: https://docs.microsoft.com/en-us/windows/win32/dlls/dynamic-link-library-search-order due to conflict with `instapi150.dll`.

Without SQL:

If SafeDllSearchMode is enabled, the search order is as follows:
1. The directory from which the application loaded. 
2. The system directory. Use the GetSystemDirectory function to get the path of this directory.
3. The 16-bit system directory. There is no function that obtains the path of this directory, but it is searched.
4. The Windows directory. Use the GetWindowsDirectory function to get the path of this directory.
5. The current directory.
6. The directories that are listed in the PATH environment variable. Note that this does not include the per-application path specified by the App Paths registry key. The App Paths key is not used when computing the DLL search path.

`LdrpComputeLazyDllPath - INFO: DLL search path computed: D:\a\1\s\NativeApp;C:\windows\SYSTEM32;C:\windows\system;C:\windows;.;%PATH%`

1. D:\a\1\s\NativeApp - The directory from which the application loaded. 
2. C:\windows\SYSTEM32 -The system directory
3. C:\windows\system - The 16-bit system directory
4. C:\windows - The Windows directory.
5. `.` - The current directory
6. %PATH%

With installed SQL Server the `.`  current directory is missing from the searching path that it breaks loading DLLs  :
`LdrpComputeLazyDllPath - INFO: DLL search path computed: D:\a\1\s\NativeApp;C:\Program Files\Microsoft SQL Server\150\Shared\;C:\windows\SYSTEM32;C:\windows\system;C:\windows; %PATH%`

#### Related issue:
https://github.com/actions/virtual-environments/issues/691
https://github.com/actions/virtual-environments/issues/692

## Check list

### DIFF
--- a/images/win/Windows2016-Azure.json
+++ b/images/win/Windows2016-Azure.json
@@ -472,14 +472,6 @@
         },
         {
             "type": "powershell",
-            "elevated_user": "SYSTEM",
-            "elevated_password": "",
-            "scripts":[
-                "{{ template_dir }}/scripts/Installers/Install-SQLExpress.ps1"
-            ]
-        },
-        {
-            "type": "powershell",
             "scripts":[
                 "{{ template_dir }}/scripts/Installers/Install-SQLPowerShellTools.ps1"
             ]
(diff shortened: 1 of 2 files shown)


ThadHouse:
So what must be happening is it replaced a base library with one that calls SetDllDirectory

PR #718, opened 2020-04-14
Fix Git installer to download directly from release

alepauly:
# Description
I'm updating the Git install script for Windows images to use the release from the official GitHub for Windows repo. The main reason for this is to be able to move the version without being attached to what chocolatey provides.

## Check list


dscho:
Looks good to me!

If it was me, I would probably squash the second commit into the first (pretending that there was never a need to fix anything up ;-)).

Issue #738, opened 2020-04-18
Installed Android SDK binaries on Ubuntu are only executable by root


kentonv:
**Describe the bug**

On `ubuntu-latest`, the binaries installed in `$ANDROID_SDK_ROOT/tools/bin` are marked executable only by root.

**Area for Triage**: Android

**Question, Bug, or Feature?**: Bug

**Virtual environments affected**
- [ ] macOS 10.15
- [x] Ubuntu 16.04 LTS
- [x] Ubuntu 18.04 LTS
- [ ] Windows Server 2016 R2
- [ ] Windows Server 2019

**Expected behavior**

Installed binaries should be executable by all users.

**Actual behavior**

From an actual action log:
```
▶ Run ls -l $ANDROID_SDK_ROOT/tools/bin
total 64
-rwxr--r-- 1 root root 6742 Sep 13  2017 apkanalyzer
-rwxr--r-- 1 root root 5253 Sep 13  2017 archquery
-rwxr--r-- 1 root root 6040 Sep 13  2017 avdmanager
-rwxr--r-- 1 root root 5250 Sep 13  2017 jobb
-rwxr--r-- 1 root root 6834 Sep 13  2017 lint
-rwxr--r-- 1 root root 3323 Sep 13  2017 monkeyrunner
-rwxr--r-- 1 root root 5613 Sep 13  2017 screenshot2
-rwxr--r-- 1 root root 6047 Sep 13  2017 sdkmanager
-rwxr--r-- 1 root root 3076 Sep 13  2017 uiautomatorviewer
```

And:
```
/home/runner/work/_temp/2fdf16f4-50b2-4c0d-adb3-024bbe42a1aa.sh: line 3: /usr/local/lib/android/sdk/tools/bin/avdmanager: Permission denied
```

Darleev:
Hi @kentonv, $ANDROIDK_SDK_ROOT path is: /usr/local/lib/android/sdk. User cannot write to `/usr/bin/` folder by default. In order to elevate required permissions, it is required to use `sudo`. For example:

`sudo lint`

or by full path

`sudo $ANDROID_SDK_ROOT/tools/bin/lint`

Also, let me provide you with link, where a [similar case](https://github.com/actions/virtual-environments/issues/534) was discussed. In case of any questions, please let me know, we will be glad to assist you further.

Darleev:
[...]

PR #747, opened 2020-04-20
Android SDK binaries insufficient permissions.

Darleev:
# Description
Bug fixing
Installed Android SDK binaries can be executed only by root, due to lack of permissions.

#### Related issue:
https://github.com/actions/virtual-environments/issues/738

## Check list
It seems it's really minor change, and nothing is changed in image build workflow.

### DIFF
--- a/images/linux/scripts/installers/1604/android.sh
+++ b/images/linux/scripts/installers/1604/android.sh
@@ -38,7 +38,7 @@ else
 fi
 
 # Add required permissions
-sudo chmod -R 755 ${ANDROID_SDK_ROOT}
+chmod -R 755 ${ANDROID_SDK_ROOT}
 
 # Install the following SDKs and build tools, passing in "y" to accept licenses.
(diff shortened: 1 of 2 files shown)


images/linux/scripts/installers/1604/android.sh
```diff
@@ -37,6 +37,8 @@ else
     exit 1
 fi
 
+# Add required permissions
+chmod -R 755 ${ANDROID_SDK_ROOT}
```
kentonv, line 41: Might be better to use `chmod -R a+X` (note capital X) instead of `chmod -R 755`. This way you won't mark files executable that aren't supposed to be executable.

(The capital X modifier says to add the executable permission only to files that already have executable set for some user. It looks like the installed Android SDK binary do have the execute bit set for root.)

Darleev: Thank you for your the suggestion @kentonv, I have tested this solution and it works perfectly fine. I will change the script according to your advice.

Issue #753, opened 2020-04-21
regression on 20200415.3?


lechen26:
**Describe the bug**
seems like there is a regression on ubuntu-18.04 version 20200415.3.
the PATH on that image doesnt includes the /snap/bin hence packages installed through snap couldnt be found.

**Area for Triage**: 
we are using Github action workflow using runs-on: ubuntu-latest
for some reason yesterday , although the latest images is still 20200406.2 , few workflows pulled the 20200415.3. version which included this issue of missing PATH.

**Virtual environments affected**
- [ ] Ubuntu 18.04 LTS

Darleev:
Hi @lechen26, thank you for the reported issue. We rolled back the latest Ubuntu release, and, I guess,  the issue should not be actual at this moment. We will check the issue further on our side and prepare fix for the issue. 
I will keep you posted about status of the issue.

dsame:
Fix provided in https://github.com/actions/virtual-environments/pull/755 and will be available for use in a week or two

lechen26:
10x!

maxim-lobanov:
@lechen26 , new image was deployed. Could you please check if snap works as expected?

lechen26:
working, 10x!

miketimofeev:
@lechen26 thanks for the update!
I'm going to close the issue, feel free to ping us if you have any concerns.

PR #755, opened 2020-04-21
Add snap/bin to /etc/environment

dsame:
# Description
Bug fixing

[snap](https://snapcraft.io/) exists on the base image before the provision has started, so there was no install script for it.

Before https://github.com/actions/virtual-environments/pull/660 the PATH variable in /etc/environment had been inherited from the build environment.

Since https://github.com/actions/virtual-environments/pull/660 the PATH variable /etc/environment it is created from scratch. This became a reason of https://github.com/actions/virtual-environments/issues/753

The inheriting run-time environment from build-time environment does not seem to be a good solution, so the provision script snap-environment.sh to to include `/snap/bin` into run-time PATH is added in this PR. This solves the issue.

The root reason of we need to manipulate with `/etc/environment` is the shell scripts running the user's tasks are launched with `-norc` option and the native profile scripts are ignored.

TODO: should we test the expected run-time PATH is the same as it is constructed in `/etc/environment`?

#### Related issue:
https://github.com/actions/virtual-environments/issues/753

## Check list


images/linux/scripts/installers/snap-environment.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
```
al-cheb, line 1: Why don't we use configure-environment.sh script?

dsame: I though about it, but decided the semantic of "Configure environment" does not match exactly the task i am doing: "Complete installation of snap".

The putting the modification of snap to "configure environment" leads to 2 potential problems:

1. the installation of snap continues to be hidden from the provision template
2. there will be inconsistency with the other installed software: why most of them have its own scripts but snap does not

Both problems are insignificant and can be ignored, please ping me if i should use  configure-environment.sh script instead

miketimofeev: Snap is available on the system by default, we don't need to create additional script
https://snapcraft.io/docs/installing-snapd
image

PR #760, opened 2020-04-21
Remove mention of Boost in PATH from Windows image docs

nightlark:
# Description
Fixes mismatch between the documentation saying Boost is part of `PATH` on Windows, and the actual Windows images not having the location of Boost in `PATH`.

## Check list

### DIFF
--- a/images/win/Windows2016-Readme.md
+++ b/images/win/Windows2016-Readme.md
@@ -478,7 +478,6 @@ _Environment:_
 #### 1.69.0 [msvc-14.1]
 
 _Environment:_
-* PATH: contains the location of Boost version 1.69.0
 * BOOST_ROOT_1_69_0: root directory of the Boost version 1.69.0 installation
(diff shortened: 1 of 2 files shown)


maxim-lobanov:
@nightlark , thank you for contribution but we should not update these files manually because they are autogenerated.
Actually, we should change file `Validate-Boost.ps1` but I see that @al-cheb has already done it here: https://github.com/actions/virtual-environments/commit/1cfa87b4a6b2950af3dc9de3c5405d3f8e50e357

So I think that documentation will be updated with next image rollout

nightlark:
It seems strange that the changed images are getting used for builds. Is pushing the new generated doc files a manual process?

maxim-lobanov:
@nightlark , usually it takes some time (3-4 days) to propagate update to all machines and after that docs are updated. We had a few issues previous weeks, that is the reason why docs are a bit outdated.

Issue #739, opened 2020-04-18
MySQL Breaking Changes


michaelhelmick:
**Describe the bug**
Not so much a bug, but more of a clearer description on how to use software on the OS.

**Area for Triage**: 
Question

**Question, Bug, or Feature?**: 
Question

**Virtual environments affected**
- [x] Ubuntu 18.04 LTS

**Expected behavior**
mysql to work out of the box without starting

**Actual behavior**
I looked across the community for solutions (including adding `services`, updating ports, trying different hosts, and it ended up being that mysql doesn't start automatically anymore (via https://github.blog/changelog/2020-02-21-github-actions-breaking-change-ubuntu-virtual-environments-will-no-longer-start-the-mysql-service-automatically/) I think it would be appreciated if that was noted somewhere other than a blog post.

maxim-lobanov:
Hello @michaelhelmick ,
Am I right that your proposal is about updating documentation to include information that MySQL service is not enabled by default and instructions how it can be enabled. Please correct me if I am wrong

michaelhelmick:
Correct! The only place I could find that MySQL is no longer started by default was the blog post disconnected from documentation. Would be super helpful!

dibir-magomedsaygitov:
Documentation has been successfully updated. Changes were merged and will be delivered to all VMs in a week.

dibir-magomedsaygitov:
Documentation has been successfully updated. All changes are delivered to VMs. Feel free to contact if you have any issues.

PR #776, opened 2020-04-23
Update MySQL info in the docs for Ubuntu 18.04 LTS image

dibir-magomedsaygitov:
In scope of this PR we added information to the documentation related to MySQL service being disabled by default and how this service can be enabled. 
Update for [this](https://github.com/actions/virtual-environments/issues/739) work-item.


images/linux/scripts/installers/mysql.sh
```diff
@@ -44,7 +44,8 @@ echo "Lastly, documenting what we added to the metadata file"
 DocumentInstalledItem "MySQL ($(mysql --version))"
 DocumentInstalledItem "MySQL Server (user:root password:root)"
 DocumentInstalledItem "MS SQL Server Client Tools"
+DocumentInstalledItem "MySQL service is disabled by default. Use the following command as a part of your job to start the service: sudo /etc/init.d/mysql start"
```
maxim-lobanov, line 47: use "`" quotes for command:
```
MySQL service is disabled by default. Use the following command as a part of your job to start the service: `sudo /etc/init.d/mysql start`
```

al-cheb, line 47: I think we should use systemctl to start the service:
`systemctl start mysql.service`

images/linux/scripts/installers/mysql.sh
```diff
@@ -44,7 +44,8 @@ echo "Lastly, documenting what we added to the metadata file"
 DocumentInstalledItem "MySQL ($(mysql --version))"
 DocumentInstalledItem "MySQL Server (user:root password:root)"
 DocumentInstalledItem "MS SQL Server Client Tools"
+DocumentInstalledItem "MySQL service is disabled by default. Use the following command as a part of your job to start the service: `sudo systemctl start mysql.service`"
```
al-cheb, line 47: You should use single quotes to escape backtick, because `sudo systemctl start mysql.service` -  is interpreted as a command.

_$ DocumentInstalledItem "MySQL service is disabled by default. Use the following command as a part of your job to start the service: `sudo systemctl start mysql.service`"_
```
System has not been booted with systemd as init system (PID 1). Can't operate.
- MySQL service is disabled by default. Use the following command as a part of your job to start the service:
```

Issue #117, opened 2019-11-28
Add Homebrew on Linux


ericallam:
**Tool information**
- Tool name: Homebrew
- Add or update? Add
- Desired version: Latest
- If this is an add request:
  - Linux dependency manager/installer
  - https://docs.brew.sh/Homebrew-on-Linux

**Virtual environments affected**
- [ ] Ubuntu 16.04 LTS
- [ ] Ubuntu 18.04 LTS

peaceiris:
For reference, It takes 20 sec to install.

```yaml
- name: Install Linuxbrew
  run: sh -c "$(curl -fsSL https://raw.githubusercontent.com/Linuxbrew/install/master/install.sh)"
```

sjackman:
FWIW there is an official Homebrew Docker image at `homebrew/brew`.
https://hub.docker.com/r/homebrew/brew
https://github.com/Homebrew/brew/packages/29947
That's not meant to be an argument against adding it to the GitHub VM image, just an FYI.

dsame:
PR https://github.com/actions/virtual-environments/pull/589

maxim-lobanov:
New image with homebrew was deployed everywhere. Closing the issue

PR #789, opened 2020-04-27
[Ubuntu] Move homebrew to the end of the $PATH

miketimofeev:
Homebrew for Linux contains internal python that overrides system python at the moment.

Changes:
- Add function that is able to add path at the end of the $PATH
- Change brew installation to use new function

### DIFF
--- a/images/linux/scripts/helpers/etc-environment.sh
+++ b/images/linux/scripts/helpers/etc-environment.sh
@@ -61,6 +61,11 @@ function prependEtcEnvironmentPath {
     prependEtcEnvironmentVariable PATH "${element}"
 }
 
+function appendEtcEnvironmentPath {
+    element="$1"
+    appendEtcEnvironmentVariable PATH "${element}"
+}
+
 # Process /etc/environment as if it were shell script with `export VAR=...` expressions
 #    The PATH variable is handled specially in order to do not override the existing PATH
 #    variable. The value of PATH variable read from /etc/environment is added to the end
--- a/images/linux/scripts/installers/homebrew.sh
+++ b/images/linux/scripts/installers/homebrew.sh
@@ -20,7 +20,7 @@ sudo chmod -R o+w $HOMEBREW_PREFIX
 brew shellenv|grep 'export HOMEBREW'|sed -E 's/^export (.*);$/\1/' | sudo tee -a /etc/environment
 # add brew executables locations to PATH
 brew_path=$(brew shellenv|grep  '^export PATH' |sed -E 's/^export PATH="([^$]+)\$.*/\1/')
-prependEtcEnvironmentPath "$brew_path"
+appendEtcEnvironmentPath "$brew_path"
 
 # Validate the installation ad hoc
 echo "Validate the installation reloading /etc/environment"


sjackman:
The Homebrew installation probably shouldn't have any packages installed, just a blank slate. Did it have Python pre-installed?

andreineculau:
I also don't understand how homebrew/linuxbrew would preinstall python.

Our CI builds started failing because "no homebrew paths in PATH" is not the same as "homebrew paths appended to PATH" (if the paths are already in PATH, we don't willy-nilly append/prepend them).

Whoever is accustomed to homebrew/linuxbrew will find this change weird IMHO, since the default behaviour (as per `brew shellenv`) is to prepend homebrew paths to PATH, not append.

miketimofeev:
@sjackman @andreidubov as you can see, python is in the `/home/linuxbrew/.linuxbrew/bin` despite we haven't installed it intentionally https://github.com/actions/virtual-environments/blob/master/images/linux/scripts/installers/homebrew.sh
image
We appreciate it if you will be able to provide a better solution to get rid of that python in the PATH.

dawidd6:
`brew ls` on clean `ubuntu-latest` VM gives:

```
aws-sam-cli
bzip2
gdbm
libffi
ncurses
openssl@1.1
patchelf
python
readline
sqlite
xz
zlib
```

sjackman:
https://github.com/aws/homebrew-tap/blob/692471155eb82ff807f12fe337511d13603b88ad/Formula/aws-sam-cli.rb#L21

miketimofeev:
@dawidd6 nice catch!
Looks like we need to move aws-cli installation after this python one
https://github.com/actions/virtual-environments/blob/master/images/linux/scripts/installers/python.sh

miketimofeev:
I've tried to move python installation without any luck — brew still downloads it's own python so we can't get rid of this workaround so simple.

andreineculau:
@miketimofeev I'm willing to put invest some time into this. But I'm having it hard to understand how I can test locally without any deps on Azure, etc. If it's not possible, then I can invest some time for ocular debugging.

miketimofeev:
@andreineculau testing it locally should be easy, you just need bare Ubuntu-18.04 image and then apply the installation scripts one by one. Some pieces of advice can be found in this issue https://github.com/actions/virtual-environments/issues/931
Feel free to contact us if you need any help.

andreineculau:
Thanks! I get the feeling it's a rabbit hole, and I'll not go down on this one.

I looked a bit at the scripts though, and afaics you're trying to have the cake and eat it too.

Shaun might slap me but `brew install aws-sam-cli` will always install `python` via `homebrew` as well, no matter if there's already a `python` installed on the system outside `homebrew`.
Depending on how things are configured, packages like aws-sam-cli (that depend on python) might end up calling homebrew's python or the first `python` executable in `$PATH`. Same goes for a developer that might call `brew install python` and then invoking `python`... They'll expect to invoke homebrew's version, but it will be the system's version instead.

With that said, homebrew was added to this codebase, but the other scripts haven't been updated to make use of homebrew, like `python.sh` or `node.sh` etc, so while we have homebrew already available we un/intentionally? still install packages via aptitude for instance, *AFTER* homebrew.
I wonder if the sequence has been properly vetted. I mean: why install homebrew before all the other things that do *NOT* need homebrew? In my linuxbrew-based bootstrapping procedures, I install whatever is required via the system's package management tool (e.g. aptitude) or by manually downloading/repo-cloning+compiling, then I install homebrew and I install all other software via homebrew.

As someone that has also bootstrapped many images via linuxbrew, my humble recommendation is that you KISS and install as many things via linuxbrew, and thus make linuxbrew central to the image bootstrapping procedure. Both homebrew maintainers and Shaun are doing a fantastic job, and the very few issues I encountered have been addressed quickly.

One possible solution, given that you have strong reasons to keep `python` (and other executables?!) to point to the system's version, instead of homebrew's version (although once again, this will be a surprise to any developer as explained above!) is to add a symlink `/home/runner/.local/bin/python -> /usr/bin/python` and making sure that `~/.local/bin` is before homebrew's bin folder in `$PATH`.

FWIW it has not been explained why is it a problem to have python installed via homebrew. Is something actually breaking?

miketimofeev:
Yes, all the packages installed via aptitude intentionally. We've added homebrew due to the request https://github.com/actions/virtual-environments/issues/117 and have no plans to switch to install all the tools with it except aws-sam-cli as [the documentation provides only this method](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/serverless-sam-cli-install-linux.html). Switching to homebrew will require a lot of efforts for building and testing and we're not 100% safe from regression.
It's hard to say now why the decision was to install brew as one of the first steps, probably due to changes in the PATH. I agree that it doesn't make sense to put homebrew at the end of the PATH.
The reason we keep system python is that it contains some pip packages like `request`, which are essential for some internal teams and we received a complaint right after python from brew was put in the PATH.

miketimofeev:
@andreineculau by the way, could you please create a bug in this repo for incorrect brew behavior?

sjackman:
Installing `aws-sam-cli` with Hombrew will also install `python` with Homebrew, as Python is a dependency of AWS SAM CLI. The simplest solution to this problem is to install AWS SAM CLI using a tool other than Homebrew if you do not want to install Python using Homebrew. Homebrew can still be provided in the AWS image as a tool for users to install dependencies. Happy to discuss other options if you like. Putting Homebrew at the end of the `PATH` is a reasonable (but perhaps not ideal) workaround. The end user could always move it to the front of the `PATH` if they prefer it there.

miketimofeev:
@sjackman thanks! We're working out a solution to install aws-cli without brew and move brew installation to the beginning of the `PATH`.

PR #791, opened 2020-04-28
Add Apache JMeter and update Apache Ant

brunoborges:
# Description
JMeter is widely used in CI/CD to perform tests.
Binary size is roughly 62MB.

Apache Ant updated to 1.10.7 on Linux and Windows.

Pending Mac OS.


AlenaSviridenko:
Hi @brunoborges, thanks for contribution!

Notes from my side: both Windows and Linux documentations are autogenerated, here is the way how to add them there:
1. for Windows: please, add JMeter here https://github.com/actions/virtual-environments/blob/master/images/win/scripts/Installers/Validate-JavaTools.ps1
2. for Linux: please, document JMeter at the bottom of the file https://github.com/actions/virtual-environments/blob/master/images/linux/scripts/installers/java-tools.sh.

brunoborges:
$ jmeter -version

This command doesn't give a nice output for parsing, so I did hardcoded for now.

images/linux/scripts/installers/java-tools.sh
```diff
@@ -29,11 +29,14 @@ echo "JAVA_HOME=/usr/lib/jvm/zulu-${DEFAULT_JDK_VERSION}-azure-amd64" | tee -a /
 echo "JAVA_TOOL_OPTIONS=-Dfile.encoding=UTF8" | tee -a /etc/environment
 
 # Install Ant
-apt-fast install -y --no-install-recommends ant ant-optional
-echo "ANT_HOME=/usr/share/ant" | tee -a /etc/environment
+curl -sL https://downloads.apache.org/ant/binaries/apache-ant-1.10.7-bin.zip -o ant.zip
```
AlenaSviridenko, line 32: could you please clarify what is the purpose of replacing`apt install` to installation by direct link? We are trying to avoid adding something that will require manual maintenance in future. 

And the same question about JMeter installatiob, there are apt packages for it for both Ubuntu versions https://packages.ubuntu.com/search?keywords=jmeter&searchon=names&suite=all&section=all

Is there any specific reason to not use apt?

images/linux/scripts/installers/java-tools.sh
```diff
@@ -73,6 +83,7 @@ DocumentInstalledItemIndent "7 ($(/usr/lib/jvm/zulu-7-azure-amd64/bin/java -show
 DocumentInstalledItemIndent "8 ($(/usr/lib/jvm/zulu-8-azure-amd64/bin/java -showversion |& head -n 1)) (default)"
 DocumentInstalledItemIndent "11 ($(/usr/lib/jvm/zulu-11-azure-amd64/bin/java -showversion |& head -n 1))"
 DocumentInstalledItemIndent "12 ($(/usr/lib/jvm/zulu-12-azure-amd64/bin/java -showversion |& head -n 1))"
-DocumentInstalledItem "Ant ($(ant -version))"
+DocumentInstalledItem "Apache Ant ($(ant -version))"
 DocumentInstalledItem "Gradle ${gradleVersion}"
-DocumentInstalledItem "Maven ($(mvn -version | head -n 1))"
+DocumentInstalledItem "Apache Maven ($(mvn -version | head -n 1))"
+DocumentInstalledItem "Apache JMeter (5.2.1)"
```
AlenaSviridenko, line 89: as for the version, something like `jmeter --version | grep -o "[0-9]\.[0-9]\.[0-9]"` should do the work and output only the Jmeter version (5.2.1 in this case)

images/linux/scripts/installers/java-tools.sh
pmouawad: Hello @AlenaSviridenko ,
Ubuntu packages are far from being up to date as of JMeter.
See https://askubuntu.com/questions/690983/how-can-i-update-jmeter-in-ubuntu

Latest version of jmeter can be found here:
https://jmeter.apache.org/download_jmeter.cgi#binaries
Would it help if we provided an additional stable name (without version suffix) for bundle ?

Regards

AlenaSviridenko: Hi @pmouawad, yes, link with non-hardcoded version would be very convenient fro us, thank you!

AlenaSviridenko: Hey @pmouawad! Are there any updates on the link with non-hardcoded version?

pmouawad: Hello @AlenaSviridenko ,
There is no agreement met on this in the project team.
http://mail-archives.apache.org/mod_mbox/jmeter-dev/202005.mbox/%3CCAH9fUpYuMkA6AXEn%2Bc97Knvc_uYnbdPZqx3ruoJaQ%3D9kU8SwnQ%40mail.gmail.com%3E

Regards

AlenaSviridenko: yeah, got it, thanks for the update

images/linux/scripts/installers/java-tools.sh
brunoborges: Done!

brunoborges:
@AlenaSviridenko changes made to the PR. Thanks for the tip on the regex with grep

images/linux/scripts/installers/java-tools.sh
AlenaSviridenko, line 32: Does Jmeter require Ant 1.10.7? 
If it is not strictly required, could you please revert Ant installation to `apt`-way? APT package is only slightly behind the official versions (1.10.5 according to https://packages.ubuntu.com/search?suite=bionic&searchon=names&keywords=ant) and we are trying to avoid manual support of versions.

AlenaSviridenko: Hey @pmouawad,
I saw in your emails that creation of latest release asset can bring more issues than help. 
I am not pushing you here, but just an idea: we could use GitHub API to get the latest release version from Jmeter GitHub releases, for example, https://api.github.com/repos/SeleniumHQ/selenium/releases/latest, but to do that latest release need to be marked as `Latest`. And then we could download the Jmeter binaries using this version.

anyway, that is not necessary, I think it is possible to use tags list https://api.github.com/repos/apache/jmeter/git/refs/tags but it just not so convenient as releases.

pmouawad: Hello @AlenaSviridenko ,
No JMeter does not require Ant.
Regards

pmouawad: @AlenaSviridenko , 
the releases in github are only source releases so it wouldn't work unfortunately.

Regards

pmouawad: Hello @AlenaSviridenko,
Thanks for your suggestion, I have shared your advice on dev mailing list with following, can you confirm I am right regarding what needs to be done on our side ?

> Hello,
Alena suggest the option to mark last release as latest:

    https://github.com/actions/virtual-environments/pull/791

Anyway we currently don't distribute binaries on github to have this option work.

So to make it work , we would need to do it and use Github API in build:
1) To push additional assets (zip and tgz binaries):

    https://developer.github.com/v3/repos/releases/#edit-a-release-asset

2)  To mark latest:

    https://developer.github.com/v3/repos/releases/#edit-a-release


Thanks for your help

AlenaSviridenko: it is not necessary to distribute binaries on GitHub, all we need is an ability to get the latest released version number. So there is no need in item 1 from your list. Just marking the latest release with `latest` tag should be enough to make it work

brunoborges: @AlenaSviridenko in the Java world, we often avoid using these common tools by installing from Linux repositories and rather prefer the binary from the main origin (apache.org).

The issue with the apt-way is the dependencies that it will bring.

I highly recommend we keep it this way.

images/win/scripts/Installers/Validate-JavaTools.ps1
```diff
@@ -84,6 +84,13 @@ if( $( $(gradle -version) | Out-String) -match  'Gradle (?<version>.*)' )
    $gradleVersion = $Matches.version.Trim()
 }
 
+#if( $(jmeter -version) -match  'Apache JMeter (?<version>.*) \(.*' )
+#{
+   $jmeterVersion = '5.2.1'
```
AlenaSviridenko, line 89: please, remove the hardcoded version

images/linux/scripts/installers/java-tools.sh
```diff
@@ -57,9 +60,16 @@ rm gradleLatest.zip
 ln -s /usr/share/gradle-"${gradleVersion}"/bin/gradle /usr/bin/gradle
 echo "GRADLE_HOME=/usr/share/gradle" | tee -a /etc/environment
 
+# Install JMeter
+curl -sL https://downloads.apache.org/jmeter/binaries/apache-jmeter-5.2.1.zip -o jmeter.zip
```
AlenaSviridenko, line 64: Please, use parsing of Git release tags to get the latest released version instead of hard-coding the version. 
Something like this should the work (probably not optimized way):
```
jmeterTags="/tmp/jmeter_tags.json"
versions=""

curl -s 'https://api.github.com/repos/apache/jmeter/git/refs/tags' >> $jmeterTags
sorted=$(jq 'sort_by(.ref)' $jmeterTags)

for row in $(echo "$sorted" | jq -c '.[]'); do
    refValue=$(echo "$row" | jq -r 'select( .ref | test("v([0-9]\\.[0-9](\\.[0-9])?$)")) | .ref')
    if [ ! -z "$refValue" ]; then
        versions="${versions}$refValue "
    fi
done

cleanVersion=$(echo "$versions" | rev | cut -d '/' -f 1 | rev)

version=$(echo ${cleanVersion//'v'})
```
the above code outputs the `5.3` which is the latest version now.

images/linux/scripts/installers/java-tools.sh
AlenaSviridenko: I see, but in my understanding `--no-install-recommends` flag tells Apt to install only the required dependencies. 
Also, changing the way how it is installed now can bring breaking changes on images that can impact someone.

AlenaSviridenko:
Hi @brunoborges,
do you need any help with it? We can can finalize this PR by ourselves but please, file an issue for that.
Thanks!

AlenaSviridenko:
Closing it up due to no updates in a long time.
Please, feel free to file new issue if JMeter is needed on image.
Thanks!

PR #793, opened 2020-04-28
Updated README with new rules of tool installation.

AlenaSviridenko:
I adde few new items to tool installation guide:

- If tool license is not MIT, Apache or GNU, the tool cannot be installed on image.
- Big tools can be pre-installed after our evaluation of the image space usage.
- If a tool requires the support of more than one version, we will consider the cost of this maintenance, how often new versions bring dangerous updates.
- Please, before creating a PR with tool installation, wait for confirmation that this tool can be installed on image.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,8 +25,8 @@ In general, these are the guidelines we consider when deciding what to pre-insta
 - Tools and versions will typically be removed 6 months after they are deprecated or have reached end-of-life.
 - If a tool can be installed during the build, we will evaluate how much time is saved
  and how much space is used by having the tool pre-installed.
-- If tool license is not MIT, Apache or GNU, we will have to check it with legal.
-- If tool takes much space we will evaluate space usage and provide a decision if this tool can be pre-installed.
+- If a tool license is not MIT, Apache or GNU, we will have to check it with legal.
+- If a tool takes much space we will evaluate space usage and provide a decision if this tool can be pre-installed.
 - If a tool requires the support of more than one version, we will consider the cost of this maintenance, how often new versions bring dangerous updates.
 
 **Note:** For new tools, please, create an issue and get an approval from us to add this tool to the image before creating the pull request.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -25,6 +25,10 @@ In general, these are the guidelines we consider when deciding what to pre-insta
 - Tools and versions will typically be removed 6 months after they are deprecated or have reached end-of-life.
 - If a tool can be installed during the build, we will evaluate how much time is saved
  and how much space is used by having the tool pre-installed.
+- If tool license is not MIT, Apache or GNU, the tool cannot be installed on image.
```
maxim-lobanov, line 28: > If tool license is not MIT, Apache or GNU, the tool cannot be installed on image

`If tool license it not MIT, GNU or Apache, we will have to check it with legal`
I think some tools with custom license may be installed

> Big tools can be pre-installed after our evaluation of the image space usage.

Looks like this point is already covered above
```
If a tool can be installed during the build, we will evaluate how much time is saved
 and how much space is used by having the tool pre-installed.
```

> Please, before creating a PR with tool installation, wait for confirmation that this tool can be installed on image.

I think we should add Note below the list.

```
**Note:** For new tools, please create an issue and get approval that we want to add this tool to the image before creating the pull-request.
```

AlenaSviridenko: Thanks, fixed.

README.md
```diff
@@ -25,6 +25,11 @@ In general, these are the guidelines we consider when deciding what to pre-insta
 - Tools and versions will typically be removed 6 months after they are deprecated or have reached end-of-life.
 - If a tool can be installed during the build, we will evaluate how much time is saved
  and how much space is used by having the tool pre-installed.
+- If a tool license is not MIT, Apache or GNU, we will have to check it with legal.
```
alepauly, line 28: MIT, Apache, and GNU licenses are ok, anything else we'll have to check with lawyers 💼

AlenaSviridenko: Thanks, updated this line

PR #796, opened 2020-04-28
Software list: "do not change" label.

Darleev:
# Description
Software list improvment. Added additional line with strong recommendation.

#### Related issue:
Internal issue.

## Check list
- [+] Documentation is updated (if applicable)


images/linux/scripts/installers/1604/preparemetadata.sh
```diff
@@ -7,6 +7,7 @@
 
 source $HELPER_SCRIPTS/document.sh
 
+WriteItem "<!--- This file is autogenerated, it is strongly recommended not to modify it. -->"
```
alepauly, line 10: I think this message can be more direct, something like:
```
DO NOT EDIT - This markdown file is autogenerated
```

Darleev: I have added the string according to provided example.

Issue #782, opened 2020-04-25
[Minor] Found a typo


alexrosenfeld10:
**Describe the bug**
Was reading through the readme for virtual environment support and found a small typo. Relevant for folks using `kubectl`.

**Area for Triage**: 
https://github.com/actions/virtual-environments/blame/ubuntu18/20200406.2/images/linux/Ubuntu1804-README.md#L159

**Question, Bug, or Feature?**: 
Bug

vsafonkin:
Pull request with fix: #799

PR #799, opened 2020-04-29
Fix kubectl installation

vsafonkin:
[Issue #782](https://github.com/actions/virtual-environments/issues/782) 
[Issue #818](https://github.com/actions/virtual-environments/issues/818)

After [kubectl release v1.18](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.18.md#v1180), command `kubectl version --short` is not working without setting configuration. It produces error `Missing or incomplete configuration info.  Please point to an existing, complete config file`. In order to get kubectl version we should use flag `--client`.

Also, kubectl installation was corrected according [kubernetes docs](https://kubernetes.io/docs/tasks/tools/install-kubectl/).

Separated scripts for 16.04 and 18.04 were merged into one.

`helm init` command was deleted, because it was deprecated in Helm v3: [helm/helm-www/pull/485](https://github.com/helm/helm-www/pull/485)

Test build: [ubuntu-1804](https://github.visualstudio.com/virtual-environments/_build/results?buildId=66973&view=results)


images/linux/scripts/installers/1604/kubernetes-tools.sh
```diff
@@ -9,10 +9,10 @@ source $HELPER_SCRIPTS/document.sh
 source $HELPER_SCRIPTS/apt.sh
 
 ## Install kubectl
```
al-cheb, line 11: What is the difference between kubernetes-tools.sh? Could we combine into one?

images/linux/scripts/installers/1604/kubernetes-tools.sh
```diff
@@ -9,10 +9,10 @@ source $HELPER_SCRIPTS/document.sh
...
-apt-get install -y apt-transport-https
+apt-get install -y apt-transport-https gnupg2
```
miketimofeev, line 12: Maybe we can move these installations to basic.sh scripts to install it at the very beginning?

images/linux/scripts/installers/1604/kubernetes-tools.sh
vsafonkin: As I see, it uses different deb packages for 16.04 and 18.04:
16.04: `deb https://apt.kubernetes.io/ kubernetes-$(lsb_release -cs) main`
18.04: `deb https://apt.kubernetes.io/ kubernetes-xenial main`

al-cheb: 16.04: echo $(lsb_release -cs) = xenial
18.04: echo $(lsb_release -cs) = bionic

I suppose we can combine if only string is different

images/linux/scripts/installers/1604/kubernetes-tools.sh
vsafonkin: Done

images/linux/scripts/installers/1604/kubernetes-tools.sh
maxim-lobanov: You can use these helpers: https://github.com/actions/virtual-environments/blob/master/images/linux/scripts/helpers/os.sh

vsafonkin: @maxim-lobanov , I believe we can use `xenial` for both OS versions.
@al-cheb , done, thank you!

images/linux/scripts/installers/1604/basic.sh
```diff
@@ -57,7 +57,9 @@ apt-fast install -y --no-install-recommends \
     rpm \
     xz-utils \
     xorriso \
-    zsync
+    zsync \
+    apt-transport-https \
```
al-cheb, line 61: https://github.com/actions/virtual-environments/blob/master/images/linux/scripts/base/repos.sh already contains apt-transport-https

images/linux/scripts/installers/1804/basic.sh
```diff
@@ -130,6 +130,12 @@ apt-get install -y --no-install-recommends curl
 echo "Install parallel"
 apt-get install -y --no-install-recommends parallel
 
+echo "Install apt-transport-https"
+apt-get install -y --no-install-recommends apt-transport-https
```
al-cheb, line 134: https://github.com/actions/virtual-environments/blob/master/images/linux/scripts/base/repos.sh already contains apt-transport-https

images/linux/scripts/installers/1604/basic.sh
vsafonkin: Fixed

images/linux/scripts/installers/1804/basic.sh
vsafonkin: Fixed

al-cheb:
@vsafonkin Could we fix https://github.com/actions/virtual-environments/issues/818 in this PR?

Issue #798, opened 2020-04-28
Ubuntu 18.04 unable to resolve cognitiveservices DNS names


mikeharder:
**Describe the bug**
Ubuntu 18.04 is unable to resolve `*.cognitiveservices.azure.com` DNS names by default.  As a workaround, we are bypassing the local (stub) DNS server using the following command:

```
sudo ln -sf /run/systemd/resolve/resolv.conf /etc/resolv.conf
```

This may be related to https://bugs.launchpad.net/ubuntu/+source/systemd/+bug/1822416 and/or https://github.com/actions/virtual-environments/issues/397.

**Area for Triage**: 
Servers

**Question, Bug, or Feature?**: 
Bug

**Virtual environments affected**
- [ ] macOS 10.15
- [ ] Ubuntu 16.04 LTS
- [x] Ubuntu 18.04 LTS
- [ ] Windows Server 2016 R2
- [ ] Windows Server 2019

**Expected behavior**
Ubuntu 18.04 should be able to resolve `*.cognitiveservices.azure.com` DNS names by default.

**Actual behavior**
If we try to resolve a `*.cognitiveservices.azure.com` DNS name, it fails with `SERVFAIL`:

https://dev.azure.com/mharder/public/_build/results?buildId=634&view=logs&j=3dc411e8-b5bf-57f2-a8a7-b25d565c86b1&t=f636eda2-37c8-5cad-c3dc-807f9e9ed0bb&l=59

However, if we bypass the local (stub) DNS server using the following command:

```
sudo ln -sf /run/systemd/resolve/resolv.conf /etc/resolv.conf
```

Then the DNS name can be resolved successfully:

https://dev.azure.com/mharder/public/_build/results?buildId=634&view=logs&j=3dc411e8-b5bf-57f2-a8a7-b25d565c86b1&t=aef3c3f0-973b-547b-f96d-ab903995d1d8&l=59

This doesn't repro on Ubuntu 16.04:

https://dev.azure.com/mharder/public/_build/results?buildId=634&view=logs&j=88c4e28e-b89e-5514-cbb8-a3c153cbe716&t=a85e08da-6704-59a1-1759-4e62f4964eb3&l=59

Pipeline Sources: https://github.com/mikeharder/AzurePipelineTests/blob/f920bf50f72fe45c1d653a3bbaac9dcaf3df7682/azure-pipelines.yml

mikeharder:
[...]

PR #819, opened 2020-05-01
Systemd-resolve workaround for Ubuntu 18.

Darleev:
# Description
Bug fixing
Sometimes, stub DNS for systemd-resolve does not work properly, and it is not possible to determine what exactly is happening. It looks like a bug in systemd-resolve service.

#### Related issue:
https://github.com/actions/virtual-environments/issues/798

## Check list
- [+] Changes are tested and related VM images are successfully generated

### DIFF
--- a/images/linux/scripts/installers/1804/basic.sh
+++ b/images/linux/scripts/installers/1804/basic.sh
@@ -139,6 +139,10 @@ for cmd in curl file ftp jq netcat ssh parallel rsync shellcheck sudo telnet tim
     fi
 done
 
+# Workaround for systemd-resolve, since sometimes stub resolver does not work properly. Details: https://github.com/actions/virtual-environments/issues/798
+echo "Create resolv.conf link."
+ln -sf /run/systemd/resolve/resolv.conf /etc/resolv.conf
+
 # Document what was added to the image
 echo "Lastly, documenting what we added to the metadata file"
 DocumentInstalledItem "Basic CLI:"


Darleev:
I have deployed Ubuntu 18 agent with the workaround and it works fine.

maxim-lobanov:
Do you know the root cause why default systemd-resolve doesn't work properly?
you mentioned that it is workaround. If so, when it can be removed?

miketimofeev:
> Do you know the root cause why default systemd-resolve doesn't work properly?
> you mentioned that it is workaround. If so, when it can be removed?

Agree, @Darleev could you please provide some more data about the issue? Maybe some similar issues links.

Darleev:
> Do you know the root cause why default systemd-resolve doesn't work properly?
> you mentioned that it is workaround. If so, when it can be removed?
@miketimofeev @maxim-lobanov 
Let me describe my findings below: 
- [There is Github issue](https://github.com/systemd/systemd/issues/10298), with exactly the same problem, which was not fixed, only workaround was provided as solution.
- Root cause is unclear.
- I'm not sure that we can remove systemd-resolve without effects. but we can disable it by default. However, in this case, it is required to add manually nameserver record to `/etc/resolv.conf `file.
- In the [official documentation](https://wiki.archlinux.org/index.php/Systemd-resolved), I didn't find misconfigurations.

So, why workaround was added as solution? - This is simple solution, without any effects and further support is not required for this workaround.

PR #830, opened 2020-05-05
add minikube to ubuntu and windows images

medyagh:
# Description

#### Related issue:
- https://github.com/actions/virtual-environments/issues/834
- https://github.com/kubernetes/minikube/issues/8009
## Check list


medyagh:
thank you @maxim-lobanov, addressed the review comments.

maxim-lobanov:
@medyagh , I think that files `images/linux/ubuntu1604.json` and `images/linux/ubuntu1804.json` should be reverted too since you don't add new file.
Remain looks good to me.

One more concern from my side is size of caching `latest kubernetes images`. Recently, we had an incident with disk space on Ubuntu images and I am not sure that we can afford it.

cc: @alepauly @al-cheb @miketimofeev to take a look and address concern about disk space

maxim-lobanov:
Could you please also pull the latest master to your branch and resolve conflicts?

medyagh:
Thanks @maxim-lobanov I merged upstream. the cached images would save users from downloading the images when minikube starts. would greatly make it faster if pre-cached the images before start.

maxim-lobanov:
@medyagh , thanks! Changes look good and I have rerun CI.
We had an incident with free space on Ubuntu images recently. So we are very concerned from the tool size side about adding new tools (especially on Ubuntu). Minikube itself is pretty small tool, I am speaking about cached images.
@alepauly , @al-cheb , could you please take a look?

images/linux/scripts/installers/kubernetes-tools.sh
```diff
@@ -32,7 +37,20 @@ if ! command -v helm; then
...
 
+# Run tests to determine that the software installed as expected
+echo "Testing to make sure that minikube was installed"
+if ! command -v minikube; then
+    echo "minikube was not installed"
+    exit 1
+fi
+
+# Cache Kubernetes images (images, tarballs) into .minikube folder.
+echo "Caching Kubernetes images"
```
alepauly, line 48: @medyagh Any idea how big these are? Might make sense to leave them off the cache for now.

maxim-lobanov: @alepauly , Based on issue description, ~500 MB.
I think we should leave cache off for now because we have only about 1.5GB of free disk space on Ubuntu. That is pretty small.

alepauly: agreed. I understand the benefit of preloading but we can't afford the space at this time. @medyagh can you update please? thanks!

medyagh:
@maxim-lobanov 
@al-cheb 
@alepauly done ! removed the cache.

maxim-lobanov:
@medyagh , sorry for inconvenience, could you please pull latest master to your branch to make CI pass.

medyagh:
@maxim-lobanov  done

PR #839, opened 2020-05-06
Add nodejs tool cache for linux.

dmitry-shibanov:
# Description
Adding toolset for node.js generation. Reduced minimum free space to 17000MB because of swap file from 8 Gb to 4 Gb
## Check list


images/linux/scripts/installers/validate-disk-space.sh
```diff
@@ -5,7 +5,7 @@
 ################################################################################
 
 availableSpaceMB=$(df / -hm | sed 1d | awk '{ print $4}')
-minimumFreeSpaceMB=18000
+minimumFreeSpaceMB=17500
```
maxim-lobanov, line 8: What is the accurate amount of free disk space will we have after your changes?
I a bit worried that we will affect customers

dmitry-shibanov: 1. success image - https://github.visualstudio.com/virtual-environments/_build/results?buildId=68046&view=results
    Available disk space: 17557 MB
2. failed image - https://github.visualstudio.com/virtual-environments/_build/results?buildId=68044&view=results
    Available disk space: 17557 MB
    Not enough disk space on the image (minimum available space: 17800 MB)

PR #489, opened 2020-03-02
Updating readme file for AzP.20200225.ubuntu18.0 version 20200225

al-cheb:

PR #848, opened 2020-05-07
Install Go to hostedtoolcache directory on Ubuntu image

vsafonkin:
Internal issue: [#489](https://github.com/actions/virtual-environments-internal/issues/489)

Moved Go location from `/usr/local` to `/opt/hostedtoolcache/go` directory to work with `setup-go` action.

Example of path:
```
/opt/hostedtoolcache/go/1.14.2/x64
```

### DIFF
--- a/images/linux/scripts/installers/go.sh
+++ b/images/linux/scripts/installers/go.sh
@@ -13,8 +13,8 @@ golangTags="/tmp/golang_tags.json"
 #   $2=IsDefaultVersion (true or false)
 function InstallGo () {
     version=$( getFullGoVersion $1 )
-    downloadVersion=$version.linux-amd64.tar.gz
-    goFolder="$AGENT_TOOLSDIRECTORY/go/$version/x64"
+    downloadVersion="go$version.linux-amd64.tar.gz"
+    goFolder="$AGENT_TOOLSDIRECTORY/Go/$version/x64"
 
     echo "Install Go $version"
     curl -sL https://dl.google.com/go/${downloadVersion} -o ${downloadVersion}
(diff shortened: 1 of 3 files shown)


images/linux/scripts/installers/go.sh
```diff
@@ -13,16 +13,16 @@ golangTags="/tmp/golang_tags.json"
 #   $2=IsDefaultVersion (true or false)
 function InstallGo () {
     version=$( getFullGoVersion $1 )
-    downloadVersion=$version.linux-amd64.tar.gz
-    goFolder=/usr/local/go$1
+    downloadVersion="go$version.linux-amd64.tar.gz"
+    goFolder="$AGENT_TOOLSDIRECTORY/go/$version/x64"
```
al-cheb, line 17: Where do we define AGENT_TOOLSDIRECTORY?

images/linux/ubuntu1804.json
```diff
@@ -296,6 +283,19 @@
             ],
             "execute_command": "sudo sh -c '{{ .Vars }} {{ .Path }}'"
         },
+        {
+            "type": "shell",
+            "scripts": [
+                "{{template_dir}}/scripts/installers/go.sh"
+            ],
+            "environment_vars": [
+                "METADATA_FILE={{user `metadata_file`}}",
```
miketimofeev, line 292: Could we move this script to the section above to avoid env_vars duplication?

images/linux/scripts/installers/go.sh
```diff
@@ -13,16 +13,16 @@ golangTags="/tmp/golang_tags.json"
 #   $2=IsDefaultVersion (true or false)
 function InstallGo () {
     version=$( getFullGoVersion $1 )
-    downloadVersion=$version.linux-amd64.tar.gz
-    goFolder=/usr/local/go$1
```
alepauly, line 17: should we setup a symlink for now so that it's not such a breaking change?

maxim-lobanov: Honestly, I don't think that symlinks are required here because we have never announced these paths anywhere.
But probably symlinks won't be worse

vsafonkin: Added symlinks.

Issue #488, opened 2020-03-02
Spectre Mitigated Libraries missing 


davidhesselbomvcc:
**Describe the bug**
`msbuild.exe` spits out warnings about missing Spectre mitigation libraries when building a v141 vcxproj, and later fails the build with `LINK : fatal error LNK1104: cannot open file 'msvcprt.lib'`. This did not happen when building the very same commit 10 days ago.

**Area for Triage**: 
C/C++

**Question, Bug, or Feature?**: 
Bug

**Virtual environments affected**
- [ ] macOS 10.15
- [ ] Ubuntu 16.04 LTS
- [ ] Ubuntu 18.04 LTS
- [ ] Windows Server 2016 R2
- [X] Windows Server 2019

**Expected behavior**
The build was fine 10 days ago, and since I didn't change anything (same commit), the expected behavior is that the build should be fine today as well.

**Actual behavior**
`msbuild.exe` spits out warnings about missing Spectre mitigation libraries when building a v141 vcxproj on `windows-latest`:

> 19>C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\MSBuild\Microsoft\VC\v150\Microsoft.CppBuild.targets(402,5): warning MSB8038: Spectre mitigation is enabled but Spectre mitigated libraries are not found.  Verify that the Visual Studio Workload includes the Spectre mitigated libraries.  See https://aka.ms/Ofhn4c for more information.

It later fails the build:

> LINK : fatal error LNK1104: cannot open file 'msvcprt.lib'

The very same commit built successfully 10 days ago, so something must have changed in the VM since then (I have made no builds in the interim, so I unfortunately can't give you a more precise timeframe, but the fact that no one else has reported this problem suggests it's a recent change). If it helps, the WindowsTargetPlatformVersion specified in the vcxproj is 10.0.17134.0.

I notice there's a `/Qspectre` option among the build flags in the log now that was *not* present in the build logs 10 days ago, and the vcxproj itself has never specified the option. According to some several months old forum threads I found while googling the issue, this flag originates from a `Microsoft.Cpp.targets` or `Microsoft.CppBuild.targets` or `.props` file somewhere in the Visual Studio install folder. Perhaps such a file was modified recently on the VM?

I also found [this](https://developercommunity.visualstudio.com/content/problem/348985/installing-wdk-1809-enabled-spectre-mitigation-fla.html) discussion, but its original post is 16 months old.

[...]

PR #849, opened 2020-05-07
Install Go to hostedtoolcache directory on Windows image

vsafonkin:
Internal issue: [#488](https://github.com/actions/virtual-environments-internal/issues/488)

Moved Go location from `C:` to `C:\hostedtoolcache\go` directory to work with `setup-go` action.

Example of path:
```
C:\hostedtoolcache\go\1.14.2\x64
```

Also it uses `7z` tool instead `Expand-Archive` to decrease expand archive time (about 20 min with `Expand-Archive` and 3 min with `7z`).

### DIFF
--- a/images/win/scripts/Installers/Install-Go.ps1
+++ b/images/win/scripts/Installers/Install-Go.ps1
@@ -29,7 +29,7 @@ function Install-GoVersion
 
     # Extract the zip archive.  It contains a single directory named "go".
     Write-Host "Extracting Go $latestVersion..."
-    $toolDirectory = Join-Path $env:AGENT_TOOLSDIRECTORY "Go\$latestVersion"
+    $toolDirectory = Join-Path $env:AGENT_TOOLSDIRECTORY "go\$latestVersion"
     7z.exe x $goArchPath -o"$toolDirectory" -y | Out-Null
 
     # Rename the extracted "go" directory to "x64" for full path "C:\hostedtoolcache\windows\Go\1.14.2\x64\..."
--- a/images/win/scripts/Installers/Install-Go.ps1
+++ b/images/win/scripts/Installers/Install-Go.ps1
@@ -29,36 +29,37 @@ function Install-GoVersion
 
     # Extract the zip archive.  It contains a single directory named "go".
     Write-Host "Extracting Go $latestVersion..."
-    Expand-Archive -Path $goArchPath -DestinationPath "C:\" -Force
+    $toolDirectory = Join-Path $env:AGENT_TOOLSDIRECTORY "Go\$latestVersion"
+    7z.exe x $goArchPath -o"$toolDirectory" -y | Out-Null
+
+    # Rename the extracted "go" directory to "x64" for full path "C:\hostedtoolcache\windows\Go\1.14.2\x64\..."
+    Rename-Item -path "$toolDirectory\go" -newName "x64"
+    $fullArchPath = "$toolDirectory\x64"
 
     # Delete unnecessary files to conserve space
     Write-Host "Cleaning directories of Go $latestVersion..."
-    if (Test-Path "C:\go\doc")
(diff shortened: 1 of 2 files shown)


maxim-lobanov:
> Also it uses 7z tool instead Expand-Archive to decrease expand archive time (about 20 min with Expand-Archive and 3 min with 7z).

That sounds really good. Found that it is [known PowerShell](https://github.com/PowerShell/Microsoft.PowerShell.Archive/issues/32) issue that is not fixed for a long time :(

Do we want to switch all `Expand-Archive` to `7zip` for all tools (in scope of separate PR)? If it gives at least 1 hour of image-generation improvement, it can be good improvement. We can implement small helper function as a wrapper for 7z invocation.
cc: @alepauly , @AlenaSviridenko for any thoughts

images/win/scripts/Installers/Install-Go.ps1
```diff
@@ -29,36 +29,37 @@ function Install-GoVersion
 
     # Extract the zip archive.  It contains a single directory named "go".
     Write-Host "Extracting Go $latestVersion..."
-    Expand-Archive -Path $goArchPath -DestinationPath "C:\" -Force
+    $toolDirectory = Join-Path $env:AGENT_TOOLSDIRECTORY "go\$latestVersion"
+    7z.exe x $goArchPath -o"$toolDirectory" -y | Out-Null
```
miketimofeev, line 33: `-o"$toolDirectory"` - space missed

AlenaSviridenko:
@maxim-lobanov `7zip` sound good to me, 1 hour of win in image generation is a good speed improvement, it will allow us to check PRs faster, for example.

images/win/scripts/Installers/Install-Go.ps1
vsafonkin: I don't know why, but this flad uses without whitespace:
https://sevenzip.osdn.jp/chm/cmdline/switches/output_dir.htm

maxim-lobanov: Yep, missed space is expected here :)

images/win/scripts/Installers/Install-Go.ps1
```diff
@@ -29,36 +29,37 @@ function Install-GoVersion
 
     # Extract the zip archive.  It contains a single directory named "go".
     Write-Host "Extracting Go $latestVersion..."
-    Expand-Archive -Path $goArchPath -DestinationPath "C:\" -Force
+    $toolDirectory = Join-Path $env:AGENT_TOOLSDIRECTORY "go\$latestVersion"
```
alepauly, line 32: similar comment as in #848 - should we setup a link from the old location to the new while we communicate with users?

vsafonkin: Done!

images/win/scripts/Installers/Install-Go.ps1
```diff
@@ -29,36 +29,40 @@ function Install-GoVersion
...
     {
-        Remove-Item -Recurse -Force "C:\go\blog"
+        Remove-Item -Recurse -Force "$fullArchPath\blog"
     }
 
-    # Rename the extracted "go" directory to include the Go version number (to support side-by-side versions of Go).
-    $newDirName = "Go$latestVersion"
-    Rename-Item -path "C:\go" -newName $newDirName
+    # Create symlink in old location
+    New-Item -Path "C:\go$latestVersion" -ItemType SymbolicLink -Value $fullArchPath
```
alepauly, line 51: isn't this missing a `\`?

vsafonkin: @alepauly , no, in old location Go was installed in folders called `go1.14.2`, `go1.13.2` and etc. I have tested symlinks on created VM manually, it works correct.

PR #851, opened 2020-05-08
Install GitHUB CLI on Ubuntu

dsame:
# Description
New tool

Installation size: 16.5Mb
Installation time:  5 sec

## Check list


images/linux/scripts/installers/github-cli.sh
```diff
@@ -0,0 +1,25 @@
...
+##  File:  github-cli.sh
+##  Desc:  Installs GitHub CLI
+##         Must be run as non-root user after homebrew
+################################################################################
+
+# Source the helpers for use with the script
+source $HELPER_SCRIPTS/document.sh
+
+# Install GitHub CLI
+wget https://github.com/cli/cli/releases/download/v0.7.0/gh_0.7.0_linux_amd64.deb
```
al-cheb, line 12: Please parse the latest version and don't hardcoded version.

images/linux/scripts/helpers/github-api.sh
```diff
@@ -0,0 +1,27 @@
...
+
+    if [[ "x$1" == *"/"* ]]; then
+        url=$1
+        format=$2
+    else
+        url="$1/$2"
+        format=$3
+    fi
+
+    jq_script=".assets[].browser_download_url|select(contains(\"linux\") and contains(\"amd64\") and contains(\"${format}\"))"
```
maxim-lobanov, line 25: I like approach with helper functions but considering this condition it is very specific and can't be re-used anywhere.
Such conditions work for GH CLI but could be very different for other tools

images/linux/scripts/helpers/github-api.sh
```diff
@@ -0,0 +1,27 @@
...
+     if [[ $# -eq 0 || "x$1" = "x-h" || "x$1" = "x--help" ]];then
+        echo "usage: " >&2
+        echo "       $0 organisation/repo format" >&2
+        echo "       $0 organisation repo format" >&2
+        echo >&2
+        echo " Format should be .deb, .tar.gz or other download file extension" >&2
+        exit 1
+    fi
+
+    if [[ "x$1" == *"/"* ]]; then
```
maxim-lobanov, line 17: Do we really need to support both ways?
I think supporting only first way `url=$1, format=$2` should be enough

images/linux/scripts/helpers/github-api.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+################################################################################
+##  File:  github-api.sh
+##  Desc:  Helper functions for github api
+################################################################################
+
+function getLatestReleaseURI {
+     if [[ $# -eq 0 || "x$1" = "x-h" || "x$1" = "x--help" ]];then
```
maxim-lobanov, line 8: This function is intended to run in CI so we don't need to support help argument for it.
To be consistent, it is better to put some synopsis comment under function

images/linux/scripts/installers/github-cli.sh
```diff
@@ -0,0 +1,27 @@
...
+################################################################################
+
+# Source the helpers for use with the script
+source $HELPER_SCRIPTS/document.sh
+source $HELPER_SCRIPTS/github-api.sh
+
+# Install GitHub CLI
+url=$(getLatestReleaseURI cli/cli .deb)
+wget $url
+sudo apt install ./gh_*_linux_amd64.deb
```
maxim-lobanov, line 15: Why we need sudo here?
The whole script is run with sudo.

images/linux/scripts/installers/github-cli.sh
```diff
@@ -0,0 +1,27 @@
...
+# Run tests to determine that the software installed as expected
+echo "Testing to make sure that script performed as expected, and basic scenarios work"
+if ! gh --version; then
+    echo "GitHub CLI was not installed"
+    exit 1
+fi
+
+# Document what was added to the image
+echo "Lastly, documenting what we added to the metadata file"
+DocumentInstalledItem "GitHub CLI $(gh --version|head -1|sed -E 's/gh version ([0-9.]+).*/\1/')"
```
al-cheb, line 27: `$(gh --version|awk 'FNR==1 {print $3}')`

images/linux/scripts/helpers/github-api.sh
dsame: Nope. It works for any github release for ubuntu

images/linux/scripts/helpers/github-api.sh
dsame: I think yes. It's annoying to remember suchs small and insignificant  differences. The cost of the mistake in a call is few hours of test unstable build while the cost of the handling both cases close to zero

images/linux/scripts/helpers/github-api.sh
dsame: In fact this code intended to provide the logs easy to find a reason of fail. But ok. "-h" keys are unneeded in fact.

images/linux/scripts/installers/github-cli.sh
dsame: I prefer to avoid awk as much as possible. Its options and capabilities very differ between different implementations and cause lot of annoying unexpected behaviour. 
 Please confirm the awk is preferable in this case.

images/linux/scripts/helpers/github-api.sh
maxim-lobanov: What about 
- `https://api.github.com/repos/julialang/julia/releases/latest`
- `https://api.github.com/repos/SeleniumHQ/selenium/releases/latest`
- `https://api.github.com/repos/mozilla/geckodriver/releases/latest`
- `https://api.github.com/repos/docker/composereleases/latest`

These conditions will work only in case if release asset naming contains linux / amd64 works but it is special case when asset name satisfy this template. The most of the releases have own naming and this function won't work for them.
So this helper is not generic one because it works correctly only with this specific repo

dsame: Ah, i planned to add more or less generalised  helpers functions on demand but that would be out of scope of this PR. 

What about renaming the function to getLatestLinuxAMDReleaseURI and let it alone till we need to handle other formats

maxim-lobanov: I don't really think that we need this helper function because if we remove help and processing of different input parameters, function can be reduced to 2 lines:
```
jq_script=".assets[].browser_download_url|select(contains(\"linux\") and contains(\"amd64\") and contains(\"${format}\"))"
curl -s https://api.github.com/repos/${url}/releases/latest | jq -r "$jq_script"
```
Considering the fact that the first line will different for any repo, function won't be re-usable.

I like the direction of reworking common logic to helper functions but in this particular case, it is not needed because process of getting download link from releases if pretty specific and can't generalized

images/linux/scripts/installers/github-cli.sh
al-cheb: I confirm that the FNR function is a POSIX one and variant with awk is shorter and easier to read.

tamalsaha:
Thanks for pre-installing `gh`. When will this be available?

maxim-lobanov:
@tamalsaha , it should be deployed at the second part of next week

tamalsaha:
@maxim-lobanov , I would like to point out that `gh` will not be usable until the next release is cut. `gh`'s latest release is v0.9.0. But the pr to respect GITHUB_TOKEN env var has not been released yet. https://github.com/cli/cli/pull/909 :cry:

PR #853, opened 2020-05-08
Install GitHub CLI on Windows images

dsame:
# Description
New tool

Installation size: 16.5Mb
Installation time:  5 sec

## Check list

### DIFF
--- a/images/win/scripts/Installers/Install-GitHub-CLI.ps1
+++ b/images/win/scripts/Installers/Install-GitHub-CLI.ps1
@@ -1,0 +1,10 @@
+################################################################################
+##  File:  Install-GitHub-CLI.ps1
+##  Desc:  Install GitHub CLI
+################################################################################
+
+Import-Module -Name ImageHelpers -Force
+
+Choco-Install -PackageName gh
+
+Add-MachinePathItem "C:\Program Files (x86)\GitHub CLI"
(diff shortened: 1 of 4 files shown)


images/win/scripts/Installers/Install-GitHub-CLI.ps1
```diff
@@ -0,0 +1,10 @@
+################################################################################
+##  File:  Install-GitHub-CLI.ps1
+##  Desc:  Install GitHub CLI
+################################################################################
+
+
+Choco-Install -PackageName gh
+
+Import-Module -Name ImageHelpers -Force
```
maxim-lobanov, line 9: I think we should move Import Module to the begin of the file

dsame: I cloned this code from the others scripts, but i agree it is more convenient to have all imports on top

images/win/scripts/Installers/Validate-GitHub-CLI.ps1
```diff
@@ -0,0 +1,27 @@
...
+else
+{
+    Write-Host 'gh is not on path'
+    exit 1
+}
+
+# Adding description of the software to Markdown
+$SoftwareName = "GitHub CLI"
+
+$version = (gh --version) -replace 'gh version (\d+\.\d.\d).*', '$1'
```
al-cheb, line 21: (gh --version).Split()[3]

PR #860, opened 2020-05-10
Update Haskell tool Stack

Ailrun:
# Description
Update the version of stack, as it fixes several issues related with haddock, Cabal 3, etc.

#### Related issue:
As far as I searched, there is no related issue.

## Check list
- [N/A] Tests are written (if applicable)
  I'm not on Windows, and I don't know how to use `ps1` under linux.

### DIFF
--- a/images/linux/scripts/installers/haskell.sh
+++ b/images/linux/scripts/installers/haskell.sh
@@ -33,8 +33,8 @@ apt-get install -y \
     cabal-install-3.0 \
     cabal-install-3.2
 
-# Install haskell stack, pinned to v2.3.1
-curl -sSL https://raw.githubusercontent.com/commercialhaskell/stack/v2.3.1/etc/scripts/get-stack.sh | sh
+# Install the latest stable release of haskell stack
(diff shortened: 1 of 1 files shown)


miketimofeev:
Hi @Ailrun! 
Thanks for your contribution! 
Could you please remove changes in the Readme files? These docs are updated automatically.

Ailrun:
Oops, I didn't know that. I will remove them.

Ailrun:
Fixed

miketimofeev:
Thanks!
By the way, what do you think about switching to install the latest version instead of specifying the exact one every time? It could be done by invoking `curl -sSL https://get.haskellstack.org/ | sh`

Ailrun:
@miketimofeev I think that would be better, but how often GitHub Actions builds virtual env against the scripts?

miketimofeev:
@Ailrun Usually, updates take place once a week if there are no delays due to found bugs

Ailrun:
OK, that sounds reasonably short enough. I will update the script so that it installs stack using get.haskellstack.org

PR #866, opened 2020-05-12
Update Boost documentation

AlenaSviridenko:
# Description
Added a special note for setting BOOST_ROOT variable since CMake requires it.

## Check list

### DIFF
--- a/images/win/scripts/Installers/Validate-Boost.ps1
+++ b/images/win/scripts/Installers/Validate-Boost.ps1
@@ -55,6 +55,10 @@ foreach ($BoostTool in $BoostTools)
 $CMakeFindBoostInfo = @"
 
 #### _Notes:_
+
+> ``BOOST_PATH`` was deprecated on images but it is required by CMake. Please make sure you set this variable
+> value to either ``BOOST_ROOT_1_69_0`` or ``BOOST_ROOT_1_72_0`` depending on the Boost version you are using.
+
 Link: https://cmake.org/cmake/help/latest/module/FindBoost.html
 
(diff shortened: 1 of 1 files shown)


images/win/scripts/Installers/Validate-Boost.ps1
```diff
@@ -55,6 +55,10 @@ foreach ($BoostTool in $BoostTools)
 $CMakeFindBoostInfo = @"
 
 #### _Notes:_
+
+> ``BOOST_PATH`` was deprecated on images but it is required by CMake. Please make sure you set this variable
```
maxim-lobanov, line 59: I don't think that we should write `was deprecated` in documentation. it is more for changelog :)
I would consider something like `BOOST_ROOT` environment variable is not set up on the images.

maxim-lobanov, line 59: `BOOST_PATH` -> `BOOST_ROOT`

AlenaSviridenko: thanks, updated

AlenaSviridenko: good catch, thank you

miketimofeev:
Do we need to provide an example how to add this env variable? Or link to the docs at least

AlenaSviridenko:
> Do we need to provide an example how to add this env variable? Or link to the docs at least

Actually, I think this is more of a warning not to forget to set it, and how to set an environment variable on Windows is a bit out of scope of this note.

PR #880, opened 2020-05-13
Added zlib to ubuntu images

egor-bryzgalov:
# Description

Tool name: zlib
Add or update? Add
Desired version: Latest
Approximate size: 170 kB
URL for tool's homepage: https://zlib.net/
Virtual environments affected:

- Ubuntu 16.04 LTS
- Ubuntu 18.04 LTS

#### Related issue:
[#12879 at azure-pipelines-tasks](https://github.com/microsoft/azure-pipelines-tasks/issues/12879)


images/linux/scripts/installers/1804/basic.sh
```diff
@@ -133,6 +133,9 @@ apt-get install -y --no-install-recommends parallel
 echo "Install gnupg2"
 apt-get install -y --no-install-recommends gnupg2
 
+echo "Install zlib"
+apt-get install -y --no-install-recommends lib32z1
```
alepauly, line 137: should we be adding the 64 bit packages as well? otherwise it seems incomplete

egor-bryzgalov: @alepauly Thank you for your review! It's only required to have this one to solve the issue mentioned in the description and furthermore size of lib64z1 is about 12 MB. So could you please confirm that it's should be installed despite its size?

maxim-lobanov: @alepauly , I think it is okay to keep only `lib32z1`. It has size 120 KB.
`lib64z1` has version 12 MB. So if it is not required for this particular issue it doesn't make sense to bring both.

Issue #645, opened 2020-03-30
Add GHC, Cabal, and Stack to macOS and Windows


jared-w:
**Tool information**
- Tool name: GHC, Cabal, Stack
- Add or update? Add
- Desired version: Latest supported GHC versions (8.10.1, 8.8.3, 8.6.5), Latest version of Cabal (3.0.0.0), Latest version of Stack (2.1.3).
- Approximate size: ~250MB per GHC. ~25MB for Cabal. ~9MB for Stack.
- If this is an add request:
  - Brief description of tool: Haskell language (already present on Linux). Stack is a cross-platform program for developing Haskell projects.
  - URL for tool's homepage: https://www.haskell.org/ https://haskellstack.org

**Area for Triage**: Haskell

**Question, Bug, or Feature?**: Feature

**Virtual environments affected**
- [x] macOS 10.15
- [ ] Ubuntu 16.04 LTS
- [ ] Ubuntu 18.04 LTS
- [x] Windows Server 2016 R2
- [x] Windows Server 2019

**Can this tool be installed during the build?**
Yes. Download and install takes approximately 2-3 minutes per GHC version.

**Are you willing to submit a PR?**
Yes

---

**Further information:**

While building cross-platform support for [actions/setup-haskell](https://github.com/actions/setup-haskell/pull/12), I've gone through and figured out how to install GHC, Stack and Cabal on Linux, Windows, and macOS as uniformly as possible. It would be really nice to have the most common 3 versions of GHC available on all 3 platforms so that most CI pipelines could shave off 1-3 minutes per run.

For the PR, how would you like GHC to be installed on macOS and Windows? The most reliable way to do so on windows is through `chocolatey`, and through macOS it's `ghcup` (which is how I've set it up in the action currently).

I'm happy to split the Stack tool request into a separate issue if desired.

jared-w:
I'm not sure where to put this, but currently for Haskell on Linux, there's a lot more installed than seems to be necessary. Is that something you're interested in changing at all? It seems excessive to have 11 versions of GHC when common projects will only use 1 (or 3 at most).

Culturally, Haskell has a convention of supporting the last 3-4 major compiler releases, so at the time of writing that would suggest only needing GHC 8.10.1, 8.8.3, and 8.6.5. As for Cabal, having only versions 3.0.1.0 and 2.4.1.0 is sufficient (even then, Cabal 2.4.1.0 is only necessary for people who need backwards compatibility with old shell scripts).

[...]

Issue #595, opened 2020-03-20
Update Java to latest LTS version (Java 11) across all images


brunoborges:
As mentioned in #405, it is imperative that all images have the same Java LTS version set as default to avoid complexity in cross-platform build workflows.

Java 11.0.7 is the latest LTS release.

alepauly:
Since switching to latest LTS as default is a breaking change for many workflows and builds we won't do it until we have support and adoption from the ecosystem (e.g. `setup-java`) and can minimize disruption to existing builds. The work to improve `setup-java` is in transit, so I'll leave this open until we can get back to it.

DiederickA:
As stated above 'switching to latest LTS as default is a breaking change' it won't happen till setup-java feature is further developed. At the same time in our organization users start requesting newer Java versions (13, 14). Therefore, we would like to see these versions to be included as well (across the three platforms). Does that also wait for the setup-java feature? As this wait is yet unknown, these versions should be made available.

maxim-lobanov:
@DiederickA , We wouldn't like to pre-cache non-LTS versions (13, 14) on image. But they can be installed via `actions/setup-java` right now. 
```
- name: Set up JDK 14
  uses: actions/setup-java@v1
  with:
    java-version: 14

```

DiederickA:
Is that action setup-java also available in Azure DevOps, or is it only a GitHUB action??

maxim-lobanov:
@DiederickA , For Azure DevOps, there is a separate action [Java Tool Installer task](https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/tool/java-tool-installer?view=azure-devops) with absolutely same functionality

maxim-lobanov:
Hello @brunoborges , thank you for your proposal. Just an update for visibility:
Per internal discussion, we wouldn't like to change default Java on existing images due to unpredictable customers impact.
We are going to setup Java 11 (LTS) as default on all new images since it won't introduce breaking changes for customers.
Just an example, `ubuntu-20.04` (aka `ubuntu-latest`) has Java 11 as default. We will follow this approach for all future images.

brunoborges:
That's excellent news, thanks @maxim-lobanov !

Issue #344, opened 2020-01-31
Add Android SDK/NDK


larvacea:
**Tool information**
- Tool name: Android SDK and NDK
- Add or update? Add
- Desired version: Android SDK 29.0.1, NDK 21.0.6113669
- Approximate size: 
- If this is an add request:
  - Brief description of tool: The full SDK/NDK combination is required to build and test (via the Android Emulator) for Android.
  - URL for tool's homepage: 

**Area for Triage**:
Android

**Question, Bug, or Feature?**:
Feature

**Virtual environments affected**
- [ ] macOS 10.15
- [ ] Ubuntu 16.04 LTS
- [ ] Ubuntu 18.04 LTS
- [ ] Windows Server 2016 R2
- [X ] Windows Server 2019

**Can this tool be installed during the build?**
I have no idea; I'm now looking into building a Docker image.

**Are you willing to submit a PR?**
Possibly, if you can suggest a starting point.

hannesa2:
I want to have the desired version: Android SDK 29.0.1, NDK 21.0.6113669 too

But I don't know if this it the solution https://github.com/actions/virtual-environments/pull/169 ?
Does some one knows it ?

NotMoni:
I an also having this issue.

larvacea:
You can arrange to download the SDK and NDK versions you want. You can download sdkmanager from the source of your choice, and then use it to download the components (such as SDK and NDK) that you need. Afaik, you will not be able to run the Android emulator in a GitHub runner VM, but you certainly can gradle-build both Java and C++ into APK or AAR, as you prefer, and you can run Java-only unit-tests (what Android documentation calls a "unit test" as opposed to an "instrumented test") in the VM.

If you want a simple solution, I like [Chocolatey](https://chocolatey.org/).

If you need instrumented tests, Google does offer testing in their Firebase Test product (and they offer a free subscription for a limited number of tests per day). It's fairly straightforward to set it up and run tests, at least when I tried it on my code.

Hope this helps. I'm sorry to say that the project I've been using this on is not yet public, so I don't have examples to share.

NotMoni:
Oh, thanks. I will test and let you know.

Darleev:
@NotMoni @larvacea @hannesa2 Hello,
Let me provide you steps to install desired versions of Android SDK/NDK in runtime:
```
"$ANDROID_HOME/tools/bin/sdkmanager.bat" "build-tools;29.0.1"
[...]

PR #883, opened 2020-05-14
Add Ubuntu 20 image

miketimofeev:
# Description
Adding Ubuntu20 generation to CI

#### Related issue:
N\A

## Check list
- [N\A] Related issue / work item is attached


images/linux/scripts/installers/2004/android.sh
```diff
@@ -0,0 +1,128 @@
...
+    "build-tools;26.0.1" \
+    "build-tools;26.0.0" \
+    "build-tools;25.0.3" \
+    "build-tools;25.0.2" \
+    "build-tools;25.0.1" \
+    "build-tools;25.0.0" \
+    "build-tools;24.0.3" \
+    "build-tools;24.0.2" \
+    "build-tools;24.0.1" \
+    "build-tools;24.0.0" \
```
maxim-lobanov, line 75: Let's deprecate all build-tools and platforms less than 27.x

images/linux/scripts/installers/2004/android.sh
```diff
@@ -0,0 +1,128 @@
...
+    "extras;android;m2repository" \
+    "extras;google;m2repository" \
+    "extras;google;google_play_services" \
+    "add-ons;addon-google_apis-google-24" \
+    "add-ons;addon-google_apis-google-23" \
+    "add-ons;addon-google_apis-google-22" \
+    "add-ons;addon-google_apis-google-21" \
+    "cmake;3.6.4111459" \
```
maxim-lobanov, line 83: Let's get rid of cmake 3.6. It is legacy that we don't need to keep

images/linux/scripts/installers/2004/mercurial.sh
```diff
@@ -0,0 +1,21 @@
...
+
+# Run tests to determine that the software installed as expected
+echo "Testing to make sure that script performed as expected, and basic scenarios work"
+if ! command -v hg; then
+    exit 1
+fi
+
+# Document what was added to the image
+echo "Lastly, documenting what we added to the metadata file"
+DocumentInstalledItem "Mercurial ($(hg --version | head -n 1))"
```
maxim-lobanov, line 21: What is the difference between this file and file from `1804/mercurial.sh`. They look identical

images/linux/scripts/installers/2004/php.sh
```diff
@@ -0,0 +1,224 @@
...
+    php7.1-mysql \
+    php7.1-odbc \
+    php7.1-opcache \
+    php7.1-pgsql \
+    php7.1-phpdbg \
+    php7.1-pspell \
+    php7.1-readline \
+    php7.1-recode \
+    php7.1-snmp \
+    php7.1-soap \
```
maxim-lobanov, line 48: Currently, we have PHP 7.1, 7.2, 7.3, 7.4.
@alepauly , do you have an idea what versions can be safety deprecated?
For me looks like that 7.1 can be deprecated because this version is pretty long and not available on official site at all.
Not sure about 7.2, 7.3

images/linux/scripts/installers/azpowershell.sh
```diff
@@ -7,6 +7,26 @@
...
+            Write-Host "Az Module was not installed"; $env:PSModulePath = $actualPSModulePath; exit 1
+        }
+        $env:PSModulePath = $actualPSModulePath'; then
+        exit 1
+    fi
+
+    # Document what was added to the image
+    echo "Lastly, documenting what we added to the metadata file"
+    DocumentInstalledItem "Az Module (3.8.0)"
+    exit 0
```
maxim-lobanov, line 27: Am I right that you want to keep Azure 3.8.0 only on Ubuntu 20.04? I worry that some AzDO tasks won't work properly after that.

images/linux/scripts/installers/haskell.sh
```diff
@@ -15,38 +15,57 @@ add-apt-repository -y ppa:hvr/ghc
...
-    cabal-install-2.4 \
-    cabal-install-3.0 \
-    cabal-install-3.2
+if isUbuntu20 ; then
+    apt-get install -y \
+        ghc-8.6.5 \
+        ghc-8.8.3 \
+        ghc-8.10.1 \
+        cabal-install-2.4 \
+        cabal-install-3.0 \
```
maxim-lobanov, line 24: Based on #645, we should keep only the latest cabal on the image. Don't need to keep 3 versions

images/linux/scripts/installers/java-tools.sh
```diff
@@ -7,10 +7,10 @@
 # Source the helpers for use with the script
 source $HELPER_SCRIPTS/document.sh
 
-DEFAULT_JDK_VERSION=8
-
 set -e
 
+DEFAULT_JDK_VERSION=8
```
maxim-lobanov, line 12: @alepauly , does it make sense to set Java 11 as default version on Ubuntu 20.04 image? (based on #595)

images/linux/ubuntu2004.json
```diff
@@ -0,0 +1,324 @@
...
+        "metadata_file": "/imagegeneration/metadatafile",
+        "installer_script_folder": "/imagegeneration/installers",
+        "helper_script_folder": "/imagegeneration/helpers",
+        "vm_size": "Standard_DS2_v2",
+        "capture_name_prefix": "packer",
+        "image_version": "dev",
+        "image_os": "ubuntu20",
+        "github_feed_token": null,
+        "go_default": "1.14",
+        "go_versions": "1.12 1.13 1.14"
```
maxim-lobanov, line 28: May be it makes sense to keep only Go 1.14 for now? Any other version can be easily installed via `setup-go`

images/linux/scripts/installers/2004/android.sh
maxim-lobanov: 27.x was released on November 2017 so they are pretty long.
We can add them later if we see requests but I think they are not popular at all

maxim-lobanov:
@miketimofeev , let's also get rid of Java 7 (confirmed with Ale)

images/linux/scripts/installers/haskell.sh
miketimofeev: Done!

images/linux/scripts/installers/2004/android.sh
miketimofeev: Done!

images/linux/scripts/installers/2004/android.sh
miketimofeev: Done!

images/linux/scripts/installers/2004/android.sh
```diff
@@ -0,0 +1,97 @@
...
+    "build-tools;27.0.2" \
+    "build-tools;27.0.1" \
+    "build-tools;27.0.0" \
+    "extras;android;m2repository" \
+    "extras;google;m2repository" \
+    "extras;google;google_play_services" \
+    "add-ons;addon-google_apis-google-24" \
+    "add-ons;addon-google_apis-google-23" \
+    "add-ons;addon-google_apis-google-22" \
+    "add-ons;addon-google_apis-google-21" \
```
maxim-lobanov, line 67: `addon-google_apis-google-*` could be removed because we removed platforms 21-24.
(Starting from platform 25, they are included to build-tools, and we don't need to install them separatelly)

miketimofeev: Done!

images/linux/scripts/installers/azpowershell.sh
miketimofeev: Yeah, only the latest. Is there any way to check?

images/linux/scripts/installers/2004/php.sh
```diff
@@ -0,0 +1,224 @@
...
+    php7.1-tidy \
+    php7.1-xml \
+    php7.1-xmlrpc \
+    php7.1-xsl \
+    php7.1-zip
+apt-get remove --purge -yq php7.1-dev
+
+# Install php7.2
+apt-fast install -y --no-install-recommends \
+    php7.2 \
```
AlenaSviridenko, line 60: it seems to me that the set of PHP packages we are installing is almost the same for all versions. 
Does it makes sense to create a list of PHP versions and iterate over it to install common packages? and then install specific packages
something like
```
php_versions=("7.1" "7.2" "7.3" "7.4")
for version in php_versions; do
    php${vesrion}\
    php${vesrion}-bcmath \
    php${vesrion}-bz2 \
    php${vesrion}-cgi \
    ...
done;

#specific packages installation below
...
```

images/linux/scripts/installers/azpowershell.sh
maxim-lobanov: Not sure. Tests look green.
@alepauly , is there anyone who can help us to clarify it?

images/linux/scripts/helpers/os.sh
```diff
@@ -6,15 +6,20 @@
 
 function isUbuntu16
 {
-    lsb_release -d|grep 'Ubuntu 16' > /dev/null
+    lsb_release -d | grep -q 'Ubuntu 16'
```
alepauly, line 9: for future cleanup maybe, should we add an "IMAGE_NAME" variable in the conf that we can use for this? seems silly that we need this.

images/linux/scripts/installers/2004/android.sh
```diff
@@ -0,0 +1,89 @@
...
+fi
+
+# Install the following SDKs and build tools, passing in "y" to accept licenses.
+echo "y" | ${ANDROID_SDK_ROOT}/tools/bin/sdkmanager \
+    "ndk-bundle" \
+    "platform-tools" \
+    "platforms;android-29" \
+    "platforms;android-28" \
+    "platforms;android-27" \
+    "build-tools;29.0.3" \
```
alepauly, line 50: can we leave only one build-tools of each major version? or are all of them necessary?

images/linux/scripts/installers/2004/basic.sh
```diff
@@ -0,0 +1,173 @@
+#!/bin/bash
+################################################################################
+##  File:  basic.sh
```
alepauly, line 3: is this file any different than 18.04? Maybe we can share it?

images/linux/scripts/installers/2004/mercurial.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+################################################################################
+##  File:  mercurial.sh
```
alepauly, line 3: Share with 18.04?

images/linux/scripts/installers/2004/php.sh
alepauly: I think we should start with only 7.4. 
7.2 is out of support, and 7.3 will be supported until December. If we need to add one later I prefer that than building dependency on the older version unnecessarily:

https://www.php.net/supported-versions.php

images/linux/scripts/installers/2004/php.sh
alepauly: I like this idea but perhaps as a separate cleanup item so it doesn't block this OS.

images/linux/scripts/installers/ansible.sh
```diff
@@ -6,11 +6,15 @@
...
 # Source the helpers for use with the script
 source $HELPER_SCRIPTS/document.sh
-source $HELPER_SCRIPTS/apt.sh
+source $HELPER_SCRIPTS/os.sh
 
-# Install Ansible PPA and latest Ansible
-add-apt-repository ppa:ansible/ansible
-apt-get update
+# ppa:ansible/ansible doesn't contain packages for Ubuntu20.04
+if isUbuntu16 || isUbuntu18 ; then
```
alepauly, line 12: this smells, why are we installing from PPA? Aren't those unofficial?

images/linux/scripts/installers/azpowershell.sh
alepauly: I wouldn't worry too much, lets go for it and see what workflows out there say. This image version starts at preview only.

images/linux/scripts/installers/java-tools.sh
alepauly: yeah I think we should set 11 as default. @brunoborges would probably like that too. And it's a good preamble to switch it in 18.04 later.

images/linux/ubuntu2004.json
```diff
@@ -0,0 +1,324 @@
...
+            "temp_resource_group_name": "{{user `temp_resource_group_name`}}",
+            "capture_container_name": "images",
+            "capture_name_prefix": "{{user `capture_name_prefix`}}",
+            "virtual_network_name": "{{user `virtual_network_name`}}",
+            "virtual_network_resource_group_name": "{{user `virtual_network_resource_group_name`}}",
+            "virtual_network_subnet_name": "{{user `virtual_network_subnet_name`}}",
+            "private_virtual_network_with_public_ip": "{{user `private_virtual_network_with_public_ip`}}",
+            "os_type": "Linux",
+            "image_publisher": "canonical",
+            "image_offer": "0001-com-ubuntu-server-focal",
```
alepauly, line 52: this is now `UbuntuServer` (they finally moved it) and sku is `18.04-LTS`. Full spec from Azure:

```
                    "imageReference": {
                        "publisher": "Canonical",
                        "offer": "UbuntuServer",
                        "sku": "18.04-LTS",
                        "version": "latest"
```

images/linux/scripts/installers/ansible.sh
miketimofeev: Ansible provides this method actually
image

alepauly: ok. i guess that's how it's hosted. that's fine then. thank you!

images/linux/scripts/installers/azpowershell.sh
miketimofeev: Let's settle on the last version then

images/linux/scripts/installers/java-tools.sh
miketimofeev: Done, 11 set as default for ubuntu20

images/linux/scripts/installers/2004/android.sh
miketimofeev: Customers ask for specific versions for some reason, like here when we have 29.0.3, but 29.0.1 needed
https://github.com/actions/virtual-environments/issues/344

images/linux/ubuntu2004.json
miketimofeev: It does not work for some reason
 InvalidParameter : The following list of images referenced from the deployment template are not found: Publisher: canonical, Offer: UbuntuServer, Sku: 20.04-LTS

miketimofeev: 18.04 is already set to the 
```
            "image_publisher": "Canonical",
            "image_offer": "UbuntuServer",
            "image_sku": "18.04-LTS",
```

images/linux/scripts/helpers/os.sh
miketimofeev: we should think of it

images/linux/ubuntu2004.json
al-cheb: @alepauly, Currently I can't find the spec for `Ubuntu 20.04`. The latest available version under that spec is 19_04/19_10.

```
PS > Get-AzVMImage -Location westus -PublisherName Canonical -Offer UbuntuServer -Sku "20.04-LTS"
Get-AzVMImage : Artifact: VMImage was not found.
ErrorCode: NotFound
ErrorMessage: Artifact: VMImage was not found.
```

images/linux/scripts/installers/2004/mercurial.sh
miketimofeev: unified!

images/linux/scripts/installers/2004/mercurial.sh
miketimofeev: unified!

images/linux/scripts/installers/2004/basic.sh
miketimofeev: They are _slightly_ different, but yeah, let's combine them

images/linux/scripts/installers/java-tools.sh
brunoborges: Thanks guys. This is perfect.

images/linux/ubuntu2004.json
alepauly: > It does not work for some reason
> ` InvalidParameter : The following list of images referenced from the deployment template are not found: Publisher: canonical, Offer: UbuntuServer, Sku: 20.04-LTS`
> image
> 

Must be my mistake, I had both 18.04 and 20.04 specs open and must have rushed. Sorry for the confusion

images/linux/scripts/helpers/os.sh
al-cheb: @alepauly, With  "IMAGE_NAME" variable using multiple conditions we have got complex syntax, instead of we have now. I think we can change the approach after deprecation Ubuntu 16.04.

```
if isUbuntu16 || isUbuntu18 ; then
    add-apt-repository ppa:ansible/ansible
    apt-get update
fi
```

images/linux/scripts/installers/java-tools.sh
miketimofeev: @brunoborges @alepauly @maxim-lobanov it turned out that Android sdk requires Java8 by default. We've faced the same issue https://stackoverflow.com/questions/46402772/failed-to-install-android-sdk-java-lang-noclassdeffounderror-javax-xml-bind-a

brunoborges: You can certainly have JAVA_HOME default to 11, and somewhere in Android SDK point to the 8 install folder.

miketimofeev: We'll try to figure out possible workaround

images/linux/scripts/installers/2004/php.sh
al-cheb: @AlenaSviridenko, Could you please review?

miketimofeev:
> This is awesome. One minor tweak to consider is to remove Java 12.
> 
> There is a new Java version every 6 months, so, if it is not a big deal for your team to keep up, then just keep the latest non-LTS pre installed (today, that's 14). Otherwise, just keep the LTS ones pre installed.
> 
> I believe that users who want a non-LTS version like 12, 13 or 14, should be fine having to install manually.

We can do that in the scope of migration to Adopt Open JDK. 
@alepauly thoughts?

images/linux/scripts/installers/2004/php.sh
AlenaSviridenko: @al-cheb new php.sh is beautiful, thanks!

images/linux/scripts/installers/2004/android.sh
```diff
@@ -0,0 +1,94 @@
...
+wget -O android-sdk.zip https://dl.google.com/android/repository/sdk-tools-linux-4333796.zip
+unzip android-sdk.zip -d ${ANDROID_SDK_ROOT}
+rm -f android-sdk.zip
+
+# Add required permissions
+chmod -R a+X ${ANDROID_SDK_ROOT}
+
+if isUbuntu20 ; then
+    # Sdk manager doesn't work with Java > 8, set version 8 explicitly
+    sed -i "2i export JAVA_HOME=${JAVA_HOME_8_X64}" /usr/local/lib/android/sdk/tools/bin/sdkmanager
```
maxim-lobanov, line 35: Am I right that sdkmanager won't work for customers too and they will have to use the same workaround?

miketimofeev: It will work because we add the setting directly to the SDK manager executable script.

images/linux/scripts/helpers/os.sh
alepauly: Yeah, let's think of this for future cleanup, shouldn't block here.

PR #892, opened 2020-05-17
Support for Az Module (3.8.0)

GabrieleCastellani:
# Description
Added support for PowerShell Az Module 3.8.0
This allows to leverage the new APIs introduced in 3.8.0
 
## Check list


maxim-lobanov:
This code file definitely should be refactored :)
cc: @alepauly

images/linux/scripts/installers/azpowershell.sh
```diff
@@ -65,3 +71,4 @@ DocumentInstalledItem "Az Module (2.6.0)"
 DocumentInstalledItem "Az Module (2.8.0)"
 DocumentInstalledItem "Az Module (3.1.0)"
```
alepauly, line 72: @maxim-lobanov as an aside, we should probably get rid of 3.1 and 3.5? Support only one revision per major version.

maxim-lobanov: I agree but I think it is breaking change for customers because they point directly to 3.1 for example in their pipelines

Issue #890, opened 2020-05-16
Misidentification of "Mingw-w64" as "MinGW"


SlySven:
Not sure if this is the best place to mention it but I will be pleases to see MSYS2 added by #30 . 

However on the https://github.com/actions/virtual-environments/blob/master/images/win/Windows2019-Readme.md page it confusingly lists MinGW but that is WRONG - as the original developers of THAT project at https://osdn.net/projects/mingw/releases/ pointedly point out:
> "MinGW is a registered trademark of Software in the Public Interest Inc., registration number 86017856; this trademark has been registered on behalf of the MinGW.org Project, and its use by any other project is unauthorized."

Therefore can that be edited to refer to the correct designation as **Mingw-w64**, ~~who are, as I understand it the developers of MSYS2 as well~~ *Edit: further research suggests that is that is perhaps not the situation - but it seems possible that they do talk to each other*!

Apart from anything else using the correct name helps as it indicates that we can use a subsystem that is actively being developed and maintained and supports 64-bit processors - the last of which is not something that can be said of the original MinGW project...

Edit: should be closed by #893 .

1480c1:
It's possible the MinGW that they mention in that readme is the actual MinGW and not a Mingw-w64 toolchain

SlySven:
I was going to say that the version number (8.1.0 IIRC) *seems* more likely to be a Mingw-w64 one (though, looking at http://mingw-w64.org/doku.php/download it seems that that is actually the GCC version - and the MSYS2 builds of Mingw-w64 are now official offering a 9.x one, though having done a sort of homunculus installation of the interesting parts of GitForWindows as per https://github.com/git-for-windows/git/wiki/Install-inside-MSYS2-proper I seem now to have 10.x to compile with) - but when I go to look for a semantic version identifier for Mingw-w64 - it is, *um*, difficult to track...

The latest *version* of MinGW (from https://osdn.net/projects/mingw/releases/72860) seems to be 5.3.1 .

Wikipedia does not give precise details either - the highest that it suggests for a part of MinGW is only **5.2.1** (for "WSL" {?}) and for Mingw-w64 is **7.0.0**.

1480c1:
I believe the versioning between MinGW, Mingw-w64, and gcc are all separate (sort of), so I think the 8.1.0 they are referring to is gcc 8.1.0 and not necessarily referring to mingw-w64 or MinGW 8.1.0, but the quickest way to confirm is to check the gcc they have in their runner.

[...]

PR #893, opened 2020-05-18
Update Mingw-w64 documenation version

al-cheb:
# Description
"MinGW is a registered trademark of Software in the Public Interest Inc., registration number 86017856; this trademark has been registered on behalf of the MinGW.org Project, and its use by any other project is unauthorized."

#### Related issue:
https://github.com/actions/virtual-environments/issues/890

## Check list


images/win/Windows2016-Azure.json
```diff
@@ -532,7 +532,7 @@
         {
             "type": "powershell",
             "scripts":[
-                "{{ template_dir }}/scripts/Installers/Install-MinGW.ps1"
+                "{{ template_dir }}/scripts/Installers/Install-Mingw64.ps1"
```
SlySven, line 535: The naming is a little confusing I agree - I think the name is *technically* "MinGW-w64" {though I confess I tend to use "Mingw-w64" - and so, sometimes, does the project itself} see, e.g.: http://mingw-w64.org/doku.php/start . Confusingly the project does use `mingw64` and `mingw32` as sub-directories/system roots in the file-system into which MSYS2 is installed - but those are used to segregate the two sets of objects from the `mingw-w64-i686-xxx`  and `mingw-w64-x86_64-xxx` package-sets {and also the packages from the third repository, the less-Windows/more P.O.S.I.X. MSYS2 core itself}.

Unless there is a compelling reason not to include another hyphen I'd gently suggest the PowerShell script to be `Install-Mingw-w64.ps1`

al-cheb: @SlySven, Thank you for description .PowerShell uses a "verb-noun" naming system and we are trying to follow this naming

SlySven: Ah, I get you.

PR #896, opened 2020-05-18
Fix Stack's installation path

dibir-magomedsaygitov:
In scope of this PR we are fixing Stack's installation path.


images/win/scripts/Installers/Install-Stack.ps1
```diff
@@ -5,16 +5,18 @@
 
 Write-Host "Get the latest Stack version..."
 $StackReleasesJson = Invoke-RestMethod "https://api.github.com/repos/commercialhaskell/stack/releases/latest"
+$Version = $StackReleasesJson.name.TrimStart("v")
 $DownloadFilePattern = "windows-x86_64.zip"
 $DownloadUrl = $StackReleasesJson.assets | Where-Object { $_.name.EndsWith($DownloadFilePattern) } | Select-Object -ExpandProperty "browser_download_url" -First 1
 
 Write-Host "Download stack archive"
-$DestinationPath = Join-Path $Env:AGENT_TOOLSDIRECTORY "stack\x64"
+$ToolcachePath = Join-Path $Env:AGENT_TOOLSDIRECTORY "stack\$Version"
```
vmapetr, line 13: nitpick: technically it isn't a Toolcache path, but a path to Stack location in Toolcache folder, so maybe it should be named something like $StackToolcachePath

dibir-magomedsaygitov: thank you @vmapetr, changed name

PR #906, opened 2020-05-20
WIndows - update Install-Msys2.ps1

MSP-Greg:
# Description

1. Use 20200517 release from GitHub (https) instead of older 20190524 release (http)
2. Git tar  - change paths to full nix style
3. Git tar - requires exe's from Git/mingw64/bin, set ENV

2 & 3 may not be needed, but if anything changes...

Script will run locally assuming one has Git for Windows installed.  Since the base install is much newer, it requires fewer updates to be installed.

## Check list

### DIFF
--- a/images/win/scripts/Installers/Install-Msys2.ps1
+++ b/images/win/scripts/Installers/Install-Msys2.ps1
@@ -10,8 +10,11 @@
 $origPath = $env:PATH
 $gitPath  = "$env:ProgramFiles\Git"
 
-# get info from https://sourceforge.net/projects/msys2/files/Base/x86_64/
-$msys2Uri  = "http://repo.msys2.org/distrib/x86_64/msys2-base-x86_64-20190524.tar.xz"
+$msys2_release = "https://api.github.com/repos/msys2/msys2-installer/releases/latest"
+
+$msys2Uri = ((Invoke-RestMethod $msys2_release).assets | Where-Object {
+  $_.name -match "x86_64" -and $_.name.EndsWith("tar.xz") }).browser_download_url
+
 $msys2File = "$env:TEMP\msys2.tar.xz"
 
 # Download the latest msys2 x86_64
@@ -19,13 +22,17 @@ Write-Host "Starting msys2 download"
 (New-Object System.Net.WebClient).DownloadFile($msys2Uri, $msys2File)
 Write-Host "Finished download"
 
-$msys2FileU = "/$msys2File".replace(':', '')
+# nix style path for tar
+$msys2FileU = "/$msys2File".replace(':', '').replace('\', '/')
+
+# Git tar needs exe's from mingw64\bin
+$env:PATH = "$gitPath\usr\bin;$gitPath\mingw64\bin;$origPath"
 
 $tar = "$gitPath\usr\bin\tar.exe"
 
 # extract tar.xz to C:\
-Write-Host "Starting msys2 extraction"
-&$tar -Jxf $msys2FileU -C /c/
+Write-Host "Starting msys2 extraction from $msys2FileU"
+&$tar -xJf $msys2FileU -C /c/
 Remove-Item $msys2File
 Write-Host "Finished extraction"
 


images/win/scripts/Installers/Install-Msys2.ps1
```diff
@@ -11,21 +11,25 @@ $origPath = $env:PATH
 $gitPath  = "$env:ProgramFiles\Git"
 
 # get info from https://sourceforge.net/projects/msys2/files/Base/x86_64/
-$msys2Uri  = "http://repo.msys2.org/distrib/x86_64/msys2-base-x86_64-20190524.tar.xz"
+$msys2Uri  = "https://github.com/msys2/msys2-installer/releases/download/2020-05-17/msys2-base-x86_64-20200517.tar.xz"
```
maxim-lobanov, line 14: @MSP-Greg , what do you think about reworking it to consume the latest release always?
```
$latestRelease = Invoke-RestMethod "https://api.github.com/repos/msys2/msys2-installer/releases/latest"
$downloadAsset = $latestRelease.assets | Where-Object { $_.name -match "x86_64" -and $_.name -match "tar.xz" }                         
$msys2Uri = $downloadAsset.browser_download_url
```

MSP-Greg: Good idea.  Used REST from a lot of languages, but never PS...  Collapsed a bit, hate needing to horizontally scroll code windows...

1480c1: An alternative is using the one explicitly marked as the latest, these should be the same 

https://github.com/msys2/msys2-installer/releases/download/nightly-x86_64/msys2-base-x86_64-latest.tar.xz

images/win/scripts/Installers/Install-Msys2.ps1
```diff
@@ -10,22 +10,29 @@
...
 Write-Host "Starting msys2 download"
 (New-Object System.Net.WebClient).DownloadFile($msys2Uri, $msys2File)
 Write-Host "Finished download"
 
-$msys2FileU = "/$msys2File".replace(':', '')
+# nix style path for tar
+$msys2FileU = "/$msys2File".replace(':', '').replace('\', '/')
+
+# Git tar needs exe's from mingw64\bin
+$env:PATH = "$gitPath\usr\bin;$gitPath\mingw64\bin;$origPath"
```
1480c1, line 29: Is 7z available by this time? You can use `7z x` to extract the tarball (although not at once like `tar`)

images/win/scripts/Installers/Install-Msys2.ps1
maxim-lobanov: I don't think that we should install nightly build to image.

1480c1: The two tarballs are the same exact files.

```bash
sha256sum msys2-base-x86_64-20200517.tar.xz msys2-base-x86_64-latest.tar.xz
c4443113497acb2d2e285d40b929fc55f33f8f669902595ecdf66a655b63dc60  msys2-base-x86_64-20200517.tar.xz
c4443113497acb2d2e285d40b929fc55f33f8f669902595ecdf66a655b63dc60  msys2-base-x86_64-latest.tar.xz
```

images/win/scripts/Installers/Install-Msys2.ps1
MSP-Greg: I think if one pipes to STDOUT, 7z will extract easily.

But, given that the tar was created by MSYS2 tar, and Git tar is MSYS2, might was well extract with GIT?

1480c1: It's up to you since both should result in the same output and the only reason why I mentioned 7z is that it supports windows style paths naturally without using msys2 style paths

images/win/scripts/Installers/Install-Msys2.ps1
maxim-lobanov: image

MSP-Greg: > The two tarballs are the same exact files.

In a month that may not be the case.  Although not common, there have been reversions in some packages.  Whether that has happened with any 'base' packages, I don't recall.  With the update code, both will yield the same set of packages?

@lazka sorry for the ping.  For the starting msys2 package for Actions Windows images, would you suggest using

msys2-base-x86_64-20200517.tar.xz

or

msys2-base-x86_64-latest.tar.xz?

1480c1: https://gitter.im/msys2/msys2?at=5ec5526e9832dd6f0464673c
image

https://gitter.im/msys2/msys2?at=5ec5539bcf55346b5fbca538
image

From lazka

maxim-lobanov: @1480c1,
I don't see `msys2-base-x86_64-latest.tar.xz` file in `latest` release: https://api.github.com/repos/msys2/msys2-installer/releases/latest
Is there any reason to use `nightly-x86_64` tag instead of `latest`?

1480c1: `msys2-base-x86_64-latest.tar.xz` is in the [`nightly-x86_64`](https://github.com/msys2/msys2-installer/releases/tag/nightly-x86_64) release

One reason to use `nightly-x86_64` is to reduce the number of packages that would need to be updated, depending on how behind the `latest` release is

maxim-lobanov: I am not familiar with MSYS2 release scheme but usually stable releases of tool are distributed as a separate GitHub releases, like this one: https://github.com/msys2/msys2-installer/releases/tag/2020-05-17
As I can see, Nightly builds are built based on master branch and just update the single release: https://github.com/msys2/msys2-installer/releases/tag/nightly-x86_64 (it means no changelog, no updates history)

And as it is mentioned in repository, those builds are not tested.
As [author mentioned](https://gitter.im/msys2/msys2?at=5ec5539bcf55346b5fbca538), they publish stable releases as separate tags after "manual testing".

It means that any bug introduced by new changes, will be picked up by nightly build and then will be consumed during image generation.

al-cheb:
@MSP-Greg, 
We are planning merge the PR due to unblock Windows Server 2019 image generation process - https://github.visualstudio.com/virtual-environments/_build/results?buildId=69714&view=logs&j=011e1ec8-6569-5e69-4f06-baf193d1351e

According to https://github.com/msys2/MSYS2-packages/issues/1965#issuecomment-631558391, we must installing the latest version of msys2 to resolve the current issue. I have checked your script on local machine and if works ax expected.

maxim-lobanov:
I will merge PR as @al-cheb mentioned, feel free to raise separate issue and discussion about switching to nightly tag

eine:
@maxim-lobanov, according to https://github.com/msys2/MSYS2-packages/issues/1965#issuecomment-631558391 (referenced by @al-cheb above), the following procedure should be updated:

https://github.com/actions/virtual-environments/blob/db702848ced824526bd2bc502f0a719b3b360cd8/images/win/scripts/Installers/Install-Msys2.ps1#L48-L57

See also https://github.com/msys2/MSYS2-packages/issues/1965#issuecomment-631567674 and https://github.com/msys2/MSYS2-packages/issues/1962#issuecomment-632061175.

/cc @MSP-Greg

MSP-Greg:
@eine

From https://github.com/msys2/MSYS2-packages/issues/1965#issuecomment-631558391, it would imply that lines 48-50 could be removed.  Note that the current code in master is using 20200517...

eine:
@MSP-Greg, the suggested procedure is two comments below: https://github.com/msys2/MSYS2-packages/issues/1965#issuecomment-631576545 and it involves:

- Merging 49-50.
- Removing 56, modifying 57 and modifying 55 accordingly.

Also, 53 should be modified according to https://github.com/msys2/MSYS2-packages/issues/1962#issuecomment-632061175, because *"situation with leftover processes can change in the future"*.

eine:
Ref #355

PR #909, opened 2020-05-20
Add flag validation diskspace and set false(skip) step by default

al-cheb:
# Description
This flag allows to skip disk space validation step during the build process. Default value of the flag is set `false`. 

#### Related issue:
https://github.com/actions/virtual-environments-internal/issues/580

## Check list


images.CI/build-image.ps1
```diff
@@ -54,6 +54,7 @@ packer build    -var "capture_name_prefix=$ResourcesNamePrefix" `
                 -var "virtual_network_name=$VirtualNetworkName" `
                 -var "virtual_network_resource_group_name=$VirtualNetworkRG" `
                 -var "virtual_network_subnet_name=$VirtualNetworkSubnet" `
+                -var "run_validation_diskspace=true" `
```
maxim-lobanov, line 57: What about customers who use our `Images.CI` scripts? Can we pass this variable from some environment variable?

PR #913, opened 2020-05-20
Add Android SDK 29.0.1 and NDK 21

Darleev:
# Description
Improvement
We have only Android NDK installed with Visual Studio, but version is too old(16.1.4479499). Need to install the latest one.
Also, by unknown reason we missed Android SDK 29.0.1 in our installation steps. So, added it.
NDK installation time is about 4-5 minutes.

#### Related issue:
https://github.com/actions/virtual-environments/issues/344
## Check list
- [+] Changes are tested and related VM images are successfully generated

### DIFF
--- a/images/win/scripts/Installers/Update-AndroidSDK.ps1
+++ b/images/win/scripts/Installers/Update-AndroidSDK.ps1
@@ -36,9 +36,6 @@ Expand-Archive -Path .\android-sdk-licenses.zip -DestinationPath 'C:\Program Fil
 
 $sdk_root = "C:\Program Files (x86)\Android\android-sdk"
 
-# NDK version to be installed.
-$ndk_version = '21.1.6352462'
-
 Push-Location -Path $sdk.FullName
 
 & '.\tools\bin\sdkmanager.bat' --sdk_root=$sdk_root `
@@ -98,10 +95,10 @@ Push-Location -Path $sdk.FullName
     "cmake;3.6.4111459" `
     "cmake;3.10.2.4988404" `
     "patcher;v4" `
-    "ndk;$ndk_version"
+    "ndk-bundle"
 
     # Android NDK root path.
-    $ndk_root = "C:\Program Files (x86)\Android\android-sdk\ndk\$ndk_version"
+    $ndk_root = "C:\Program Files (x86)\Android\android-sdk\ndk-bundle"
 
     if(Test-Path $ndk_root){
     
--- a/images/win/scripts/Installers/Update-AndroidSDK.ps1
+++ b/images/win/scripts/Installers/Update-AndroidSDK.ps1
@@ -102,12 +102,9 @@ Push-Location -Path $sdk.FullName
 
     if(Test-Path $ndk_root){
     
-        $androidNDKs = Get-ChildItem -Path $ndk_root | Sort-Object -Property Name -Descending | Select-Object -First 1
-        $latestAndroidNDK = $androidNDKs.FullName;
-    
         setx ANDROID_HOME $sdk_root /M
-        setx ANDROID_NDK_HOME $latestAndroidNDK /M
-        setx ANDROID_NDK_PATH $latestAndroidNDK /M
+        setx ANDROID_NDK_HOME $ndk_root /M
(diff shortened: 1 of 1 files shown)


images/win/scripts/Installers/Update-AndroidSDK.ps1
```diff
@@ -111,7 +97,25 @@ Push-Location -Path $sdk.FullName
     "add-ons;addon-google_apis-google-21" `
     "cmake;3.6.4111459" `
     "cmake;3.10.2.4988404" `
-    "patcher;v4"
+    "patcher;v4" `
+    "ndk;$ndk_version"
```
maxim-lobanov, line 101: I don't think that we should install ndk this way. I believe it is better to just use `ndk-bundle`: https://github.com/actions/virtual-environments/blob/master/images/linux/scripts/installers/1804/android.sh#L45

It will picked up the latest version and we won't need to hardcode it

Darleev: Agree, I have tested that ndk-bundle contains only lastest version of NDK. I have changed it.

images/win/scripts/Installers/Update-AndroidSDK.ps1
```diff
@@ -111,7 +94,25 @@ Push-Location -Path $sdk.FullName
...
-    "patcher;v4"
+    "patcher;v4" `
+    "ndk-bundle"
+
+    # Android NDK root path.
+    $ndk_root = "C:\Program Files (x86)\Android\android-sdk\ndk-bundle"
+
+    if(Test-Path $ndk_root){
+    
+        $androidNDKs = Get-ChildItem -Path $ndk_root | Sort-Object -Property Name -Descending | Select-Object -First 1
```
maxim-lobanov, line 105: I am not sure that we need it. What folder will be chosen here?
I think env variables should point to $ndk_root

Darleev: I think so, thank you. changed this.

images/win/scripts/Installers/Update-AndroidSDK.ps1
```diff
@@ -111,7 +94,22 @@ Push-Location -Path $sdk.FullName
...
     "cmake;3.6.4111459" `
     "cmake;3.10.2.4988404" `
-    "patcher;v4"
+    "patcher;v4" `
+    "ndk-bundle"
+
+    # Android NDK root path.
+    $ndk_root = "C:\Program Files (x86)\Android\android-sdk\ndk-bundle"
+
+    if(Test-Path $ndk_root){
```
maxim-lobanov, line 103: Could you please fix style a bit to be consistent with other code:
```
if (Test-Path $ndk_root) {
...
} else {
}
```
Mostly spaces and new line symbols

Darleev: thank you, I have changed this.

StephenHodgson:
This change broke Android builds using the Unity Editor

Darleev:
@StephenHodgson could you please create the issue for us to investigate it?

StephenHodgson:
Sure. Although I did mitigate the issue on my end by utilizing the Unity Hub to install any dependencies #1013

Darleev:
@StephenHodgson thank you for the issue, but it seems we are not able to install Unity Hub on images because the software is not distributed for free. It seems only [paid licenses](https://unity3d.com/legal/terms-of-service) are available for Unity hub.

StephenHodgson:
That's untrue. The images already contain Unity installs from Visual Studio, but adding the Hub is fairly simple. See this Powershell script I wrote [here](https://github.com/StephenHodgson/UnityCI/blob/master/InstallUnityHub.ps1).

I am able to install any version as needed, but really I'd like to see the Hub itself installed on the image so it's one less step.

Darleev:
@StephenHodgson thank you very much for the script provided, we will discuss the possibility to install Unity Hub on agent machines, but it takes some time to make a final decision. We will keep you informed in scope of the issue created by you.

StephenHodgson:
Much appreciated, thank you.

Issue #895, opened 2020-05-18
Old cargo cache left over from the build in Windows and Linux


mzabaluev:
**Describe the bug**
Parts of the `~/.cargo` file tree baked into the virtual environments `windows-latest` and `ubuntu-latest` contain cached crate registry data left over from the build.

**Area for Triage**: 
Rust

**Question, Bug, or Feature?**: 
Bug

**Virtual environments affected**
- [ ] macOS 10.15
- [ ] Ubuntu 16.04 LTS
- [x] Ubuntu 18.04 LTS
- [ ] Windows Server 2016 R2
- [x] Windows Server 2019

**Expected behavior**
The only contents of the `~/.cargo` directory that should be present in the pristine virtual environment are the Rust toolchain binaries in `bin/` and the `env` script, and possibly some configuration dotfiles. The subdirectories `registry` and `git` cache dependencies for package builds and should either be left empty or not be present.

**Actual behavior**
A description with steps to reproduce the issue.

1. Run this GitHub workflow:

```yaml
name: Test cargo behavior

on: push

jobs:
  generate-lockfile:
    strategy:
      matrix:
        expunge-index: [true, false]
    runs-on: windows-latest
    steps:

      - uses: actions/checkout@v2

      - name: Install Rust toolchain
        uses: actions-rs/toolchain@v1
        with:
          profile: minimal
          toolchain: stable
          override: true

      - if: ${{ matrix.expunge-index }}
        name: Clear cargo registry index
        run: |
          rm -r -fo ~\.cargo\registry\index

      - run: cargo generate-lockfile -vv
```

2. Check the time it takes to run `cargo generate-lockfile` in each of the two jobs.

Cargo attempts to check out the head of the [crates.io registry index repository](https://github.com/rust-lang/crates.io-index) into a predictably named directory under `~/.cargo/registry/index`.
Apparently, if the working copy directory is already present, but its local head commit is way back in the current history of the frequently updated index, the git backend ends up doing something that works way slower on Windows than a clean clone. This should probably be fixed in cargo, but it will take more effort than doing a bit more post-build cleanup here.

There are also crate sources left in `~/.cargo/registry/cache` and `~/.cargo/registry/src`; the whole thing looks like a leftover from a `cargo install` performed during the build of the virtual environment. These files need to be removed, as they only bloat the images.

mzabaluev:
[...]

PR #914, opened 2020-05-20
Cleanup Cargo registry data folders

vmapetr:
# Description
This PR add cleanup of Cargo cached registry data folders `Index` and `Src` in Rust provisioners for Windows and Linux images.

#### Related issue:
[Old cargo cache left over from the build in Windows and Linux #895](https://github.com/actions/virtual-environments/issues/895) 

## Check list


images/win/scripts/Installers/Install-Rust.ps1
```diff
@@ -44,5 +44,10 @@ if exist $env:RUSTUP_HOME (
 $cmdPath = "C:\Rust\rustsym.bat"
 $cmdRustSymScript | Out-File -Encoding ascii -FilePath $cmdPath
 
+# Cleanup Cargo crates cache
+$cargoRegistryPath = Join-Path $env:CARGO_HOME "registry"
+Remove-Item "$cargoRegistryPath/index/*" -Recurse -Force
```
maxim-lobanov, line 49: Use `Windows`-style paths `$cargoRegistryPath\index\*`

images/linux/scripts/installers/rust.sh
```diff
@@ -32,6 +32,10 @@ for cmd in rustup rustc rustdoc cargo rustfmt cargo-clippy bindgen cbindgen; do
     fi
 done
 
+# Cleanup Cargo cache
+rm -rf "${CARGO_HOME}/registry/index/*"
+rm -rf "${CARGO_HOME}/registry/src/*"
```
mzabaluev, line 37: I think this should remove the entire `${CARGO_HOME}/registry`.
`${CARGO_HOME}/registry/cache` will only contain crate archives that happened to be downloaded to build and install `bindgen`/`cbindgen`. These can be only occasionally useful in crate builds in a workflow, which may well use different versions of the same crates depending on what their `Cargo.lock` has settled on.
I think it's better to clean these up entirely from the virtual environment and direct the users to [caching schemes](https://github.com/actions/cache/blob/master/examples.md#rust---cargo).

mzabaluev:
Please be aware that removing a pre-populated `~/.cargo/registry/index` may make cargo jobs run slower due to the need to clone the index repository's master branch from scratch. However, when the baked-in index checkout gets too old, it may actually run slower on Windows than a clean clone. The maintainers of `crates.io-index` also periodically reroot the master branch on a squashed commit, so a single-branch clone or fetch performed by cargo should not get exceedingly large.

But again, the right answer to speeding up cargo index sync is caching.

images/linux/scripts/installers/rust.sh
vmapetr: Thanks!

mzabaluev:
[This workflow run](https://github.com/mzabaluev/github-actions-playground/actions/runs/111944173) simulates the effects of the cleanup on cargo build speeds, against a caching scheme that has all dependencies for the build restored prior to running cargo. The crate's `Cargo.lock` lists 71 dependencies, a medium-sized amount. There is surprisingly little difference in build speeds. Furthermore, on Windows the cache-assisted build actually takes longer. This is probably a bug in cargo.

mzabaluev:
> There is surprisingly little difference in build speeds. Furthermore, on Windows the cache-assisted build actually takes longer.

[Solved this](https://github.com/mzabaluev/github-actions-playground/actions/runs/112187488) by adding the `--offline` flag to build commands.

mzabaluev:
I've got [a test case](https://github.com/mzabaluev/github-actions-playground/blob/ccbdc0e6d1a649516b158575095777fa1b18b422/.github/workflows/test-tar.yml) that can be used to verify the fix.

mzabaluev:
Here's [a workflow](https://github.com/mzabaluev/github-actions-playground/blob/1f6d9fbab2699000ef9a51015e08ce41c15418b4/.github/workflows/test-tar.yml) to verify the fix with. It's poorly suitable for a regression test though, as it spends a lot of time on cargo syncing the index over the network.

Issue #905, opened 2020-05-20
7z binary (7zip) is not present on windows-latest runner beginning 2020-05-19


lopopolo:
**Describe the bug**

Invoking `7z` with `shell: bash` on `windows-latest` os produces this error:

```
/c/msys64/usr/bin/7z: line 2: /usr/lib/p7zip/7z: No such file or directory
##[error]Process completed with exit code 127.
```

https://github.com/actions/virtual-environments/issues/898 mentions an msys upgrade. Is this possibly related?

Failing run: https://github.com/artichoke/nightly/runs/691297319

This is the first failing run of a job that has not changed. This failing run has the following environment:

```
Current runner version: '2.262.1'
Operating System
  Microsoft Windows Server 2019
  10.0.17763
  Datacenter
Virtual Environment
  Environment: windows-2019
  Version: 20200517.1
  Included Software: https://github.com/actions/virtual-environments/blob/win19/20200517.1/images/win/Windows2019-Readme.md
```

Yesterday's successful run: https://github.com/artichoke/nightly/runs/687237362?check_suite_focus=true

This run used the following environment:

```
Current runner version: '2.262.1'
Operating System
  Microsoft Windows Server 2019
  10.0.17763
  Datacenter
Virtual Environment
  Environment: windows-2019
  Version: 20200430.2
  Included Software: https://github.com/actions/virtual-environments/blob/win19/20200430.2/images/win/Windows2019-Readme.md
```

**Area for Triage**: 

**Question, Bug, or Feature?**:  Bug

**Virtual environments affected**
- [ ] macOS 10.15
- [ ] Ubuntu 16.04 LTS
- [ ] Ubuntu 18.04 LTS
- [ ] Windows Server 2016 R2
- [x] Windows Server 2019

**Expected behavior**
I expect invoking `7z` to produce a zip file.

**Actual behavior**
A description with steps to reproduce the issue. If your have a public example or repo to share,
please provide the link.

```
  staging="artichoke-nightly-x86_64-pc-windows-msvc"
  mkdir -p "$staging"/
  cp artichoke/{README.md,LICENSE} "$staging/"
  if [ "windows-latest" = "windows-latest" ]; then
    cp "artichoke/target/x86_64-pc-windows-msvc/release/artichoke.exe" "$staging/"
    cp "artichoke/target/x86_64-pc-windows-msvc/release/airb.exe" "$staging/"
    7z a "$staging.zip" "$staging"
[...]

PR #916, opened 2020-05-21
[Bug] Windows - Install-Msys2.ps1 - remove msys2 p7zip / 7z package

MSP-Greg:
# Description
New tool, Bug fixing, or Improvement?  Bug

Due to conflicts with two installed versions

See https://github.com/actions/virtual-environments/issues/905

### DIFF
--- a/images/win/scripts/Installers/Install-Msys2.ps1
+++ b/images/win/scripts/Installers/Install-Msys2.ps1
@@ -18,7 +18,7 @@ $msys2Uri = ((Invoke-RestMethod $msys2_release).assets | Where-Object {
 $msys2File = "$env:TEMP\msys2.tar.xz"
 
 # Download the latest msys2 x86_64
-Write-Host "Starting msys2 download"
+Write-Host "Starting msys2 download using $($msys2Uri.split('/')[-1])"
 (New-Object System.Net.WebClient).DownloadFile($msys2Uri, $msys2File)
 Write-Host "Finished download"
 
@@ -45,9 +45,9 @@ bash.exe -c "pacman-key --init 2>&1"
 Write-Host "bash pacman-key --populate msys2"
 bash.exe -c "pacman-key --populate msys2 2>&1"
 
-Write-Host "pacman --noconfirm -Sy pacman"
-pacman --noconfirm -Sy pacman
-pacman --noconfirm -Su
+Write-Host "pacman -Sy --noconfirm --needed pacman"
+pacman -Sy --noconfirm --needed pacman
+pacman -Su --noconfirm
 
 # Force stop gpg-agent to continue installation
 Get-Process gpg-agent -ErrorAction SilentlyContinue | Stop-Process -Force
@@ -58,6 +58,9 @@ pacman.exe -Syuu --noconfirm
 
 Write-Host "Install msys2 packages"
 pacman.exe -S --noconfirm --needed --noprogressbar base-devel compression
+
+Write-Host "Remove p7zip/7z package due to conflicts"
+pacman.exe -R --noconfirm --noprogressbar p7zip
 
 # mingw package list
 $tools = "___clang ___cmake ___llvm  ___toolchain ___ragel"


MSP-Greg:
Rebased (duh) and added two minor changes.  Worked locally...

MSP-Greg:
**Reminder**

Once MSYS2 has been rolled out, the current Windows Rubies can be replaced with standard installs.

The 'vendored' code in RubyInstaller2 Rubies will find the MSYS2 install, so there won't be an issues.  JFYI, the code looks for an embedded MSYS2 first, then checks C:/msys64.

In the most recent release:
https://github.com/oneclick/rubyinstaller2/releases/tag/RubyInstaller-2.7.1-1

The current download, rubyinstaller-devkit-2.7.1-1-x64.exe, is 131 MB, while the standard install, rubyinstaller-2.7.1-1-x64.7z, is 11 MB.  Given that there are four Rubies installed (2.4 thru 2.7), that should free up a lot of disk space, and remove a lot of duplicated files.

I'd do a PR for it, but I can't find the code...

al-cheb:
@MSP-Greg , I think we don't need to remove p7zip to unblock only one particular customer case.

Try to count how many apps contains raise the same issue in the msys64 directory:
```
PS > dir C:\msys64\usr\bin | measure | % count
778
```

We can easily reproduce the issue for example with  `make` util:

```
runs-on: [windows-latest]
    steps:
    - name: Install Ruby toolchain
      uses: ruby/setup-ruby@v1
      with:
          ruby-version: "2.6.3"
    - name: make
      run: |
         which make
         make --version
      shell: bash 
```
m - Copy

Normal behavior:
```
runs-on: [windows-latest]
    steps:
    - name: make
      run: |
         which make
         make --version
      shell: bash
```

s - Copy

As I know you provided the request to change `git -> bash` to `msys64 -> bash` that can resolve this  issue with correct initialization `msys2` environment when we use bash shell.

MSP-Greg:
@al-cheb

Everything you said is true.  I thought an exception for p7zip was warranted because:

1. I've used MSYS@ locally for a few years, and I've never installed the package.
2. It's an old version of 7z, which also implies other users of MSYS2 don't have a need for it.  Current 7zip version is 19.00, p7zip is 16.02.
3. 7z is a 'native' windows app, and there's already two copies on the image.

With Windows CI, there's always going to be issues with GNU dev/build tools.  Or, I think 7z is in a different category than things like gcc, make, openssl, etc...

My POV is also biased from helping macOS/nix types get Windows CI working, which probably shows in how ruby/setup-ruby works with Windows.

Long story short, if you want to close, that's ok by me...

eregon:
The `7z` from MSYS2 seems clearly broken though, right?
```
/c/msys64/usr/bin/7z: line 2: /usr/lib/p7zip/7z: No such file or directory
##[error]Process completed with exit code 127.
```
What causes that?
Is there a way to make it work or is it completely unusable? If the latter I'd argue it should not be part of the installed packages.

Why does `make` fail above? I understand it will pick MSYS2's `make`, but shouldn't that make work too?

IIRC it seems most software already has to care to use `make.exe` explicitly on Windows to not get some half-broken gmake (by setting the `MAKE` env var). Which still seems unfortunate and confusing.

Since GitHub Actions is adding a MSYS2 installation, I would assume adding it to PATH should work reasonably well. If basic tools like `7z` or `make` fail no matter the arguments it seems we need a way to fix that. AFAIK we had no such issues with the builtin Ruby MSYS2 DevKits from RubyInstaller.

MSP-Greg:
@eregon

The problem is using the bash shell, which is from Git, it's not the MSYS2 bash shell.  See https://github.com/actions/runner/issues/497

eregon:
Because the Git Bash does some initialization that MSYS2 binaries don't like, or conversely those binaries need some setup that MSYS2 bash does?

Why don't we see this issue with the RubyInstaller DevKit but we see them with GitHub Actions MSYS2? Simply because it has more packages?

MSP-Greg:
Do you know of any repos using `shell: bash` in any workflow steps?

Puma, Ruby, RubyGems, & ruby-loco are all working fine?  I had to update setup-ruby-pkgs because of a GPG key issue (ruby 2.4 OpenSSL 1.0.2), and it builds the OpenSSL gem as a test, also ok...

MSP-Greg:
BTW:

1. I got the version number of the MSYS2 p7zip package locally, so it at least that much works.

2. Action custom shells - They appear to not work, as I just tried.  Googled it, and found an issue I filed that I  forgot about: https://github.com/actions/runner/issues/460.  It seems to not be affected by changes to PATH by previous steps?

eine:
> As I know you provided the request to change `git -> bash` to `msys64 -> bash` that can resolve this issue with correct initialization `msys2` environment when we use bash shell.

Note that MSYS2 provides three different shells: MSYS, MINGW64 and MINGW32. Replacing `git -> bash` with `msys64 -> bash` is not enough, unless it defaults to MINGW64.

> Since GitHub Actions is adding a MSYS2 installation, I would assume adding it to PATH should work reasonably well.

Ref #908.

MSYS2 provides a built-in mechanism to either inherit the PATH or set a new "clean" env for the shell. For reference, `path-type` is an option in [eine/setup-msys2](https://github.com/eine/setup-msys2) and it defaults to `strict`. [eine/setup-msys2](https://github.com/eine/setup-msys2) reuses the "default installation" when `update: true`. Examples of how the action is used can be found in https://github.com/ghdl/ghdl/actions, https://github.com/ghdl/ghdl-cosim/actions or https://github.com/ghdl/setup-ghdl-ci/actions.

eine:
@al-cheb, see also actions/toolkit#318.

MSP-Greg:
I made changes to the script and ran it on Actions.  Tried to bullet-proof it, changed the output a bit, including the logging of packages installed.

See [Actions step here](https://github.com/MSP-Greg/actions-msys2-pkgs/runs/698489954?check_suite_focus=true#step:3:1)

I can:
1. Do nothing
2. Push it here
3. Remove the the removal of p7zip and open a new PR

images/win/scripts/Installers/Install-Msys2.ps1
```diff
@@ -45,9 +45,9 @@ bash.exe -c "pacman-key --init 2>&1"
 Write-Host "bash pacman-key --populate msys2"
 bash.exe -c "pacman-key --populate msys2 2>&1"
 
-Write-Host "pacman --noconfirm -Sy pacman"
-pacman --noconfirm -Sy pacman
-pacman --noconfirm -Su
+Write-Host "pacman -Sy --noconfirm --needed pacman"
+pacman -Sy --noconfirm --needed pacman
```
eine, line 49: As discussed in https://github.com/actions/virtual-environments/pull/906#issuecomment-632409785, pacman should NOT be updated first.

al-cheb: @eine, Thank you for providing details. I think we can include these changes to the PR.

MSP-Greg:
> I think we can include these changes to the PR.

I've got changes that I can push as a separate commit here, the changes I'm referring to were in the CI I mentioned above, see See [Actions step here](https://github.com/MSP-Greg/actions-msys2-pkgs/runs/698489954?check_suite_focus=true#step:3:1).

The file is at https://github.com/MSP-Greg/actions-msys2-pkgs/blob/master/Install-Msys2.ps1#L50-L55, highlighted are the changes /additions, along with multiple `taskkill /f /fi "MODULES eq msys-2.0.dll"` statements.

Note that additional changes were made to logging, especially the package lists.

In general, the main issue here is that MSYS2 got to a state where 'standard update' code did not work (which is fixed in 20200517)).  I've been using MSYS2 for several years, and I think this is only the second time that has happened.

So, I'm not sure we can be certain that whatever code we use can be guaranteed to install/update MSYS2 at every moment in time in the future.  I haven't looked to see if MSYS2 has code checking this, if not, I'll suggest/contribute there.

@eine, any recollection of past issues?

@al-cheb - Thanks, can you have a look at the links?

al-cheb:
@MSP-Greg, Thank you for updates. I have run the  [Instal-Msys2.ps1]( https://github.com/MSP-Greg/actions-msys2-pkgs/blob/master/Install-Msys2.ps1#L50-L55) script on a local machine. It works as expected without pre-installed 7z.


123 - Copy

MSP-Greg:
@al-cheb

Should I move the changes here with another commit?  Are you ok with the other changes (logging, taskkill, etc)?

It's Friday, be nice for all of us if we can resolve this today...

al-cheb:
@MSP-Greg , LGTM. I think we can include all changes in this PR. 

@maxim-lobanov, What do you think?

eine:
> @eine, any recollection of past issues?

Overall, I believe that whoever that will take care of maintaining MSYS2 support in this virtual environment, should watch the main repos of the project and should have close contact with the maintainers (through issues, chat or private channels).

As you said, updating MSYS2 does NOT fail often, but it is a rolling distribution, and it is expected to have issues more frequently than frozen distributions. This is specially so because the (your) decision was to auto-update the environment by default and to include additional packages, as opposed to providing a simple and stable solution. I already explained in our previous discusions that such an approach would imply complications in the maintenance.

To be precise, @lazka is working hard these last weeks, and the only mechanism to be aware of it is to get partially involved: https://github.com/msys2/MSYS2-packages/issues/1962#issuecomment-632703947

MSP-Greg:
I haven't pushed another commit, but below is patch file:
```patch
diff --git a/images/win/scripts/Installers/Install-Msys2.ps1 b/images/win/scripts/Installers/Install-Msys2.ps1
index 8c5f859..90971b4 100644
--- a/images/win/scripts/Installers/Install-Msys2.ps1
+++ b/images/win/scripts/Installers/Install-Msys2.ps1
@@ -1,12 +1,14 @@
 ##  File:  Install-Msys2.ps1
-##  Desc:  Install Msys2 and 64-bit gcc, cmake, & llvm (32-bit commented out)
+##  Desc:  Install Msys2 and 64 & 32 bit gcc, cmake, & llvm
 
 # References
 # https://github.com/msys2/MINGW-packages/blob/master/azure-pipelines.yml
 # https://packages.msys2.org/group/
 
+$dash = "—" * 40
+
 $origPath = $env:PATH
 $gitPath  = "$env:ProgramFiles\Git"
 
@@ -17,7 +19,7 @@ $msys2Uri = ((Invoke-RestMethod $msys2_release).assets | Where-Object {
 
 $msys2File = "$env:TEMP\msys2.tar.xz"
 
-# Download the latest msys2 x86_64
+# Download the latest msys2 x86_64, filename includes release date
 Write-Host "Starting msys2 download using $($msys2Uri.split('/')[-1])"
 (New-Object System.Net.WebClient).DownloadFile($msys2Uri, $msys2File)
 Write-Host "Finished download"
@@ -31,7 +33,7 @@ $env:PATH = "$gitPath\usr\bin;$gitPath\mingw64\bin;$origPath"
 $tar = "$gitPath\usr\bin\tar.exe"
 
 # extract tar.xz to C:\
-Write-Host "Starting msys2 extraction from $msys2FileU"
+Write-Host "Starting msys2 extraction"
 &$tar -xJf $msys2FileU -C /c/
 Remove-Item $msys2File
 Write-Host "Finished extraction"
@@ -39,55 +41,52 @@ Write-Host "Finished extraction"
 $env:PATH = "C:\msys64\mingw64\bin;C:\msys64\usr\bin;$origPath"
 
-Write-Host "bash pacman-key --init"
+Write-Host "`n$dash bash pacman-key --init"
 bash.exe -c "pacman-key --init 2>&1"
 
 Write-Host "bash pacman-key --populate msys2"
 bash.exe -c "pacman-key --populate msys2 2>&1"
 
-Write-Host "pacman -Sy --noconfirm --needed pacman"
-pacman -Sy --noconfirm --needed pacman
-pacman -Su --noconfirm
-
-# Force stop gpg-agent to continue installation
-Get-Process gpg-agent -ErrorAction SilentlyContinue | Stop-Process -Force
-
-Write-Host "pacman --noconfirm -Syyuu"
+Write-Host "`n$dash pacman --noconfirm -Syyuu"
 pacman.exe -Syyuu --noconfirm
-pacman.exe -Syuu --noconfirm
+taskkill /f /fi "MODULES eq msys-2.0.dll"
+Write-Host "`n$dash pacman --noconfirm -Syuu (2nd pass)"
+pacman.exe -Syuu  --noconfirm
+taskkill /f /fi "MODULES eq msys-2.0.dll"
 
-Write-Host "Install msys2 packages"
+Write-Host "`n$dash Install msys2 packages"
 pacman.exe -S --noconfirm --needed --noprogressbar base-devel compression
+taskkill /f /fi "MODULES eq msys-2.0.dll"
 
-Write-Host "Remove p7zip/7z package due to conflicts"
+Write-Host "`n$dash Remove p7zip/7z package due to conflicts"
 pacman.exe -R --noconfirm --noprogressbar p7zip
 
 # mingw package list
 $tools = "___clang ___cmake ___llvm  ___toolchain ___ragel"
 
 # install mingw64 packages
-Write-Host "Install mingw64 packages"
+Write-Host "`n$dash Install mingw64 packages"
 $pre = "mingw-w64-x86_64-"
 pacman.exe -S --noconfirm --needed --noprogressbar $tools.replace('___', $pre).split(' ')
 
 # install mingw32 packages
-Write-Host "Install mingw32 packages"
+Write-Host "`n$dash Install mingw32 packages"
 $pre = "mingw-w64-i686-"
 pacman.exe -S --noconfirm --needed --noprogressbar $tools.replace('___', $pre).split(' ')
 
 # clean all packages to decrease image size
-Write-Host "Clean packages"
+Write-Host "`n$dash Clean packages"
 pacman.exe -Scc --noconfirm
 
-Write-Host "Installed mingw64 packages"
-pacman.exe -Qs --noconfirm mingw-w64-x86_64-
+Write-Host "`n$dash Installed mingw64 packages"
+pacman.exe -Q | grep ^mingw-w64-x86_64-
 
-Write-Host "Installed mingw32 packages"
-pacman.exe -Qs --noconfirm mingw-w64-i686-
+Write-Host "`n$dash Installed mingw32 packages"
+pacman.exe -Q | grep ^mingw-w64-i686-
 
-Write-Host "Installed msys2 packages"
-pacman.exe -Qs --noconfirm
+Write-Host "`n$dash Installed msys2 packages"
+pacman.exe -Q | grep -v ^mingw-w64-
 
-Write-Host "MSYS2 installation completed"
+Write-Host "`nMSYS2 installation completed"
 
-exit 0
\ No newline at end of file
+exit 0

```

MSP-Greg:
@eine

See https://github.com/msys2/msys2-installer/issues/5.

Re rolling vs release, etc, etc --

I work a lot with Ruby master.  Given that, there is an interest in always using the most recent gcc available.  The most recent packages are zst files.  Without a full MSYS2 update, those can't even be installed.  One example, but there are also security updates (openssl, etc) and other examples.

Also, as I've mentioned before, the home page of [msys.org](https://www.msys2.org/) shows updating as part of the normal user install process.  We disagree on this, but we can certainly work together and benefit.  Ok?

maxim-lobanov:
@MSP-Greg let's merge this PR for now to include it to next week deployment.
And we can make one more PR with remain changes.
Any concerns / objections?

eine:
@MSP-Greg, sure! I believe that tight integration between upstream MSYS2, this repo and setup-msys2 (or any similar Action) is required in order to provide a pleasant UX. Since we are users, we'd better work together to benefit all of us! Let's move the discussion about higher-level strategies to msys2/msys2-installer#5.

MSP-Greg:
@maxim-lobanov 

Just to clarify:

> Any concerns / objections?

What, me worry?

Sorry.  No, progress is good, and thanks.

> And we can make one more PR with remain changes.

So the patch file above I'll open a new PR?  Not sure if it will conflict with this, but I can wait until this is accepted...

images/win/scripts/Installers/Install-Msys2.ps1
eine: @maxim-lobanov ^

MSP-Greg:
@maxim-lobanov

Thanks.  See https://github.com/actions/virtual-environments/pull/928

eine:
Ref #355

PR #933, opened 2020-05-23
Add retries to get PyPy version

al-cheb:
# Description
The pypy.sh script failed due to unable to get PyPy version from downloads.python.org related to network issue.
```
==> azure-arm: Provisioning with shell script: C:\agent\_work\2\s\images\linux/scripts/installers/pypy.sh
==> azure-arm: gzip: stdin: not in gzip format
    azure-arm: Failed to get PyPy version '2.7'
```

Changes :
- Add curl parameter -4 to use IPv4 only
- Max attempts 20
## Check list

### DIFF
--- a/images/linux/scripts/installers/pypy.sh
+++ b/images/linux/scripts/installers/pypy.sh
@@ -7,9 +7,6 @@
 # Source the helpers for use with the script
 source $HELPER_SCRIPTS/document.sh
 
-# Fail out if any setups fail
-set -e
-
 # This function installs PyPy using the specified arguments:
 #   $1=PACKAGE_URL
 function InstallPyPy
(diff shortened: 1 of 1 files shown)


images/linux/scripts/installers/pypy.sh
```diff
@@ -75,14 +72,31 @@ function InstallPyPy
 function getPyPyVersions
 {
     uri="https://downloads.python.org/pypy/"
-    wget -q -O - $uri | gunzip -c | grep 'linux64' | awk -v uri="$uri" -F'>|<' '{print uri$5}'
+    i=20
+
+    while [ $i -gt 0 ]; do
```
maxim-lobanov, line 77: I suggest to implement helper function to download file.
You implement good logic so it will let us reuse it in any other places if needs

miketimofeev: Let's skip this for now to unblock the rollout. I'll create another issue for that.

PR #935, opened 2020-05-25
Update Windows2019 image to use 2004 SDK & WDK

jpark37:
# Description
Need newer Windows SDK to compile against GraphicsCaptureSession.IsCursorCaptureEnabled property. I personally don't care about WDK, but it seems like they need to be updated as a set.

Verified correct version of SDK, WDK, and WDK VS extension were all installed to generated VM. My application compiles successfully.

Didn't touch Windows2019-Readme.md because it looks auto-generated.

#### Related issue:
Should I create an issue?

## Check list

### DIFF
--- a/images/win/scripts/Installers/Install-WDK.ps1
+++ b/images/win/scripts/Installers/Install-WDK.ps1
@@ -9,8 +9,8 @@ Import-Module -Name ImageHelpers -Force
 
 if (Test-IsWin19)
 {
-    $winSdkUrl = "https://go.microsoft.com/fwlink/p/?linkid=2083338"
-    $wdkUrl = "https://go.microsoft.com/fwlink/?linkid=2085767"
+    $winSdkUrl = "https://go.microsoft.com/fwlink/p/?linkid=2120843"
+    $wdkUrl = "https://go.microsoft.com/fwlink/?linkid=2128854"
     $FilePath = "C:\Program Files (x86)\Windows Kits\10\Vsix\VS2019\WDK.vsix"
     $VSver = "2019"
 }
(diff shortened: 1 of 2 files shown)


images/win/scripts/Installers/Install-WDK.ps1
```diff
@@ -9,8 +9,8 @@ Import-Module -Name ImageHelpers -Force
 
 if (Test-IsWin19)
 {
-    $winSdkUrl = "https://go.microsoft.com/fwlink/p/?linkid=2083338"
-    $wdkUrl = "https://go.microsoft.com/fwlink/?linkid=2085767"
+    $winSdkUrl = "https://go.microsoft.com/fwlink/p/?linkid=2120843"
```
alepauly, line 12: @jpark37 is this backwards compatible with previous version that was being installed or will it break existing workflows?

jpark37: Can you be more specific? In general, having a newer SDK should still allow you to target older Windows versions.

maxim-lobanov: It is update from `10.0.18346.0` to `10.0.19041.1`. I think should be safe enough

miketimofeev:
Hey @nibanks!
We've noticed that after these changes you can't specify version 10.0.18362.0 as discussed here 
https://github.com/actions/virtual-environments/issues/259#issuecomment-586757009
The error is
` ##[error]C:\Program Files (x86)\Windows Kits\10\build\WindowsDriver.common.targets(222,5): Error : A WDK corresponding to target version '10.0.18362.0' was not found. Please install this WDK version before building.`
Will simple version changing work for you or we need to take care of two WDK versions on the image?

nibanks:
@miketimofeev I just created https://github.com/microsoft/msquic/pull/439 to try removing the explicit version specified in the project files. Assuming this builds Ok, I think MsQuic will be Ok with this change. That being said, it might be a good idea to always support two version (current & previous) just in case.

miketimofeev:
@nibanks thanks for the answer! We'll consider adding the previous version.

PR #945, opened 2020-05-27
Refactor AzModules installation

dibir-magomedsaygitov:
In scope of this PR we are refactoring AzModules installation to avoid code duplication. [Link](https://github.com/actions/virtual-environments-internal/issues/577) to the work-item.


images/linux/scripts/installers/azpowershell.sh
```diff
@@ -6,92 +6,33 @@
 
 # Source the helpers for use with the script
 source $HELPER_SCRIPTS/document.sh
-source $HELPER_SCRIPTS/os.sh
 
-# Install Azure CLI (instructions taken from https://docs.microsoft.com/en-us/cli/azure/install-azure-cli)
-if isUbuntu20 ; then
```
miketimofeev, line 12: We'd like to install only the latest version for Ubuntu20. Why did you remove this section?

images/linux/scripts/installers/azpowershell.sh
```diff
@@ -6,92 +6,33 @@
...
-sudo pwsh -Command 'Save-Module -Name Az -LiteralPath /usr/share/az_1.0.0 -RequiredVersion 1.0.0 -Force'
-sudo pwsh -Command 'Save-Module -Name Az -LiteralPath /usr/share/az_1.6.0 -RequiredVersion 1.6.0 -Force'
-sudo pwsh -Command 'Save-Module -Name Az -LiteralPath /usr/share/az_2.3.2 -RequiredVersion 2.3.2 -Force'
-sudo pwsh -Command 'Save-Module -Name Az -LiteralPath /usr/share/az_2.6.0 -RequiredVersion 2.6.0 -Force'
-sudo pwsh -Command 'Save-Module -Name Az -LiteralPath /usr/share/az_2.8.0 -RequiredVersion 2.8.0 -Force'
-sudo pwsh -Command 'Save-Module -Name Az -LiteralPath /usr/share/az_3.1.0 -RequiredVersion 3.1.0 -Force'
-sudo pwsh -Command 'Save-Module -Name Az -LiteralPath /usr/share/az_3.5.0 -RequiredVersion 3.5.0 -Force'
-sudo pwsh -Command 'Save-Module -Name Az -LiteralPath /usr/share/az_3.8.0 -RequiredVersion 3.8.0 -Force'
+for version in ${versions[@]}; do
+    sudo pwsh -Command "Save-Module -Name Az -LiteralPath /usr/share/az_$version -RequiredVersion $version -Force"
```
al-cheb, line 15: Why do we need sudo?

images/linux/scripts/installers/azpowershell.sh
```diff
@@ -6,92 +6,33 @@
...
-    if (!(get-module -listavailable -name Az.accounts)) {
-        Write-Host "Az Module was not installed"; $env:PSModulePath = $actualPSModulePath; exit 1
-    }
-    $env:PSModulePath = $actualPSModulePath'; then
-    exit 1
-fi
+for version in ${versions[@]}; do
+    pwsh -Command '
+        $actualPSModulePath = $env:PSModulePath;
+        $env:PSModulePath = "/usr/share/az_${version}:" + $env:PSModulePath;
```
maxim-lobanov, line 23: ${version} won't be resolved properly in single quote

images/linux/scripts/installers/azpowershell.sh
dibir-magomedsaygitov: Sorry, it was done by mistake. I reverted it back.

images/linux/scripts/installers/azpowershell.sh
dibir-magomedsaygitov: Agreed, there's no need for sudo.

images/linux/scripts/installers/azpowershell.sh
dibir-magomedsaygitov: Fixed, should be ok now.

images/linux/scripts/installers/azpowershell.sh
miketimofeev: You don't need this block then
image

images/linux/scripts/installers/azpowershell.sh
```diff
@@ -30,68 +30,36 @@ if isUbuntu20 ; then
     exit 0
 fi
 
+# List of versiion
+if isUbuntu20 ; then
```
miketimofeev, line 34: Seems like a block on the top is not needed anymore

dibir-magomedsaygitov: Yes, agree. Thank you.

images/linux/scripts/installers/azpowershell.sh
```diff
@@ -8,90 +8,37 @@
...
-            Write-Host 'Az Module was not installed'
-            exit 1
-        }"; then
-        exit 1
-    fi
-
-    # Document what was added to the image
-    DocumentInstalledItem "Az Module ($latestVersion)"
-    exit 0
+    versions=($latestVersion)
```
miketimofeev, line 14: why we need two variables instead just
```
# Take the latest version only for Ubuntu20
versions=$(pwsh -Command '(Find-Module -Name Az).Version')
```

images/linux/scripts/installers/azpowershell.sh
```diff
@@ -8,90 +8,36 @@
...
-            Write-Host 'Az Module was not installed'
-            exit 1
-        }"; then
-        exit 1
-    fi
-
-    # Document what was added to the image
-    DocumentInstalledItem "Az Module ($latestVersion)"
-    exit 0
+    versions=(pwsh -Command '(Find-Module -Name Az).Version')
```
miketimofeev, line 13: Have you tested? It doesn't work this way, you've missed `$`

images/linux/scripts/installers/azpowershell.sh
maxim-lobanov: @miketimofeev , I don't think that installing the latest version is good idea here.
For example, 3.5.0 and 3.8.0 contain breaking changes. If we install the latest only, it will break customers

images/linux/scripts/installers/azpowershell.sh
```diff
@@ -8,90 +8,36 @@
 source $HELPER_SCRIPTS/document.sh
 source $HELPER_SCRIPTS/os.sh
 
-# Install Azure CLI (instructions taken from https://docs.microsoft.com/en-us/cli/azure/install-azure-cli)
+# List of version
```
AlenaSviridenko, line 11: nitpick: version**s**

dibir-magomedsaygitov: fixed, thank you!

images/linux/scripts/installers/azpowershell.sh
```diff
@@ -8,90 +8,36 @@
...
-fi
+for version in ${versions[@]}; do
+    modulePath="/usr/share/az_$version"
+    pwsh -Command "
+        \$env:PSModulePath = '${modulePath}:' + \$env:PSModulePath;
+        if ( -not (Get-Module -ListAvailable -Name Az.Accounts)) {
+            Write-Host 'Az Module was not installed'
+            exit 1
+        }"
+    if [ $? -ne 0 ]; then
```
miketimofeev, line 33: exit 1 from powershell doesn't break the build?

dibir-magomedsaygitov: Good point, but `set -e` is not commonly used, so let's leave IF condition as it is.

PR #946, opened 2020-05-27
Added support for JDK13

ewanklomp:
# Description
Added support JDK13 for linux to align it with the Windows environment. 
The windows environment never had JDK12 but kept it in here.

### DIFF
--- a/images/linux/scripts/installers/java-tools.sh
+++ b/images/linux/scripts/installers/java-tools.sh
@@ -31,11 +31,13 @@ fi
 apt-get -y install zulu-8-azure-jdk=\*
 apt-get -y install zulu-11-azure-jdk=\*
 apt-get -y install zulu-12-azure-jdk=\*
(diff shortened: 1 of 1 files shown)


maxim-lobanov:
Hello @ewanklomp ,
We had internal conversation about this pull-request and we have decided not to add Java 13 to our images.
We would like to follow strategy to pre-install only LTS versions of Java. Despite the fact that we already have Java 12 on all images and Java 13 on Windows image, we would like to follow this strategy for future updates.
Sorry for inconvenience.
You can use the following tasks to install any version of Java in runtime (it is pretty quick):
1) [setup-java](https://github.com/actions/setup-java) (GitHub Actions)
2) [Java Tool Installer task](https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/tool/java-tool-installer?view=azure-devops) (Azure DevOps)

Please let us know if you have any additional questions

PR #955, opened 2020-05-28
Add download retry helper on Ubuntu

vsafonkin:
Added helper function for retry to download file.

Internal issue: [#588](https://github.com/actions/virtual-environments-internal/issues/588)

### DIFF
--- a/images/linux/scripts/helpers/install.sh
+++ b/images/linux/scripts/helpers/install.sh
@@ -11,15 +11,20 @@ download_with_retries() {
     local URL="$1"
     local DEST="${2:-.}"
     local NAME="${3:-${URL##*/}}"
+
     echo "Downloading $URL..."
-    wget $URL   --output-document="$DEST/$NAME" \
-                --tries=30 \
-                --wait 30 \
-                --retry-connrefused \
-                --no-verbose
-    if [ $? != 0 ]; then
-        echo "Could not download $URL; Exiting build!"
-        exit 1
-    fi
-    return 0
+    i=20
+    while [ $i -gt 0 ]; do
+        ((i--))
+        wget $URL   --output-document="$DEST/$NAME" \
+                    --no-verbose
+        if [ $? != 0 ]; then
+            sleep 30
+        else
+            return 0
+        fi
+    done
+
+    echo "Could not download $URL"
+    return 1
 }
--- a/images/linux/scripts/helpers/install.sh
+++ b/images/linux/scripts/helpers/install.sh
@@ -5,7 +5,7 @@
 ################################################################################
 
 download_with_retries() {
-# Due to restrictions of bash functions, positional arguments is used here.
+# Due to restrictions of bash functions, positional arguments are used here.
 # In case if you using latest argument NAME, you should also set value to all previous parameters.
 # Example: download_with_retries $ANDROID_SDK_URL "." "android_sdk.zip"
     local URL="$1"
--- a/images/linux/scripts/helpers/install.sh
+++ b/images/linux/scripts/helpers/install.sh
@@ -16,8 +16,6 @@ download_with_retries() {
                 --tries=30 \
                 --wait 30 \
                 --retry-connrefused \
-                --retry-on-host-error \
-                --retry-on-http-error=429,500,503 \
                 --no-verbose
     if [ $? != 0 ]; then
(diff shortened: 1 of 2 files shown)


images/linux/scripts/helpers/install.sh
```diff
@@ -0,0 +1,27 @@
...
+# Example: download_with_retries $ANDROID_SDK_URL "." "android_sdk.zip"
+    local URL="$1"
+    local DEST="${2:-.}"
+    local NAME="${3:-${URL##*/}}"
+    echo "Downloading $URL..."
+    wget $URL   --output-document="$DEST/$NAME" \
+                --tries=30 \
+                --wait 30 \
+                --retry-connrefused \
+                --retry-on-host-error \
```
al-cheb, line 19: wget: unrecognized option '--retry-on-host-error'

vsafonkin: Fixed, thanks!

images/linux/scripts/helpers/install.sh
```diff
@@ -0,0 +1,25 @@
...
+# In case if you using latest argument NAME, you should also set value to all previous parameters.
+# Example: download_with_retries $ANDROID_SDK_URL "." "android_sdk.zip"
+    local URL="$1"
+    local DEST="${2:-.}"
+    local NAME="${3:-${URL##*/}}"
+    echo "Downloading $URL..."
+    wget $URL   --output-document="$DEST/$NAME" \
+                --tries=30 \
+                --wait 30 \
+                --retry-connrefused \
```
al-cheb, line 18: We don't need to remove `--retry-on-http-error=429,500,503` parameter

miketimofeev:
@vsafonkin shouldn't we switch all scripts with wget to use this function in the scope of the PR?

images/linux/scripts/helpers/install.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+################################################################################
+##  File:  install.sh
+##  Desc:  Helper functions for installing tools
+################################################################################
+
+download_with_retries() {
+# Due to restrictions of bash functions, positional arguments is used here.
```
AlenaSviridenko, line 8: nitpick: "**are** used here"

AlenaSviridenko:
@vsafonkin as far as I remember, this issue was created based on this PR https://github.com/actions/virtual-environments/pull/933, should we then use this new method for PyPy downloading?

vsafonkin:
> @vsafonkin shouldn't we switch all the scripts with wget to use this function in the scope of the PR?

> @vsafonkin as far as I remember, this issue was created based on this PR #933, should we then use this new method for PyPy downloading?

I think, it will be better to separate adding retry logic for tools to another workitem , because there will be a lot of changes if it will be done here.

images/linux/scripts/helpers/install.sh
vsafonkin: Fixed, thank you!

maxim-lobanov:
Let's switch only PyPy for now

vsafonkin:
> Let's switch only PyPy for now

Done

images/linux/scripts/installers/pypy.sh
```diff
@@ -69,28 +69,11 @@ function InstallPyPy
...
-    done
-
-    echo "There are no more attempts to retrive PyPy version"
-    exit 1
-}
-
 # Installation PyPy
-pypyVersions=$(getPyPyVersions)
+uri="https://downloads.python.org/pypy/"
+download_with_retries $uri "." "pypyUrls.html"
```
maxim-lobanov, line 74: Can we put this file to `/tmp` to make sure that it will be clean up?

images/linux/scripts/installers/pypy.sh
```diff
@@ -69,28 +69,11 @@ function InstallPyPy
...
+pypyVersions="$(cat pypyUrls.html | grep 'linux64' | awk -v uri="$uri" -F'>|<' '{print uri$5}')"
```
maxim-lobanov, line 75: Do we want the separate retry function for curl?
Looks like file is not useful here?

miketimofeev:
> Let's switch only PyPy for now

As I know we switched from wget to curl because oof the issues, so you can't just switch it back
cc @al-cheb

PR #966, opened 2020-05-29
update docker-compose version

fguisso:
Update to new stable version of docker-compose:
https://github.com/docker/compose/releases/tag/1.25.5

### DIFF
--- a/images/linux/scripts/installers/docker-compose.sh
+++ b/images/linux/scripts/installers/docker-compose.sh
@@ -7,10 +7,10 @@
 source $HELPER_SCRIPTS/apt.sh
 source $HELPER_SCRIPTS/document.sh
 
-version="1.25.4"
+URL=$(curl -s https://api.github.com/repos/docker/compose/releases/latest | jq -r '.assets[].browser_download_url | select(contains("docker-compose-Linux-x86_64"))' | head -1)
 
 # Install latest docker-compose from releases
-curl -L "https://github.com/docker/compose/releases/download/$version/docker-compose-$(uname -s)-$(uname -m)" -o /usr/local/bin/docker-compose
+curl -L $URL -o /usr/local/bin/docker-compose
 chmod +x /usr/local/bin/docker-compose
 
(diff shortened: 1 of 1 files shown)


miketimofeev:
Hi @fguisso!
Would you mind changing the script to take the latest version instead of a hardcoded one?
Something like:
```
URL=$(curl -s https://api.github.com/repos/docker/compose/releases/latest | jq -r '.assets[].browser_download_url | select(contains("docker-compose-Linux-x86_64"))' | head -1)
curl -L $URL -o /usr/local/bin/docker-compose
```

fguisso:
Awesome, I didn't know that they had this endpoint.
I can change, but I don't know if this is the best option, because using the latest version sometimes broken things. What you, contributors, and maintainers think about it?

miketimofeev:
@fguisso we're ok with the latest, it helps reducing maintenance amount.

maxim-lobanov:
@fguisso , @miketimofeev , I prefer to install the latest one as Mikhail mentioned.
We do the same strategy for other tools and it works for us.

fguisso:
Hey guys!
Can you explain to me how it works? because one build using: `ubuntu-latest` keep fail, and looking the next commit, docker-compose are at the same version: https://github.com/actions/virtual-environments/commit/5c334052e782ac68f57961055a45b92e7a7bf25a#

I made something wrong, or just need to wait more time to the registry updates the image?

miketimofeev:
@fguisso ubuntu-latest is pointed at ubuntu-18.04 and it will be updated next week. As a workaround, you can simply add the same step to your workflow:
```
    steps:
      - uses: actions/checkout@v2
      - name: update docker compose
        run: |
          URL=$(curl -s https://api.github.com/repos/docker/compose/releases/latest | jq -r '.assets[].browser_download_url | select(contains("docker-compose-Linux-x86_64"))' | head -1)
          sudo curl -L $URL -o /usr/local/bin/docker-compose
```

PR #972, opened 2020-06-01
Switch provisioners to install Boost from GitHub releases on Windows

dibir-magomedsaygitov:
In scope of this PR we are switching provisioners to install Boost from GitHub releases on Windows. [Link](https://github.com/actions/virtual-environments-internal/issues/528) to the work-item.


images/win/toolset-2019.json
```diff
@@ -47,6 +47,16 @@
...
                 "14.*"
             ]
+        },
+        {
+            "name": "Boost",
+            "url" : "https://raw.githubusercontent.com/akv-platform/boost-hostedtoolcache/master/versions-manifest.json",
+            "arch": "x64",
+            "platform" : "win32",
+            "versions": [
+                "1.69.0",
```
maxim-lobanov, line 57: Version `1.71.0` has two different toolsets `msvc14.1` and `msvc14.2`. How will we choose between them?

images/win/toolset-2019.json
```diff
@@ -47,6 +47,16 @@
                 "12.*",
                 "14.*"
             ]
+        },
+        {
+            "name": "Boost",
+            "url" : "https://raw.githubusercontent.com/akv-platform/boost-hostedtoolcache/master/versions-manifest.json",
+            "arch": "x64",
```
maxim-lobanov, line 54: I don't see file with this architecture for Windows in manifest json

dibir-magomedsaygitov: Changed to correct version

images/win/toolset-2016.json
```diff
@@ -47,6 +47,26 @@
...
+        },
+        {
+            "name": "Boost",
+            "url" : "https://raw.githubusercontent.com/akv-platform/boost-hostedtoolcache/master/versions-manifest.json",
+            "arch": "x86_64",
+            "platform" : "win32",
+            "toolset": "msvc14.2",
+            "versions": [
+                "1.71.0"
+            ]
```
maxim-lobanov, line 69: I think both versions on Win2016 should be with msvc14.1 toolset. See toolcache-2016.json

images/win/toolset-2016.json
```diff
@@ -47,6 +47,26 @@
...
+        },
+        {
+            "name": "Boost",
+            "url" : "https://raw.githubusercontent.com/akv-platform/boost-hostedtoolcache/master/versions-manifest.json",
+            "arch": "x86_64",
+            "platform" : "win32",
+            "toolset": "msvc14.1",
+            "versions": [
+                "1.71.0"
+            ]
```
maxim-lobanov, line 69: Why we can't merge both objects to the single one?
`versions` is array and you can put more than 1 version here. See Python as example

images/win/scripts/Installers/Validate-Toolset.ps1
```diff
@@ -67,8 +67,8 @@ foreach($tool in $tools) {
 
     foreach ($version in $tool.versions) {
         # Add wildcard if missing
-        if (-not $version.Contains('*')) {
-            $version += '.*'
+        if ($version.Split(".").Length -lt 3) {
```
al-cheb, line 70: What if we would like to use `x.*` pattern?

images/win/scripts/Installers/Install-Toolset.ps1
```diff
@@ -60,7 +66,7 @@ foreach ($tool in $tools) {
         $asset = $assets | Where-Object version -like $toolVersion `
                          | Sort-Object -Property {[version]$_.version} -Descending `
                          | Select-Object -ExpandProperty files `
-                         | Where-Object { ($_.platform -eq $tool.platform) -and ($_.arch -eq $tool.arch) } `
+                         | Where-Object { ($_.platform -eq $tool.platform) -and ($_.arch -eq $tool.arch) -and ($_.toolset -eq $tool.toolset) } `
```
vmapetr, line 69: `-and ($_.toolset -eq $tool.toolset)` are you sure about that? As I see, only Boost have toolset field, and this function is used for all toolset tools.

images/win/scripts/Installers/Validate-Toolset.ps1
dibir-magomedsaygitov: It would still work, because all folders have x.x.x pattern (ex: 10.* it would change it to 10.*.*, so folder would be like 10.0.0).
We had to change old check for wildcard because it was adding extra * to versions (1.69.0 -> 1.69.0.*) and it was causing issues with validation part for boost.

al-cheb: What if we would to use x.x without third number?

images/win/scripts/Installers/Install-Toolset.ps1
dibir-magomedsaygitov: yes, it works fine with `-and ($_.toolset -eq $tool.toolset)`.

images/win/scripts/Installers/Validate-Toolset.ps1
dibir-magomedsaygitov: It would still work: `10.2` -> `10.2.*`, `10` -> `10.*.*`, `10.2.2 ` -> `10.2.2`

images/win/toolset-2016.json
dibir-magomedsaygitov: Thank you, fixed.

images/win/scripts/Installers/Install-Toolset.ps1
vmapetr: Surely, it will work, but it's just little messy. Maybe we should move platform, arch and additional checks in some separate loop.

PR #999, opened 2020-06-03
Specify environment details for Inno Setup.

vslynko:
# Description
Inno Setup is installed, but whether its location is added to PATH was not mentioned. This PR fixes that.

I did this PR just from Github, let me know if mixed line-ending should be addressed too.


AlenaSviridenko:
Hi @vslynko,
we are updating the way of generating README markdown and its structure, and current README will be deprecated approximately the next week.

Issue #882, opened 2020-05-14
Add vercel


brunowowk:
**Tool information**
- Tool name: Formely `now`, renamed to `vercel`
- Tool license:  Apache-2.0
- Add or update? Add
- Desired version: latest
- Approximate size: ?

ref: https://github.com/zeit/now/releases/tag/vercel%4019.0.0
> We will continue publishing now and vercel in tandem until January 2021 at which point, new updates will only be published to vercel.

it's probably best to keep both `now` and `vercel` until then. Maybe install `vercel` and create a symlink for `now` because it's the same package

**Area for Triage**: JavaScript and Node.js

**Virtual environments affected**
- [ ] macOS 10.15
- [x] Ubuntu 16.04 LTS
- [x] Ubuntu 18.04 LTS
- [ ] Windows Server 2016 R2
- [ ] Windows Server 2019

miketimofeev:
Hi @brunowowk!
Can it be installed during runtime? How much time does it take?

brunowowk:
yes, it's an npm package. It's actually an update to the `now` bin that's already included. 

Then only change would be to install `vercel` instead of `now`, or to install both over here:
https://github.com/actions/virtual-environments/blob/51d02f26f42e53bb963bb6bc9754501b28a3d2ba/images/linux/scripts/installers/zeit-now.sh#L11

maxim-lobanov:
if they are fully compatible (you mentioned that it is identical package), installing `vacel` instead of `now` and creating symlink sounds good

andy-mishechkin:
Hello!
We have made these changes, as we discussed above. Please wait for the next Ubuntu Images deployment, which will bring Zeit Vercel CLI and symlibk "now" to this.

andy-mishechkin:
Hello, @styfle 
Thank you for your note.
I've fixed this issue. The right name should be present after the next Ubuntu VM updating.
If it will not be fixed - please feel free to ping me here

andy-mishechkin:
Deployment has been successful. The valid name is present.

PR #1032, opened 2020-06-11
Removing the "Zeit" term from Vercel deployment

andy-mishechkin:
# Description
Improvement. 
"Zeit" term has been removed from all Vercel deployment scripts
#### Related issue:
https://github.com/actions/virtual-environments/issues/882

## Check list


images/linux/scripts/installers/vercel.sh
```diff
@@ -1,31 +1,31 @@
...
 if ! command -v now; then
-    echo "Now link to Zeit Vercel CLI was not created"
+    echo "[Now] symlink to Vercel CLI was not created"
     exit 1
 fi
 
 # Document the installed version
 echo "Document the installed version"
-DocumentInstalledItem "Zeit Vercel CLI ($(vercel --version))"
+DocumentInstalledItem "Vercel CLI ($(vercel --version))"
```
AlenaSviridenko, line 31: I believe tool should be documented according to this guide: https://github.com/actions/virtual-environments/blob/master/CONTRIBUTING.md#submitting-a-pull-request

> add changes to document tool name and version to the software report generator: images/win/scripts/SoftwareReport/SoftwareReport.Generator.ps1 that is used to generate software README file, e.g. Windows2019-Readme.md

miketimofeev: It's for windows only at the moment.

vmapetr: There is no separate software report generator on linux images, and it is a common practice that we install, validate, and create documentation in the same sh script.

AlenaSviridenko: oh, that's linux 🤦‍♀️. sorry for the confusion

andy-mishechkin: images/win/scripts/SoftwareReport/SoftwareReport.Generator.ps1 using in Windows deployments only

images/linux/scripts/installers/vercel.sh
```diff
@@ -1,31 +1,31 @@
...
 
 # Source the helpers for use with the script
 source $HELPER_SCRIPTS/document.sh
 
-# Install the Zeit Vercel CLI
+# Install the Vercel CLI
 npm i -g vercel
 
 # Validate the installation
 echo "Validate the installation"
```
vmapetr, line 14: Perhaps, it's make sense to move this block below symlink creation, to separate installation from validation.

andy-mishechkin: IMHO In this case, if Vercel installation will be failed we can get additional error in symlink creation.
I think, better to check Vercel installation before symlink creation as it already done

Issue #986, opened 2020-06-02
Add explicit dependencies on essential packages


favonia:
**Describe the bug**
Essential tools should be explicitly included in Ubuntu images. This is a resubmission of #985.

**Area for Triage**: 
Packages

**Question, Bug, or Feature?**: 
Feature

**Virtual environments affected**
- [ ] macOS 10.15
- [x] Ubuntu 16.04 LTS
- [x] Ubuntu 18.04 LTS
- [ ] Windows Server 2016 R2
- [ ] Windows Server 2019

**Expected behavior**
Essential packages should be included regardless of the installation of other packages. For example, issues in PHP installation (#981) should not cause `m4` to dissapear. I was asked to provide a list, and perhaps we could start with the following one. I probably missed many packages because it was made in a hurry, but I hope the idea of explicitly including essential packages is clear. Thank you for your consideration.

- autoconf
- automake
- bash
- binutils
- bison
- bzip2
- coreutils
- fakeroot
- file
- findutils
- flex
- gawk
- gcc
- gettext
- glibc
- grep
- groff
- gzip
- iproute2
- iputils
- libtool
- m4
- make
- patch
- ~pciutils~ (This was removed because this might not be too useful for CI? I don't know.)
- pkgconf
- sed
- shadow
- sudo
- tar
- texinfo
- util-linux
- which
- xz

vsafonkin:
Hi @favonia , thank you for your help! I've checked your list and created PR to add some packages that are absent on the Ubuntu images, thank you!

PR #1034, opened 2020-06-11
Add linux utils

vsafonkin:
# Description
Add a few linux utilities.

Size: ~ 3 Mb
Time: ~ 1 min

#### Related issue:
#986 

## Check list

### DIFF
--- a/images/linux/scripts/installers/basic.sh
+++ b/images/linux/scripts/installers/basic.sh
@@ -26,7 +26,8 @@ common_packages="dnsutils
                  libxss1
                  libgconf-2-4
                  dbus
-                 xvfblibgbm-dev
+                 xvfb
+                 libgbm-dev
                  libgtk-3-0
                  tk
                  fakeroot
@@ -36,7 +37,8 @@ common_packages="dnsutils
                  xorriso
                  zsync
                  gnupg2
-                 lib32z1"
+                 lib32z1
+                 texinfo"
 
 cmd_packages="curl
               file
@@ -55,8 +57,7 @@ cmd_packages="curl
               wget
               m4
               bison
-              flex
-              texinfo"
+              flex"
 
 for package in $common_packages $cmd_packages; do
     echo "Install $package"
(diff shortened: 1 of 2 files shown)


al-cheb:
Can we put all utils in array to update them only at one place (currently we have two )?

images/linux/scripts/installers/basic.sh
```diff
@@ -150,6 +150,18 @@ apt-get install -y --no-install-recommends gnupg2
...
+
+echo "Install flex"
+apt-get install -y --no-install-recommends flex
+
+echo "Install texinfo"
+apt-get install -y --no-install-recommends texinfo
+
 # Run tests to determine that the software installed as expected
 echo "Testing to make sure that script performed as expected, and basic scenarios work"
 for cmd in curl file ftp jq netcat ssh parallel rsync shellcheck sudo telnet time unzip wget zip; do
```
al-cheb, line 167: Please add these utils to the test loop

vsafonkin: Done

images/linux/scripts/installers/basic.sh
```diff
@@ -45,114 +77,9 @@ fi
...
-echo "Install gnupg2"
-apt-get install -y --no-install-recommends gnupg2
-
-echo "Install lib32z1"
-apt-get install -y --no-install-recommends lib32z1
-
 # Run tests to determine that the software installed as expected
 echo "Testing to make sure that script performed as expected, and basic scenarios work"
-for cmd in curl file ftp jq netcat ssh parallel rsync shellcheck sudo telnet time unzip wget zip; do
+for cmd in curl file ftp jq netcat ssh parallel rsync shellcheck sudo telnet time unzip wget zip m4 bison flex texinfo; do
```
maxim-lobanov, line 82: Is it possible to use to use array `common_packages` here?
And at the bottom with `DocumentInstalledItemIndent`.

vsafonkin: Done.

Issue #1039, opened 2020-06-12
Update/Add NuGet.exe


jcansdale:
**Tool information**
- Tool name: NuGet
- Tool license: MIT license
- Add or update? Add
- Desired version: https://dist.nuget.org/win-x86-commandline/latest/nuget.exe
- Approximate size: 6.5 MD
- If this is an add request:
  - Brief description of tool: NuGet is the package manager for .NET
  - URL for tool's homepage: https://www.nuget.org/

**Area for Triage**:
.NET Framework

**Question, Bug, or Feature?**:
Feature

**Virtual environments affected**
- [ ] macOS 10.15
- [ ] Ubuntu 16.04 LTS
- [ ] Ubuntu 18.04 LTS
- [ ] Ubuntu 20.04 LTS
- [x] Windows Server 2016 R2
- [x] Windows Server 2019

**Can this tool be installed during the build?**

**Tool installation time in runtime**

**Are you willing to submit a PR?**

**Notes**

NuGet.exe is required to consistently install packages to GitHub Packages. Using `dotnet nuget` often fails. See https://github.com/github/c2c-package-registry/issues/1129#issuecomment-643370557

al-cheb:
Hello, @jcansdale 
The deployment with Nuget has been done. Could you please check your pipeline?

**_NuGet 5.6.0.6591:_**
Windows Server 2019 - https://github.com/actions/virtual-environments/blob/releases/win19/20200621/images/win/Windows2019-Readme.md
Windows Server 2016 - https://github.com/actions/virtual-environments/blob/releases/win16/20200621/images/win/Windows2016-Readme.md

jcansdale:
Boom! 💥 

![image](https://user-images.githubusercontent.com/11719160/85763482-983f5780-b70c-11ea-9bda-0d58b70c8cae.png)

I think we can close this issue. 👍

PR #1049, opened 2020-06-15
Add NuGet.exe on Windows Server 2016/2019

al-cheb:
# Description
The macOS 10.15 image includes nuget.exe (using Mono) but our Windows images doesn’t. This is unfortunate because publishing NuGet packages doesn’t work consistently with dotnet nuget. Users need to figure out a way to download nuget.exe manually if they want to use it,NuGet.exe is required to consistently install packages to GitHub Packages.

#### Related issue:
https://github.com/actions/virtual-environments/issues/1039

## Check list

### DIFF
--- a/images/win/scripts/Installers/Install-Nuget.ps1
+++ b/images/win/scripts/Installers/Install-Nuget.ps1
@@ -5,7 +5,6 @@
 
 Choco-Install -PackageName NuGet.CommandLine
 
-$env:Path = $env:Path + ";$env:ChocolateyInstall\bin"
 if (Get-Command -Name 'nuget.exe')
 {
     Write-Host 'nuget on path'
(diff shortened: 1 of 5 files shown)


images/win/scripts/Installers/Install-Nuget.ps1
```diff
@@ -0,0 +1,17 @@
+################################################################################
+##  File:  Install-Nuget.ps1
+##  Desc:  Install NuGet.CommandLine
+################################################################################
+
+Choco-Install -PackageName NuGet.CommandLine
+
+$env:Path = $env:Path + ";$env:ChocolateyInstall\bin"
```
maxim-lobanov, line 8: I believe this line should already been in PATH after installation of Choco in previous steps

Issue #1061, opened 2020-06-16
Refactor apt/ulimit commands in ubuntu image packer templates


gswallow:
Hi,

I'm toying around with this project as I am building an AWS EC2 AMI, which I can fire up in my own VPCs.  I have a pull request where I take the commands in https://github.com/actions/virtual-environments/blob/40caaf3e2c11a9e4289482cce72e14ed97fa64fe/images/linux/ubuntu2004.json#L80-L93, and move them into their own scripts in the base folder.

Not much value, other than it eases my mind when I look at the code :)

I hope I can submit another PR soon with a packer template that builds AMIs via a separate pipeline.

miketimofeev:
@gswallow Ubuntu images with the changes were deployed, thanks for your contribution!

PR #1062, opened 2020-06-16
Paragraphs (or stanzas) should cover 1 topic only.

gswallow:
# Description

I propose a slight improvement: Each stanza, like a paragraph or a function, should cover one topic.  Move the apt commands and ulimit adjustments into separate scripts and invoke them separately.

See https://github.com/actions/virtual-environments/issues/1061


maxim-lobanov:
Thank you for contribution!

PR #1072, opened 2020-06-18
Had issues where android.sh just hangs on a packer build

gswallow:
# Description
The android sdkmanager command would just hang when installing packages.  I fought with it all day and finally got it to work by directging stderr and stdout to /dev/null.  If that's not ideal, I can test redirecting stdout and stderr to a file and sending the contents of that file to the packer log.  I'll create an issue then edit this comment.


maxim-lobanov:
Closing with the same reason like related issue. Feel free to reopen if you have additional context

Issue #206, opened 2019-12-23
Add Google Cloud SDK for Windows


scothis:
**Tool information**
- Tool name: Google Cloud SDK
- Add or update? Add
- Desired version: latest
- Approximate size: 
- If this is an add request:
  - Brief description of tool: Command-line interface for Google Cloud Platform products and services
  - URL for tool's homepage: https://cloud.google.com/sdk/

**Area for Triage**: Scripting and command line

**Question, Bug, or Feature?**: Feature

**Virtual environments affected**
- [x] macOS 10.15
- [x] Ubuntu 16.04 LTS
- [x] Ubuntu 18.04 LTS
- [X] Windows Server 2016 R2
- [X] Windows Server 2019

**Can this tool be installed during the build?**
This tool is already available on the Mac and Ubuntu environments. There is a chocolatey package available, although it is currently failing. 

**Are you willing to submit a PR?**

maxim-lobanov:
Hello,
Google Cloud SDK can be installed in runtime via the following commands:
```
Invoke-WebRequest "https://dl.google.com/dl/cloudsdk/channels/rapid/GoogleCloudSDKInstaller.exe" -OutFile "GoogleCloudSDKInstaller.exe"
Start-Process -FilePath "GoogleCloudSDKInstaller.exe" -ArgumentList "/S" -Wait
```

al-cheb:
Hello, @scothis 
Sorry for long delay. The https://github.com/actions/virtual-environments/pull/1102 PR was merged and the Google Cloud SDK will be available in the next rollout one/two weeks on Windows Server 2016/2019.

al-cheb:
@scothis,The Google Cloud SDK is now available on Windows Server 2016/2019. Could you please check from your side?

scothis:
@al-cheb looks good. Thanks

PR #1102, opened 2020-06-23
Add Google Cloud SDK for Windows

al-cheb:
# Description
Google Cloud SDK is a set of tools that you can use to manage resources and applications hosted on Google Cloud Platform. These include the gcloud, gsutil, and bq command line tools. Google Cloud SDK is already available on the Mac and Ubuntu environments.

Google Cloud SDK:
- Size ~800mb
- Installation time ~ 3 -5 minutes

#### Related issue:
https://github.com/actions/virtual-environments/issues/206
## Check list

### DIFF
--- a/images/win/scripts/Installers/Install-GoogleCloudSDK.ps1
+++ b/images/win/scripts/Installers/Install-GoogleCloudSDK.ps1
@@ -7,19 +7,3 @@
 $googleCloudSDKInstaller = "https://dl.google.com/dl/cloudsdk/channels/rapid/GoogleCloudSDKInstaller.exe"
 $argumentList = @("/S", "/allusers", "/noreporting")
 Install-Binary -Url $googleCloudSDKInstaller -Name "GoogleCloudSDKInstaller.exe" -ArgumentList $argumentList
-
-# Simple validation gcloud, gsutil, and bq command line tools
-$env:Path += ";${env:ProgramFiles(x86)}\Google\Cloud SDK\google-cloud-sdk\bin"
-$validateTools = @("bq", "gcloud", "gsutil")
-foreach($tool in $validateTools)
-{
-    if (Get-Command -Name $tool)
-    {
-        Write-Host "$tool on path"
-    }
-    else
-    {
-        Write-Host "$tool is not on path"
-        exit 1
-    }
-}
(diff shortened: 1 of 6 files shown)


images/win/scripts/Installers/Install-GoogleCloudSDK.ps1
```diff
@@ -0,0 +1,25 @@
...
+################################################################################
+
+# https://cloud.google.com/sdk/docs/downloads-interactive
+$googleCloudSDKInstaller = "https://dl.google.com/dl/cloudsdk/channels/rapid/GoogleCloudSDKInstaller.exe"
+$argumentList = @("/S", "/allusers", "/noreporting")
+Install-Binary -Url $googleCloudSDKInstaller -Name "GoogleCloudSDKInstaller.exe" -ArgumentList $argumentList
+
+# Simple validation gcloud, gsutil, and bq command line tools
+$env:Path += ";${env:ProgramFiles(x86)}\Google\Cloud SDK\google-cloud-sdk\bin"
+$validateTools = @("bq", "gcloud", "gsutil")
```
miketimofeev, line 13: Shall we use a separate Validation script?

al-cheb: I don't think so, we need a third place to validate tools in path, another one is Get-GoogleCloudSDKVersion.

AlenaSviridenko: But as a common practice we have an Install-Tool.ps1 and Validate-Tool.ps1 scripts, so for the sake of consistency I would suggest to move validation to a separate script.

al-cheb: Agree, but starting since `Install-AWS.ps1` we don't follow strictly this practice.

miketimofeev: @maxim-lobanov let's set the policy because I don't mind if we get rid of separate validation script, but we should follow that strategy for all new tools

maxim-lobanov: Discussed offline.
Let's create the separate `Validate-<tool>` file for now. It will simplify future migration to Pester

PR #1104, opened 2020-06-24
Remove  ppa:ondrej/php from the list of apt repositories

dsame:
# Description
Bug fixing.

ppa:ondrej/php repository is used to install PHP and the modules. Currently it contains libzip-dev package corresponding to libzip4 shared libraries. Ubuntu 20.04 does not contain libzip4 so, instead it has libzip5 libraries. As a result the binaries built with pipelines are linked against to libzip4 and crash on the real life Ubuntu 20.04 

As a solution in this PR i removed the ppa:ondrej/php right after PHP has been installed. 

## Check list

### DIFF
--- a/images/linux/scripts/installers/php.sh
+++ b/images/linux/scripts/installers/php.sh
@@ -93,6 +93,10 @@ apt-get remove --purge -yq php7.2-dev
 
 apt-fast install -y --no-install-recommends snmp
 
+if isUbuntu20 ; then
+    php_versions="7.4"
+fi
+
 # Install composer
 php -r "copy('https://getcomposer.org/installer', 'composer-setup.php');"
 php -r "if (hash_file('sha384', 'composer-setup.php') === file_get_contents('https://composer.github.io/installer.sig')) { echo 'Installer verified'; } else { echo 'Installer corrupt'; unlink('composer-setup.php'); } echo PHP_EOL;"
@@ -133,6 +137,12 @@ echo "Lastly, documenting what we added to the metadata file"
 for version in $php_versions; do
     DocumentInstalledItem "PHP $version ($(php$version --version | head -n 1))"
 done
+
+# ubuntu 20.04 libzip-dev is libzip5 based and is not compatible libzip-dev of ppa:ondrej/php
+if isUbuntu20 ; then
+  rm /etc/apt/sources.list.d/ondrej-ubuntu-php-focal.list
+  apt-get update
+fi
 
 DocumentInstalledItem "Composer  ($(composer --version))"
 DocumentInstalledItem "PHPUnit ($(phpunit --version))"
--- a/images/linux/scripts/installers/php.sh
+++ b/images/linux/scripts/installers/php.sh
@@ -93,10 +93,6 @@ apt-get remove --purge -yq php7.2-dev
 
 apt-fast install -y --no-install-recommends snmp
 
-if isUbuntu20 ; then
-    php_versions="7.4"
-fi
-
 # Install composer
 php -r "copy('https://getcomposer.org/installer', 'composer-setup.php');"
 php -r "if (hash_file('sha384', 'composer-setup.php') === file_get_contents('https://composer.github.io/installer.sig')) { echo 'Installer verified'; } else { echo 'Installer corrupt'; unlink('composer-setup.php'); } echo PHP_EOL;"
@@ -139,6 +135,7 @@ for version in $php_versions; do
 done
 
 # ubuntu 20.04 libzip-dev is libzip5 based and is not compatible libzip-dev of ppa:ondrej/php
+# see https://github.com/actions/virtual-environments/issues/1084
 if isUbuntu20 ; then
   rm /etc/apt/sources.list.d/ondrej-ubuntu-php-focal.list
   apt-get update
@@ -146,3 +143,9 @@ fi
 
 DocumentInstalledItem "Composer  ($(composer --version))"
 DocumentInstalledItem "PHPUnit ($(phpunit --version))"
+
+if isUbuntu20 ; then
+  AddBlockquote "To use ppa:ondrej/php APT repository On Ubuntu 20.04 it is necessary to add it to the APT sources"
+apt-add-repository ppa:ondrej/php -y
+apt-get update
+fi
--- a/images/linux/scripts/installers/php.sh
+++ b/images/linux/scripts/installers/php.sh
@@ -139,13 +139,12 @@ done
 if isUbuntu20 ; then
   rm /etc/apt/sources.list.d/ondrej-ubuntu-php-focal.list
   apt-get update
+  AddBlockquote "To use ppa:ondrej/php APT repository On Ubuntu 20.04 it is necessary to add it to the APT sources"
+  StartCode
+  WriteItem apt-add-repository ppa:ondrej/php -y
+  WriteItem apt-get update
+  EndCode
 fi
 
 DocumentInstalledItem "Composer  ($(composer --version))"
 DocumentInstalledItem "PHPUnit ($(phpunit --version))"
-
-if isUbuntu20 ; then
-  AddBlockquote "To use ppa:ondrej/php APT repository On Ubuntu 20.04 it is necessary to add it to the APT sources"
-apt-add-repository ppa:ondrej/php -y
-apt-get update
-fi
--- a/images/linux/scripts/helpers/document.sh
+++ b/images/linux/scripts/helpers/document.sh
@@ -28,3 +28,15 @@ function DocumentInstalledItem {
 function DocumentInstalledItemIndent {
     WriteItem "  - $1"
 }
+
+function AddBlockquote {
+    WriteItem "> $1"
+}
+
+function StartCode {
+    WriteItem "``````"
+}
+
+function EndCode {
+    WriteItem "``````"
+}


images/linux/scripts/installers/php.sh
```diff
@@ -93,6 +93,10 @@ apt-get remove --purge -yq php7.2-dev
 
 apt-fast install -y --no-install-recommends snmp
 
+if isUbuntu20 ; then
+    php_versions="7.4"
+fi
```
maxim-lobanov, line 98: Why you copy-paste these lines from 27-29 lines of this file?

images/linux/scripts/installers/php.sh
```diff
@@ -134,5 +138,11 @@ for version in $php_versions; do
     DocumentInstalledItem "PHP $version ($(php$version --version | head -n 1))"
 done
 
+# ubuntu 20.04 libzip-dev is libzip5 based and is not compatible libzip-dev of ppa:ondrej/php
```
miketimofeev, line 141: let's put the link to the issue here

shivammathur:
This is not a good approach to fix this dependency issue and this PR should not be merged. PHP projects rely on ondrej's PPA to install packages and removing it would create unnecessary confusion when extensions stop installing.

After PHP scripts are installed a fix can be added to install `libzip-dev` from the `ubuntu` archives and thus installing `libzip5`.

images/linux/scripts/installers/php.sh
dsame: For sake of further evolution. There will be ubuntu21 some day and thought it would be nice to have the suggestion for a developer it is possible to branch

dsame:
>  a fix can be added to install libzip-dev from the ubuntu archives

This approach was investigated as wall and seemed to be worst of all.

1. any actions user must be informed he must explicitly point its apt to focal repo 
2. libzip is not the only package that overwrites the ubuntu ones,  ondrej's PPA has lot of them

The "right way" is to update the PPA to do not have Ubuntu packages but the corresponding issue has been rejected https://github.com/oerdnj/deb.sury.org/issues/1417

dsame:
> Yes, the ondrej's PPA overwrites ubuntu packages, that is how PPAs work and the PHP projects depend on the packages in the PPA. Your argument to remove the PPAs because they overwrite ubuntu packages does not make sense to me. Are you saying that PPAs should not be used in the runner images.

The image we produce are used not only for the PHP development and the other guys who link their binaries against PPAs are definitely no happy their built binaries do not work in their prod environment because of  PPA absence. 

But i understand your concern, so, as a php expert, please help us with your thoughts about:

1. if we switch to Ubuntu PHP - is there predictable  downgrades comparing to PPA? Can we detect them with test case and fix somehow? Is there probability of the unpredictable problems?

2. What if we install PHP from PPA, next remove the repo from the apt sources list and make a note for PHP devs they need to add this repo back if needed? Could this work in theory?

shivammathur:
Option 1 is not a good one.
- It will install older versions of PHP packages on focal while bionic, xenial and other OS (windows and macOS) will have newer versions. Not good for CI, will require upgrading.
- It will also not provide monthly patch updates to PHP and other PHP extensions.
- It will not be possible to install older PHP versions like 7.3 and future PHP releases like 8.0 without adding back the PPA.
- The PPA has some PHP extensions which are not in ubuntu archives.
Therefore many projects will have to add the PPA and upgrade PHP packages or stick with ubuntu-18.04.

Option 2 is better, it will work with proper documentation in theory as adding back the PPA without needing to upgrade any package is not costly in runtime, but again it has the same problem you are trying to solve, the reach of the documentation. So you need to decide if more PHP users are there or people packaging software with libzip are there. This problem as I said before would amplify when you point `ubuntu-latest` to `ubuntu-20.04`

I would still suggest to keep the PPA and install both versions of the library and `libzip4` and `libzip5`. Both versions can be added as software installed in the docs. Anyone packaging software should have the know-how to add or remove PPA on their project or pick the right dependency. The patch in this PR can be applied on the project basis.

shivammathur:
@dsame Just increasing the priority of `libzip-dev` from `focal/universe` in the runner image should solve the issue. Then installing `libzip-dev` in a workflow would install the official one along with `libzip5` and thus `libzip.so` would be linked to `libzip.so.5`. Anyone building a binary from this would correctly link to `libzip.so.5`. This solution won't require removing the PPA or changing any docs.
Here is a [test workflow](https://github.com/shivammathur/test-setup-php/actions/runs/152543875/workflow)

dsame:
@shivammathur Thank you very much for info about ubuntu repo. Now it is clear the approach is not acceptable. 

So we have a choice between

1. disable PPA repo as a whole and make php devs to restore it back as whole 
2. Pin the only libzip-dev to focal repo and keep everything else as it is now

The second seems to be promising but my concern is there are many others common libs (including the notorious openssl) in PPA repo and having them in apt sources might cause unpredictable, not easy to detect, problems, so for sake of safety i still prefer turning PPA on and off. Probably we need to add an action task to do so...

Do you have any thoughts?

shivammathur:
@dsame Both the PPA and focal ubuntu archives have same patch version of `openssl` i.e `1.1.1` and both depend on the same version of `libssl` i.e `1.1`. So there should not be any linking issue there.

Anyway it would be fine if you prioritise all libraries in the PPA to the official versions after PHP setup. Just don't remove the PPA.
Here is the a list of packages you can prioritise.
`libargon2-dev`, `libicu-dev`, `icu-devtools`, `libgd-dev`, `libgd-tools`, `libidn2-dev`, `libxml2-dev`, `libzip-dev` and `openssl`

shivammathur:
@dsame Both the PPA and focal ubuntu archives have same patch version of `openssl` i.e `1.1.1` and both depend on the same version of `libssl` i.e `1.1`. So there should not be any linking issue there.

Anyway it would be fine if you prioritise all libraries in the PPA to the official versions after PHP setup which have different versions. Just don't remove the PPA.
Here is the a list of packages which are common.
`libargon2-dev`, `libicu-dev`, `icu-devtools`, `libgd-dev`, `libgd-tools`, `libidn2-dev`, `libxml2-dev`, `libzip-dev` and `openssl`

dsame:
@shivammathur prioritising the libraries can make problem for those PHP devs who will try to develop the php native modules and link them to the ubuntu ones instead of PPA. Well, it is not good to have a mix of libraries. Let's use "all or nothing" approach and let end user to switch from Ubuntu to PPA on his will. 

We just have to think out the smoothest way to do it

shivammathur:
I suggest pinning libraries only for those with different versions like `libzip-dev`.
If someone is using libraries in the PPA they would be already installed on the images and can be linked to. Also it would be trivial to remove pinning given the documentation.

If you are removing the PPA remove it after the PHP setup as in this PR, so that it can be added in PHP workflows and document the same. 

Installing PHP without the PPA would be a problem as I explained here https://github.com/actions/virtual-environments/pull/1104#issuecomment-651234850. This would require PHP devs to not only add the PPA but also upgrade PHP version along with all the PHP extensions. I would not recommend doing this.

images/linux/scripts/installers/php.sh
dsame: done

dsame:
@shivammathur 
We've found the action task https://github.com/shivammathur/setup-php your are maintaining - would it be possible to add 
it this task on ubuntu 20.04?

```
apt-add-repository ppa:ondrej/php -y
apt-get update
```

images/linux/scripts/installers/php.sh
```diff
@@ -134,5 +134,18 @@ for version in $php_versions; do
...
+  apt-get update
+fi
+
 DocumentInstalledItem "Composer  ($(composer --version))"
 DocumentInstalledItem "PHPUnit ($(phpunit --version))"
+
+if isUbuntu20 ; then
+  AddBlockquote "To use ppa:ondrej/php APT repository On Ubuntu 20.04 it is necessary to add it to the APT sources"
+apt-add-repository ppa:ondrej/php -y
+apt-get update
```
AlenaSviridenko, line 150: Should these two lines be in `StartCode `, `EndCode`?

shivammathur:
> @shivammathur
> We've found the action task https://github.com/shivammathur/setup-php your are maintaining - would it be possible to add
> it this task on ubuntu 20.04?
> 
> ```
> apt-add-repository ppa:ondrej/php -y
> apt-get update
> ```

Yes, I will add this.

images/linux/scripts/installers/php.sh
AlenaSviridenko: also, can this be merged with `if isUbuntu20 ; then` block above?

images/linux/scripts/installers/php.sh
```diff
@@ -134,5 +134,17 @@ for version in $php_versions; do
     DocumentInstalledItem "PHP $version ($(php$version --version | head -n 1))"
 done
 
+# ubuntu 20.04 libzip-dev is libzip5 based and is not compatible libzip-dev of ppa:ondrej/php
+# see https://github.com/actions/virtual-environments/issues/1084
+if isUbuntu20 ; then
+  rm /etc/apt/sources.list.d/ondrej-ubuntu-php-focal.list
+  apt-get update
+  AddBlockquote "To use ppa:ondrej/php APT repository On Ubuntu 20.04 it is necessary to add it to the APT sources"
```
miketimofeev, line 142: @dsame is this supposed to be in the Readme file? I'm asking because I can't see it there

Issue #1066, opened 2020-06-17
Build fails on issue in pypy.sh downloading compressed file


tkwaks:
Hi,

our build is failing after the last commit 330e62a on this file.
/virtual-environments/images/linux/scripts/installers/pypy.sh

After some investigation, I found a problem in line 76

In the pypy.sh install bash script there is an issue downloading the https://downloads.python.org/pypy/ file. The server responds by giving a compressed gzip file. 

on line 76 the code states:
pypyVersions="$(cat /tmp/pypyUrls.html | grep 'linux64' | awk -v uri="$uri" -F'>|<' '{print uri$5}')"

the grep command receives the compressed gzip file through the standard input from the cat command standard out. This can not work as the grep command cannot find any pattern of linux64 in the compressed input.

line 76 should be changed to decompress the /tmp/pypyUrls.html file and pipe its outcome to the grep command, like:
pypyVersions="$(gzip -cd /tmp/pypyUrls.html | grep 'linux64' | awk -v uri="$uri" -F'>|<' '{print uri$5}')"

I hope this can be fixed as this brakes our releases of new build agents.

Best regards,
Thomas Kwaks

vsafonkin:
Hi @tkwaks , thank you for your report! I've checked this code and it works correctly:
https://github.com/vsafonkin/check-pypy-versions/runs/780207913?check_suite_focus=true

We get html content from `pypyUrls.html` that contains urls for downloading. `Cat` command send simple html text to standard output that `grep` handle correctly.

tkwaks:
We still have the same issue:
 
![image](https://user-images.githubusercontent.com/26485925/84895286-64c45380-b0a2-11ea-82eb-0a1fdd602f20.png)

running your code on ubuntu 18.04 I see this issue still occurring
![image](https://user-images.githubusercontent.com/26485925/84898319-59275b80-b0a7-11ea-8a76-46ab6a3efd2b.png)

![image](https://user-images.githubusercontent.com/26485925/84898530-b91e0200-b0a7-11ea-895e-04b860ce91d7.png)

DiederickA:
In [this commit ](https://github.com/actions/virtual-environments/commit/330e62af9dc6e6d5f54c65b31ed4ff1871b5499b) the code to fetch this data changed from 
`curl -4 -s --compressed https://downloads.python.org/pypy/`
to
`wget https://downloads.python.org/pypy/  --output-document=pypy.html --no-verbose`

[...]

PR #1110, opened 2020-06-24
Fix download retry helper

vsafonkin:
# Description
Bug fixing: incorrect flag for `curl`  

#### Related issue:
#1066 

## Check list


images/linux/scripts/helpers/install.sh
```diff
@@ -12,11 +12,17 @@ download_with_retries() {
     local DEST="${2:-.}"
     local NAME="${3:-${URL##*/}}"
 
+    if [ $4 ]; then
```
miketimofeev, line 15: maybe use something like this to stay on a safe side?
if [ $4="-compressed" ]

images/linux/scripts/installers/pypy.sh
```diff
@@ -72,7 +72,7 @@ function InstallPyPy
 
 # Installation PyPy
 uri="https://downloads.python.org/pypy/"
-download_with_retries $uri "/tmp" "pypyUrls.html"
+download_with_retries $uri "/tmp" "pypyUrls.html" --compressed
```
miketimofeev, line 75: do we really need these `--`?

images/linux/scripts/helpers/install.sh
vsafonkin: Done

images/linux/scripts/installers/pypy.sh
vsafonkin: Fixed

Issue #1108, opened 2020-06-24
[Announcement] Update AWS CLI to V2 on Ubuntu 20.04 image on July 6


maxim-lobanov:
Hello,
Currently, AWS CLI V1 is installed on all Ubuntu images. Since V2 was released at the begin of this year, we would like to update our images.

Please pay attention that this upgrade could be breaking change for some customers. Please take a look at [migration guide for more details](https://docs.aws.amazon.com/cli/latest/userguide/cliv2-migration.html).
For now, we would like to update AWS CLI to V2 only on Ubuntu 20.04 image but we also consider the same updating on Ubuntu 18.04 and 16.04 images.
Please feel free to share your thoughts and opinion on it.

maxim-lobanov:
Hello!
AWS-CLI was updated to V2 on Ubuntu 20.04: `aws-cli/2.0.29 Python/3.7.3 Linux/5.4.0-1020-azure botocore/2.0.0dev33`

PR #1112, opened 2020-06-24
Update AWS CLI V1 to V2 on Ubuntu 20.04 Image

dibir-magomedsaygitov:
# Description
Bug fixing  
In scope of this PR we are updating AWS CLI V1 to V2 on Ubuntu 20.04 image to fix issue that occurs with AWS CLI V1 ('ascii' codec can't decode byte 0xc3 in position 66: ordinal not in range(128)). 
We are planning to merge this PR on 7/6/2020. Related announcement: [**#1108**](https://github.com/actions/virtual-environments/issues/1108) 

## Check list

### DIFF
--- a/images/linux/scripts/installers/aws.sh
+++ b/images/linux/scripts/installers/aws.sh
@@ -15,7 +15,9 @@ if isUbuntu20 ; then
     ./aws/install -i /usr/local/aws-cli -b /usr/local/bin
     rm awscliv2.zip
     rm -rf aws
-else
+fi
+
+if isUbuntu16 || isUbuntu18 ; then
     curl "https://s3.amazonaws.com/aws-cli/awscli-bundle.zip" -o "awscli-bundle.zip"
     unzip awscli-bundle.zip
     ./awscli-bundle/install -i /usr/local/aws -b /usr/local/bin/aws
--- a/images/linux/scripts/installers/aws.sh
+++ b/images/linux/scripts/installers/aws.sh
@@ -6,6 +6,7 @@
 
 # Source the helpers
 source $HELPER_SCRIPTS/document.sh
+source $HELPER_SCRIPTS/os.sh
 
 # Install the AWS CLI v1 on Ubuntu16 and Ubuntu18, and AWS CLI v2 on Ubuntu20
 if isUbuntu20 ; then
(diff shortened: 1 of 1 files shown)


images/linux/scripts/installers/aws.sh
```diff
@@ -7,12 +7,20 @@
 # Source the helpers
 source $HELPER_SCRIPTS/document.sh
```
al-cheb, line 8: `source $HELPER_SCRIPTS/os.sh` should be added before start using `IsUbuntu20`

dibir-magomedsaygitov: Thank you @al-cheb! Fixed.

images/linux/scripts/installers/aws.sh
```diff
@@ -6,13 +6,24 @@
...
 
-# Install the AWS CLI
-curl "https://s3.amazonaws.com/aws-cli/awscli-bundle.zip" -o "awscli-bundle.zip"
-unzip awscli-bundle.zip
-./awscli-bundle/install -i /usr/local/aws -b /usr/local/bin/aws
-rm awscli-bundle.zip
-rm -rf awscli-bundle
+# Install the AWS CLI v1 on Ubuntu16 and Ubuntu18, and AWS CLI v2 on Ubuntu20
+if isUbuntu20 ; then
+    curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip" -o "awscliv2.zip"
```
vmapetr, line 13: There is much sense to have two condition blocks with almost identical steps. Perhaps you can move installation logic to separate function?
```
if isUbuntu20; then
    install_aws "https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip" "/usr/local/aws-cli" "/usr/local/bin"
fi

if isUbuntu20; then
    install_aws "https://s3.amazonaws.com/aws-cli/awscli-bundle.zip" "/usr/local/aws" "/usr/local/bin/aws"
fi
```

dibir-magomedsaygitov: Unfortunately, we can't use same approach for both AWL CLI V1 and V2 installation. V2 contains some changes in source code, that's why we were forced to use two condition blocks.

PR #1121, opened 2020-06-25
[Ubuntu] Merge several toolset provisioners into single Configure-Toolset script

al-cheb:
# Description
This PR merges go.sh and boost.sh provisioners into single Configure-Toolset.ps1 script.

# Details
Configure-Toolset.ps1 is used to prepare system environment for toolset tools such as Boost and Go, and set environment variables for default tool versions. This PR also moves toolset.json config files into separate folder toolsets in image template root.

#### Related issue:
https://github.com/actions/virtual-environments-internal/issues/686

## Check list


images/linux/scripts/installers/Configure-Toolset.ps1
```diff
@@ -0,0 +1,84 @@
...
+}
+
+$toolset = Get-Content -Path "$env:INSTALLER_SCRIPT_FOLDER/toolset.json" -Raw | ConvertFrom-Json
+
+foreach ($tool in $toolset.toolcache)
+{
+    $toolName = $tool.name
+    $toolEnvironment = $toolsEnvironment[$toolName]
+
+    if ($toolEnvironment)
```
miketimofeev, line 57: asking just out of curiosity — do we have a plan to move Node and PyPy toolset installations to this script?

maxim-lobanov: I don't think that we should. Installation is still in `Install` script. This script just sets environment variables. As I know, we don't have any env vars for Node and PyPy

maxim-lobanov, line 57: nitpick:
may be we can replace `if ($toolEnvironment)` with `if (-not $toolEnvironment) continue;`
Just to reduce nesting and improve readability

images/linux/scripts/installers/Configure-Toolset.ps1
```diff
@@ -0,0 +1,84 @@
...
+                {
+                    Add-EnvironmentVariable -Name $toolEnvironment.defaultVariable -Value $toolPath
+                }
+
+                if ($toolEnvironment.command)
+                {
+                    $command = $toolEnvironment.command -f $toolPath
+                    Invoke-Expression -Command $command
+                }
+            }
```
maxim-lobanov, line 81: nitpick: May be we may move logic for default version outside loop and does something like that:
```
$toolPath = Get-ToolsetToolFullPath -ToolName $toolName -ToolVersion $tool.default-ToolArchitecture $tool.arch
if ($toolEnvironment.defaultVariable)
                {
                    Add-EnvironmentVariable -Name $toolEnvironment.defaultVariable -Value $toolPath
                }

                if ($toolEnvironment.command)
                {
                    $command = $toolEnvironment.command -f $toolPath
                    Invoke-Expression -Command $command
                }
```

just to reduce nesting and simplify readability

images/linux/scripts/installers/Configure-Toolset.ps1
```diff
@@ -0,0 +1,84 @@
...
+            $toolPath = Get-ToolsetToolFullPath -ToolName $toolName -ToolVersion $version -ToolArchitecture $tool.arch
+            $envName = $toolEnvironment.variableTemplate -f $version.split(".")
+
+            # Add environment variable name=value
+            Add-EnvironmentVariable -Name $envName -Value $toolPath
+
+            # Invoke command and add env variable for the default tool version
+            if ($version -eq $tool.default)
+            {
+                if ($toolEnvironment.defaultVariable)
```
maxim-lobanov, line 71: Let's add small `Write-Host` that we set default version

PR #721, opened 2020-04-15
Improve choco stability on Windows images

al-cheb:
# Description
We would like to implement installation retries for choco  when installing packages to reduce the number of transient errors. At the time of writing choco doesn't have logic to support wait/retry attempts -  https://github.com/chocolatey/choco/issues/385

Changes:
- Update `Install-Choco` function set by default `5 retries with  30 sec timeout` between attempts
- All `choco`  occurences have been replaced with `Install-Choco` function
- Files have been changed:

1. images/win/Windows2016-Azure.json
2. images/win/scripts/ImageHelpers/ChocoHelpers.ps1
3. images/win/scripts/Installers/Install-7zip.ps1
4. images/win/scripts/Installers/Install-AWS.ps1
5. images/win/scripts/Installers/Install-AliyunCli.ps1
6. images/win/scripts/Installers/Install-AzureCli.ps1
7. images/win/scripts/Installers/Install-Bazel.ps1
8. images/win/scripts/Installers/Install-Cmake.ps1
9. images/win/scripts/Installers/Install-Docker.ps1
10. images/win/scripts/Installers/Install-Edge.ps1
11. images/win/scripts/Installers/Install-Git.ps1
12. images/win/scripts/Installers/Install-GitVersion.ps1
13. images/win/scripts/Installers/Install-InnoSetup.ps1
14. images/win/scripts/Installers/Install-JavaTools.ps1
15. images/win/scripts/Installers/Install-Jq.ps1
16. images/win/scripts/Installers/Install-Julia.ps1
17. images/win/scripts/Installers/Install-KubernetesCli.ps1
18. images/win/scripts/Installers/Install-Mercurial.ps1
19. images/win/scripts/Installers/Install-MinGW.ps1
20. images/win/scripts/Installers/Install-NSIS.ps1
21. images/win/scripts/Installers/Install-NodeLts.ps1
22. images/win/scripts/Installers/Install-OpenSSL.ps1
23. images/win/scripts/Installers/Install-PHP.ps1
24. images/win/scripts/Installers/Install-Packer.ps1
25. images/win/scripts/Installers/Install-Perl.ps1
26. images/win/scripts/Installers/Install-Sbt.ps1
27. images/win/scripts/Installers/Install-Svn.ps1
28. images/win/scripts/Installers/Install-VSWhere.ps1
29. images/win/scripts/Installers/Install-Wix.ps1
30. images/win/scripts/Installers/Install-Zstd.ps1
[...]

PR #1123, opened 2020-06-25
Add workflows to create software PRs and GitHub releases

vsafonkin:
# Description
Improvement: added workflows that allow to use GitHub system token to create pull requests and releases.

#### Related issue:
Internal issue: [#721](https://github.com/actions/virtual-environments-internal/issues/721)

## Check list


.github/workflows/merge_pull_request.yml
```diff
@@ -0,0 +1,22 @@
+# This is a basic workflow to help you get started with Actions
```
AlenaSviridenko, line 1: I believe this could be safely removed

vsafonkin: Deleted, thanks!

.github/workflows/create_or_update_release.yml
```diff
@@ -0,0 +1,24 @@
...
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Create release for ${{ github.event.client_payload.ReleaseBranchName }}
+      uses: ncipollo/release-action@v1
```
alepauly, line 16: can we not use: https://github.com/actions/create-release?

.github/workflows/create_pull_request.yml
```diff
@@ -0,0 +1,21 @@
...
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Create pull request for ${{ github.event.client_payload.ReleaseBranchName }}
+      uses: repo-sync/pull-request@v2
```
alepauly, line 16: I'm surprised there isn't an easier way to create a PR since all the code is already pushed. This Action clones the whole repo locally. I think you can probably create one much easier by using https://github.com/actions/github-script to access https://octokit.github.io/rest.js/v18#pulls

.github/workflows/merge_pull_request.yml
```diff
@@ -0,0 +1,20 @@
...
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Merge pull request for ${{ github.event.client_payload.ReleaseBranchName }}
+      uses: juliangruber/merge-pull-request-action@v1
```
alepauly, line 16: same as above, you can probably just use octokit for this.

.github/workflows/create_pull_request.yml
vsafonkin: I didn't know about this way, it's great, thank you! PR is reworked.

.github/workflows/merge_pull_request.yml
vsafonkin: Done.

.github/workflows/create_or_update_release.yml
vsafonkin: Сreate-release action task not allows to update existing release, so, I've added another workflow for updating via github-script

PR #1126, opened 2020-06-26
Migrate php to 20.04

dsame:
# Description

## Check list

### DIFF
--- a/images/linux/scripts/installers/2004/php.sh
+++ b/images/linux/scripts/installers/2004/php.sh
@@ -10,24 +10,11 @@ source $HELPER_SCRIPTS/document.sh
 source $HELPER_SCRIPTS/os.sh
 
 set -e
-# since 20.04 php is installed from Ubuntu repository
-# see https://github.com/actions/virtual-environments/issues/1084
-if isUbuntu20 ; then
-    echo "This script must not be used for Ubuntu 20.04"
-    exit 1
-fi
-
-# add repository
-apt-add-repository ppa:ondrej/php -y
-apt-get update
 
 # Install PHP
-if isUbuntu16 ; then
-    php_versions="5.6 7.0 7.1 7.2 7.3 7.4"
-fi
 
-if isUbuntu18 ; then
-    php_versions="7.1 7.2 7.3 7.4"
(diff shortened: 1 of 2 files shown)


images/linux/scripts/installers/2004/php.sh
```diff
@@ -0,0 +1,127 @@
...
+# Source the helpers for use with the script
+source $HELPER_SCRIPTS/etc-environment.sh
+source $HELPER_SCRIPTS/document.sh
+source $HELPER_SCRIPTS/os.sh
+
+set -e
+
+# Install PHP
+
+if isUbuntu20 ; then
```
AlenaSviridenko, line 16: I think this condition is not needed because this file in Ubuntu 20 folder

dsame: Agree

images/linux/scripts/installers/php.sh
```diff
@@ -10,6 +10,12 @@ source $HELPER_SCRIPTS/document.sh
 source $HELPER_SCRIPTS/os.sh
 
 set -e
+# since 20.04 php is installed from Ubuntu repository
+# see https://github.com/actions/virtual-environments/issues/1084
+if isUbuntu20 ; then
```
al-cheb, line 15: Why do you split `php.sh` if you can simple add `if` statement?

```
if isUbuntu16 || isUbuntu18; then
    # add repository
    apt-add-repository ppa:ondrej/php -y
    apt-get update
fi
```

shivammathur:
This approach is even worse than #1104, as installing PHP 7.4 without ondrej's PPA on focal would mean different patch releases for PHP 7.4 between bionic and focal. Please see https://github.com/actions/virtual-environments/pull/1104#issuecomment-650711664.

dsame:
>  between bionic and focal

but why don't they stay with bionic if they depends on ondrej's PPA so much?

images/linux/scripts/installers/php.sh
dsame: Using the other repo seems to be a very significant change which should not be handled by "if" branches. Different package names and such. I just created the VM and testing

PR #1148, opened 2020-06-29
PostgreSQL: Validation and Installer parameters fix.

Darleev:
# Description
Bug fixing
Validation fails due to PGBIN local parameter absence, need to specify it as environment variables. /NoPath parameters was ignored since called unproperly.

#### Related issue:
No issue/Reported by @miketimofeev 

## Check list


al-cheb:
We should fix display version of PostGreSQL, now it's 11.3, but in the log is - https://get.enterprisedb.com/postgresql/postgresql-12.3-1-windows-x64.exe'

PR #1149, opened 2020-06-29
Tests for Windows.

Darleev:
# Description
Improvement
It is required to improve tools validation on Windows images. So, several pester tests have been added to show how it should work.

#### Related issue:
222

## Check list


images/win/scripts/Tests/Validate-Java.ps1
```diff
@@ -0,0 +1,50 @@
...
+        {
+            Write-Host "Java $(java -version) on path"
+            Write-Host "Maven $(mvn -version) on path"
+            Write-Host "Ant $(ant -version) on path"
+            Write-Host "Gradle $(gradle -version) on path"
+        }
+        else
+            {
+            Write-Host "one of Java,Maven,Ant,Gradle is not on path."
+            exit 1
```
maxim-lobanov, line 30: We must not do `exit 1` in Pester tests. This command will cause script failure but the idea of Pester tests is running all test cases from file and then report summary.
Also every `It` block must contain at least one `Should` statement. And ideally, every `It` should test only one thing.

I suggest to rework `It "Java, Gradle, Maven, Ant added to the PATH:"` to something like that:
```
@("java", "gradle", "maven", "ant") | ForEach-Object {
  It "$_" {
    Validate-ZeroExitCode "$($_) --version" | Should -Be 0
  }
}
```

images/win/scripts/Tests/Validate-Java.ps1
```diff
@@ -0,0 +1,50 @@
...
+        if ($version -In @(7, 8)) {
+            # old Java versions look like "1.7", "1.8"
+            $version = "1." + $version
+        }
+
+        It "Java $version" {
+            $javaenv = "JAVA_HOME_${_}_X64"
+            $javavar = Get-EnvironmentVariable $javaenv
+            $javapath = Join-Path $javavar "bin\java"
+            Validate-ZeroExitCode "Start-Process $javapath -Argument version"
```
maxim-lobanov, line 46: This statement doesn't test anything and will pass even if exit code is not 0. We should add `Should -Be 0` at the end

images/win/scripts/Tests/Validate-MongoDB.ps1
```diff
@@ -0,0 +1,27 @@
...
+##  Desc:  Validate MongoDB
+################################################################################
+Describe "MongoDB" {
+
+    It "mongod is on path." {
+        if (Get-Command -Name 'mongod') {
+            Write-Host 'mongod is on path'
+        }
+        else {
+            Write-Host 'mongod not on path'
```
maxim-lobanov, line 12: We should not to print anything in Pester tests. Pester produces enough logging by itself

maxim-lobanov: Also we should not write `exit 1`. See my message above.
I suggest to rework test to 
```
It "mongod" {
   Validate-ZeroExitCode "mongod --version" | Should -Be 0
}
```

maxim-lobanov: The same for all other tests

images/win/scripts/Tests/run-tests.ps1
```diff
@@ -0,0 +1,6 @@
+################################################################################
+##  File:  run-tests.ps1
+##  Desc:  run all pester tests
+################################################################################
+
+Get-ChildItem  | % {  & $_.FullName }
```
maxim-lobanov, line 6: I believe this command won't run Pester tests. It will just run powershell files as usual scripts.
You should run tests like `Invoke-Pester -Path <test_folder>\Validate-*.ps1`

images/win/Windows2016-Azure.json
```diff
@@ -498,12 +498,24 @@
                 "{{ template_dir }}/scripts/Installers/Install-JavaTools.ps1"
             ]
         },
+        {
+            "type": "powershell",
+            "scripts":[
+                "{{ template_dir }}/scripts/Tests/Validate-Java.ps1"
```
maxim-lobanov, line 504: This command won't run Pester. It will just run ps1 file as usual script. We should use syntax `Invoke-Pester Tests/Validate-MongoDB.ps1` to run pester tests on file.
Also I suggest to remove test invocations from template at all and run them from Installation files directly

PR #1158, opened 2020-07-01
[Windows] Add available disk space info to markdown

vmapetr:
# Description
This PR add info about available disk space to generated markdown.

**Example:**
```
# Microsoft Windows Server 2019 Datacenter
- OS Version: 10.0.17763 Build 17763
- Image Version: 20200630.0

### Available disk space
| Drive Label | Available space (GB) |
| ----------- | -------------------- |
| C:          | 90                   |
*Size of provided mounted SSD may vary, please check [documentation.](https://docs.github.com/en/actions/reference/virtual-environments-for-github-hosted-runners#supported-runners-and-hardware-resources)

## Installed Software
### Language and Runtime
- Java 1.7.0_232 

```

#### Related issue:
https://github.com/actions/virtual-environments-internal/issues/759

## Check list

### DIFF
--- a/images/win/scripts/SoftwareReport/SoftwareReport.Common.psm1
+++ b/images/win/scripts/SoftwareReport/SoftwareReport.Common.psm1
@@ -17,8 +17,8 @@ function Get-AvailableDiskSpace {
     {
         $availableDiskSpace = [math]::Round((Get-PSDrive -Name $driveLabel).Free / $sizeMultiplier)
         [PSCustomObject]@{
-            "Drive Label" = $driveLabel
-            "Available space ${sizeMultiplierLabel}" = $availableDiskSpace
+            "Drive Label" = "${driveLabel}:"
+            "Available space (${sizeMultiplierLabel})" = $availableDiskSpace
         }
     }
 }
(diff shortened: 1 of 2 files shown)


al-cheb:
I don't think we should provide unreliable information about disk space it will create a lot of question.

AlenaSviridenko:
I think we should use Gb measurement instead of Mb, it is more readable than Mb.

maxim-lobanov:
@al-cheb , what do you mean "unreliable"? The idea is adding free space to documentation so we will be able to track it in dynamic.

images/win/scripts/SoftwareReport/SoftwareReport.Common.psm1
```diff
@@ -8,6 +8,21 @@ function Get-OSVersion {
     return "OS Version: $OSVersion Build $OSBuild"
 }
 
+function Get-AvailableDiskSpace {
+    $drives = @("C", "D")
```
maxim-lobanov, line 12: I thought that D drive doesn't exist during image-generation and it is attached only during deployment? But I could be wrong

vmapetr: Yep, I've forgot about it. I don't think it's a good idea to hardcode expectable mount size, so seems like, we should remove info about it completely.

vmapetr: Fixed.

Issue #1136, opened 2020-06-26
homebrew executables are superseded by system ones


andreineculau:
**Describe the bug**
Homebrew (or specifically linuxbrew) was added to the mix in https://github.com/actions/virtual-environments/pull/589 .
Then came https://github.com/actions/virtual-environments/pull/789 which turned off the expected behaviour, which is brew executables take precedence over the system ones. This is the expected behaviour, no matter what some might think, and it's what `brew shellenv` outputs as well.

I wrote more on the topic in https://github.com/actions/virtual-environments/pull/789#issuecomment-650087417 .

What is not clear is the reason why the default behaviour (prepending homebrew's paths to PATH) does work well with the Github Actions images.

As per @miketimofeev's comment https://github.com/actions/virtual-environments/pull/789#issuecomment-650121339 , the reason was: `we keep system python is that it contains some pip packages like request, which are essential for some internal teams and we received a complaint right after python from brew was put in the PATH.`

I fear the argument doesn't stand on its own. 

First of all `apt-get install python python-pip && pip list` will show that there are only 3 default packages, namely: pip setuptools wheel. These happens in all 3 Ubuntu distros (`docker run -it ubuntu:XX.04`), both for python (2) and python3 aptitude packages. Thus, whatever packages the internal teams expect to be installed are coming from some obscure intermediate steps, rather than the default installation of python.

Second of all, if the internal teams require a dependency, then they need to make sure that they actually install that dependency. As it is now, the complaint sounds like "Before #589, we had access to our hidden dependency. Now we don't, and we don't want to make the hidden dependency explicit."

PS: if someone takes this bug report with a 30000 feet perspective of fixing the homebrew installation as a whole, note that since homebrew executable will take precedence over the system ones, homebrew installer should be called as one of the very last things, if not the very last thing. Example: installing homebrew, than python (via aptitude) and calling `python` to see if the aptitude installation succeeded is going to yield false results (in its current form) because you will test for and invoke homebrew's python, not the system's python.

**Area for Triage**: 

Deployment/Release ?? 🤷 

**Question, Bug, or Feature?**: 

Bug

**Virtual environments affected**
- [ ] macOS 10.15
- [x] Ubuntu 16.04 LTS
- [x] Ubuntu 18.04 LTS
- [x] Ubuntu 20.04 LTS
- [ ] Windows Server 2016 R2
- [ ] Windows Server 2019

**Expected behavior**

[...]

PR #1160, opened 2020-07-02
[Ubuntu] Install aws-sam-cli without brew

miketimofeev:
# Description
Aws-sam-cli is supposed to be installed via homebrew yet that approach leads to install separate python3, which overrides system one and causes issues for some customers. The workaround was implemented here https://github.com/actions/virtual-environments/pull/789 but it turned out that brew functionality is partly broken after the changes.
The solution is to get-rid of brew aws-sam cli installation and use `setup.py` script that comes along with aws-sam sources. After these changes, another PR will be made to move homebrew to the beginning of the PATH as it was initially.

Changes:
- Aws-sam-cli installation switched to build from sources via setup.py script
- Installation script moved to the end of the template in order to run after toolset installation

#### Related issue:
https://github.com/actions/virtual-environments/issues/1136

## Check list

### DIFF
--- a/images/linux/scripts/installers/aws-sam-cli.sh
+++ b/images/linux/scripts/installers/aws-sam-cli.sh
@@ -2,7 +2,7 @@
 ################################################################################
 ##  File:  aws-sam-cli.sh
 ##  Desc:  Installs AWS SAM CLI
-##         Requires Pyton >=3.6, must be run as non-root user after toolset installation
+##         Requires Python >=3.6, must be run as non-root user after toolset installation
 ################################################################################
 
 # Source the helpers for use with the script
--- a/images/linux/scripts/installers/aws-sam-cli.sh
+++ b/images/linux/scripts/installers/aws-sam-cli.sh
@@ -13,20 +13,19 @@ TarballUrl=$(curl -s https://api.github.com/repos/awslabs/aws-sam-cli/releases/l
 TarballPath="/tmp/aws-sam-cli.tar.gz"
 wget $TarballUrl -O $TarballPath
 tar -xzvf $TarballPath -C /tmp
-SourcesDir=$(echo /tmp/awslabs-aws-sam-cli*)
+cd /tmp/awslabs-aws-sam-cli*
 
 # Use python 3.7 from toolcache to install aws sam, setuptools package required for the installation
 Python3Dir=$(echo ${AGENT_TOOLSDIRECTORY}/Python/3.7*/x64)
 Python3BinDir="${Python3Dir}/bin"
 export PATH="$Python3Dir:$Python3BinDir:$PATH"
 python3 -m pip install setuptools
-cd $SourcesDir
 python3 setup.py install
 sudo ln -sf ${Python3BinDir}/sam /usr/local/bin/sam
 
 # Cleanup downloaded files
 rm -rf $TarballPath
-rm -rf $SourcesDir
+rm -rf /tmp/awslabs-aws-sam-cli*
 
 # Run tests to determine that the software installed as expected
 echo "Testing to make sure that script performed as expected, and basic scenarios work"
--- a/images/linux/scripts/installers/aws-sam-cli.sh
+++ b/images/linux/scripts/installers/aws-sam-cli.sh
@@ -23,10 +23,6 @@ python3 -m pip install setuptools
 python3 setup.py install
 sudo ln -sf ${Python3BinDir}/sam /usr/local/bin/sam
 
-# Cleanup downloaded files
-rm -rf $TarballPath
-rm -rf /tmp/awslabs-aws-sam-cli*
-
 # Run tests to determine that the software installed as expected
 echo "Testing to make sure that script performed as expected, and basic scenarios work"
 if ! sam --version; then
--- a/images/linux/scripts/installers/aws-sam-cli.sh
+++ b/images/linux/scripts/installers/aws-sam-cli.sh
@@ -17,7 +17,7 @@ SourcesDir=$(echo /tmp/awslabs-aws-sam-cli*)
 
 # Use python 3.7 from toolcache to install aws sam, setuptools package required for the installation
 Python3Dir=$(echo ${AGENT_TOOLSDIRECTORY}/Python/3.7*/x64)
-Python3BinDir=$(echo ${Python3Dir}/bin)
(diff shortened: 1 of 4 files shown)


images/linux/scripts/installers/aws-sam-cli.sh
```diff
@@ -2,15 +2,31 @@
...
 # Source the helpers for use with the script
 source $HELPER_SCRIPTS/document.sh
 
-# Install aws sam cli
-brew tap aws/tap
-brew install aws-sam-cli
+# Download latest aws sam cli sources
+TarballUrl=$(curl -s https://api.github.com/repos/awslabs/aws-sam-cli/releases/latest | jq -r '.tarball_url')
+TarballPath="/tmp/aws-sam-cli.tar.gz"
+wget $TarballUrl -O $TarballPath
```
vsafonkin, line 14: Does it make sense to use `download_with_retries()` helper?

miketimofeev: I've tried but it looks like curl can't download from such links with redirection or maybe some flag is supposed to be specified

images/linux/ubuntu1604.json
```diff
@@ -288,6 +276,18 @@
...
         },
+        {
+            "type": "shell",
+            "scripts": [
+                "{{template_dir}}/scripts/installers/aws-sam-cli.sh"
+            ],
+            "environment_vars": [
+                "METADATA_FILE={{user `metadata_file`}}",
+                "HELPER_SCRIPTS={{user `helper_script_folder`}}",
+                "DEBIAN_FRONTEND=noninteractive"
```
al-cheb, line 287: Do we need it?

images/linux/scripts/installers/aws-sam-cli.sh
```diff
@@ -2,15 +2,31 @@
...
+tar -xzvf $TarballPath -C /tmp
+SourcesDir=$(echo /tmp/awslabs-aws-sam-cli*)
+
+# Use python 3.7 from toolcache to install aws sam, setuptools package required for the installation
+Python3Dir=$(echo ${AGENT_TOOLSDIRECTORY}/Python/3.7*/x64)
+Python3BinDir=$(echo ${Python3Dir}/bin)
```
al-cheb, line 20: nipick: "${Python3Dir}/bin"

images/linux/scripts/installers/aws-sam-cli.sh
```diff
@@ -2,15 +2,31 @@
...
 
 # Source the helpers for use with the script
 source $HELPER_SCRIPTS/document.sh
 
-# Install aws sam cli
-brew tap aws/tap
-brew install aws-sam-cli
+# Download latest aws sam cli sources
+TarballUrl=$(curl -s https://api.github.com/repos/awslabs/aws-sam-cli/releases/latest | jq -r '.tarball_url')
+TarballPath="/tmp/aws-sam-cli.tar.gz"
```
AlenaSviridenko, line 13: I think our bash variables naming convention is either ALL_UPPER_CASE or all_lower_case symbols with underscores.

images/linux/scripts/installers/aws-sam-cli.sh
miketimofeev: Thanks!

images/linux/ubuntu1604.json
miketimofeev: Nope, forgot to remove it, thanks!

images/linux/scripts/installers/aws-sam-cli.sh
miketimofeev: Judging by the other scripts we don't have any convention 😀
I've seen both your options and Julia.sh contains the same name types
image

images/linux/scripts/installers/aws-sam-cli.sh
```diff
@@ -2,15 +2,31 @@
...
+wget $TarballUrl -O $TarballPath
+tar -xzvf $TarballPath -C /tmp
+SourcesDir=$(echo /tmp/awslabs-aws-sam-cli*)
```
al-cheb, line 16: I think we can rid of SourcesDir variable and move cd from 23 line:
`cd /tmp/awslabs-aws-sam-cli*`

images/linux/scripts/installers/aws-sam-cli.sh
```diff
@@ -2,15 +2,31 @@
...
+# Use python 3.7 from toolcache to install aws sam, setuptools package required for the installation
+Python3Dir=$(echo ${AGENT_TOOLSDIRECTORY}/Python/3.7*/x64)
+Python3BinDir=$(echo ${Python3Dir}/bin)
+export PATH="$Python3Dir:$Python3BinDir:$PATH"
+python3 -m pip install setuptools
+cd $SourcesDir
+python3 setup.py install
+sudo ln -sf ${Python3BinDir}/sam /usr/local/bin/sam
+
+# Cleanup downloaded files
```
al-cheb, line 27: nitpick: we can remove this section and leave it the `cleanup.sh` script

images/linux/scripts/installers/aws-sam-cli.sh
AlenaSviridenko: ok than, disregard my comment

images/linux/scripts/installers/aws-sam-cli.sh
miketimofeev: Cool, thanks!

images/linux/scripts/installers/aws-sam-cli.sh
```diff
@@ -2,15 +2,26 @@
 ################################################################################
 ##  File:  aws-sam-cli.sh
 ##  Desc:  Installs AWS SAM CLI
-##         Must be run as non-root user after homebrew and clang
+##         Requires Pyton >=3.6, must be run as non-root user after toolset installation
```
maxim-lobanov, line 5: typo: `Pyton`

images/linux/scripts/installers/aws-sam-cli.sh
miketimofeev: yeah, -L flag should be specified, which is not implemented in the download_with_retries function

PR #1168, opened 2020-07-03
[Ubuntu] Prepend brew directories to the PATH

miketimofeev:
# Description
This is the second PR to resolve the initial issue. We don't use brew to install tools anymore after https://github.com/actions/virtual-environments/pull/1160 so brew executables can be safely placed at the beginning of the `PATH` in order to install tools via brew **in runtime** correctly
`/usr/share/rust/.cargo/bin:/home/runner/.config/composer/vendor/bin:/home/runner/.dotnet/tools:/snap/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/home/linuxbrew/.linuxbrew/bin:/home/linuxbrew/.linuxbrew/sbin` -> `/usr/share/rust/.cargo/bin:/home/runner/.config/composer/vendor/bin:/home/runner/.dotnet/tools:/home/linuxbrew/.linuxbrew/bin:/home/linuxbrew/.linuxbrew/sbin:/snap/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games`
As a note: clean brew installation doesn't contain any packages thus it's supposed to be safe to install brew at any time of the image build.

#### Related issue:
https://github.com/actions/virtual-environments/issues/1136

## Check list
- [-] Tests are written (if applicable)
- [-] Documentation is updated (if applicable)

### DIFF
--- a/images/linux/scripts/installers/aws-sam-cli.sh
+++ b/images/linux/scripts/installers/aws-sam-cli.sh
@@ -16,7 +16,7 @@ tar -xzvf $TarballPath -C /tmp
 cd /tmp/awslabs-aws-sam-cli*
 
 # Use python 3.7 from toolcache to install aws sam, setuptools package required for the installation
-Python3Dir=$(echo ${AGENT_TOOLSDIRECTORY}/Python/3.7*/x64)
+Python3Dir=$(echo /opt/hostedtoolcache/Python/3.7*/x64)
 Python3BinDir="${Python3Dir}/bin"
 export PATH="$Python3Dir:$Python3BinDir:$PATH"
 python3 -m pip install setuptools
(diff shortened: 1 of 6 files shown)


maxim-lobanov:
Does it also make sense to move brew installation at the end of image-generation template? And put some addittional comments that we mustn't use brew to install different tools during image-generation

miketimofeev:
> Does it also make sense to move brew installation at the end of image-generation template? And put some addittional comments that we mustn't use brew to install different tools during image-generation

I've added a not for that question in the description 😀 Brew can be installed at any place, I can't see any **dangerous** executables in the clean brew installation

Issue #1133, opened 2020-06-26
Add/Update Xcode 12.0 beta


ryanwilson:
**Tool information**
- Tool name: Xcode 12
- Tool license: https://www.apple.com/legal/sla/docs/xcode.pdf
- Add or update? Add
- Desired version: 12.0 beta 1
- Approximate size: ~15GB
- If this is an add request:
  - Brief description of tool: The latest beta of Apple's Xcode IDE
  - URL for tool's homepage: https://developer.apple.com/xcode/

**Area for Triage**:
Apple

**Question, Bug, or Feature?**:
Feature

**Virtual environments affected**
- [X] macOS 10.15
- [ ] Ubuntu 16.04 LTS
- [ ] Ubuntu 18.04 LTS
- [ ] Ubuntu 20.04 LTS
- [ ] Windows Server 2016 R2
- [ ] Windows Server 2019

**Can this tool be installed during the build?**
I'm not entirely sure.

**Tool installation time in runtime**
I'm not entirely sure.

**Are you willing to submit a PR?**
I can try my best, would likely need some guidance though.

maxim-lobanov:
Hello, we have already started to work on this request. We are planning to deploy new image with Xcode 12 at the middle next week (if there is no any issues)

ryanwilson:
Awesome, thanks for the update! 😄 very happy to hear. Didn't see an open issue so thought I'd create one just in case.

lewisdonofrio:
is anyone working on xcode IDE for ubuntu?  I mean if you want to be cross-platform, be cross platform ;)

maxim-lobanov:
@lewisdonofrio , Currently, Xcode doesn't support other platforms except MacOS. Feel free to create feature request for Apple

MaxDesiatov:
@lewisdonofrio if you're referring to a Swift IDE here, you can use Visual Studio code with [SourceKit-LSP](https://marketplace.visualstudio.com/items?itemName=pvasek.sourcekit-lsp--dev-unofficial) to get auto-complete working.

markavitale:
@maxim-lobanov is the current plan to support/add both the Xcode 12 beta (build 12A6159) and the Xcode 12 for macOS Universal Apps beta (build 12A8158a)? Our macOS pipelines will likely require the "macOS universal apps beta" to start testing ARM support.

maxim-lobanov:
[...]

PR #1173, opened 2020-07-04
Update documented Xcode 12 path

kiliankoe:
# Description
It appears the install path of Xcode 12 (or rather the name of the application) has been updated. This changes the documentation in the readme to reflect that.

#### Related issue:
https://github.com/actions/virtual-environments/issues/1133

### DIFF
--- a/images/macos/macos-10.15-Readme.md
+++ b/images/macos/macos-10.15-Readme.md
@@ -171,7 +171,7 @@
 ### Xcode
 | Version        | Build   | Path                              |
 | -------------- | ------- | --------------------------------- |
-| 12.0 (beta)    | 12A6159 | /Applications/Xcode_12_beta.app   |
+| 12.0 (beta)    | 12A6159 | /Applications/Xcode_12.app   |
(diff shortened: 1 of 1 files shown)


maxim-lobanov:
Hello, we should not update documentation manually since it is autogenerated.
Actually, Xcode 12 is available on both paths: `/Applications/Xcode_12.app` and `/Applications/Xcode_12_beta.app, we just add one more symlink. In documentation, we usually put `beta` symlink for beta versions. Please see Xcode 11.6 as example.

kiliankoe:
Interesting, I was using Xcode_12_beta.app successfully until xcode-select started failing yesterday saying that no such application was available. It appears to only be available as Xcode_12.app now.

maxim-lobanov:
@kiliankoe , wow, I just found that not all VM machines were updated properly. For some machines, neither `/Applications/Xcode_12.app` nor `/Applications/Xcode_12_beta.app` are available. They have image version `20200618.1`. The correct version is `20200625.2`. We will take care about fixing it on Monday. Thank you for report

kiliankoe:
Ah, thanks for the update! And thanks for the upcoming fix as well 😊

Issue #1145, opened 2020-06-29
Add AWS CLI Session Manager Plugin


nicolasnoel:
**Tool information**
- Tool name: AWS CLISession Manager Plugin
- Tool license: 
- Add or update? Add
- Desired version: latest
- Approximate size: 5 MB
- If this is an add request:
  - Brief description of tool: Allows ssh tunneling to ec2 instance through AWS System Manager
  - URL for tool's homepage: https://docs.aws.amazon.com/systems-manager/latest/userguide/session-manager-working-with-install-plugin.html

**Area for Triage**:

**Question, Bug, or Feature?**:

**Virtual environments affected**
- [x] macOS 10.15
- [x] Ubuntu 16.04 LTS
- [x] Ubuntu 18.04 LTS
- [x] Ubuntu 20.04 LTS
- [x] Windows Server 2016 R2
- [x] Windows Server 2019

**Can this tool be installed during the build?**
No, not on ubuntu 18 : need to install a system package

**Tool installation time in runtime**
< 1 mn

**Are you willing to submit a PR?**
No

nicolasnoel:
Hi,
thanks a lot, i could test it today : runned an action that uses Packer to create an AMI in a VPC _**without**_ openning any port on the packer builder instance, ssh connection going through aws ssm, thanks to session manager plugin. 
That's great !

PR #1178, opened 2020-07-06
Add aws session manager windows

dmitry-shibanov:
# Description
 - Add aws cli session manager.
### images:
 - [windows19](https://github.visualstudio.com/virtual-environments/_build/results?buildId=74607&view=results)
 - [windows16](https://github.visualstudio.com/virtual-environments/_build/results?buildId=74608&view=results)
#### Related issue:
https://github.com/actions/virtual-environments/issues/1145
## Check list


images/win/scripts/Installers/Validate-AWS-CLI-SM.ps1
```diff
@@ -0,0 +1,14 @@
+################################################################################
+##  File:  Validate-AWS-CLI-SM.ps1
+##  Desc:  Validate aws cli session manager
+################################################################################
+
+$sessionMessage = session-manager-plugin
+
+if ($sessionMessage -Match "*plugin was installed successfully*") {
+    Write-Host "$sessionMessage"
```
maxim-lobanov, line 9: Can we put `Write-Host "$sessionMessage"` before `if`. So we won't need to duplicate it.
Also I believe `exit 0` doesn't make more sense.

images/win/scripts/Installers/Validate-AWS-CLI-SM.ps1
```diff
@@ -0,0 +1,14 @@
+################################################################################
+##  File:  Validate-AWS-CLI-SM.ps1
```
maxim-lobanov, line 2: Do we really need the separate script for validation if we install in the single script?

images/win/scripts/Installers/Validate-AWS-CLI-SM.ps1
dmitry-shibanov: Yes, sure. Should we totally remove if check because if command not found it will fail.

dmitry-shibanov: Done.

images/win/scripts/Installers/Validate-AWS-CLI-SM.ps1
dmitry-shibanov: Done.

PR #1208, opened 2020-07-09
[Ubuntu] Fix 'Unable to acquire the dpkg frontend lock' error

miketimofeev:
# Description
About 30% of Ubuntu builds fail with the error `E: Unable to acquire the dpkg frontend lock (/var/lib/dpkg/lock-frontend), is another process using it?` in some random places in the first 10-30 minutes of the build.
image
It looks like the root cause is in the `source $HELPER_SCRIPTS/apt.sh` invocation in some of the installation scripts. That `apt.sh` script contains several `apt-get` commands and probably sometimes `/var/lib/dpkg/lock-frontend` is still blocked when installation script tries to launch `apt-get` to install something. Moreover, that doesn't make sense to invoke `apt-get update` at the beginning of the installation script because we run `apt-get update` after `add-apt-repository` anyway.

Changes:
- Logic from `helpers/apt.sh` moved to base/apt.sh(to invoke it only once at the beginning of the build) and to helpers/install.sh to use `IsInstalled` function when needed
- `source $HELPER_SCRIPTS/apt.sh` were removed from all installation scripts
- `apt-get -yqq` changed to `apt-get -yq` because it's not recommended to use `qq` without a no-action modifier
https://manpages.debian.org/stretch/apt/apt-get.8.en.html
image

#### Related issue:

## Check list


images/linux/scripts/helpers/install.sh
```diff
@@ -33,4 +33,13 @@ download_with_retries() {
...
     echo "Could not download $URL"
     return 1
+}
+
+## Use dpkg to figure out if a package has already been installed
+## Example use:
+## if ! IsInstalled packageName; then
+##     echo "packageName is not installed!"
+## fi
+function IsInstalled {
```
maxim-lobanov, line 43: The function name looks too generic for me.
`IsInstalled` sounds like we check if particular tool is installed on machine but actually this function checks only dpkg.

miketimofeev: we can rename it to IsPackageInstalled, does it sound better?

maxim-lobanov: Yes, I think so, thanks

PR #1228, opened 2020-07-13
[Windows] Remove WSL from software output on Windows 2016

miketimofeev:
# Description
WSL was added to Windows 2019 only in the scope of this PR, but it seems we forget to add a condition in the software report generator.

#### Related issue:
https://github.com/actions/virtual-environments/issues/50

## Check list

### DIFF
--- a/images/win/scripts/SoftwareReport/SoftwareReport.Generator.ps1
+++ b/images/win/scripts/SoftwareReport/SoftwareReport.Generator.ps1
@@ -22,10 +22,13 @@ $markdown += New-MDList -Style Unordered -Lines @(
     "Image Version: $env:ImageVersion"
 )
 
-$markdown += New-MDHeader "Enabled windows optional features" -Level 2
-$markdown += New-MDList -Style Unordered -Lines @(
-    "Windows Subsystem for Linux"
-)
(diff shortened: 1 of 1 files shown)


images/win/scripts/SoftwareReport/SoftwareReport.Generator.ps1
```diff
@@ -22,10 +22,13 @@ $markdown += New-MDList -Style Unordered -Lines @(
     "Image Version: $env:ImageVersion"
 )
 
-$markdown += New-MDHeader "Enabled windows optional features" -Level 2
-$markdown += New-MDList -Style Unordered -Lines @(
-    "Windows Subsystem for Linux"
-)
+if ($OSName -match "Microsoft Windows Server 2019 Datacenter")
```
maxim-lobanov, line 25: There is a function: ` Test-IsWin19`. I believe we should use it
https://github.com/actions/virtual-environments/blob/master/images/win/scripts/ImageHelpers/InstallHelpers.ps1#L440

miketimofeev: Thanks! Replaced.

PR #1232, opened 2020-07-14
Replace current validates with pester testing approach #5

al-cheb:
# Description
Replace current Validate scripts with pester implementation. Tests:
- PostgreSQL
- PowershellCore
- Rust
- Sbt
- Selenium
- ServiceFabricSDK

#### Related issue:
https://github.com/actions/virtual-environments-internal/issues/810
## Check list


images/win/scripts/Tests/Rust.Tests.ps1
```diff
@@ -0,0 +1,35 @@
+Describe "Rust" {
+    Push-Location -Path $env:UserProfile
+    if ( -not ((Test-Path -Path ".rustup") -and (Test-Path -Path ".cargo")))
+    {
+        $null = New-Item -Name ".rustup" -Value "C:\Rust\.rustup" -ItemType Junction
+        $null = New-Item -Name ".cargo" -Value "C:\Rust\.cargo" -ItemType Junction
+    }
+    Pop-Location
```
maxim-lobanov, line 8: I don't think that we should change the image in tests. Can we avoid it somehow?

al-cheb: Without temporary `Junction` point we can't execute tools from Rust. I fixed Rust test to remove invoke the commands and to stay only validate binary path.

Issue #1243, opened 2020-07-15
Add Az Powershell Module vs. 4.4.0


juteasl:
**Tool information**
- Tool name: Az Powershell
- Tool license:  Default
- Add or update? Add
- Desired version: 4.4.0
- Approximate size: 
- If this is an add request:
  - Brief description of tool: Latest version of Az Powershell to use with Event Grid parameters that aren't in 4.3.0
  - URL for tool's homepage:
https://docs.microsoft.com/en-us/powershell/module/az.eventgrid/new-azeventgridsubscription?view=azps-4.4.0
  - Provide a basic test case to validate the tool's functionality:
Execute New-AzEventGridSubscription using the AzureActiveDirectoryTenantId and AzureActiveDirectoryApplicationIdOrUri parameters

**Area for Triage**:
'Scripting and command line'

**Question, Bug, or Feature?**:
Feature

**Virtual environments affected**
- [ ] macOS 10.15
- [ ] Ubuntu 16.04 LTS
- [ ] Ubuntu 18.04 LTS
- [ ] Ubuntu 20.04 LTS
- [X] Windows Server 2016 R2
- [X] Windows Server 2019

al-cheb:
Hello, @juteasl 
We have deployed new Windows Server 2016/2019 images with Az 4.4.0 PowerShell module. Feel free to reopen the thread if you have any concerns.

PR #1245, opened 2020-07-16
[Windows] Add Az powershell module 4.4.0

al-cheb:
# Description
Add the latest 4.4.0 version of Az Powershell module to use with Event Grid parameters that aren't in Az 4.3.0

#### Related issue:
https://github.com/actions/virtual-environments/issues/1243

## Check list


images/win/toolsets/toolset-2016.json
```diff
@@ -109,7 +109,7 @@
                 "3.1.0",
                 "3.5.0",
                 "3.8.0",
-                "4.3.0"
+                "4.4.0"
```
maxim-lobanov, line 112: I think such update will be a breaking change:
Previously, folder `C:\Modules\az_4.3.0` existed and customers pointed to it.
After your changes, this folder will be removed and replaced with `C:\Modules\az_4.4.0`.

maxim-lobanov: I think we should add it as a separate version

Issue #1220, opened 2020-07-11
Add R CLI tools


fnocera:
**Tool information**
- Tool name: R
- Tool license: GNU
- Add or update? Add
- Desired version: latest 3.x and 4.x versions
- Approximate size: 150MB
- If this is an add request:
  - Brief description of tool: R is a language and environment for statistical computing and graphics. R provides a wide variety of statistical (linear and nonlinear modelling, classical statistical tests, time-series analysis, classification, clustering, …) and graphical techniques, and is highly extensible.
  - URL for tool's homepage: https://www.r-project.org/

**Area for Triage**:
Scripting and command line

**Question, Bug, or Feature?**:
Feature

**Virtual environments affected**
- [x] macOS 10.15
- [x] Ubuntu 16.04 LTS
- [x] Ubuntu 18.04 LTS
- [x] Ubuntu 20.04 LTS
- [x] Windows Server 2016 R2
- [x] Windows Server 2019

**Can this tool be installed during the build?**
Yes, though specifying specific versions can be tricky across platforms. Basic installations are:
```
### Windows
choco install R.Project
choco install rtools
 
### Ubuntu
sudo apt-get install -y r-base
 
### macOS
brew install r
```

**Tool installation time in runtime**
Windows: ~4 minutes
Ubuntu: ~1 minute
macOS: ~1 minute

**Are you willing to submit a PR?**
No

cc @margaretmeehan @jeremydelacruz

maxim-lobanov:
Hello, thank you for your request.
We have added this issue to our backlog and will handle it soon

gaborcsardi:
There is also an action that installs a requested R version, and it is cross platform: https://github.com/r-lib/actions/tree/master/setup-r#setup-r

jimhester:
Is there any chance of reconsidering this? The versions of R available via the platform package managers are generally quite old, so they are often not useful in CI settings.

In addition the version of R installed by homebrew is not generally the best for macOS, rather installing the CRAN binary from the cask would be preferred, e.g. `brew cask install r`. There are some details on why this is preferred at https://rstats.wtf/set-up-an-r-dev-environment.html#what-about-homebrew.

Having an existing R on the `PATH` causes our [setup-r](https://github.com/r-lib/actions/tree/master/setup-r) action to avoid installing R, which breaks builds relying on it.

[...]

PR #1263, opened 2020-07-20
Add R tool for linux

dmitry-shibanov:
# Description
- Add R language and tools
#### Related issue:
https://github.com/actions/virtual-environments/issues/1220
## Check list

### DIFF
--- a/images/linux/scripts/installers/r.sh
+++ b/images/linux/scripts/installers/r.sh
@@ -1,5 +1,9 @@
+#!/bin/bash
+
 # Source the helpers for use with the script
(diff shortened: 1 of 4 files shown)


images/linux/scripts/installers/r.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+################################################################################
+##  File:  r.sh
+##  Desc:  Installs R
+################################################################################
```
maxim-lobanov, line 5: Could you please remove this header because it looks like it doesn't contain any important information?

Issue #1203, opened 2020-07-08
Add Swig


dot-asm:
**Tool information**
- Tool name: swig
- Tool license: GPL
- Add or update? add
- Desired version: 4.0.x
- Approximate size: 6MB
- If this is an add request:
  - Brief description of tool: connects programs written in C and C++ with a variety of high-level programming languages
  - URL for tool's homepage: http://swig.org/

**Area for Triage**: Scripting and command line

**Question, Bug, or Feature?**: Feature

**Virtual environments affected**
- [ ] macOS 10.15
- [x] Ubuntu 16.04 LTS
- [x] Ubuntu 18.04 LTS
- [x] Ubuntu 20.04 LTS
- [ ] Windows Server 2016 R2
- [ ] Windows Server 2019

**Can this tool be installed during the build?** It's available pre-packaged for Ubuntu, so it's possible to simply apt-get it.

**Tool installation time in runtime** negligible, in fact one can apt-get it on per-job basis, having it pre-installed simply makes life a little easier.

**Are you willing to submit a PR?**

nikita-bykov:
Hello @dot-asm! Your desired version of Swing is 4.0.x. We investigated that when installing Swing using apt-get, we got the following results:

- On Ubuntu 16.04 - we got SWIG Version 3.0.8
- On Ubuntu 18.04 - we got SWIG Version 3.0.12
- On Ubuntu 20.04 - we got SWIG Version 4.0.1

Is it expected for you?

dot-asm:
Yes. As mentioned in request, pre-packaged version as available through apt-get is what I'm looking for. I suppose I could have filled "3.0.x-4.0.x" as "desired version" in request form... Sorry about confusion :-) Thanks!

nikita-bykov:
We have added Swig installation to Ubuntu.
Images with it will be rolled out around next week.

dot-asm:
Cool! Much appreciated:+1: Thanks!

maxim-lobanov:
@dot-asm , I have reopened.
We usually keep issue in opened state until new image is deployed

dot-asm:
Oh! You'll handle the closing yourself. Got it! Thanks again!

miketimofeev:
@dot-asm Ubuntu 16&18 images were deployed with swig 3 and Ubuntu 20 with swig 4.
I'm going to close the issue, but feel free to contact us if you have any concerns.
Thank you!

zllovesuki:
[...]

PR #1265, opened 2020-07-20
Add swig support

nikita-bykov:
# Description
This PR adds on all Ubuntu images new tool - Swig. 
`apt-get` was used for installation.

#### Related issue:
#1203
## Check list

### DIFF
--- a/images/linux/scripts/installers/swig.sh
+++ b/images/linux/scripts/installers/swig.sh
@@ -8,7 +8,6 @@
 source $HELPER_SCRIPTS/document.sh
 
 # Install Swig
-apt-get update -y
 sudo apt-get install -y swig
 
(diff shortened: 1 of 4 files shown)


images/linux/scripts/installers/swig.sh
```diff
@@ -0,0 +1,23 @@
...
+################################################################################
+##  File:  swig.sh
+##  Desc:  Installs Swig
+################################################################################
+
+# Source the helpers for use with the script
+source $HELPER_SCRIPTS/document.sh
+
+# Install Swig
+apt-get update -y
```
maxim-lobanov, line 11: We don't need do `update` command if we don't add new repo

nikita-bykov: Fixed it

PR #1285, opened 2020-07-23
Add announcements to software documentation

vsafonkin:
# Description
Improvement: Added markdown files with announcements for each platform. Content of these files is added to software reports on top part of report.

#### Related issue:
Internal issue: [#830](https://github.com/actions/virtual-environments-internal/issues/830)

## Check list


images/linux/scripts/installers/preparemetadata.sh
```diff
@@ -8,6 +8,10 @@
 source $HELPER_SCRIPTS/document.sh
 
 WriteItem "<!--- DO NOT EDIT - This markdown file is autogenerated. -->"
+
+Announcements=$(cat $HELPER_SCRIPTS/UbuntuAnnouncements.md)
```
maxim-lobanov, line 12: I think it is better to name file just `announcements.md` because it is located in `linux` folder and it is obvious that it is for ubuntu.
The same for windows

vsafonkin: Done

AlenaSviridenko:
Do we need 2 different announcements files? Aren't they the same?

vsafonkin:
> Do we need 2 different announcements files? Aren't they the same?

@AlenaSviridenko , yes, it will be easier to maintain, because such platform report should content announcements for itself platform. If it one file, we need to implement logic to parse announcements for defined platform.

maxim-lobanov:
> Do we need 2 different announcements files? Aren't they the same?

@AlenaSviridenko  I think it is expected case when the table can be different for different OSs. It doesn't makes sense to put macOS announcements to Ubuntu or Windows docs

PR #1310, opened 2020-07-28
Remove cached ghc versions from Ubuntu docs

vsafonkin:
# Description
There are three versions of GHC are installed on Ubuntu (the latest version and two previous versions). Only latest version shoud be added to software documentation.

#### Related issue:
Internal issue: [#890](https://github.com/actions/virtual-environments-internal/issues/890)

## Check list


images/linux/scripts/installers/haskell.sh
```diff
@@ -53,10 +55,11 @@ if ! command -v stack; then
     exit 1
 fi
 
+# Add the latest ghc to path
+prependEtcEnvironmentPath "/opt/ghc/$defaultGHCVersion/bin"
```
maxim-lobanov, line 59: May be it is better to make symlink to `/usr/bin/ghc`?

images/linux/scripts/installers/haskell.sh
```diff
@@ -53,10 +55,11 @@ if ! command -v stack; then
...
+
 # Document what was added to the image
 echo "Lastly, documenting what we added to the metadata file"
 DocumentInstalledItem "Haskell Cabal ($(/opt/cabal/$cabalVersion/bin/cabal --version))"
-for version in ${ghcInstalledVersions[@]}; do
-    DocumentInstalledItem "GHC ($(/opt/ghc/$version/bin/ghc --version))"
-done
+DocumentInstalledItem "GHC ($(/opt/ghc/$defaultGHCVersion/bin/ghc --version))"
```
maxim-lobanov, line 64: If we put ghc to PATH and remove other versions, why we need to print path here?
I just worry that this PATH will be changed on every version release

vsafonkin: @maxim-lobanov , the path to default ghc will not be shown on software report, it is just invoke ghc to take version. On software report it looks like:
```
- GHC (The Glorious Glasgow Haskell Compilation System, version 8.10.1)
```

maxim-lobanov: Oh, sorry, I missed it

images/linux/scripts/installers/haskell.sh
vsafonkin: I suppose this symlink may cause some issues for `setup-haskell` action task.

maxim-lobanov: What issue do you mean?
Adding something to the PATH is always dangerous. `/opt/ghc/$defaultGHCVersion/bin` contains a lot of binaries:
image
And all of them will be added to the PATH that could cause customers impact.

vsafonkin: Yes, but `setup-haskell` task add `/opt/ghc/<version>/bin` directory to path for installed ghc and it works correctly. I'm afraid using symlink in `/usr/bin` may cause some conflict with ghc version that will be installed by `setup-haskell`

maxim-lobanov: `setup-haskell` does it only for customers who need ghc in path so it is expected.
Symlink should not cause issues because `setup-haskell` will continue to add directory to PATH and this directory will be before `/usr/bin` in PATH

vsafonkin: In other words, I decided to act similarly to the `setup-haskell` task.

vsafonkin: Done, what about other haskell tool `cabal`? We need to create symlink for it?

maxim-lobanov: Yes, I think we should. Good point

PR #1315, opened 2020-07-28
Add libc++ and libc++abi for all Ubuntu images

dpaoliello:
# Description
Modifies the existing script to add libc++ and libc++abi to all Ubuntu images, rather than just 16.04.


dpaoliello:
Ping @al-cheb: can I please get a review? This is blocking our ability to test on Linux (we're adding a new feature that uses coroutines, which is not supported by the default STL on Ubuntu).

Issue #1269, opened 2020-07-20
windows-latest environment running much slower since 20200714.1


beechtom:
**Describe the bug**
This is similar to an issue I filed two weeks ago: https://github.com/actions/virtual-environments/issues/1204
Workflows running on windows-latest environment are running about 2.5x slower than usual since the release of 20200714.1

**Area for Triage**: 
Image Administration
Testing and code coverage

**Question, Bug, or Feature?**: 
Bug

**Virtual environments affected**
- [ ] macOS 10.15
- [ ] Ubuntu 16.04 LTS
- [ ] Ubuntu 18.04 LTS
- [ ] Ubuntu 20.04 LTS
- [ ] Windows Server 2016 R2
- [x] Windows Server 2019

**Expected behavior**
New releases of Windows environments should not result in significantly slower workflows.

**Actual behavior**
Since the release of 20200714.1, workflows running on windows-latest have been running much slower than usual. Compare the following jobs run for the same pull request with minor changes between commits:

**20200706.1 (~13m):** https://github.com/puppetlabs/bolt/runs/883584515?check_suite_focus=true
**20200714.1 (~35m):** https://github.com/puppetlabs/bolt/runs/891066891?check_suite_focus=true

Several steps saw a noticeable increase in time to complete:
- Checkout takes 3x longer, from 9s to 20s
- Running a custom setup script takes 2x longer, from 14s to 23s
- Running tests with rspec takes 3x longer, from 12m to 34m
- Post-checkout takes 4.5x longer, from 3s to 14s

The overall time for job completion increased ~2.5x from an average of ~13m to an average of ~35m.

maxim-lobanov:
Hello @beechtom , checkout time and scripts time are very close on both images. The real difference is tests step.
I see that you configure connection to remote machine in runtime. It could be potential reason for this delay. Can we check it somehow?
Interesting point that `Agentful` job takes the same time on both images (11minutes vs 12 minutes) and the difference is only for `Agentless` job. What is the difference between these jobs?

beechtom:
@maxim-lobanov The script only configures a new user on the runner and tests a connection to itself, not a remote machine. You can check it by just running the script on the runner.

While the `Agentful` job takes roughly the same time, if you look at the individual steps they experienced a similar increase in time to complete. Checkout takes over twice as long, post-checkout takes three times as long, etc. Even the tests themselves take about a minute longer. While that's not significant, it _is_ indicative that the environment itself is running slower than it used to be.

[...]

PR #1316, opened 2020-07-29
[Windows] .NET Framework  Performance Issue

al-cheb:
# Description
https://support.microsoft.com/en-us/help/2570538/installing-updates-for-the-microsoft-net-framework-4-can-take-longer-t - When you install an update for the Microsoft .NET Framework 4, the Native Image Generator (NGen.exe) uses a high percentage of the CPU cycles on the computer for a long time. This time varies with how many Native Images are installed on the computer.

#### Related issue:
https://github.com/actions/virtual-environments/issues/1269

## Check list

### DIFF
--- a/images/win/Windows2019-Azure.json
+++ b/images/win/Windows2019-Azure.json
@@ -773,6 +773,12 @@
         },
         {
             "type": "powershell",
+            "scripts":[
+                "{{ template_dir }}/scripts/Installers/Run-NGen.ps1"
+            ]
+        },
+        {
+            "type": "powershell",
             "inline": [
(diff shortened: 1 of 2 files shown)


miketimofeev:
Maybe do the same for win2016? Just in case

dragon788:
The linked article mentions this should be done before updates to the .NET Framework 4 are applied, does that mean that the NGen.ps1 step should go before the `Install-WindowsUpdates.ps1` to speed up that and future steps rather than having NGen.exe churning in the background for all the steps following the updates?

Issue #1142, opened 2020-06-29
Add yamllint (or any similar tool to lint YAML)


yvele:
**Tool information**
- Tool name: yamllint
- Tool license: GNU General Public License v3.0
- Add or update? Add
- Desired version: Latest
- Approximate size: 
- If this is an add request:
  - Brief description of tool: A linter for YAML files
  - URL for tool's homepage: https://github.com/adrienverge/yamllint

**Area for Triage**:

Testing and code coverage

**Question, Bug, or Feature?**:

Feature

**Virtual environments affected**
- [x] macOS 10.15
- [x] Ubuntu 16.04 LTS
- [x] Ubuntu 18.04 LTS
- [x] Ubuntu 20.04 LTS
- [x] Windows Server 2016 R2
- [x] Windows Server 2019

**Can this tool be installed during the build?**

```
sudo apt-get install yamllint
```

**Tool installation time in runtime**

run in 23s on `ubuntu-latest`

```txt
 Set up yamllint 21s
Run sudo apt-get install yamllint
  sudo apt-get install yamllint
  shell: /bin/bash -e {0}
(Reading database ... 100%
(Reading database ... 254558 files and directories currently installed.)
Preparing to unpack .../python3-pathspec_0.5.5-1_all.deb ...
Unpacking python3-pathspec (0.5.5-1) ...
Selecting previously unselected package yamllint.
Preparing to unpack .../yamllint_1.10.0-1_all.deb ...
Unpacking yamllint (1.10.0-1) ...
Setting up python3-pathspec (0.5.5-1) ...
Setting up yamllint (1.10.0-1) ...
Processing triggers for man-db (2.8.3-2ubuntu0.1) ...
```

But there are many other ways to install it, like `pip install`, see https://yamllint.readthedocs.io/en/stable/quickstart.html#installing-yamllint 

**Are you willing to submit a PR?**

Yes (but I don't master Windows Server)

maxim-lobanov:
Hello @yvele , thank you for feature request. We would like to keep this work item opened for awhile to see popularity of this request.

al-cheb:
Hello, @yvele 
All images now contain yamllint package:

macOS - https://github.com/actions/virtual-environments/blob/main/images/macos/macos-10.15-Readme.md#linters
Ubuntu 16.04 - https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu1604-README.md
[...]

PR #1320, opened 2020-07-29
[Windows] Add yamllint support for the default Python version only

al-cheb:
# Description
Add yamllint support for the default Python version only

#### Related issue:
https://github.com/actions/virtual-environments/issues/1142

## Check list


images/win/scripts/Installers/Install-YAMLLint.ps1
```diff
@@ -0,0 +1,3 @@
+pip install yamllint
```
maxim-lobanov, line 1: I recently found pipx: https://github.com/pipxproject/pipx
That is the tool that allows to install pip packages globally and don't depend on any Python.
I am okay with current way. JFYI

images/win/scripts/SoftwareReport/SoftwareReport.Common.psm1
```diff
@@ -270,3 +270,14 @@ function Get-PacmanVersion {
     $pacmanVersion = $matches[0]
     return "- Pacman $pacmanVersion"
 }
+
+function Get-YAMLLintVersion {
+    $yamlVersion = (yamllint --version).Split()[1]
+    $yamlPath = (Get-Command -Name yamllint).Path
```
maxim-lobanov, line 276: Why we want to print tool path to the documentation if tool is in PATH?

al-cheb: Fixed.

PR #1330, opened 2020-07-30
Fix error in Install-Toolset.ps1 and added cleanup

toutougabi:
# Description
Bug fixing & Improvement

At line 17 the assetFolderPath is equal to the name of the file that was just downloaded.
This causes line 18 to fail since we try to create a directory that has the exact same name as the file downloaded.
So i added a _tmp to the end of the path so that it can create the folder.
Also added at line 25 a cleanup so we remove these install folders once the work is finished

## Check list

### DIFF
--- a/images/linux/scripts/installers/Install-Toolset.ps1
+++ b/images/linux/scripts/installers/Install-Toolset.ps1
@@ -14,7 +14,7 @@ Function Install-Asset {
     wget $ReleaseAsset.download_url -nv --retry-connrefused --tries=10
 
     Write-Host "Extract $($ReleaseAsset.filename) content..."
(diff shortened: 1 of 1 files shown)


MaksimZhukov:
Hello @toutougabi !
We create an `assetFolderPath` folder in a different location (in the `INSTALLER_SCRIPT_FOLDER` directory) where a release asset with the same name does not exist. It does not cause any issues for us. We also clear the `INSTALLER_SCRIPT_FOLDER` in the [images templates](https://github.com/actions/virtual-environments/blob/main/images/linux/ubuntu1604.json#L313).
Could you please provide additional information about how it affects your builds?
Thanks!

MaksimZhukov:
Currently we see no reason to merge your changes. I am closing the PR. Please let us know if have any questions or objections

PR #1338, opened 2020-07-31
Add Business Central Generic

freddydk:
# Description
Add Business Central Generic Docker image for the current Windows Version

Microsoft Dynamics 365 Business Central uses Docker in Azure DevOps pipelines and Github Actions.
Today a lot of our App Developing partners are forced to use Self hosted agents because downloading the docker image takes too much time.
The image is built on top of the 4.8 dotnet framework image and contains all what is needed in order to run Dynamics NAV and Business Central docker containers.
The image is built and maintained by the Business Central R&D team out of MDCC (Microsoft Development Center Copenhagen)

## Check list

### DIFF
--- a/images/win/scripts/ImageHelpers/ImageHelpers.psm1
+++ b/images/win/scripts/ImageHelpers/ImageHelpers.psm1
@@ -26,6 +26,7 @@ Export-ModuleMember -Function @(
     'Install-VsixExtension'
     'Get-VSExtensionVersion'
     'Get-WinVersion'
+    'Get-WinVersionNo'
     'Test-IsWin19'
     'Test-IsWin16'
     'Choco-Install'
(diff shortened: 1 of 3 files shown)


maxim-lobanov:
Docker image size: 10.1 GB
Installation time: 3 minutes
Screen Shot 2020-07-31 at 7 45 24 PM

freddydk:
Of the 10Gb, 7,8Gb are reused from the dotnet framework runtime image.
In our pipelines, we typically see around 5 minutes, which ends up being 25-33% of the full pipeline runtime.
image

alepauly:
Hi @freddydk, thanks for bringing this up but I don't think we can cache that image at this time due to space (even it's effectively only using 2.2 GB).  We are continuously looking for ways to provide for more custom needs like this so I'll make sure we keep it in mind as we figure out what we're making available next in our product.

Issue #1332, opened 2020-07-31
The latest vcpkg is broken, please don't take it


snnn:
**Describe the bug**
see: https://github.com/microsoft/vcpkg/issues/12565

I know your team usually generates a new set of images every week. But, the build script always take vcpkg from master, it is kind of dangerous.  Please consider to use tags.  The issue 12565 is an example, if you rebuild the image now, you'll have the bug, and would break all the vcpkg Windows users. 

**Area for Triage**: 
C/C++

**Question, Bug, or Feature?**: 
Bug

**Virtual environments affected**
- [ ] Windows Server 2016 R2
- [ ] Windows Server 2019

**Expected behavior**
Rebuild the image
Try to use vcpkg to install protobuf/libffi or some other open source libraries
The should work fine

**Actual behavior**
It fails, even through it's not your fault. Here just kindly remind you there is a problem.  Please feel free to close this issue without taking extra actions.

maxim-lobanov:
Hello @snnn , Thank you for reaching us.
Am I right that your suggestion to use latest release instead of master branch during installation?
https://github.com/microsoft/vcpkg/releases/latest

snnn:
Yes!

larshg:
It seems that CMake 18 is the culprint and it already hit the pipelines like here:

https://dev.azure.com/PointCloudLibrary/pcl/_build/results?buildId=17096&view=logs&j=d944f354-b9eb-5c76-5ade-3c70deb27a9e&t=fb7e0546-b0fe-5e7c-2a44-5710f4ba7243

So please make a hotfix when VCPKG has solved its bug.

kcgen:
This has tanked our Windows CI since Monday.

maxim-lobanov:
Just to merge discussions in #1340 and this issue:
Looks like rolling back vcpkg to the latest release will cause additional impact because of cmake incompatibility.
Installing master branch looks like the better solution for now.

cc: @strega-nil

kcgen:
@maxim-lobanov , drop and update when the new Windows GitHub images are regenerated w/ the upated vcpkg.

Going on day 8 now with Windows CI broken, so can't wait for the fix :raised_hands: 

![2020-08-03_08-32](https://user-images.githubusercontent.com/1557255/89206500-3a552980-d56e-11ea-95d4-ed4232ab2953.png)

maxim-lobanov:
Hello,
[...]

PR #1340, opened 2020-07-31
[Windows] Fix vcpkg installation

dibir-magomedsaygitov:
# Description
We need to install vcpkg from the latest release instead of master branch commit.

#### Related issue:
https://github.com/actions/virtual-environments/issues/1332

## Check list

### DIFF
--- a/images/win/scripts/Installers/Install-Vcpkg.ps1
+++ b/images/win/scripts/Installers/Install-Vcpkg.ps1
@@ -3,21 +3,23 @@
 ##  Desc:  Install vcpkg
 ################################################################################
 
-Import-Module -Name ImageHelpers -Force
+Write-Host "Download the latest vcpkg version..."
+$vcpkgDownloadUrl = (Invoke-RestMethod "https://api.github.com/repos/microsoft/vcpkg/releases/latest").tarball_url
+$installDir = 'C:\vcpkg'
+New-Item $installDir -type directory
+$tempPath = Join-Path -Path "${env:Temp}" -ChildPath "vcpkg.tar.gz"
+Invoke-WebRequest $vcpkgDownloadUrl -OutFile $tempPath
 
-$Uri = 'https://github.com/Microsoft/vcpkg.git'
-$InstallDir = 'C:\vcpkg'
-$VcpkgExecPath = 'vcpkg.exe'
-
-git clone --depth=1 $Uri $InstallDir -q
+Write-Host "Expand vcpkg archive"
+tar -xzvf $tempPath -C $installDir --strip-components=1
 
 # Build and integrate vcpkg
-Invoke-Expression "$InstallDir\bootstrap-vcpkg.bat"
-Invoke-Expression "$InstallDir\$VcpkgExecPath integrate install"
+Invoke-Expression "$installDir\bootstrap-vcpkg.bat"
+Invoke-Expression "$installDir\vcpkg.exe integrate install"
 
 # Add vcpkg to system environment
-Add-MachinePathItem $InstallDir
+Add-MachinePathItem $installDir
 $env:Path = Get-MachinePath
-setx VCPKG_INSTALLATION_ROOT $InstallDir /M
+setx VCPKG_INSTALLATION_ROOT $installDir /M
 
 Invoke-PesterTests -TestFile "Tools" -TestName "Vcpkg"


images/win/scripts/Installers/Install-Vcpkg.ps1
```diff
@@ -3,21 +3,23 @@
 ##  Desc:  Install vcpkg
 ################################################################################
 
-Import-Module -Name ImageHelpers -Force
+Write-Host "Download the latest vcpkg version..."
+$vcpkgDownloadUrl = (Invoke-RestMethod "https://api.github.com/repos/microsoft/vcpkg/releases/latest").tarball_url
+$installDir = 'C:\vcpkg'
+New-Item $installDir -type directory
+$tempPath = Join-Path -Path "${env:Temp}" -ChildPath "vcpkg.tar.gz"
+Invoke-WebRequest $vcpkgDownloadUrl -OutFile $tempPath
```
vmapetr, line 11: Maybe use `Start-DownloadWithRetry`?

vmapetr: And for releases manifest too.

images/win/scripts/Installers/Install-Vcpkg.ps1
```diff
@@ -3,21 +3,23 @@
 ##  Desc:  Install vcpkg
 ################################################################################
 
-Import-Module -Name ImageHelpers -Force
+Write-Host "Download the latest vcpkg version..."
+$vcpkgDownloadUrl = (Invoke-RestMethod "https://api.github.com/repos/microsoft/vcpkg/releases/latest").tarball_url
```
miketimofeev, line 7: maybe zipball? And then use `Extract-7Zip` function instead of tar since I'm not so confident in Windows tar

dibir-magomedsaygitov: We can't unzip archive without first directory with `Extract-7Zip`.  That's why i'm using `tar`.

images/win/scripts/Installers/Install-Vcpkg.ps1
vsafonkin: @vmapetr , it looks strange, but `Start-DownloadWithRetry` helper return unexpected response for this link:
```
The remote server returned an error: (403) Forbidden
```

dibir-magomedsaygitov: Thank you for your point, but when `Start-DownloadWithRetry` is being used we are getting `Exception calling "DownloadFile" with "2" argument(s): "The remote server returned an error: (403) Forbidden."`.
error

kcgen:
This week's `master` branch of vspkg was broken due to their lack of strong CI gating on their part to handle a recent change in cmake.  This break [rippled across vspkg users](https://github.com/dosbox-staging/dosbox-staging/issues/524#issuecomment-667112140), impacting many projects include CI workflows.

However, "fixing" this by falling back to vspkg's 2-month-release cycle will cause all of vspkg's hotfixes to resurface for GitHub CI users.  These will be hidden from you though - and instead a new flood of issues will hit vspkg.

So to address one issue today, we will be [re-opening approximately ~300 closed issues](https://github.com/microsoft/vcpkg/issues?page=1&q=is%3Aissue+is%3Aclosed) that have been fixed by vspkg since their last stable release.

I believe a more robust solution is to include vspkg as a pre-req in GitHub's VM environment CI gating. 

Run the candidate branch of vspkg through a handful of low-level packages installations - like png, zlib, and gpg.   If any fail, then rollback to the prior passing vspkg.  In the worst case here, you will only undo the minimum amount of vspkg progress as possible to proceed with a stable build - and not undo months of fixes.

maxim-lobanov:
@kcgen , thank you for your feedback. That all makes sense.
Unfortunately, we don't have capacity to test the latest master of vcpkg, before accepting it. So we have to trust VCPKG team here.
Actually, We have two solutions:
- Always install latest master
- Always install latest release

Since we don't have much experience with VCPKG, probably you have better context what is preferred.

@strega-nil, Hello!
Could you please take a look at the discussion in this PR and in #1332 and advice the best solution?

kcgen:
> Unfortunately, we don't have capacity to test the latest master of vcpkg, before accepting it. So we have to trust VCPKG team here.

Thanks for explaining the dilemma @maxim-lobanov - yes, the problem of _"we don't have capacity to test"_ completely dwarfs this vspkg issue.  Are you telling me the VM images that (millions?) of users rely on every week are loaded with software that hasn't undergone even the most basic of sanity checks?

Surely Microsoft and GitHub have Windows VMs where all of the candidate packages can be installed and some dead-simple powershell commands could execute the software (starting with the most simplest of tests `vspkg install zlib` is all we're talking about) prior to accepting that software? 

You could additionally have a virus scanner and behavior/network scanner inspect the process to, and include that in your CI gate. 

This is so fundamental that I hope it's someone's full time job there. It's only a matter of time before just one of these "trusted-but-not-tested" packages are trojaned with a GitHub SECRETs harvester.

Back to this vspkg issue - this issue alone has blown up for them, causing apologies and broken workflows; they see the immediate need for strong CI gating to prevent this from ever happening again.  I suspect that will be in-place in the coming days or weeks. Whatever trust you currently had in their master branch, you will be able to trust it a whole lot more going forward.  

In the meantime, I understand they've already fixes this issue too.  So rolling back to their last release will reintroduce already fixed bugs since then.

Perhaps the safest course of action would be to temporarily roll back to the stable release - and inform your users ( which might be other GitHub teams, like Workflows, etc) who in turn can inform their end-users (like myself) that vspkg is being reverted and prior fixes might be broken again.

This will give vspkg time to implement their strong CI gating and get any kinks worked out. Once that's stable, I would then roll forward to their master branch again as breaking-changed will be blocked before they hit master.

maxim-lobanov:
@kcgen, 
As for the validation, we have validation gates that contain a bunch of smoke (very very simple) tests to validate that the tool is installed on image successfully and antivirus checks. (Unfortunately, vcpkg is not covered properly right now and we are planning to do that soon).

When I said "we don't have capacity to validate latest master of VCPKG", I meant a bit different: we need to generate image, do validation of image, if vcpkg doesn't work properly, we have to do PR to rollback it and hardcode some old version, regenerate image and etc. It will take much time and could block and delay image rollout and other important updates. Considering the fact that we have ton of tools installed and every could be affected by the same (release contains bugs). We could risk never deploy image in this case because we hang in rolling back versions :)
I understand your point and we are trying to improve our test-coverage constantly but unfortunately, it is not possible to follow up all tools and all release notes. And as you mentioned above, it is always dilemma between good test-coverage from our side and trusting tool owners. (Absolutely, agree that this particular issue is pretty obvious and could be caught from our side).
It is what I mentioned "We have to trust tool owners" that their releases don't contain bugs (at least critical).

As for the Vcpkg, we do even worse thing - always pick up master branch. It means every day a lot of commits could break everything (especially, if tool doesn't have own strong validation as you said) 
Personally, I prefer pickup the latest release instead of latest master as you mentioned. In this case, vcpkg version could be a bit outdated but the risk of critical bug is lower.

kcgen:
@maxim-lobanov  - thanks for the clarification! Great to hear that smoke tests are in place for major components and is improving for lesser components like `vspkg`.

> Personally, I prefer pickup the latest release instead of latest master as you mentioned.

With testing of vspkg weak on both sides, I think falling back to the stable release is the best you can do. Once automatic CI gating is in place, then you can safely move up to master and users will benefit from the post-relaese fixes.

kcgen:
@maxim-lobanov: some late breaking news on this - https://github.com/microsoft/vcpkg/pull/12569#issuecomment-667318339

The issue is that vxpkg (all prior versions including latest master) will blindly use the host-installed cmake (if detected) instead of its own internally proven-and-tested version. 

If the github VM is rolled back to an older vxpkg but you continue to install the latest cmake version (1.18 I believe), then that older vxpkg will attempt to use cmake 1.18 and again fail. 

This can be solved by:
 - not installing cmake (vxpkg will then use its internal version), or
- install cmake, but only up to the version that vxpkg has been tested with (1.17 for prior versions, or 1.18 as of today's latest master)
- only permit new cmake versions once vxpkg has been tested with them

kcgen:
@maxim-lobanov: some new information that changes the above assumptions:

If vcpkg finds cmake installed on the host, then it will blindly use that instead ofits own internally-bundled proven-and-tested version. 

This exposes a version race-condition: if cmake on the host has new API breaking-changes versus the latest version that the vcpkg devs have tested with (and embedded), then vcpkg will fail when it tries to interface with the host cmake.

So our proposed solution above: rolling back to the last stable release-build of vxpkg won't solve the problem, because its "knowledge" of cmake only goes up to 1.17, however the version of cmake separately installed on the GitHub VMs is 1.18 -- so this version race-condition will continue to exist.

This can be solved by:
 - not installing cmake on the host (vxpkg will then use its internal version), or
- install cmake, but only up to the version that vxpkg has been tested with (1.17 for prior versions, or 1.18 as of today's latest vcpkg master)
- only permit new cmake versions once vxpkg has been tested with them

I've proposed to the vcpkg team that vcpkg be changed to check the host cmake version, and only use that if its version is equal-to or less-than what vcpkg is using internally.  This will eliminate that race condition.

https://github.com/microsoft/vcpkg/pull/12569#issuecomment-667318339

strega-nil:
@kcgen I will say that breaking changes in cmake are _exceedingly_ rare, and are mostly bugfixes; our code was buggy, and was only allowed by cmake also being buggy.

kcgen:
> breaking changes in cmake are exceedingly rare

So to prevent this from happening again, the answer is ... Do nothing?

- This issue knocked out vcpkg for almost a week.

- GitHub's CI virtual machines are still knocked out because the infrastructure team re-generates the images once a week. My project still cannot merge PRs.

- Airgap deployments bundling today's "fixed" vcpkg that are cmake-1.18 compatible carry a constant risk that any /future/ cmake version will break vxpkg. 

wreckage aside,  why are we still trying to avoid the seat belt? What benefit does vxpkg gain by -not- limiting the version of cmake used to that which vxpkg has been tested against?  

Good design should avoid leaps of faith, and use hard proofs instead (when possible).  This is one case where a trivial version check avoids the leap of faith. 

> our code was buggy, and was only allowed by cmake also being buggy.

All the more reason to only trust up-to the version tested against.
Bugs will continue to be added by both projects; that's the one guarantee we all know :-)

strega-nil:
Perhaps. We'll have to look at it, especially because we _cannot_ download tools on some platforms.

kcgen:
Thanks @strega-nil.

maxim-lobanov:
Thank you for your discussion, that is helpful.

Honestly, I don't think that synchronize cmake and vcpkg version on image suites for us:
- It will add maintenance work to check compatibility and update cmake manually if it is allowed on weekly basis
- Customers who don't care about vcpkg but want to use the latest cmake won't be happy to wait additional weeks until new cmake is tested and approved by vcpkg team (usually, we get requests to update cmake right after releasing new version - that is the reason why we just pickup the latest version automatically)

If vcpkg has strong dependency on cmake version, it makes sense to include cmake to vcpkg bundle and update it regularly and use host version of cmake only if it is compatible as it was proposed above. The current approach doesn't suit for using in CI systems.

We are going to hold off / close current PRs to switch vcpkg to latest release because as we have discussed above it will cause more impact and will cause issue with cmake.
Let's move further discussion to #1332

kcgen:
Any progress when the new Windows GitHub images are going to be regenerated w/ the fixed vcpkg?

(going on day 8 now with Windows CI broken)

Issue #804, opened 2020-04-29
Update Windows config for error reporting


vtbassmatt:
[Original request](https://developercommunity.visualstudio.com/idea/1012461/hosted-build-agents-have-error-reporting-set-to-re.html)

When something crashes on a hosted Windows agent, Windows hangs waiting for consent to send telemetry. Consider either disable error reporting, or offer a way to auto-consent.

**Area for Triage**:

**Question, Bug, or Feature?**:
Feature

**Virtual environments affected**
- [ ] macOS 10.15
- [ ] Ubuntu 16.04 LTS
- [ ] Ubuntu 18.04 LTS
- [X] Windows Server 2016 R2
- [X] Windows Server 2019

vtbassmatt:
@jessehouwing

jessehouwing:
I consider it a bug ;).

alepauly:
closing in lieu of #806 thanks @vtbassmatt and @jessehouwing !

PR #1352, opened 2020-08-03
Windows templates: Merge installation scripts.

Darleev:
# Description
Improvment

#### Related issue:
Add single task for all installation scripts in Packer template. #804
## Check list


images/win/Windows2019-Azure.json
```diff
@@ -143,49 +143,27 @@
...
-        {
-            "type": "powershell",
-            "environment_vars": [
-                "IMAGE_VERSION={{user `image_version`}}",
-                "IMAGEDATA_FILE={{user `imagedata_file`}}"
-            ],
-            "scripts":[
-                "{{ template_dir }}/scripts/Installers/Update-ImageData.ps1"
+              "setx ImageOS {{user `image_os` }} /m",
+              "setx ImageFile {{user `imagedata_file`}} /m"
```
maxim-lobanov, line 155: Previously, we didn't set this environments variable (we set it only for step, not the whole VM).
Do we really need to set it for whole machine?

Darleev: Thank you, I have changed this back.

Issue #915, opened 2020-05-21
Random IO failures during nightly actions


sixlettervariables:
**Describe the bug**
We receive random IO failures on a nightly matrix job that runs a build across two JDK versions on both Windows and Linux. The failures are always during an IO operation and vary which JDK and OS depending on the night.

**Area for Triage**: 

**Question, Bug, or Feature?**: 
Bug

**Virtual environments affected**
- [ ] macOS 10.15
- [ ] Ubuntu 16.04 LTS
- [X] Ubuntu 18.04 LTS
- [ ] Windows Server 2016 R2
- [X] Windows Server 2019

**Expected behavior**
Consistent IO performance when executing workflow.

**Actual behavior**
From [MekHQ's repo](
https://github.com/MegaMek/mekhq/actions?query=workflow%3A%22MekHQ+Nightly+CI%22), 3 of the last 5 nightlies have failed:
- https://github.com/MegaMek/mekhq/actions/runs/109858317
    - ubuntu-latest JDK8 is the only failure with File Not Found: `> Directory '/home/runner/work/mekhq/mekhq/megamek/megamek/build/resources/main' specified for property '$2' does not exist.`
- https://github.com/MegaMek/mekhq/actions/runs/106916147
    - ubuntu-latest JDK8 failed with File Not Found: `Could not set UNIX mode on /home/runner/work/mekhq/mekhq/megamek/megamek/build/resources/main/megamek/client/bot/messages_ru.properties: could not chmod file (errno 2: No such file or directory)`
    - windows-latest JDK8 failed for two reasons:
        - `> Couldn't read file content: 'D:\a\mekhq\mekhq\megamek\megamek\build\resources\main\megamek\MegaMek.properties'.`
        - `> java.io.IOException: Unable to delete directory D:\a\mekhq\mekhq\megamek\megamek\build\resources\main\megamek\client.`
- https://github.com/MegaMek/mekhq/runs/669049985
    - windows-latest JDK11 failed: `> java.io.IOException: Unable to delete directory D:\a\mekhq\mekhq\megamek\megamek\build\resources\main\megamek\common\templates\tro`

From [MegaMekLab's repo](https://github.com/MegaMek/megameklab/actions?query=workflow%3A%22MegaMekLab+Nightly+CI%22), we have fewer failures:
- https://github.com/MegaMek/megameklab/actions/runs/108718941
[...]

PR #1355, opened 2020-08-04
Add generation of announcements tables

vsafonkin:
Added logic to get announcement's content from environment variables and generate tables.

#### Related issue:
Internal issue: [#915](https://github.com/actions/virtual-environments-internal/issues/915)

## Check list


images/linux/scripts/installers/preparemetadata.sh
```diff
@@ -9,8 +9,18 @@ source $HELPER_SCRIPTS/document.sh
...
+    Announcements="| Announcements |\n|-|\n"
+    IFS=';;' read -ra rawAnnouncements <<< "$UBUNTU_ANNOUNCEMENTS"
+    for rawAnnouncement in "${rawAnnouncements[@]}"; do
+        if [ $rawAnnouncement ]; then
+            IFS="===" read -ra announcement <<< "$rawAnnouncement"
+            Announcements+="| [${announcement[0]}](${announcement[3]}) |\n"
+        fi
+    done
+    WriteItem $Announcements
+fi
```
maxim-lobanov, line 22: Why can't we pass ready table to the packer template? 
Why we need to generate it? I think it will cause a lot of duplicated logic between macos, linux, windows

vsafonkin: Done, thank you!

PR #1394, opened 2020-08-11
Resolve possible conflicts in software update PR

vsafonkin:
# Description
Sometimes, merge-pull-request workflow fails because of conflicts in PR. It happens when current deployment was started earlier then previous one was finished.
 
Changes:
- Added step to create special branch for pull request
- Added step to resolve PR's conflicts in this branch before merging to main
- Added step to delete this additional branch

#### Related issue:
Internal issue: [#908](https://github.com/actions/virtual-environments-internal/issues/908)

## Check list

### DIFF
--- a/.github/workflows/merge_pull_request.yml
+++ b/.github/workflows/merge_pull_request.yml
@@ -18,9 +18,9 @@ jobs:
       run: |
         git config --global user.email "no-reply@github.com"
         git config --global user.name "Actions service account"
-        git checkout ${{ github.event.client_payload.ReleaseBranchName }}-pull-request
+        git checkout ${{ github.event.client_payload.ReleaseBranchName }}-docs
         git merge --no-edit --strategy-option=ours main
-        git push origin ${{ github.event.client_payload.ReleaseBranchName }}-pull-request
+        git push origin ${{ github.event.client_payload.ReleaseBranchName }}-docs
         sleep 30
 
     - name: Merge pull request for ${{ github.event.client_payload.ReleaseBranchName }}
--- a/.github/workflows/merge_pull_request.yml
+++ b/.github/workflows/merge_pull_request.yml
@@ -33,3 +33,14 @@ jobs:
             repo: context.repo.repo,
             pull_number: ${{ github.event.client_payload.PullRequestNumber }}
           })
+
+    - name: Delete docs branch ${{ github.event.client_payload.ReleaseBranchName }}-docs
+      uses: actions/github-script@v2
+      with:
+        github-token: ${{secrets.GITHUB_TOKEN}}
+        script: |
+          github.git.deleteRef({
+            owner: context.repo.owner,
+            repo: context.repo.repo,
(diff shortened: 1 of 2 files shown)


.github/workflows/merge_pull_request.yml
```diff
@@ -12,6 +12,12 @@ jobs:
     steps:
     - uses: actions/checkout@v2
 
+    - name: Resolve possible conflicts ${{ github.event.client_payload.ReleaseBranchName }} with main
+      run: |
+        git checkout ${{ github.event.client_payload.ReleaseBranchName }}
+        git merge --no-edit --strategy-option=ours main
```
maxim-lobanov, line 18: Don't we need to push these changes?

vsafonkin: I believe you're right

maxim-lobanov: Please test this change on fork properly. It is committing to main branch and we need to be extra careful here.
I also believe push will fail if no new commits in master

vsafonkin: Done, thank you!

.github/workflows/merge_pull_request.yml
```diff
@@ -11,6 +11,17 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+
+    - name: Resolve possible conflicts ${{ github.event.client_payload.ReleaseBranchName }} with main
```
alepauly, line 17: Why would there be conflicts? I rather not blindly do this, seems dangerous.

maxim-lobanov: Repro steps:
1) Create branch `test` from `main`, update readme in that branch, create PR to `main`
2) Update readme in main branch
3) PR can't be merged due to conflicts.

It is often case, when we still deploy `Week 1` on Ring 3 and starting to deploy `Week 2` to Ring 0.
Docs from Week 1 updates `main` branch on step 2.

Resolving conflicts this way looks safety because:
1) `PR branch` has only the single change in comparison with `main`: change in documentation file.
2) We trigger this workflow when we have finished deployment of particular week. It means that we 100% would like to take new documentation from our branch and replace any other docs in `main`. This process is pretty similar to what we had previously, when we just updated documentation file directly in master blindly

maxim-lobanov: Oh, discussed with @vsafonkin offline and realized issue with this approach. This approach pull `main` branch to release branch and this way we modify release branch and change state from what image was generated. We shouldn't do that

vsafonkin: So, I've reworked this approach. The special branch for pull request is creating as clone of release branch. It allows to resolve conflicts on PRs safely, without depends for release branch.
@maxim-lobanov @alepauly

maxim-lobanov (approved):

PR #1419, opened 2020-08-15
Add current commit into logs for custom builds

vmapetr:
# Description
This PR add latest commit hash and name into build logs for Nightly CI builds and custom builds queued via CreateAzureVMFromPackerTemplate.ps1 script.
It also remove unused variables from Windows and Linux templates.

#### Related issue:
https://github.com/actions/virtual-environments-internal/issues/882

## Check list


images.CI/download-customrepo.ps1
```diff
@@ -0,0 +1,14 @@
...
+)
+
+Write-Host "Clean up default repository"
+Remove-Item -path './*' -Recurse -Force
+
+Write-Host "Download ${CustomRepoBranch} branch from ${CustomRepoUrl}"
+$env:GIT_REDIRECT_STDERR = '2>&1'
+git clone $CustomRepoUrl . -b $CustomRepoBranch --single-branch --depth 1
+
+Write-Host "Latest commit:"
```
AlenaSviridenko, line 13: Can we reuse Get-LatestCommit here?

vmapetr: We can, but Get-LatestCommit function was created only in helper scripts that are completely separated from our Images.CI definitions, and it will doesn't make sense to have common helpers for both images.CI and helpers.

images.CI/download-customrepo.ps1
```diff
@@ -0,0 +1,14 @@
+param(
+    [String] [Parameter (Mandatory=$true)] $CustomRepoUrl,
```
maxim-lobanov, line 2: Let's get rid of `Custom` in parameters naming. Also I suggest to rename script to `download-repo.ps1`

images.CI/download-customrepo.ps1
```diff
@@ -0,0 +1,14 @@
...
+    [String] [Parameter (Mandatory=$true)] $CustomRepoBranch
+)
+
+Write-Host "Clean up default repository"
+Remove-Item -path './*' -Recurse -Force
```
maxim-lobanov, line 7: I wonder if we can set condition on `checkout` task instead of removing repo

images.CI/download-customrepo.ps1
vmapetr: Done. Thanks!

helpers/CreateAzureVMFromPackerTemplate.ps1
```diff
@@ -1,10 +1,10 @@
 Function CreateAzureVMFromPackerTemplate {
     <#
         .SYNOPSIS
-            Creates an Azure VM from a template. Also generates network resources in Azure to make the VM accessible.
+            A helper function to deploy a VM from generated image.
```
miketimofeev, line 4: from **a** generated image.

images.CI/download-customrepo.ps1
vmapetr: AFAIK, `checkout` task do not support stage conditions. 
However, we can avoid removing repo, but it will lead to additional workarounds, so let's just leave it that way, at least for now.

helpers/CreateAzureVMFromPackerTemplate.ps1
vmapetr: Thanks, fixed!

maxim-lobanov (approved):

AlenaSviridenko (approved):

al-cheb (approved):

PR #1420, opened 2020-08-17
Move android components list to toolsets on Ubuntu image

dsame:
# Description
Improvement

Replace 3 installation script with one controlled by .json toolset files

## Check list


images/linux/scripts/installers/android.sh
```diff
@@ -45,56 +45,54 @@ else
     exit 1
 fi
 
+# Add required permissions
```
maxim-lobanov, line 48: Looks like permissions are already set in lien 31

dsame: fixed

PR #1428, opened 2020-08-18
Add Pulumi to Ubuntu images

LeonidLapshin:
# Description
This installs Pulumi on Ubuntu images, the latest available version will be installed.
The total size of the package is 57MB and it takes approximately 10 seconds to install.

## Check list

### DIFF
--- a/images/linux/scripts/installers/pulumi.sh
+++ b/images/linux/scripts/installers/pulumi.sh
@@ -10,16 +10,8 @@ source $HELPER_SCRIPTS/document.sh
 # Install Pulumi
 VERSION=$(curl --fail --silent -L "https://www.pulumi.com/latest-version")
 TARBALL_URL="https://get.pulumi.com/releases/sdk/pulumi-v${VERSION}-linux-x64.tar.gz"
-TARBALL_DEST=$(mktemp -t pulumi.tar.gz.XXXXXXXXXX)
-curl -L -o "${TARBALL_DEST}" "${TARBALL_URL}"
-EXTRACT_DIR=$(mktemp -d pulumi.XXXXXXXXXX)
-tar zxf "${TARBALL_DEST}" -C "${EXTRACT_DIR}"
-if [ -d "${EXTRACT_DIR}/pulumi/bin" ]; then
-    mv "${EXTRACT_DIR}/pulumi/bin" "/usr/local/bin/"
-else
-    cp -r "${EXTRACT_DIR}/pulumi/." "/usr/local/bin/"
-fi
-rm -rf ${TARBALL_DEST} ${EXTRACT_DIR}
+download_with_retries ${TARBALL_URL} "/tmp" pulumi-v${VERSION}.tar.gz
+tar --strip=1 -xvf /tmp/pulumi-v${VERSION}.tar.gz -C /usr/local/bin
 
 # Run tests to determine that the software installed as expected
 echo "Testing to make sure that script performed as expected, and basic scenarios work"
@@ -29,6 +21,5 @@ if ! command -v pulumi ; then
 fi
 
 # Document what was added to the image
-pulumi_version="$(pulumi version)"
 echo "Lastly, documenting what we added to the metadata file"
-DocumentInstalledItem "Pulumi ($pulumi_version)"
+DocumentInstalledItem "Pulumi $(pulumi version)"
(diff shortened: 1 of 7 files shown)


images/linux/scripts/installers/pulumi.sh
```diff
@@ -0,0 +1,24 @@
...
+##  File:  pulumi.sh
+##  Desc:  Installs Pulumi
+################################################################################
+
+# Source the helpers for use with the script
+source $HELPER_SCRIPTS/document.sh
+
+# Install Pulumi
+curl -fsSL https://get.pulumi.com | sh
+cp -r ~/.pulumi/bin/* /usr/local/bin/
```
maxim-lobanov, line 12: Currently, `$HOME/.pulumi` has the following structure:
```
$HOME/.pulumi
---- bin
-------- pulumi
-------- pulumi-analyzer-policy
-------- pulumi-analyzer-policy-python
-------- pulumi-language-dotnet
-------- pulumi-language-go
-------- pulumi-language-nodejs
-------- pulumi-language-python
-------- pulumi-language-python-exec
-------- pulumi-resource-pulumi-nodejs
-------- pulumi-resource-pulumi-python
```
And this code work as expected for now.
But I wonder why `$HOME/.pulumi` is created by pulumi installer:
- can it be used somehow during pulumi run?
- can something be changed in future and this folder will contain additional folders to `bin`

cc: @al-cheb @miketimofeev

LeonidLapshin: Hi, Maxim!
Pulumi documentation for manual installation on Linux
https://www.pulumi.com/docs/get-started/install/
tells: “Extract the tarball and move the binaries in the pulumi directory to a directory included in your system’s $PATH.”. 
So I guess that copy binaries to /usr/local/bin (that path is included in $PATH for all users) is correct in this case. 
Anyway, I’ll be happy to change anything if I’m wrong, thank you!

miketimofeev: In that case, maybe it makes sense to use manual installation?
https://www.pulumi.com/docs/get-started/install/#manual-installation
Just download the latest tar and unpack it to `/usr/local/bin`?

maxim-lobanov: @LeonidLapshin, thank you for explanation. In this case, your code looks good to me.
But I think I agree with @miketimofeev suggestion

LeonidLapshin: @miketimofeev, thank you for your suggestion, I made some changes within script, the installation logic is no more hidden behind a downloadable file. 
@maxim-lobanov, I hope, it looks clearer than before.

images/linux/scripts/installers/pulumi.sh
```diff
@@ -0,0 +1,34 @@
...
+################################################################################
+
+# Source the helpers for use with the script
+source $HELPER_SCRIPTS/document.sh
+
+# Install Pulumi
+VERSION=$(curl --fail --silent -L "https://www.pulumi.com/latest-version")
+TARBALL_URL="https://get.pulumi.com/releases/sdk/pulumi-v${VERSION}-linux-x64.tar.gz"
+TARBALL_DEST=$(mktemp -t pulumi.tar.gz.XXXXXXXXXX)
+curl -L -o "${TARBALL_DEST}" "${TARBALL_URL}"
```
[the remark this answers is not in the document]
miketimofeev: It seems you forgot to add `source $HELPER_SCRIPTS/install.sh` at the beginning of the script

maxim-lobanov:
Builds are failing due to known issue with image-generation. Fix is on way: https://github.com/actions/virtual-environments/pull/1482

images/linux/scripts/installers/pulumi.sh
```diff
@@ -0,0 +1,26 @@
...
+##  Desc:  Installs Pulumi
+################################################################################
+
+# Source the helpers for use with the script
+source $HELPER_SCRIPTS/document.sh
+source $HELPER_SCRIPTS/install.sh
+
+# Install Pulumi
+VERSION=$(curl --fail --silent -L "https://www.pulumi.com/latest-version")
+TARBALL_URL="https://get.pulumi.com/releases/sdk/pulumi-v${VERSION}-linux-x64.tar.gz"
```
alepauly, line 13: @cnunciato are you with Pulumi? does this look like a reliable way to get the latest?

maxim-lobanov: This way is provided by official install process so it should be reliable enough

cnunciato: @alepauly I am, and yes, that's the way.

PR #1550, opened 2020-09-06
DKPG/APT retry logic for Ubuntu.

Darleev:
# Description
Improvement
In order to avoid dpkg lock issue, it is required to implement retry logic.

#### Related issue:

## Check list

### DIFF
--- a/images/linux/scripts/helpers/install.sh
+++ b/images/linux/scripts/helpers/install.sh
@@ -35,6 +35,17 @@ download_with_retries() {
     return 1
 }
 
+function wait_for_apt_lock() {
+    while [ "" = "" ]; do
+        eval "$1" 2>/dev/null
+        if [ $? -eq 0 ]; then
+            break
+        fi
+        sleep 1
+        echo "Waiting for apt lock..."
+    done
+}
+
 ## Use dpkg to figure out if a package has already been installed
 ## Example use:
 ## if ! IsPackageInstalled packageName; then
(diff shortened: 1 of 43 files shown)


images/linux/scripts/helpers/install.sh
```diff
@@ -35,6 +35,17 @@ download_with_retries() {
     return 1
 }
 
+function wait_for_apt_lock() {
+    while [ "" = "" ]; do
```
maxim-lobanov, line 39: Honestly, I don't like this approach because you will have to wrap any command in `wait_for_apt_lock` like `wait_for_apt_lock "apt-get update"`. I don't think it is good from UX side because absolutely all apt command should be wrapped to this function and it decreases code readability.

I see two possible improvements here:
1) `wait_for_apt_lock` should be without arguments and should be called once at the begin of script where apt is needed.
Something like 
```
function wait_for_apt_lock() {
  while fuser /var/lib/dpkg/lock >/dev/null 2>&1 ; do
    sleep 1
  done 
}
But this approach still has some drawbacks and we shouldn't forget to include it to begin of all files.
```

2) replace "apt-get" command with wrapper of `apt-get`.
We can use solution described in this issue: https://askubuntu.com/a/375031

```
#!/bin/bash
while fuser /var/lib/dpkg/lock >/dev/null 2>&1 ; do
     sleep 1
done
/usr/bin/apt-get "$@"
```
and put this code to `/usr/local/sbin/apt-get`. It should replace default `apt-get` with your wrapper and fix should be applied to all apt-get calls without changes in every file. Personally, I prefer this approach, not sure if it has any drawbacks

PR #1556, opened 2020-09-08
[Ubuntu] move apt pkgs to the toolset.json

al-cheb:
# Description
Changes:
- move jq installation to the apt.sh
- move apt package list to the toolset.json from basic.sh
- move docker images from containercache.sh to the docker-moby.sh
- remove containercache.sh script
- remove some sections from packer templates

#### Related issue:
https://github.com/actions/virtual-environments-internal/issues/1080
## Check list


maxim-lobanov (approved):

images/linux/scripts/installers/basic.sh
```diff
@@ -4,90 +4,13 @@
...
-
-apt-get install -y --no-install-recommends $libcurelVer
-
-# install additional packages only for Ubuntu16.04
-if isUbuntu16; then
-    common_packages="$common_packages
-            libicu55"
-fi
 
+toolsetJson="$INSTALLER_SCRIPT_FOLDER/toolset.json"
```
AlenaSviridenko, line 11: Maybe rename to `toolset`? I think it is not important to specify that it is json

images/linux/scripts/installers/docker-moby.sh
```diff
@@ -49,19 +49,28 @@ else
...
-docker pull node:12-alpine
-docker pull debian:8
-docker pull debian:9
-docker pull alpine:3.7
-docker pull alpine:3.8
-docker pull alpine:3.9
-docker pull alpine:3.10
-docker pull ubuntu:14.04
+# Pull images
+images=(
```
AlenaSviridenko, line 53: does it make sense to move docker images to toolset too?

images/linux/scripts/installers/basic.sh
al-cheb: @AlenaSviridenko, Should we rename toolsetJson -> toolset in android.sh and azpowershell.sh too?

images/linux/scripts/installers/docker-moby.sh
al-cheb: cc @maxim-lobanov

images/linux/scripts/installers/basic.sh
AlenaSviridenko: @al-cheb why not? Looks like an easy update

al-cheb: Done.

vsafonkin (approved):

images/linux/scripts/installers/docker-moby.sh
maxim-lobanov: I like @AlenaSviridenko's suggestion but I don't think that it should be in scope of this PR.
Could you please create work item on internal board to improve docker installation?

PR #1576, opened 2020-09-11
add containerized build

droopy4096:
# Description

## Summary
Add packer manifest for building Azure Agent container. 

## Rationale
Using Azure Agent in containerized form allows for deployment into K8S environments (or simply Docker/Docker Swarm) with greater utilization of host resources and better latency.

Some modules have been disabled to keep container size down as well as some fundamental problems with running those components as part of container (like additional MySQL/PostgreSQL databases running etc.)

### DIFF
--- a/images.CI/azure-pipelines/ubuntu2004-docker.yml
+++ b/images.CI/azure-pipelines/ubuntu2004-docker.yml
@@ -13,8 +13,14 @@ pr:
     include:
     - main
 
-${{ if variables['pool_name'] }}:
-  pool: $(pool_name)
+pool:
+  ${{ if variables['pool_name'] }}:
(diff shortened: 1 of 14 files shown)


droopy4096:
@v-kyljon

v-kyljon:
@maxim-lobanov

AlenaSviridenko:
Hi @droopy4096,
First, please, let me admit that this a really good contribution.
Sad to say that, but we cannot accept it as Docker is not our use case, we are not experts here, and cannot support it in the future if anything goes wrong. We deploy Azure images and familiar with them and can provide a full support on them, but not on Docker containers.

I am closing this PR. Please, let us know if you have any other questions.
Thank you!

PR #1581, opened 2020-09-14
Update Miniconda version on Windows from 4.6.14 to 4.8.3/latest

weiji14:
# Description

Updates miniconda on Windows from 4.6.13 to 4.8.3 (actually to 'latest') to make it in line with the macOS and Ubuntu versions.

## Check list

### DIFF
--- a/images/win/Windows2019-Readme.md
+++ b/images/win/Windows2019-Readme.md
@@ -29,7 +29,7 @@
 - NPM 6.14.6
 - Yarn 1.22.4
 - pip 20.2.2 (python 3.7)
-- Miniconda 4.8.3
+- Miniconda 4.6.14
 - RubyGems 3.1.4
 - Helm 3.3.0
 - Composer 1.10.10
--- a/images/win/Windows2019-Readme.md
+++ b/images/win/Windows2019-Readme.md
@@ -29,7 +29,7 @@
 - NPM 6.14.6
 - Yarn 1.22.4
 - pip 20.2.2 (python 3.7)
-- Miniconda 4.6.14
+- Miniconda 4.8.3
 - RubyGems 3.1.4
 - Helm 3.3.0
 - Composer 1.10.10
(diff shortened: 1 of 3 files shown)


images/win/Windows2019-Readme.md
```diff
@@ -29,7 +29,7 @@
 - NPM 6.14.6
 - Yarn 1.22.4
 - pip 20.2.2 (python 3.7)
-- Miniconda 4.6.14
+- Miniconda 4.8.3
```
maxim-lobanov, line 32: documentation files (*.md) are auto-generated and updated automatically during deployment. Please revert them

weiji14: Ok, will do.

maxim-lobanov (approved):

maxim-lobanov:
Will try to get more context about 
```
# Lock to Miniconda 4.6 until we do the work to run `conda init` for the vsts user
# Then we can go back to installing the latest Miniconda
```

maxim-lobanov:
cc: @brcrista hello Brian! I have found that some time ago, you have hardcoded Miniconda 4.6.14 by https://github.com/microsoft/azure-pipelines-image-generation/commit/ffc77a7776aaba6e1c96a2d2d1deeef1ca11d50a
Related PR: https://github.com/microsoft/azure-pipelines-image-generation/pull/1151

Do you remember any context around it?
We would like to return back to approach of installing latest version but would like to understand if it can break any pipelines and how we can validate it.

brcrista:
@maxim-lobanov yeah, we used to install latest but then we hit an issue as noted in the comment.  I think what happens is that you just can't activate environments anymore because `conda` will complain that you haven't run `conda init`.

That's a step we would have to add during image generation.  Should be pretty easy.  It just adds some lines to the shell config, so we would have to do it for Bash, PowerShell, and CMD (on Windows).  Then, we can go back to latest on all OSes.

al-cheb (approved):

maxim-lobanov:
@brcrista , I have deployed the image based on current changes (without `conda init` in image-generation) and run canary tests. They passed successfully.
Any concerns to merge this one? Any other tests that we need to run?

brcrista (approved):

PR #1600, opened 2020-09-16
add sccache 

bmc-msft:
# Description
New tool, Bug fixing, or Improvement?  New tool

This adds [sccache](https://github.com/mozilla/sccache), a popular compiler caching tool used by Rust, Mozilla, and a large number of other projects.  Compiling from scratch in a standard Github Actions pipeline takes roughly 9 minutes.  Coupling sccache with `actions/cache@v1` drastically improves build times for many projects.

## Check list

### DIFF
--- a/images/linux/scripts/installers/rust.sh
+++ b/images/linux/scripts/installers/rust.sh
@@ -21,13 +21,14 @@ rustup component add rustfmt clippy
 cargo install bindgen cbindgen
 cargo install cargo-audit
 cargo install cargo-outdated
+cargo install sccache
 
 echo "Test installation of the Rust toochain"
 
 # Permissions
 chmod -R 777 $(dirname $RUSTUP_HOME)
 
-for cmd in rustup rustc rustdoc cargo rustfmt cargo-clippy bindgen cbindgen 'cargo audit' 'cargo outdated'; do
+for cmd in rustup rustc rustdoc cargo rustfmt cargo-clippy bindgen cbindgen 'cargo audit' 'cargo outdated' sccache; do
     if ! command -v $cmd --version; then
         echo "$cmd was not installed or is not found on the path"
         exit 1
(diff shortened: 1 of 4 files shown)


maxim-lobanov:
@bmc-msft , Hello!
Could you please tell a bit more about use-cases of using `actions/cache` with `sccache`?
Since `sccache` is cargo package, can we cache it via `actions/cache` instead of baking into the image? https://github.com/actions/cache/blob/main/examples.md#rust---cargo

bmc-msft:
1. sccache is a build-time cache tool that supports rust, c/c++, etc, to drastically reduce compile-time builds.  Basically, it is a compiler wrapper uses a hash of the compiler arguments (source files & command line arguments) to refer into a cache.  Using sccache coupled with an `actions/cache` directly enables saving the compiler cache.
2. While compiled binaries can be cached via `actions/cache`, it has issues.  It is my understanding that the cache key generation doesn't provide for using on-build-node information.   This means tying the cache key to prerequisite shared library versions isn't possible.  If any of the prerequisite libraries (libgcc, libssl, libc, libcrypto, etc) include breaking changes, the cached binaries are effectively broken without a reasonable path to correct the issue.

maxim-lobanov:
As I can see, [sccache](https://github.com/mozilla/sccache) provides prebuilt binaries. As I understand, it is portable binaries that include all necessary dependencies.
It means that we can cache them via `actions/cache` without any issues.

So I believe we can use `actions/cache` to [cache sccache](https://github.com/actions/cache/blob/main/examples.md#rust---cargo) and then cache everything what we need using this tool.

We have concerns about baking additional cargo packages into the image. In spite of very small size, installation of cargo packages take too much time like 9 minutes in this case that is huge time for single tool. Image-generation time is pretty important factor that we need to consider during adding new tools.

maxim-lobanov:
Hello,
Considering the installation time, we wouldn't like to bake this tool into the image for now.
Please consider using `actions/cache` to cache `sccache` on image

PR #1635, opened 2020-09-21
refactor: auto format JSON using VSC

hutson:
# Description

Reformat JSON Packer files using the default JSON formatter baked into Visual Studio Code.

Allows taking advantage of automatic JSON formatting to ensure consistent formatting across all JSON files, allowing me to focus on features instead.

#### Related issue:

None

## Check list

### DIFF
--- a/images/linux/ubuntu1604.json
+++ b/images/linux/ubuntu1604.json
@@ -24,7 +24,10 @@
         "run_validation_diskspace": "false",
         "announcements": "{{env `ANNOUNCEMENTS`}}"
     },
-    "sensitive-variables": ["client_secret", "github_feed_token"],
+    "sensitive-variables": [
+        "client_secret",
+        "github_feed_token"
+    ],
     "builders": [
         {
             "type": "azure-arm",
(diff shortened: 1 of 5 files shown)


maxim-lobanov:
@hutson , thank you for contribution!

I like this PR but I worry that templates' styling will be broken again in future because we have a lot of contributions.
@alepauly @AlenaSviridenko what do you think about configuring some linter for json templates? I think we can use official GH linter tool: https://github.com/github/super-linter

AlenaSviridenko:
@maxim-lobanov I am supporting this idea for templates, but probably not for the main code as it will bring a lot of noise that may spread our focus from the code to formatting issues.

hutson:
> I like this PR but I worry that templates' styling will be broken again in future because we have a lot of contributions.

Agreed! 👍 

> @alepauly @AlenaSviridenko what do you think about configuring some linter for json templates? I think we can use official GH linter tool: https://github.com/github/super-linter

Ideally I would like to see this integrated into Packer itself. For example, I wish `packer validate` supported plugins that we could use to integrate a _linter_ plugin into the command. We could then use `packer validate` to verify the template matched the project's preferred format. Unfortunately `packer` [only supports plugins](https://www.packer.io/docs/extending/plugins#installing-plugins) for `builder`, `post-processor`, and `provisioner`.

alepauly (approved):
@hutson - thank you!

alepauly:
@maxim-lobanov - I agree with @AlenaSviridenko  - just the templates to keep them readable for now. If we want to enforce formatting in the scripts, then I would say we do it through code reviews.

maxim-lobanov (approved):

maxim-lobanov:
Resolved conflict and rerun builds

PR #1647, opened 2020-09-22
Add attempt count to macOS VM name

dmitry-shibanov:
# Description
For now we can't rerun failed build, because it fails due to new virtual machine has the same name as previous VM. Adding attempt counter we create new virtual machine name on rerun failed build. 

## Check list

### DIFF
--- a/images.CI/macos/azure-pipelines/image-generation.yml
+++ b/images.CI/macos/azure-pipelines/image-generation.yml
@@ -13,6 +13,10 @@ jobs:
     clean: true
     fetchDepth: 1
 
+  - pwsh: |
+      Get-ChildItem Env:
+      exit 1
+
   - task: PowerShell@2
     displayName: 'Validate contributor permissions'
     condition: startsWith(variables['Build.SourceBranch'], 'refs/pull/')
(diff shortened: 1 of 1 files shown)


images.CI/macos/azure-pipelines/image-generation.yml
```diff
@@ -7,6 +7,8 @@ jobs:
   variables:
   - group: Mac-Cloud Image Generation
   - group: Mac-Cloud Image Generation Key Vault
+  - name: VirtualMachineName
+    value: $(Build.BuildNumber).$(System.JobAttempt)
```
maxim-lobanov, line 11: I don't think that it is good option. With this approach we will VM name like `macOS-10.15_20200922.9_unstable.1`

maxim-lobanov: It is better to keep `_unstable` prefix at the end

dmitry-shibanov: I think we can keep this variant.

al-cheb (approved):

Darleev (approved):

miketimofeev (approved):

images.CI/macos/azure-pipelines/image-generation.yml
maxim-lobanov: Sounds good

maxim-lobanov (approved):

Issue #1525, opened 2020-09-01
Switch Unix tools from Git internal tools to MSYS2 tools on Windows images


AlenaSviridenko:
### Breaking changes
"Git For Windows" Unix tools will be replaced by "MSYS2" Unix tools on Windows images.
Currently, folders `C:\Program Files\Git\bin` and `C:\Program Files\Git\usr\bin` and `C:\Program Files\Git\mingw64\bin` are added to PATH and it expose a ton of Unix tools (Git For Windows Internal tools) to PATH. We would like to remove those folders from PATH because one day they can be removed by Git maintainers.
A lot of people already depend on those tools so we would like to replace Git Unix tools by MSYS2 Unix tools to reduce impact.

### Target date
~~Deployment will start on November, 2 and changes propagation will take 2-3 days.~~
This change was rolled back and should be reevaluated 

### The motivation for the changes
Per communication with Git For Windows maintainers, we should not add Git Internal tools to PATH and rely on it. They could be removed in future and break a lot of people.

More details can be found in this PR by @dscho https://github.com/actions/virtual-environments/pull/1435.

All discussions can be found in this issue, also in MSYS2 issue https://github.com/actions/virtual-environments/issues/1572 and in the final PR https://github.com/actions/virtual-environments/pull/1648

### Possible impact
The most of the tools are identical between "Git For Windows" Internal and MSYS2 but some minor versions could be different and it can affect some builds.

- [x] Windows 2016
- [x] Windows 2019

### Mitigation ways
You will be able to return "Git For Windows" Unix tools to PATH using the following powershell step:
- GitHub Actions:
```yaml
echo "C:\Program Files\Git\mingw64\bin" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
echo "C:\Program Files\Git\usr\bin" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
```
- Azure Pipelines
```yaml
echo "##vso[task.prependpath]C:\Program Files\Git\mingw64\bin"
echo "##vso[task.prependpath]C:\Program Files\Git\usr\bin"
```

maxim-lobanov:
These changes will remove some Unix tools (that come with Git) from PATH on Windows machines. See the full list of tools in post above. This change won't affect such tools like bash and tar.

[...]

Issue #1572, opened 2020-09-10
Improve delivery of MSYS2


alepauly:
As discussed in #1525, there is probably room to improve on how and what we install with MSYS2. 

I believe what we want is for the install of MSYS2 that's available to be reliable and flexible enough to allow others to add the packages they need without unexpected results. Ideally, it would behave as a cached install so that `setup-msys2` type Actions can interoperate and take advantage of whatever is already laid out on the image to avoid having to do unecessary and repetitive work across workflow runs.

MSP-Greg:
As mentioned elsewhere, most (but not all) of my experience with MSYS2 is related to compiling Ruby extension gems.  As such, the main concern is the availability of MSYS2 gcc and associated tools.  Outside of Ruby, more tools are needed.  You may recall that the PR's I submitted to install MSYS2 started with only gcc, but as others asked for more tools, they were added to the install script.

Like a lot of software, people are happy to use whatever is installed, especially given the frequency of image updates.  Most of the repos I'm involved with I update the gcc tools.

What seems to be the central issue is whether an MSYS2 installation can be partially updated, or whether it always requires a full update.

With a full update, every package is updated, and can take quite a while.  Conversely, partial updates can fail, although it's not common.  But, given the frequency of runner updates, partial updates can happen in a few seconds, as often the needed packages are up to date, and the only thing done is the package database.

Re the issues with Git bash tools vs MSYS2 bash tools, they're the same, so I don't think anyone will be bothered if the bash shell changes.

But since MSYS2 has the tools installed, there will be conflicts if the `C:/msys64/mingw64/bin` folder is included in Path.

See:
https://github.com/MSP-Greg/actions-image-testing/runs/1082867165?check_suite_focus=true#step:13:5

Note that make & gcc exist in a t least two of MSYS2, `C:\ProgramData\Chocolatey\bin`, and `C:\Strawberry\c\bin` folders.

An option might be to only add `C:/msys64/mingw64/bin` to Path if the MSYS2 bash shell is used.

eine:
Hi @alepauly! First off, there is a very relevant issue to bear in mind: MSYS2 is a rolling environment for development. Until very few months ago, it was maintained by a single person, who built and tested every single package on his own computer.  Fortunately, other maintainers stepped in, and a partially automated workflow has been set up, which uses Actions and Releases: https://github.com/msys2/msys2-autobuild.

As a result, there are several constraints that you need to be aware of:

[...]

PR #1648, opened 2020-09-22
[Windows]  Removed Git internal tools from PATH

al-cheb:
# Description
Topic discussion: https://github.com/actions/virtual-environments/issues/1525

#### Related issue:
https://github.com/actions/virtual-environments/issues/1525
## Check list


images/win/Windows2016-Azure.json
```diff
@@ -235,6 +235,14 @@
                 "{{ template_dir }}/scripts/Installers/Update-AndroidSDK.ps1"
             ]
         },
+        {
+            "type": "powershell",
+            "elevated_user": "SYSTEM",
```
maxim-lobanov, line 240: Just for curiosity, why we need `elevated_user` here?

images/win/Windows2016-Azure.json
```diff
@@ -385,7 +384,8 @@
             "type": "powershell",
             "scripts":[
                 "{{ template_dir }}/scripts/Installers/Configure-Antivirus.ps1",
-                "{{ template_dir }}/scripts/Installers/Disable-JITDebugger.ps1"
+                "{{ template_dir }}/scripts/Installers/Disable-JITDebugger.ps1",
+                "{{ template_dir }}/scripts/Installers/Post-Deployment.ps1"
```
maxim-lobanov, line 388: I suggest to name this script `Configure-Shell.ps1`. It is pretty big so I don't think that we would like to add something to it.
Also I think both `Configure-Shell.ps1` and `Disable-JITDebugger.ps1` should be run before `Finalize-VM.ps1`

images/win/scripts/Installers/Post-Deployment.ps1
```diff
@@ -0,0 +1,19 @@
+# Create shells folder
+$shellPath = "C:\shells"
+$null = New-Item -Path $shellPath -ItemType Directory
+
+# MSYS2 <--> C:\msys64\usr\bin\bash.exe
+$null = New-Item -ItemType SymbolicLink -Path "$shellPath\msys2.exe" -Target "C:\msys64\usr\bin\bash.exe"
+
+# sh and bash <--> C:\msys64\usr\bin\bash.exe
+$null = New-Item -ItemType SymbolicLink -Path "$shellPath\bash.exe" -Target "C:\msys64\usr\bin\bash.exe"
```
maxim-lobanov, line 9: nitpick: May be `$shellPath\bash.exe` symlink should point to `$shellPath\msys2.exe`? Let's put this line after creation all other symlinks

images/win/scripts/Installers/Post-Deployment.ps1
```diff
@@ -0,0 +1,19 @@
+# Create shells folder
+$shellPath = "C:\shells"
+$null = New-Item -Path $shellPath -ItemType Directory
```
maxim-lobanov, line 3: nitpick: may be use `<command> | Out-Null` instead of `$null = <command>`?

images/win/scripts/Tests/Git.Tests.ps1
```diff
@@ -1,5 +1,5 @@
 Describe "Git" {
-    $gitTools = 'bash', 'awk', 'git', 'git-lfs'
+    $gitTools = 'git', 'git-lfs'
```
maxim-lobanov, line 2: Could you please add additional tests to validate:
1) default bash / sh come from MSYS
2) shell folder exists and contains all necessary symlinks
3) both shell folder and MSYS folder are in PATH and MSYS folder located after System32

images/win/scripts/Installers/Post-Deployment.ps1
al-cheb: Done.

images/win/Windows2016-Azure.json
al-cheb: Some components can't be installed via winrm in that case we should use elevated_user parameter[ If specified, the PowerShell script will be run with elevated privileges using the given Windows user.]

images/win/scripts/Tests/Git.Tests.ps1
al-cheb: Done.

images/win/scripts/Installers/Configure-Shell.ps1
```diff
@@ -0,0 +1,19 @@
...
+New-Item -ItemType SymbolicLink -Path "$shellPath\sh.exe" -Target "C:\msys64\usr\bin\sh.exe" | Out-Null
+
+# winbash <--> C:\Windows\System32\bash.exe
+New-Item -ItemType SymbolicLink -Path "$shellPath\winbash.exe" -Target "$env:SystemRoot\System32\bash.exe" | Out-Null
+
+# git4bash <--> C:\Program Files\Git\bin\bash.exe
+New-Item -ItemType SymbolicLink -Path "$shellPath\git4bash.exe" -Target "$env:ProgramFiles\Git\bin\bash.exe" | Out-Null
+
+# MSYS2 <--> C:\msys64\usr\bin\bash.exe
+New-Item -ItemType SymbolicLink -Path "$shellPath\msys2.exe" -Target "C:\msys64\usr\bin\bash.exe" | Out-Null
```
eine, line 16: This might break Action setup-msys2. Currently, the Action adds `msys2.cmd` to the PATH, and then `shell: msys2 {0}` is used. I'm afraid that `msys2.exe` might be picked instead of `msys2.cmd` or the other way round. There are several possibilities:

- I might be wrong. It might be that `shell: msys2` corresponds to `msys2.exe` (the symbolic link) and `shell: msys2 {0}` corresponds to `msys2.cmd`.
- Is this symbolic link actually needed? If both `sh` and `bash` do already point to the same target, I feel this to be redundant.
- We can deal with it introducing a breaking change and bumping setup-msys2 to v3, should it be really needed. However, unless the `msys2.exe` symlink is required ATM, I would defer it. I'd say, let's let other changes in this PR stabilize before going forward with "custom" msys2 entrypoint details.

/cc @lazka

images/win/scripts/Installers/Install-Msys2.ps1
```diff
@@ -9,35 +9,22 @@
...
-$origPath = $env:PATH
-$gitPath  = "$env:ProgramFiles\Git"
-
-$msys2_release = "https://api.github.com/repos/msys2/msys2-installer/releases/latest"
-
-$msys2Uri = ((Invoke-RestMethod $msys2_release).assets | Where-Object {
+# Downloading msys2
+$msys2Release = "https://api.github.com/repos/msys2/msys2-installer/releases/latest"
+$msys2Uri = ((Invoke-RestMethod $msys2Release).assets | Where-Object {
   $_.name -match "x86_64" -and $_.name.EndsWith("tar.xz") }).browser_download_url
```
eine, line 15: MSYS2 provides a self-extracting executable, apart from `tar.xz` tarballs. The content is the same. Hence, I'd suggest to use the self-extracting solution, which avoids the need for 7z.  See https://github.com/msys2/setup-msys2/blob/master/main.js#L12 and https://github.com/msys2/setup-msys2/blob/master/main.js#L222-L223. This is the commit where it was changed: https://github.com/msys2/setup-msys2/commit/03af082796cc8420caff534b796ff2c4830ee83c.

images/win/scripts/Installers/Install-Msys2.ps1
```diff
@@ -9,35 +9,22 @@
...
-# Git tar needs exe's from mingw64\bin
-$env:PATH = "$gitPath\usr\bin;$gitPath\mingw64\bin;$origPath"
-
-$tar = "$gitPath\usr\bin\tar.exe"
 
 # extract tar.xz to C:\
 Write-Host "Starting msys2 extraction"
-&$tar -xJf $msys2FileU -C /c/
-Remove-Item $msys2File
+$tempPath = Join-Path $env:Temp msys
```
eine, line 20: See comment about using the self-extracting executable (above).

images/win/scripts/Installers/Install-Msys2.ps1
```diff
@@ -9,35 +9,22 @@
...
+Extract-7Zip -Path $msys2File -DestinationPath $tempPath
+$tarPath = Resolve-Path $tempPath\msys*.tar
+Extract-7Zip -Path $tarPath -DestinationPath C:\
 Write-Host "Finished extraction"
 
 # Add msys2 bin tools folders to PATH temporary
-$env:PATH = "C:\msys64\mingw64\bin;C:\msys64\usr\bin;$origPath"
+$env:PATH = "C:\msys64\mingw64\bin;C:\msys64\usr\bin;$env:PATH"
```
eine, line 27: Why is it required to add MSYS2 to the PATH temporaly? MSYS2 terminals can have different PATH inheritance settings. The idiomatic way to achieve this would be setting it to `inherit`.

Nonetheless, in setup-msys2 a couple of helper functions are used: https://github.com/msys2/setup-msys2/blob/master/main.js#L175-L183.

images/win/scripts/Installers/Install-Msys2.ps1
```diff
@@ -9,35 +9,22 @@
...
 
 Write-Host "`n$dash bash pacman-key --init"
 bash.exe -c "pacman-key --init 2>&1"
```
eine, line 30: See the update procedure in https://github.com/msys2/setup-msys2/blob/master/main.js#L249-L261.

images/win/scripts/Installers/Install-Msys2.ps1
```diff
@@ -88,4 +75,15 @@ pacman.exe -Q | grep -v ^mingw-w64-
 
 Write-Host "`nMSYS2 installation completed"
 
+# Environment
+# add C:\msys64\mingw64\bin and C:\msys64\usr\bin to PATH
```
eine, line 79: I'm not sure about adding both of these being a good idea. IIRC, the former corresponds to MINGW64 packages and the later to MSYS packages. Those are different terminals, and many packages might be duplicated.

Moreover, I don't think that adding any of these to the PATH is required. The shells defined in `Configure-Shell.ps1` (above) do need to be available in the PATH. But, then, users can enter the desired terminal, and the corresponding ecosystem will be available already. Adding these PATHs will expose tools which might not work in cmd/powershell due to missing libs/dlls.

/cc @lazka

images/win/scripts/Installers/Install-Msys2.ps1
```diff
@@ -88,4 +75,15 @@ pacman.exe -Q | grep -v ^mingw-w64-
...
+# C:\msys64\mingw64\bin add after C:\Windows\System32 to not replace built-in tar.exe
+$regEnvKey = 'HKLM:\SYSTEM\CurrentControlSet\Control\Session Manager\Environment\'
+$pathValue = Get-ItemPropertyValue -Path $regEnvKey -Name 'Path'
+$pathValue += ";C:\msys64\mingw64\bin;C:\msys64\usr\bin"
+Set-ItemProperty -Path $regEnvKey -Name 'Path' -Value $pathValue
+
+# Rename python
+Rename-Item "C:\msys64\usr\bin\python.exe" -NewName "_python.exe"
```
eine, line 87: For example, this is an undesired effect of adding `C:\msys64\*\user\bin` to the PATH. Similar conflicts will exist for each tool that can be installed "natively" or through MSYS2.

eine (commented):
Thanks @al-cheb! I'm really grateful for seeing this moving forward. Please, see comments below, specially regarding the manipulation of the PATH.

images/win/scripts/Installers/Install-Msys2.ps1
lazka: MSYS2 also add those two to PATH (among others..), mingw tools will win anyway.

images/win/scripts/Installers/Install-Msys2.ps1
maxim-lobanov: @al-cheb Why we need to rename python here?
Can we just put MSYS folders at the end of the PATH? I think it was working this way previously with Git

al-cheb: Python fixed.

images/win/scripts/Installers/Configure-Shell.ps1
al-cheb: Fixed. Renamed to msys2bash.exe.

images/win/scripts/Installers/Install-Msys2.ps1
al-cheb: Fixed.

images/win/scripts/Installers/Install-Msys2.ps1
al-cheb: Thank you. Done.

images/win/scripts/Installers/Install-Msys2.ps1
al-cheb: After removing C:\Program Files\Git\mingw64\bin and C:\Program Files\Git\usr\bin folders from PATH, binaries which are resolved from them, will become unavailable. This change can cause issues and impact since some customers / tools depend on those binaries. We decided to replace:
```
C:\Program Files\Git\usr\bin -> C:\msys64\usr\bin
C:\Program Files\Git\mingw64\bin -> C:\msys64\mingw64\bin
```

This folders contain pretty the same list of binaries like C:\Program Files\Git\usr\bin with minimal differences and mostly the same versions. If we add C:\msys64\usr\bin at the end of PATH, removing Git tools from PATH should be pretty smoothly since the most tools will continue to work as previously. (Unix tools that are resolved from System32 folder right now, will continue to work as expected since system32 is located before in PATH)

images/win/scripts/Installers/Install-Msys2.ps1
al-cheb: Currently we don't have any issues with msys2 installation to rework the whole process.

images/win/scripts/Installers/Configure-Shell.ps1
```diff
@@ -0,0 +1,23 @@
...
+}
+
+# gitbash <--> C:\Program Files\Git\bin\bash.exe
+New-Item -ItemType SymbolicLink -Path "$shellPath\gitbash.exe" -Target "$env:ProgramFiles\Git\bin\bash.exe" | Out-Null
+
+# msysbash <--> C:\msys64\usr\bin\bash.exe
+New-Item -ItemType SymbolicLink -Path "$shellPath\msysbash.exe" -Target "C:\msys64\usr\bin\bash.exe" | Out-Null
+
+# Add shells to PATH
+Add-MachinePathItem $shellPath
```
miketimofeev, line 23: Shall we run Shell.Tests.ps1 in this file or it doesn't make sense since it's the latest provisioner before the full test run?

al-cheb: I don't think we should twice run the same test if the next step runs all tests.

images/win/scripts/Installers/Install-Msys2.ps1
eine: @al-cheb, my point is that, after this PR is merged, users of `shell: sh`, `shell: bash`, `shell: msys2bash` or `shell: gitbash` will already have all those tools available, by default, without the need to add them explicitly. That's how MSYS2 shells work, as @lazka said.

Hence, adding these locations to the PATH is only useful for users of `shell: powershell` or `shell: cmd` which rely on unix tools. That feels like a corner case to me, specially given that unix tools from System32 are resolved first, and those are the ones that users of powershell/cmd should expect only.

However, my concern is with python, ruby, etc. which might be installed as "native" tools (by downloading installers from their websites) or through MSYS2's `pacman`. Adding MSYS2 locations to the PATH will automatically expose those tools. However, if used from powershell/cmd, chances are that e.g. shared libs will be missing. Hence, the expected behaviour is only guaranteed for statically compiled tools.

Should the concern be that a relevant subset of users expects some Unix tools to be available in powershell/cmd, but those don't exist in System32; then, I'd recommend symlinking those tools explicitly, instead of adding the location of all MSYS2/MINGW64 binaries. Otherwise, we are reproducing the mistake of exposing binaries out of context (i.e. not expected to be used that way).

images/win/scripts/Installers/Install-Msys2.ps1
eine: Fair enough. That can be dealt with in #1572.

images/win/scripts/Installers/Install-Msys2.ps1
maxim-lobanov: The problem is that `powershell` is default shell on Windows so the most of users use it by default.
For example, toolkit `which` function [searches through PATH entries to find tool](https://github.com/actions/toolkit/blob/a5e05630d8e1d89cd75b5a12a972daae8abc6e1a/packages/io/src/io.ts#L236). Some default actions like `actions/cache` (and probably, some custom tasks) depend on it.

I agree that it is not good to expose binaries out of context but since they are already exposed right now, we have to keep them for backward compatibility.
As for the option with symlinks, we don't really know what tools should be symlinked right now. Finally, we can come to point where we symlink all tools + maintenance concerns (hardcoded list of symlinked tools).

If we add MSYS folder at the end of PATH, it should safe enough because it will be resolved only in case when tool was not resolved from all other places.

eine: @maxim-lobanov, fair enough. Then, I would strongly suggest to have this clearly explained in the docs. If the toolkit or some other Actions rely on it, it's ok not to break backward compatibility. Yet, in the mid-long term, the those Actions should be updated. I believe the better approach is to start with making it explicit that this is a temporal corner case solution.

al-cheb:
https://github.community/t/intl-datetimeformat-wrong-locale/133337/6?u=al-cheb

images/win/scripts/Installers/Install-Msys2.ps1
maxim-lobanov: Agree, @al-cheb could you please also update documentation to include info about it?

MSP-Greg:
OpenSSL just released 1.1.1h, and I was reminded of an issue.  From a related issue discussion:

> Unix tools that are resolved from System32 folder right now, will continue to work as expected since system32 is located before in PATH)

The popular MSVC OpenSSL build places OpenSSL dll's in System32.  I can't test it right now, but this may cause conflicts with MSYS2/MinGW tools loading those files?  I know in the past that I've renamed them in various Actions/AppVeyor scripts.  If users reshuffle Path, it isn't an issue.

And, I don't recall if it was a matter of MSYS2 OpenSSL failing install and then 'CI' resolving the dll's to the System32 files.  That isn't your problem, but can produce very confusing results.

al-cheb:
@MSP-Greg , @eine, Hello.

Could you please review the first part of PR to remove git internal tools from PATH?

Changes:
- Remove git internal tools from PATH `"/o:PathOption=Cmd"`
- Add symlinks  to shells
### Shells
| Name               | Target                            |
| ------------------ | --------------------------------- |
| bash.exe (Default) | C:\msys64\usr\bin\bash.exe        |
| gitbash.exe        | C:\Program Files\Git\bin\bash.exe |
| msysbash.exe       | C:\msys64\usr\bin\bash.exe        |
| sh.exe             | C:\msys64\usr\bin\sh.exe          |
| winbash.exe        | C:\windows\System32\bash.exe      |

- Use sfx.exe to install msys2
- Replace git internal tools to msys2 and add small notes
##### Notes:
```
Location: C:\msys64

1. MSYS2 is pre-installed on image
2. C:\msys64\mingw64\bin is added to PATH and has lower precedence than C:\Windows\System32
3. C:\msys64\usr\bin is added to PATH and has lower precedence than C:\Windows\System32
4. Default bash.exe shell is set to the C:\msys64\usr\bin\bash.exe
```
- Add bash.exe to PATH -> `C:\msys64\usr\bin\bash.exe`

eine:
@al-cheb, as commented in https://github.com/actions/virtual-environments/pull/1648#discussion_r493806806, I believe there should be a stronger warning about relying on msys64 being added to the PATH by default. The proposed notes are informative only. Apart from that, LGTM.

MSP-Greg (approved):
LGTM

images/win/scripts/Installers/Install-Git.ps1
```diff
@@ -37,7 +37,7 @@ Choco-Install -PackageName hub
 [Environment]::SetEnvironmentVariable("GCM_INTERACTIVE", "Never", [System.EnvironmentVariableTarget]::Machine)
 
 if (Test-IsWin16) {
-    $env:Path += ";$env:ProgramFiles\OpenSSH-Win64"
+    $env:Path += ";$env:ProgramFiles\Git\usr\bin\"
```
eine, line 40: This is a Git for Windows internal tool, isn't it?

al-cheb: @eine, The `ssh-keyscan.exe`  is not available by default on Windows Server 2016. We decided to use `ssh-keyscan.exe` ,only during installation phrase, from git for windows as long as the util will not be deleted.

eine: > The `ssh-keyscan.exe` is not available by default on Windows Server 2016. We decided to use `ssh-keyscan.exe` ,only during installation phrase, from git for windows as long as the util will not be deleted.

@al-cheb, I understand that, but adding `Git\usr\bin\` defeats the purpose of this issue. `ssh-keyscan.exe` is available in MSYS2's `/usr/bin` too:

```sh
eine@DESKTOP MINGW64 /t/
# which ssh-keyscan
/usr/bin/ssh-keyscan
```

al-cheb: Thank you. Fixed.

maxim-lobanov (approved):

PR #1653, opened 2020-09-23
APT/DPKG retry logic to avoid locks

Darleev:
# Description
Improvement 
We need to implement retry logic for apt, to avoid possible dpkg lock issues

#### Related issue:

## Check list


images/linux/scripts/base/apt-get-retry.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
```
al-cheb, line 1: Looks like duplicate the apt-retry.sh

images/linux/scripts/base/apt-get-retry.sh
```diff
@@ -0,0 +1,22 @@
...
+
+i=0
+tput sc
+echo JustForTestingPurposes
```
al-cheb, line 5: Why do we need it?

images/linux/scripts/base/apt-get-retry.sh
```diff
@@ -0,0 +1,22 @@
...
+        0 ) j="-" ;;
+        1 ) j="\\" ;;
+        2 ) j="|" ;;
+        3 ) j="/" ;;
+    esac
+    if [ -n "$TERM" ] && [ "$TERM" = "unknown" ] ; then
+        TERM=rc
+    fi
+    tput rc
+    echo -en "\r[$j] Waiting for other software managers to finish..."
```
maxim-lobanov, line 17: Why do we need this logging?

images/linux/scripts/base/apt-retry.sh
```diff
@@ -0,0 +1,22 @@
...
+        0 ) j="-" ;;
+        1 ) j="\\" ;;
+        2 ) j="|" ;;
+        3 ) j="/" ;;
+    esac
+    if [ -n "$TERM" ] && [ "$TERM" = "unknown" ] ; then
+        TERM=rc
+    fi
+    tput rc
+    echo -en "\r[$j] Waiting for other software managers to finish..."
```
maxim-lobanov, line 17: Why do we need this logging?

images/linux/scripts/base/apt.sh
```diff
@@ -18,3 +18,9 @@ apt-get install jq
 add-apt-repository -y ppa:apt-fast/stable
 apt-get update
 apt-get install apt-fast
+
+# Install retry logic for apt and apt-get
+sudo mv apt-get-retry.sh /usr/local/sbin/apt-get
```
al-cheb, line 23: Why do we need to mv operation if we already copied this file?
        {
            "type": "file",
            "source": "{{template_dir}}/scripts/base/apt-get-retry.sh",
            "destination": "/usr/local/sbin/apt-get"
        },

images/linux/scripts/base/apt-retry.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
```
miketimofeev, line 1: `#!/bin/bash -e` or `set -e`?

images/linux/ubuntu1804.json
```diff
@@ -54,10 +54,28 @@
...
+        {
+            "type": "file",
+            "source": "{{template_dir}}/scripts/base/apt-retry.sh",
+            "destination": "/usr/local/sbin/apt"
+        },
         {
             "type": "shell",
             "inline": [
                 "mkdir {{user `image_folder`}}",
+                "chmod -R 755 /usr/local/sbin",
```
al-cheb, line 78: Why do we apply the second time at this line and line 60?

Darleev: Just to return permissions back to default one from 777

images/linux/scripts/base/apt-retry.sh
Darleev: thank you, I have added it

images/linux/scripts/base/apt-retry.sh
Darleev: Just for troubleshooting purposes

images/linux/scripts/base/apt-get-retry.sh
Darleev: the first one for apt;
the second for apt-get;

images/linux/scripts/base/apt-get-retry.sh
Darleev: for troubleshooting purposes.

images/linux/scripts/base/apt-retry.sh
maxim-lobanov: I don't think that we should bring such logging to main branch. It looks redundant and unclear. If you prefer, you can leave echo line everything else should be removed

images/linux/scripts/base/apt-get-retry.sh
al-cheb: Why don't we use symlinks?

images/linux/ubuntu1804.json
al-cheb: Why do we need to change permissions with sudo?

PR #1707, opened 2020-10-01
Fail Ubuntu image building on error in generation scripts

dsame:
# Description
Improvement

Add the snippet to a top of every .sh script in order to make this script to fail if any command returns not zero exit code
```
set -e
set -o pipefail
```

## Check list


images/linux/scripts/installers/7-zip.sh
```diff
@@ -4,6 +4,8 @@
 ##  Desc:  Installs 7-zip
 ################################################################################
 
+set -e
+set -o pipefail
```
maxim-lobanov, line 8: Does `packer` provides any way to set default `erroraction` behavior for all script on template level?
I worry that we can just forget to add this line to new script so it would be great to apply this behavior for all scripts without putting code to every script

images/linux/scripts/installers/7-zip.sh
```diff
@@ -4,6 +4,8 @@
 ##  Desc:  Installs 7-zip
 ################################################################################
 
+set -e
```
maxim-lobanov, line 7: Why can't we set it in shebang on the first line?

```
#!/bin/bash -e
```

images/linux/scripts/installers/7-zip.sh
miketimofeev: It looks like the packer checks only the last exit code from the script, but if something went wrong in the middle we can't see it. I've faced with a lot of such cases during macOS provision debugging

images/linux/scripts/installers/7-zip.sh
miketimofeev: Agree, it looks better. The same for -o

images/linux/scripts/installers/7-zip.sh
maxim-lobanov: Let's investigate another ways to do it. Including `set -e` to every script looks like bad solution here.
It will require additional maintenance and supporting. We will have to validate manually that all future scripts will contain this line.
Ideally, we need something like default error action parameter for all image provisioners

miketimofeev: @maxim-lobanov we should either get rid of` /bin/bash` at all or put `-e` judging by the packer docks
image

miketimofeev: However, getting rid of `/bin/bash` doesn't help me

maxim-lobanov: For Windows, I saw solution: `Add-Content -Path $PSHOME/Profile.ps1 -Value "$ErrorActionPreference='Stop'"`

images/linux/scripts/installers/7-zip.sh
dsame: I believe the packer runs the script as `/bin/bash script.sh` instead of `./script.sh` so the tag is ignored

miketimofeev: no, it's not. I've checked on macos — `#!/bin/bash -e `works!

PR #1714, opened 2020-10-01
Move linux provision scripts to virtual-environments

LeonidLapshin:
# Description
Scripts that are related to system configuration moved from MMS to virtual-environments repository.

## Check list


images/linux/scripts/installers/cleanup.sh
```diff
@@ -27,4 +27,15 @@ after=$(df / -Pm | awk 'NR==2{print $4}')
...
+ # Remove quotes if found around PATH
+if [[ $PATH == \"*\" ]]
+then
+    PATH=${PATH#"\""}
+    PATH=${PATH%"\""}
+    echo "PATH=$PATH" | sudo tee -a /etc/environment
+    echo "Updated path to remove quotes: $PATH"
+else
+    echo "No path update needed: $PATH"
+fi
```
maxim-lobanov, line 41: Do we really need this hack now?
It looks like a hotfix to fix some issue with PATH. Can we check if it is still actual?

images/linux/scripts/installers/git.sh
```diff
@@ -52,3 +52,7 @@ else
     echo "[!] Hub CLI was not installed"
     exit 1
 fi
+
+# Add well-known SSH host keys to known_hosts
+ssh-keyscan -t rsa github.com >> /etc/ssh/ssh_known_hosts
+ssh-keyscan -t rsa ssh.dev.azure.com >> /etc/ssh/ssh_known_hosts
```
maxim-lobanov, line 58: I suggest to put this logic to `https://github.com/actions/virtual-environments/blob/main/images/linux/scripts/installers/configure-environment.sh`

maxim-lobanov: Discussed offline. Decided to leave here

al-cheb (approved):

images/linux/scripts/tests/Run-Tests.ps1
```diff
@@ -0,0 +1,23 @@
...
+    Output = @{ Verbosity = "Detailed" }
+    TestResult = @{
+        Enabled = $true
+        OutputPath = $OutputFile;
+        OutputFormat = "NUnitXML"
+    }
+}
+
+Invoke-Pester -Configuration $configuration
+Invoke-PesterTests "*"
```
maxim-lobanov, line 23: Where this function is defined?

images/linux/scripts/tests/Run-Tests.ps1
```diff
@@ -0,0 +1,23 @@
...
+Set-Location -Path $WorkingDirectory
+
+$configuration = [PesterConfiguration] @{
+    Output = @{ Verbosity = "Detailed" }
+    TestResult = @{
+        Enabled = $true
+        OutputPath = $OutputFile;
+        OutputFormat = "NUnitXML"
+    }
+}
```
maxim-lobanov, line 20: I think we should use Windows version of this code because we should fail build if tests are not passed: https://github.com/actions/virtual-environments/blob/main/images/win/scripts/ImageHelpers/TestsHelpers.ps1#L67

images/linux/scripts/tests/Common.Tests.ps1
```diff
@@ -0,0 +1,5 @@
+Describe "PATH check" {
+    It "PATH is not quoted" {
+        $PATH | Should -Not -Be '"*"'
```
maxim-lobanov, line 3: May be `$PATH | Should -Not -Match '"'`?

images/linux/scripts/installers/cleanup.sh
LeonidLapshin: hack removed :)

images/linux/scripts/tests/Run-Tests.ps1
LeonidLapshin: added Invoke-PesterTests function

images/linux/scripts/tests/Common.Tests.ps1
LeonidLapshin: I guess PATH can contain '"' but should not be in "", please, correct me if I'm wrong

maxim-lobanov (approved):

Yanson:
This change seems to have broken our build.

```
==> azure-arm: Provisioning with shell script: /agent/_work/1/s/images/linux/scripts/installers/post-deployment.sh
==> azure-arm: mv: cannot stat '/imagegeneration/post-generation': No such file or directory
==> azure-arm: Provisioning step had errors: Running the cleanup provisioner, if present...
```

Are there any additional settings I need for this to work? In case it is related, my build uses modified ubuntu2004.json to publish to SIG.

```diff
+ "managed_image_name": "{{user `image_name`}}",
+ "managed_image_resource_group_name": "{{user `sig_resource_group_name`}}",
+ "shared_image_gallery_destination": {
+     "resource_group": "{{user `sig_resource_group_name`}}",
+     "gallery_name": "{{user `sig_name`}}",
+     "image_name": "{{user `image_name`}}",
+     "image_version": "{{user `image_version`}}",
+     "replication_regions": ["{{user `location`}}"]
+ },
```

maxim-lobanov:
@Yanson , please make sure that you have [this line](https://github.com/actions/virtual-environments/pull/1714/files#diff-229e82c6e7150feda8f8382a43e255c2f893d8f9f895e5183966e6e34ec6c4ceR97) in your template. May be some conflicts

Yanson:
Thanks. I do have that but note ubuntu 16 and 18 show ```"destination": "{{user `image_folder`}}"``` but 20 shows ```"destination": "{{user `installer_script_folder`}}"```. Is that intentional? Could it be a problem? (I only build 20).

maxim-lobanov:
Good point, not sure why we have missed Ubuntu 20.04 validation.
@LeonidLapshin could you please take a look?

Yanson:
Can confirm https://github.com/actions/virtual-environments/pull/1812 fixes this issue for me.

PR #1720, opened 2020-10-02
[Windows] Fail Windows image building on error in generation scripts

al-cheb:
# Description
Add $ErrorActionPreference = "Stop" to all image generation scripts to stop generation process in case of any error.

Changes:
- Remove all $ErrorActionPreference = "Stop" statements from scripts
- Set $ErrorActionPreference="Stop" for AllUsersAllHosts [Initialize-VM.ps1]
- Remove at the end of image generation [Finalize-VM.ps1]
- Remove Import-Module -Name ImageHelpers

#### Related issue:
https://github.com/actions/virtual-environments-internal/issues/1226

## Check list

### DIFF
--- a/images/win/scripts/Installers/Install-Sbt.ps1
+++ b/images/win/scripts/Installers/Install-Sbt.ps1
@@ -2,7 +2,6 @@
 ##  File:  Install-Sbt.ps1
 ##  Desc:  Install sbt for Windows
 ################################################################################
-Import-Module -Name ImageHelpers
 
 # Install the latest version of sbt.
 # See https://chocolatey.org/packages/sbt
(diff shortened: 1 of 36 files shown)


images/win/scripts/Installers/Install-Sbt.ps1
```diff
@@ -2,8 +2,6 @@
 ##  File:  Install-Sbt.ps1
 ##  Desc:  Install sbt for Windows
 ################################################################################
-$ErrorActionPreference = "Stop"
-
 Import-Module -Name ImageHelpers
```
maxim-lobanov, line 5: May be we should also remove `Import-Module -Name ImageHelpers`?
It doesn't make much sense since all functions are auto-imported

al-cheb: Done.

maxim-lobanov (approved):

dibir-magomedsaygitov (approved):

vsafonkin (approved):

miketimofeev (approved):

PR #1741, opened 2020-10-06
[macOS] Fix haskell installation

miketimofeev:
# Description
Ghcup output has changed recently and awk doesn't work for us anymore, it returns the wrong data either in $3 or $4
image
image

The solution is switching to grep usage and search the version by pattern.

#### Related issue:
https://github.com/actions/virtual-environments-internal/issues/1261

## Check list

### DIFF
--- a/images/macos/provision/core/haskell.sh
+++ b/images/macos/provision/core/haskell.sh
@@ -4,7 +4,7 @@ curl --proto '=https' --tlsv1.2 -sSf https://get-ghcup.haskell.org | sh
 export PATH="$HOME/.ghcup/bin:$PATH"
 echo 'export PATH="$PATH:$HOME/.ghcup/bin"' >> "$HOME/.bashrc"
 
-# ghcup output looks like this "x ghc 8.8.3 base-4.13.0.0 hls-powered", need to take all the first versions only(8.8.3 in that case) and avoid pre-release ones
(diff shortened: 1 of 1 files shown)


images/macos/provision/core/haskell.sh
```diff
@@ -4,7 +4,10 @@ curl --proto '=https' --tlsv1.2 -sSf https://get-ghcup.haskell.org | sh
 export PATH="$HOME/.ghcup/bin:$PATH"
 echo 'export PATH="$PATH:$HOME/.ghcup/bin"' >> "$HOME/.bashrc"
 
-availableVersions=$(ghcup list | grep -v "prerelease" | grep "ghc " | awk '{print $3}')
+# ghcup output looks like this "x ghc 8.8.3 base-4.13.0.0 hls-powered", need to take all the first versions only(8.8.3 in that case) and avoid pre-release ones
+availableVersions=$(ghcup list | grep -v "prerelease" | grep "ghc " | grep -Eo ' ([0-9]*\.){2}[0-9]*' | tr -d ' ')
```
al-cheb, line 8: nitpick: `ghcup list -t ghc -r | awk '{print $2}'`

al-cheb (approved):

images/macos/provision/core/haskell.sh
miketimofeev: I should have checked `ghcup` flags, thank you!

maxim-lobanov (approved):

PR #1755, opened 2020-10-07
Fail Ubuntu image building on error in generation scripts

dsame:
# Description
Improvement

Added set -e to all image generation scripts to fail generation process in case of any error within the shell script

## Check list
- [x ] Related issue / work item is attached
- [x ] Changes are tested and related VM images are successfully generated


images/linux/scripts/installers/dpkg-config.sh
```diff
@@ -7,6 +7,7 @@
 # automatic installs. If you are using this front-end, and require
 # non-default  answers  to questions, you will need to preseed the
 # debconf database
+
```
al-cheb, line 10: Why do we need empty line here?

images/linux/scripts/installers/swift.sh
```diff
@@ -1,15 +1,20 @@
...
 
-
 # Install
 image_label="$(lsb_release -rs)"
-swift_version=$(curl -s https://swift.org/download/ | grep -m1 "id=\"swift-" | awk -F"[ <]" '{print $4}')
+
+# curl would compline we do not read the stream till the very end (result code 23)
+# so let's turn off pipefail for a while
+set +o pipefail
+swift_version=$(curl -s -L https://swift.org/download|awk -F"[ <]" '/id="swift-/ {print $4; exit}')
```
al-cheb, line 13: What about `-N, --no-buffer` parameter and remove set statements?

vsafonkin (approved):

maxim-lobanov (approved):

AlenaSviridenko (approved):

PR #1756, opened 2020-10-07
[macOS] Fail builds on errors during the image generation

miketimofeev:
# Description
Currently, the build isn't failed if there is an error somewhere in the middle of the provision script, and that leads to unpredictable results because we can miss tool installation failure.
This PR adds `-e` flag to exit the provision script immediately and fail a build. In addition `-o pipelinefail` flag was also added because the bash shell normally only looks at the exit code of the last command of a pipeline. This behavior is not ideal as it causes the `-e` option to only be able to act on the exit code of a pipeline's last command, `-o pipelinefail` option sets the exit code of a pipeline to that of the rightmost command to exit with a non-zero status or zero if all commands of the pipeline exit successfully.
During the tests, a few scripts had to be reworked since they contained redundant logic or had bugs, which causes a non-zero exit code. Here is the full list:

1. `provision/configuration/configure-machine.sh`get rid of the no operational thus redundant commands:
```
==> vsphere-clone: 2020-09-29 12:37:23.590 defaults[833:4810] Could not write domain /Library/Preferences/com.apple.alf; exiting
    vsphere-clone: Configuring system time to GMT...
==> vsphere-clone: rm: /etc/localtime: Permission denied
```
2. `provision/core/ruby.sh` ruby 2.6 switched to ruby 2.7 due to gem installation errors with 2.6. Previously we decided not to use 2.7 because of Fastlane issues, but now it should be fine with 2.7 since brew Fastlane formula uses ruby 2.7 for several months already.
```
==> vsphere-clone: ERROR:  While executing gem ... (NoMethodError)
    vsphere-clone: Successfully installed webrobots-0.1.2
==> vsphere-clone:     undefined method `[]' for nil:NilClass
```
3. `provision/core/vsmac.sh`remove `vstools` invocation because it does nothing without any arguments and moreover doesn't work on 10.13 & 10.14 due to outdated default Mono
```
    vsphere-clone: Launching vstools...
==> vsphere-clone: 2020-09-30 03:48:39.144 vstool[32105:90804] Could not launch: Visual Studio
==> vsphere-clone: This application requires a newer version (6.4.0.94+) of the Mono framework.
```
4. `provision/core/xamarin.sh` workaround for big sur is not needed since there is no mono 6.4 there.
```
    vsphere-clone: Installing nuget 5.3.1 for Mono 6.4.0
==> vsphere-clone: /Users/runner/utils/xamarin-utils.sh: line 192: cd: /Library/Frameworks/Mono.framework/Versions/6.4.0/lib/mono/nuget: No such file or directory
```
5. `provision/core/node.sh` and `provision/core/nvm.sh` get rid of `npm config set prefix /usr/local` for 10.13 & 10.14 to avoid errors during nvm installation and instead set default node version via nvm as we do for 10.15.
```
==> vsphere-clone: nvm is not compatible with the npm config "prefix" option: currently set to "/usr/local"
==> vsphere-clone: Run `npm config delete prefix` or `nvm use --delete-prefix v6.17.1` to unset it.
```
6. `provision/core/postgresql.sh` postgre can't accept connections immediately after start so retry logic implemented.
```
==> vsphere-clone: Cloning into '/usr/local/Homebrew/Library/Taps/homebrew/homebrew-services'...
==> vsphere-clone: Tapped 1 command (40 files, 343.2KB).
    vsphere-clone: ==> Successfully started `postgresql` (label: homebrew.mxcl.postgresql)
    vsphere-clone: /tmp:5432 - no response
```
7. `provision/core/xcode-postbuild.sh` Xcode 12.2_beta installed not in the beta directory at the moment, had to implement a workaround.
```
    vsphere-clone: Erase a device's contents and settings:
    vsphere-clone:     Xcode Version: 10.3
    vsphere-clone:     Xcode Version: 11
    vsphere-clone:     Xcode Version: 11.1
    vsphere-clone:     Xcode Version: 11.2.1
    vsphere-clone:     Xcode Version: 11.3.1
    vsphere-clone:     Xcode Version: 11.4
    vsphere-clone:     Xcode Version: 11.4.1
    vsphere-clone:     Xcode Version: 11.5
    vsphere-clone:     Xcode Version: 11.6
    vsphere-clone:     Xcode Version: 11.7
    vsphere-clone:     Xcode Version: 12_beta
    vsphere-clone:     Xcode Version: 12
    vsphere-clone:     Xcode Version: 12.2_beta
==> vsphere-clone: xcode-select: error: invalid developer directory '/Applications/Xcode_12.2_beta.app/Contents/Developer'
```
8. `provision/configuration/finalize-vm.sh` big sur can't into ignoring updates 😫

#### Related issue:
https://github.com/actions/virtual-environments-internal/issues/1245

## Check list


images/macos/provision/configuration/finalize-vm.sh
```diff
@@ -1,12 +1,14 @@
...
+#!/bin/bash -e -o pipefail
 
 # Close all finder windows because they can interfere with UI tests
 osascript -e 'tell application "Finder" to close windows'
 
-# Ignore available updates to prevent system pop-ups
-updateName=$(softwareupdate -l | grep "Title: " | awk -F[:,] '{print $2}' | awk '{$1=$1};1')
-if [ ! -z "$updateName" ]; then
-    sudo softwareupdate --ignore "$updateName"
+if is_Less_BigSur; then
```
al-cheb, line 6: Should we add source utils.sh to import this function?

miketimofeev: looks like! Thanks!

images/macos/provision/core/postgresql.sh
```diff
@@ -1,18 +1,26 @@
...
+    echo "Check PostgreSQL service status"
+    eval $COMMAND && break
+    ((i--))
+    if [ $i == 0 ]; then
+        echo "PostgreSQL service not ready, all attempts exhausted"
+        exit 1
+    fi
+    echo "PostgreSQL service not ready, wait 10 more sec, attempts left: $i"
+    sleep 10
+done
```
maxim-lobanov, line 23: Do we really need this validation in installation script?
Can we move test to tests files?

PowerShell command `Start-Service` can wait until service is started fully

images/macos/provision/core/xcode-postbuild.sh
```diff
@@ -15,6 +15,10 @@ do
     #add sleep to let CoreSimulatorService to exit
     sleep 3
 
+    # Version 12.2_beta installed into 12.2 directory
+    if [[ $XCODE_VERSION == "12.2_beta" ]] ; then
+        XCODE_VERSION="12.2"
```
maxim-lobanov, line 20: It is not workaround but expected behavior for all future Xcode versions. I suggest replacing this logic to just `XCODE_VERSION=$(echo $XCODE_VERSION | cut -d"_" -f 1)`

images/macos/provision/core/postgresql.sh
miketimofeev: The check was initially implemented with some intent for it 😃
Does it make sense to do a test if we need postgre being stopped at the end of the build anyway?

al-cheb (approved):

images/macos/provision/core/xcode-postbuild.sh
miketimofeev: cool, didn't know that, thanks!

images/macos/provision/core/postgresql.sh
AlenaSviridenko: I think we need this script as an indicator that postgres was installed correctly

AlenaSviridenko (approved):

PR #1803, opened 2020-10-13
Remove mobile-center-cli package from MacOS images

MaksimZhukov:
# Description
We've run into an issue with the `mobile-center-cli` package during the image generation. We've noticed that the `mobile-center-cli` package is no longer supported. In scope of this PR we removed installation of the `mobile-center-cli` from all MacOS images.

## Check list

### DIFF
--- a/images/macos/provision/core/node.sh
+++ b/images/macos/provision/core/node.sh
@@ -1,10 +1,5 @@
 source ~/utils/utils.sh
 
-node_common_modules=(
-  node-gyp
-  mobile-center-cli
-)
-
 node_catalina_modules=(
   appcenter-cli
   newman
@@ -40,8 +35,6 @@ echo Installing yarn...
 curl -o- -L https://yarnpkg.com/install.sh | bash
 
 if is_Less_BigSur; then
-    for module in ${node_common_modules[@]}; do
-        echo "Install $module"
-        npm install -g $module
-    done
+  echo "Install node-gyp"
+  npm install -g node-gyp
 fi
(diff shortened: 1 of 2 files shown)


images/macos/provision/core/node.sh
```diff
@@ -1,10 +1,5 @@
 source ~/utils/utils.sh
 
-node_common_modules=(
-  node-gyp
-  mobile-center-cli
-)
-
 node_catalina_modules=(
```
miketimofeev, line 3: let's change the name since we're going to install these modules for Big Sur as well

maxim-lobanov (approved):

images/macos/provision/core/node.sh
```diff
@@ -40,8 +35,6 @@ echo Installing yarn...
 curl -o- -L https://yarnpkg.com/install.sh | bash
 
 if is_Less_BigSur; then
```
miketimofeev, line 37: what's wrong with Big Sur and node-gyp?

al-cheb (approved):

images/macos/provision/core/node.sh
maxim-lobanov: We wouldn't like to bring useless module to future images and can't remove it from existing ones

images/macos/provision/core/node.sh
MaksimZhukov: Thanks!

AlenaSviridenko (approved):

vsafonkin (approved):

PR #1815, opened 2020-10-14
[macOS] Separate xcode command line tools installation & Xcode 12.1 for Big Sur

miketimofeev:
# Description
Sometimes homebrew can't install xcode command line tools because fetching from apple software update endpoint returns nothing. This PR adds separate command line tools installation using the same method as implemented in brew, but with a simple retry logic. 
Also, we can get rid of separate xcode clt installation for macOS Big Sur since it's not working anymore.
And the final addition is Xcode 12.1 for Big Sur.

#### Related issue:
https://github.com/actions/virtual-environments-internal/issues/1031

## Check list


images/macos/provision/core/homebrew.sh
```diff
@@ -1,23 +1,39 @@
...
+    if [[ $retries -eq 0 ]]; then
+        echo "Unable to find the Command Line Tools, all the attempts exhausted"
+        exit 1
+    fi
+    echo "Command Line Tools not found, trying to install them via softwareupdates, $retries attempts left"
+    install_clt
+    ((retries--))
+    echo "Wait $sleepInterval seconds before the next check for installed Command Line Tools"
+    sleep $sleepInterval
+done
```
maxim-lobanov, line 21: could you please keep this logic in the separate file `xcode-ctl.sh`. I don't think it makes sense to mix it with homebrew installation. Actually, they are not related

maxim-lobanov: In this case, you can put both `is_clt_installed` and `install_clt` directly to `xcode-ctl.sh` and don't put them to helper file because they are not re-usable

AlenaSviridenko (approved):

al-cheb (approved):

PR #1832, opened 2020-10-16
added a swiftc symlink

LeonidLapshin:
# Description
There is a need to have a swiftc symlink point to swift binary for some tools (such as bazel's swift rules) 

## Check list

### DIFF
--- a/images/linux/scripts/installers/swift.sh
+++ b/images/linux/scripts/installers/swift.sh
@@ -16,11 +16,17 @@ mv swift-$swift_version-RELEASE-ubuntu$image_label /usr/share/swift
 SWIFT_PATH="/usr/share/swift/usr/bin"
 SWIFT_BIN="$SWIFT_PATH/swift"
 ln -s "$SWIFT_BIN" /usr/local/bin/swift
+ln -s "$SWIFT_BIN" /usr/local/bin/swiftc
 echo "SWIFT_PATH=$SWIFT_PATH" | tee -a /etc/environment
 
 # Run tests to determine that the software installed as expected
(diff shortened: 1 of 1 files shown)


images/linux/scripts/installers/swift.sh
```diff
@@ -16,6 +16,7 @@ mv swift-$swift_version-RELEASE-ubuntu$image_label /usr/share/swift
 SWIFT_PATH="/usr/share/swift/usr/bin"
 SWIFT_BIN="$SWIFT_PATH/swift"
 ln -s "$SWIFT_BIN" /usr/local/bin/swift
+ln -s "$SWIFT_BIN" /usr/local/bin/swiftc
```
maxim-lobanov, line 19: Why we create symlink from `swift` -> `swiftc`? It is the different tools.
- swift is an interactive environment (REPL) for Swift
- swiftc is the Swift compiler

https://stackoverflow.com/questions/57777091/whats-the-difference-between-swift-and-swiftc

LeonidLapshin: It's the same binary I guess, the provided link contains comment:

> The compiler isn't delegating anything to a REPL. They're just the same binary. This is very common in the unix world.

maxim-lobanov: I see, that makes sense.
Anyway, IMHO, better to make consistent symlinks:
- swift -> swift
- swiftc -> swiftc
Just in case if it changes in future

maxim-lobanov (approved):

al-cheb (approved):

PR #1876, opened 2020-10-22
add spotlight reindex

LeonidLapshin:
# Description
Added a step for indexing before VM shutdown. 
Spotlight indexing consume a lot of CPU time on VM start because the indexing process is not finished.
Currently there is no obvious way to to find out when the indexing process ends, so 180 seconds timeout is provided.

## Check list


images/macos/provision/configuration/finalize-vm.sh
```diff
@@ -17,4 +17,9 @@ cp $HOME/image-generation/output/software-report/systeminfo.txt $HOME/image-gene
...
 
 # Clean up temporary directories
-rm -rf ~/utils ~/image-generation
\ No newline at end of file
+rm -rf ~/utils ~/image-generation
+
+# Erase all indexes and wait until the rebuilding process ends,
+# for now there is no way to get status of indexing process, it takes around 3 minutes to accomplish
+sudo mdutil -E /
+sleep 180
```
maxim-lobanov, line 25: I don't think that it is good idea to use hardcoded 3 minutes.
What if we add more tools tomorrow, and it will take 4 minutes?
What if VM performance degrade tomorrow and it will take 10 minutes?

maxim-lobanov (approved):

AlenaSviridenko (approved):

al-cheb (approved):

miketimofeev (approved):

PR #1761, opened 2020-10-07
Disable automatic updates to avoid apt lock issue.

Darleev:
# Description
Improvement 
We need to implement retry logic for apt, to avoid possible dpkg lock issues

#### Related issue:
https://github.com/actions/virtual-environments/issues/1120

## Check list
- [+] Related issue / work item is attached
- [+] Tests are written (if applicable)

PR #1883, opened 2020-10-23
Workaround for apt db locked by WALinuxAgent

dsame:
# Description
Bug fixing

WALinuxAgent running on agent hosts updates omsagent with apt related tools and locks apt database for a while. All the others attempts to instal/udate/remove apt packages fail. 

The PR adds catch and retry logic by replacing the the real apt tools with custom scripts placed before the real binaries in the search path

## Check list


tranb3o0920-hub (approved):

miketimofeev (approved):

images/linux/scripts/base/apt-mock.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
```
maxim-lobanov, line 1: Could you please set correct shebang for this script?

images/linux/scripts/base/apt-mock.sh
```diff
@@ -0,0 +1,24 @@
...
+
+i=1
+while [ \$i -le 10 ];do
+  $real_tool "\$@"
+  result=\$?
+  if [ \$result -eq  0 ];then
+    break
+  else
+    sleep 5
+    echo "...retry \$i"
```
maxim-lobanov, line 18: Do we want to add more logging here? What tool / command did fail?

images/linux/scripts/base/apt-mock.sh
```diff
@@ -0,0 +1,24 @@
...
+prefix=/usr/local/bin
+
+for tool in apt apt-get apt-fast;do
+  real_tool=`which $tool`
+  cat >$prefix/$tool <<EOT
+#!/bin/sh
+
+i=1
+while [ \$i -le 10 ];do
+  $real_tool "\$@"
```
maxim-lobanov, line 12: Looks like this logic will also retry commands like `apt install tool1`. Do we really want to do it?
I think we should be more picky for apt retrying and don't retry any command blindly

images/linux/scripts/base/apt-mock.sh
dsame: No, it is obvious from the previous message issued by the real tool.The apt is already verbose there's no need to add garbage

images/linux/scripts/base/apt-mock.sh
dsame: it is intended to retry `apt install tool1` in fact. Any apt related command attempts to lock the database and fails if the updater already locked it. 

There's nothing serious affected by this retry, so there nothing to care about.

AlenaSviridenko (approved):

dmitry-shibanov (approved):

images/linux/scripts/base/apt-mock.sh
maxim-lobanov: I disagree.
Actually, impact can be huge.
You leave these files on machine after build. It means that it will be applied to all customers builds on image.
It means that all customers' builds will try to rerun any apt command 10 times instead of immediate failure that will affect customers' build duration and build cost.

maxim-lobanov: When customer uses image, they expect `apt` to work as `apt` on their local machine. These changes override their apt tool and replace it with our retry-able implementation. I don't think that it is correct behavior.

maxim-lobanov: I still don't understand why we apply this fix to absolutely all apt commands if actually only `apt clean` fails in the single place.
As I understand, @Darleev is going to fix any other random apt failures in https://github.com/actions/virtual-environments/pull/1761 so I am trying to understand if you are fixing the same issue or different one.

maxim-lobanov (changes requested):
We shouldn't replace customer's apt command

Darleev (approved):

images/linux/scripts/base/apt-mock.sh
dsame: yes, it was intentionally developed to be a workaround till this issue https://github.com/Azure/azure-linux-extensions/issues/1238 resolved. 

The problem with apt is not limited to clean task, it happens sporadically during any apt use that interferes with omsagent update.

maxim-lobanov (approved):

PR #1884, opened 2020-10-23
[macOS] Az powershell modules fix

LeonidLapshin:
# Description
Az modules installation is flaky at the moment. It uses PowershellGet module for downloading dependencies so it should be updated before actual installation

## Check list

### DIFF
--- a/images/macos/provision/core/powershell.sh
+++ b/images/macos/provision/core/powershell.sh
@@ -9,6 +9,10 @@ brew cask install powershell
 # A dummy call of `az` to initialize ~/.azure directory before the modules are installed
 az -v
 
+# Update PowerShellGet to eliminate download errors
+pwsh -Command "Install-Module -Name PowerShellGet -Force"
+pwsh -Command "Update-Module -Name PowerShellGet -Force"
+
 # Install PowerShell modules
 psModules=$(get_toolset_value '.powershellModules[].name')
 for module in ${psModules[@]}; do
--- a/images/macos/provision/core/powershell.sh
+++ b/images/macos/provision/core/powershell.sh
@@ -21,7 +21,8 @@ for module in ${psModules[@]}; do
     else
         for version in ${moduleVersions[@]}; do
             echo " - $version"
-            pwsh -command "& {[Net.ServicePointManager]::SecurityProtocol = [Net.ServicePointManager]::SecurityProtocol -bor [Net.SecurityProtocolType]::Tls12; Install-Module $module -RequiredVersion $version -Force -Scope AllUsers}"
+            tlsCommand="[Net.ServicePointManager]::SecurityProtocol = [Net.ServicePointManager]::SecurityProtocol -bor [Net.SecurityProtocolType]::Tls12"
+            pwsh -command "& {'${tlsCommand}'; Install-Module $module -RequiredVersion $version -Force -Scope AllUsers}"
         done
     fi
 done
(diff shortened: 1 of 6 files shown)


images/macos/provision/core/powershell.sh
```diff
@@ -9,6 +9,10 @@ brew cask install powershell
 # A dummy call of `az` to initialize ~/.azure directory before the modules are installed
 az -v
 
+# Update PowerShellGet to eliminate download errors
+pwsh -Command "Install-Module -Name PowerShellGet -Force"
+pwsh -Command "Update-Module -Name PowerShellGet -Force"
```
al-cheb, line 14: Why do we need right after the install Update-Module?

images/macos/provision/core/powershell.sh
```diff
@@ -9,6 +9,10 @@ brew cask install powershell
 # A dummy call of `az` to initialize ~/.azure directory before the modules are installed
 az -v
 
+# Update PowerShellGet to eliminate download errors
+pwsh -Command "Install-Module -Name PowerShellGet -Force"
```
maxim-lobanov, line 13: Can we just add it to toolset?
https://github.com/actions/virtual-environments/blob/main/images/macos/toolsets/toolset-10.15.json#L113

vsafonkin (approved):

images/macos/provision/core/powershell.sh
```diff
@@ -17,11 +17,11 @@ for module in ${psModules[@]}; do
     echo "Installing $module module"
     moduleVersions="$(get_toolset_value ".powershellModules[] | select(.name==\"$module\") | .versions[]?")"
     if [[ -z "$moduleVersions" ]];then
-        pwsh -command "& {Install-Module $module -Force -Scope AllUsers}"
+        pwsh -command "& {[Net.ServicePointManager]::SecurityProtocol = [Net.ServicePointManager]::SecurityProtocol -bor [Net.SecurityProtocolType]::Tls12; Install-Module $module -Force -Scope AllUsers}"
```
maxim-lobanov, line 20: Do we really need to invoke this command in every PowerShell session?
Can we call it once before?
```
pwsh -command "& {[Net.ServicePointManager]::SecurityProtocol = [Net.ServicePointManager]::SecurityProtocol -bor [Net.SecurityProtocolType]::Tls12"
```

LeonidLapshin: no, it should be invoked on each session that should use tls 1.2 for sure

al-cheb: No need at all:

```
PS > ${function:Save-Module}
    Begin {
        # Change security protocol to TLS 1.2
        $script:securityProtocol = [Net.ServicePointManager]::SecurityProtocol
        [Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12

        Install-NuGetClientBinaries -CallerPSCmdlet $PSCmdlet -Proxy $Proxy -ProxyCredential $ProxyCredential

        # Module names already tried in the current pipeline for InputObject parameterset
        $moduleNamesInPipeline = @()
    }
```

images/macos/provision/core/powershell.sh
```diff
@@ -17,11 +17,12 @@ for module in ${psModules[@]}; do
...
     else
         for version in ${moduleVersions[@]}; do
             echo " - $version"
-            pwsh -command "& {Install-Module $module -RequiredVersion $version -Force -Scope AllUsers}"
+            tlsCommand="[Net.ServicePointManager]::SecurityProtocol = [Net.ServicePointManager]::SecurityProtocol -bor [Net.SecurityProtocolType]::Tls12"
```
al-cheb, line 24: The same as above it's already set:

```
PS > ${function:Install-Module}
   Begin {
        # Change security protocol to TLS 1.2
        $script:securityProtocol = [Net.ServicePointManager]::SecurityProtocol
        [Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
```

LeonidLapshin: yes, true, find out that while save-module is executing, tls if forced to 1.2 and is set to default at the end of execution

PR #1885, opened 2020-10-23
[linux] Az powershell modules fix

LeonidLapshin:
# Description
Az modules installation is flaky at the moment. It uses PowershellGet module for downloading dependencies so it should be updated before actual installation

## Check list

### DIFF
--- a/images/linux/scripts/installers/azpowershell.sh
+++ b/images/linux/scripts/installers/azpowershell.sh
@@ -7,6 +7,9 @@
 # Source the helpers for use with the script
 source $HELPER_SCRIPTS/os.sh
 
+# Force Tls 1.2 for web requests
+[Net.ServicePointManager]::SecurityProtocol = [Net.ServicePointManager]::SecurityProtocol -bor [Net.SecurityProtocolType]::Tls12
+
 # List of versions
 if isUbuntu20 ; then
     versions=$(pwsh -Command '(Find-Module -Name Az).Version')
--- a/images/linux/scripts/installers/azpowershell.sh
+++ b/images/linux/scripts/installers/azpowershell.sh
@@ -7,9 +7,6 @@
 # Source the helpers for use with the script
 source $HELPER_SCRIPTS/os.sh
 
-# Force Tls 1.2 for web requests
-[Net.ServicePointManager]::SecurityProtocol = [Net.ServicePointManager]::SecurityProtocol -bor [Net.SecurityProtocolType]::Tls12
-
 # List of versions
 if isUbuntu20 ; then
     versions=$(pwsh -Command '(Find-Module -Name Az).Version')
@@ -20,7 +17,8 @@ fi
 
 # Install Azure CLI (instructions taken from https://docs.microsoft.com/en-us/cli/azure/install-azure-cli)
 for version in ${versions[@]}; do
-    pwsh -Command "Save-Module -Name Az -LiteralPath /usr/share/az_$version -RequiredVersion $version -Force"
+    pwsh -Command "[Net.ServicePointManager]::SecurityProtocol = [Net.ServicePointManager]::SecurityProtocol -bor [Net.SecurityProtocolType]::Tls12;
+    Save-Module -Name Az -LiteralPath /usr/share/az_$version -RequiredVersion $version -Force"
 done
 
 # Run tests to determine that the software installed as expected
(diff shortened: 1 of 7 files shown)


images/linux/scripts/installers/azpowershell.sh
```diff
@@ -15,6 +15,10 @@ else
     versions=$(jq -r '.azureModules[] | select(.name | contains("az")) | .versions[]' $toolset)
 fi
 
+# Update PowerShellGet to eliminate download errors
+pwsh -Command "Install-Module -Name PowerShellGet -Force"
+pwsh -Command "Update-Module -Name PowerShellGet -Force"
```
al-cheb, line 20: Why do we need right after the install Update-Module?

images/linux/scripts/installers/azpowershell.sh
```diff
@@ -15,6 +15,10 @@ else
     versions=$(jq -r '.azureModules[] | select(.name | contains("az")) | .versions[]' $toolset)
 fi
 
+# Update PowerShellGet to eliminate download errors
+pwsh -Command "Install-Module -Name PowerShellGet -Force"
```
maxim-lobanov, line 19: Why can't we use toolset?
https://github.com/actions/virtual-environments/blob/main/images/linux/toolsets/toolset-1804.json#L88

images/linux/scripts/installers/azpowershell.sh
LeonidLapshin: true, Install-module updates it to the latest version itself, I'll remove unnecessary Update-Module

images/linux/scripts/installers/azpowershell.sh
LeonidLapshin: Powershell modules installation using toolset performs on 294 string of packer template, but az modules are installed on 216. As far as I know PowerShellGet should be updated before the az installation to eliminate errors, may be I should move pwsh modules installation to the earlier stage?

images/linux/scripts/installers/azpowershell.sh
```diff
@@ -17,7 +17,8 @@ fi
 
 # Install Azure CLI (instructions taken from https://docs.microsoft.com/en-us/cli/azure/install-azure-cli)
 for version in ${versions[@]}; do
-    pwsh -Command "Save-Module -Name Az -LiteralPath /usr/share/az_$version -RequiredVersion $version -Force"
+    pwsh -Command "[Net.ServicePointManager]::SecurityProtocol = [Net.ServicePointManager]::SecurityProtocol -bor [Net.SecurityProtocolType]::Tls12;
```
maxim-lobanov, line 20: Do we really need to invoke this command in every PowerShell session?
Can we call it once before?
```
pwsh -command "& {[Net.ServicePointManager]::SecurityProtocol = [Net.ServicePointManager]::SecurityProtocol -bor [Net.SecurityProtocolType]::Tls12"
```

LeonidLapshin: no, it set tls forcing just for active session

images/linux/scripts/installers/azpowershell.sh
```diff
@@ -17,7 +17,8 @@ fi
 
 # Install Azure CLI (instructions taken from https://docs.microsoft.com/en-us/cli/azure/install-azure-cli)
 for version in ${versions[@]}; do
-    pwsh -Command "Save-Module -Name Az -LiteralPath /usr/share/az_$version -RequiredVersion $version -Force"
+    tlsCommand="[Net.ServicePointManager]::SecurityProtocol = [Net.ServicePointManager]::SecurityProtocol -bor [Net.SecurityProtocolType]::Tls12"
```
al-cheb, line 20: We don't need to set up SecurityProtocol, because it's already set to Tls123:

PS > ${function:Save-Module}:

```
 Begin {
                      # Change security protocol to TLS 1.2
                      $script:securityProtocol = [Net.ServicePointManager]::SecurityProtocol
                      [Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12

                      Install-NuGetClientBinaries -CallerPSCmdlet $PSCmdlet -Proxy $Proxy -ProxyCredential
              $ProxyCredential

                      # Module names already tried in the current pipeline for InputObject parameterset
                      $moduleNamesInPipeline = @()
 }
```

LeonidLapshin: yes, true, find out that while save-module is executing, tls if forced to 1.2 and is set to default at the end of execution

PR #1907, opened 2020-10-26
fix user on rust and brew scripts

LeonidLapshin:
# Description
Fixed post-deployment scripts on Linux. 
Scripts are executed via sudo and $USER environment variable value is "root", but these script should set permissions according to actual user, the $SUDO_USER variable is an appropriate option.

## Check list

### DIFF
--- a/images/linux/post-generation/homebrew-permissions.sh
+++ b/images/linux/post-generation/homebrew-permissions.sh
@@ -5,7 +5,7 @@
 brew_folder="/home/linuxbrew/"
 if [ -d "$brew_folder" ]; then
     brew_folder_owner=$(ls -ld $brew_folder | awk '{print $3}')
-    if [ "$USER" != "$brew_folder_owner" ]; then
-        chown "$USER":docker -R $brew_folder
+    if [ "$SUDO_USER" != "$brew_folder_owner" ]; then
+        chown "$SUDO_USER":docker -R $brew_folder
     fi
 fi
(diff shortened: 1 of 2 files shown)


images/linux/post-generation/homebrew-permissions.sh
```diff
@@ -5,7 +5,7 @@
 brew_folder="/home/linuxbrew/"
 if [ -d "$brew_folder" ]; then
     brew_folder_owner=$(ls -ld $brew_folder | awk '{print $3}')
-    if [ "$USER" != "$brew_folder_owner" ]; then
-        chown "$USER":docker -R $brew_folder
+    if [ "$SUDO_USER" != "$brew_folder_owner" ]; then
```
al-cheb, line 8: I am not sure and need to validate it. As I know we use runneradmin under provisioner process not the runner.

images/linux/post-generation/homebrew-permissions.sh
```diff
@@ -2,10 +2,13 @@
 
 # Fix permissions for Homebrew
 # https://github.com/actions/virtual-environments/issues/1568
+
 brew_folder="/home/linuxbrew/"
+homebrew_user=$(cut -d: -f1 /etc/passwd | tail -1)
```
miketimofeev, line 7: probably just call it `main_user` or something like this?

LeonidLapshin: I guess variables set to "homebrew_user"/"rust_user" to point out that "runner" or "vsts" are actually homebrew/rust users respectively and I hope it should not confuse anyone.

al-cheb (approved):

AlenaSviridenko (approved):

PR #1937, opened 2020-10-30
[Windows] Add bash wrapper from git

al-cheb:
# Description
In scope of the https://github.com/actions/virtual-environments/pull/1648 PR we removed Git internal tools from PATH. In the migration process we found an issue related to how to correct prepare environment before running commands in the bash shell.

Issue:
If we try to invoke a find command, it runs the default `C:\Windows\System32\find.exe` Windows program instead of  `C:\msys64\usr\bin\find.exe` and  `PATH` order is incorrect( `/mingw64/bin:/usr/bin` in the PATH should be before C:\Windows\System32):
```
PS > bash -c 'echo $PATH'
/c/shells:/c/Program Files/MongoDB/Server/4.4/bin:/c/aliyun-cli:/c/ProgramData/kind:/c/vcpkg:/c/cf-cli:

PS >bash -c 'where find'
C:\Windows\System32\find.exe
C:\msys64\usr\bin\find.exe

PS > bash -c 'find'
FIND: Parameter format not correct
```
We should explicitly invoke `bash` shell with the parameter `-login` to set up PATH ordering, because the `C:\msys64\etc\profile` script is not sourced if a parameter -login is not  explicitly set. That's why we need a launcher with predefined parameters.

Fix:
According to the article - https://www.msys2.org/wiki/Launchers/, we should use follow command or use a launcher:
`C:\\msys64\\usr\\bin\\env MSYSTEM=MINGW64 MSYS2_PATH_TYPE=inherit /usr/bin/bash -l -c 'find'`

We choose the  "git wrapper" launcher with predefine settings in source code by default - https://github.com/git-for-windows/MINGW-packages/tree/main/mingw-w64-git:

1. bash -login
2. MSYSTEM=MINGW64
3. MSYS2_PATH_TYPE=inherit

Invoke find command:
```
# check PATH
PS > C:\msys64\bin\bash.exe -c 'echo $PATH'
/mingw64/bin:/usr/bin:/c/Users/testAdm2/bin:/c/shells:/c/Program Files/MongoDB/Server/4.4/bin

PS > C:\msys64\bin\bash.exe -c 'where find'
C:\msys64\usr\bin\find.exe
C:\Windows\System32\find.exe

PS > C:\msys64\bin\bash.exe -c 'find /bin | tail -n 5'
/bin/zstdcat.exe
/bin/zstdgrep
/bin/zstdless
/bin/zstdmt.exe
/bin/[.exe
```

#### Related issue:
https://github.com/actions/virtual-environments-internal/issues/1397

## Check list

### DIFF
--- a/images/win/scripts/Installers/Configure-Shell.ps1
+++ b/images/win/scripts/Installers/Configure-Shell.ps1
@@ -3,8 +3,8 @@ $shellPath = "C:\shells"
 New-Item -Path $shellPath -ItemType Directory | Out-Null
 
 # sh and bash <--> C:\msys64\usr\bin\bash.exe
-New-Item -ItemType SymbolicLink -Path "$shellPath\bash.exe" -Target "C:\msys64\usr\bin\bash.exe" | Out-Null
-New-Item -ItemType SymbolicLink -Path "$shellPath\sh.exe" -Target "C:\msys64\usr\bin\sh.exe" | Out-Null
+New-Item -ItemType SymbolicLink -Path "$shellPath\bash.exe" -Target "C:\msys64\bin\bash.exe" | Out-Null
+New-Item -ItemType SymbolicLink -Path "$shellPath\sh.exe" -Target "C:\msys64\bin\sh.exe" | Out-Null
 
 # WSL is available on Windows Server 2019
 if (Test-IsWin19)
--- a/images/win/scripts/Installers/Install-Msys2.ps1
+++ b/images/win/scripts/Installers/Install-Msys2.ps1
@@ -95,4 +95,10 @@ if (Test-Path "C:\Program Files\Git\etc\ssh")
   ssh-keyscan -t rsa ssh.dev.azure.com >> "C:\Program Files\Git\etc\ssh\ssh_known_hosts"
 }
 
+# Copy bash wrapper from git
+$wrapperPath = "C:\msys64\bin"
+New-Item -Path $wrapperPath -ItemType Directory -Force | Out-Null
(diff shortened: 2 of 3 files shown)


al-cheb:
@eine, @MSP-Greg , Hey, folks.
Could you give me a piece of advice or better approach how to fix the issue?

eine:
As discussed before, adding either Git for Windows' or MSYS2's `bin` paths to the system PATH by default is looking for problems. It might work in some cases, but chances are that it will fail regularly. See https://github.com/actions/virtual-environments/pull/1648#issuecomment-703819075. If users execute PowerShell (as in your example) or cmd, I believe that no MSYS2 tool should be found. Then, it's ok that the windows program is found first in PS. Should users want to use MSYS2 tools, they should start the corresponding shell, which will setup the environment.

This is the wrapper generated by msys2/setup-msys2: https://github.com/msys2/setup-msys2/blob/master/main.js#L163-L169. So, `gitbash.exe`, `msysbash.exe` and `sh.exe` (https://github.com/actions/virtual-environments/pull/1648#issuecomment-703543679) should be replaced with `gitbash.cmd` and `msysbash.cmd`.

eine:
> We should explicitly invoke `bash` shell with the parameter `-login` to set up PATH ordering

Note that all other supported shells are actually wrapped: https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#using-a-specific-shell.

AlenaSviridenko (approved):

Issue #1939, opened 2020-10-30
Deleting files on ubuntu based runner is sometimes extremely slow, possibly applies to all disk IO


lhotari:
**Description**  

In some cases, deleting a lot of files is extremely slow on ubuntu-latest based runners. The disk IO operations seem to become slow after this point.

2 examples from apache/pulsar workflows:
"clean disk" step took 7 minutes 25 seconds
https://github.com/apache/pulsar/pull/8402/checks?check_run_id=1323968384#step:8:1
"clean disk" took 6 minutes 47 seconds
https://github.com/apache/pulsar/runs/1331699395?check_suite_focus=true

Sometimes "clean disk" completes in 26-27 seconds.

This is the "clean disk" step used in apache/pulsar workflows:
```
      - name: clean disk
        if: steps.docs.outputs.changed_only == 'no'
        run: |
          sudo swapoff -a
          sudo rm -rf /swapfile /usr/share/dotnet /usr/local/lib/android /opt/ghc
          sudo apt clean
          docker rmi $(docker images -q) -f
          df -h
```

**Virtual environments affected**  
- [x] Ubuntu 16.04
- [x] Ubuntu 18.04
- [x] Ubuntu 20.04
- [ ] macOS 10.15
- [ ] macOS 11.0
- [ ] Windows Server 2016 R2
- [ ] Windows Server 2019

**Expected behavior**  

Deleting files should be fast and disk IO should be consistent. The "clean disk" step should take less than 30 seconds to complete.

**Actual behavior**  

Deleting files is slow and "clean disk" step takes sometimes over 7 minutes. After this the disk IO seems to be very slow.

**Repro steps**  

See https://github.com/apache/pulsar workflows, for example https://github.com/apache/pulsar/blob/master/.github/workflows/ci-integration-process.yaml . History of previous runs that failed in https://github.com/apache/pulsar/actions?query=is%3Afailure .
A lot of the failed ones fail because the performance is so bad.

1. Create a workflow that contains a similar "clean disk" step as https://github.com/apache/pulsar/blob/master/.github/workflows/ci-integration-process.yaml .
2. Run multiple times, at different times of the day. Sometimes it's very slow.

lhotari:
I looked on the community forum for similar reports. I did find one about inconsistent CPU and IOPS performance, that was https://github.community/t/github-actions-have-inconsistent-cpu-and-iops/123185/7

lhotari:
[...]

PR #1942, opened 2020-10-31
[Ubuntu] Optimize Linux OS settings for better performance

lhotari:
### Motivation

The goal of this PR is to optimize Linux OS settings for better performance.

#1939 is a related issue, but this is PR is not directly addressing that issue. These changes are about general improvements for tuning Linux OS settings for CI use cases where data durability isn't the main concern.

### Modifications

* tune `vm.swappiness` to a reasonable value (`10`). Currently using the default `60`, which isn't optimal for CI environments where more RAM might be needed for running the application and dependent services. 
* tune root filesystem (/) and resource disk (/mnt) mount options for performance (ref [Azure docs about barrier=0](https://docs.microsoft.com/en-us/azure/virtual-machines/linux/optimization#adding-disks-for-size-and-performance-targets))
  * `data=writeback,commit=999999,barrier=0,nodiscard,relatime`
  * disable `discard` which is enabled by default for the root filesystem. [`discard` might have performance implications, as stated in Azure documentation](https://docs.microsoft.com/en-us/azure/virtual-machines/linux/attach-disk-portal#trimunmap-support-for-linux-in-azure): "In some cases, the discard option may have performance implications."
* Set `transparent_hugepage` boot option to `madvise` (sets `/sys/kernel/mm/transparent_hugepage/enabled=madvise`)
* Disable `fstrim.timer` and `fstrim.service` so that SSD discard/trim isn't run at all

### DIFF
--- a/images/linux/scripts/base/reboot.sh
+++ b/images/linux/scripts/base/reboot.sh
@@ -4,5 +4,8 @@
 ##  Desc:  Reboot VM
 ################################################################################
 
+echo "Update grub before reboot"
+update-grub
+
 echo "Reboot VM"
 sudo reboot
--- a/images/linux/scripts/installers/configure-environment.sh
+++ b/images/linux/scripts/installers/configure-environment.sh
@@ -30,3 +30,8 @@ echo 'vm.max_map_count=262144' | tee -a /etc/sysctl.conf
 
 # tune swappiness to 10, docs: https://www.kernel.org/doc/Documentation/sysctl/vm.txt
 echo 'vm.swappiness=10' | tee -a /etc/sysctl.conf
+
+# grub settings
+mkdir -p /etc/default/grub.d
+# configure transparent hugepages (thp) to be used when opted in with "madvise" instead of enabling by default
+echo 'GRUB_CMDLINE_LINUX="$GRUB_CMDLINE_LINUX transparent_hugepage=madvise"' >> /etc/default/grub.d/99-thp.cfg
--- a/images/linux/scripts/installers/configure-environment.sh
+++ b/images/linux/scripts/installers/configure-environment.sh
@@ -13,7 +13,7 @@ sed -i 's/ResourceDisk.Format=n/ResourceDisk.Format=y/g' /etc/waagent.conf
 sed -i 's/ResourceDisk.EnableSwap=n/ResourceDisk.EnableSwap=y/g' /etc/waagent.conf
 sed -i 's/ResourceDisk.SwapSizeMB=0/ResourceDisk.SwapSizeMB=4096/g' /etc/waagent.conf
 # Use performance optimized mount options, docs: https://www.kernel.org/doc/Documentation/filesystems/ext4.txt
-sed -i 's/ResourceDisk.MountOptions=None/ResourceDisk.MountOptions=nodiscard,barrier=0,commit=999999,data=writeback/g' /etc/waagent.conf
+sed -i 's/ResourceDisk.MountOptions=None/ResourceDisk.MountOptions=data=writeback,commit=999999,barrier=0,nodiscard,relatime/g' /etc/waagent.conf
 
 # Add localhost alias to ::1 IPv6
 sed -i 's/::1 ip6-localhost ip6-loopback/::1     localhost ip6-localhost ip6-loopback/g' /etc/hosts
@@ -35,6 +35,8 @@ echo 'vm.swappiness=10' | tee -a /etc/sysctl.conf
 mkdir -p /etc/default/grub.d
 # configure transparent hugepages (thp) to be used when opted in with "madvise" instead of enabling by default
 echo 'GRUB_CMDLINE_LINUX="$GRUB_CMDLINE_LINUX transparent_hugepage=madvise"' >> /etc/default/grub.d/99-thp.cfg
+# configure default mount options for the root filesystem
+echo 'GRUB_CMDLINE_LINUX="$GRUB_CMDLINE_LINUX rootflags=data=writeback,commit=999999,barrier=0,nodiscard,relatime"' >> /etc/default/grub.d/99-rootflags.cfg
 
 # disable fstrim.timer & fstrim.service if services exist, there's no need to run discard/trim for the disk
 systemctl disable fstrim.timer || true
--- a/images/linux/scripts/installers/configure-environment.sh
+++ b/images/linux/scripts/installers/configure-environment.sh
@@ -35,3 +35,7 @@ echo 'vm.swappiness=10' | tee -a /etc/sysctl.conf
 mkdir -p /etc/default/grub.d
 # configure transparent hugepages (thp) to be used when opted in with "madvise" instead of enabling by default
 echo 'GRUB_CMDLINE_LINUX="$GRUB_CMDLINE_LINUX transparent_hugepage=madvise"' >> /etc/default/grub.d/99-thp.cfg
+
+# disable fstrim.timer & fstrim.service if services exist, there's no need to run discard/trim for the disk
+systemctl disable fstrim.timer || true
+systemctl disable fstrim.service || true
--- a/images/linux/scripts/installers/configure-environment.sh
+++ b/images/linux/scripts/installers/configure-environment.sh
@@ -12,6 +12,8 @@ echo 'export XDG_CONFIG_HOME=$HOME/.config' | tee -a /etc/skel/.bashrc
 sed -i 's/ResourceDisk.Format=n/ResourceDisk.Format=y/g' /etc/waagent.conf
 sed -i 's/ResourceDisk.EnableSwap=n/ResourceDisk.EnableSwap=y/g' /etc/waagent.conf
 sed -i 's/ResourceDisk.SwapSizeMB=0/ResourceDisk.SwapSizeMB=4096/g' /etc/waagent.conf
+# Use performance optimized mount options, docs: https://www.kernel.org/doc/Documentation/filesystems/ext4.txt
+sed -i 's/ResourceDisk.MountOptions=None/ResourceDisk.MountOptions=nodiscard,nobarrier,commit=999999,data=writeback/g' /etc/waagent.conf
 
 # Add localhost alias to ::1 IPv6
 sed -i 's/::1 ip6-localhost ip6-loopback/::1     localhost ip6-localhost ip6-loopback/g' /etc/hosts
--- a/images/linux/scripts/installers/configure-environment.sh
+++ b/images/linux/scripts/installers/configure-environment.sh
@@ -13,7 +13,7 @@ sed -i 's/ResourceDisk.Format=n/ResourceDisk.Format=y/g' /etc/waagent.conf
 sed -i 's/ResourceDisk.EnableSwap=n/ResourceDisk.EnableSwap=y/g' /etc/waagent.conf
 sed -i 's/ResourceDisk.SwapSizeMB=0/ResourceDisk.SwapSizeMB=4096/g' /etc/waagent.conf
 # Use performance optimized mount options, docs: https://www.kernel.org/doc/Documentation/filesystems/ext4.txt
-sed -i 's/ResourceDisk.MountOptions=None/ResourceDisk.MountOptions=nodiscard,nobarrier,commit=999999,data=writeback/g' /etc/waagent.conf
+sed -i 's/ResourceDisk.MountOptions=None/ResourceDisk.MountOptions=nodiscard,barrier=0,commit=999999,data=writeback/g' /etc/waagent.conf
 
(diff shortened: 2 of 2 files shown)


lhotari:
I found some discussion about `nobarrier` in https://github.com/MicrosoftDocs/azure-docs/issues/5607#issuecomment-372851939 . The issue was originally about removing the recommendation to use `barrier=0`. However the recommendation was kept after re-evaluation. 

One detail that is interesting is that in one of the comments, it says:
> When I mount my OS disk (which has ReadWrite caching enabled) with barrier=0, I get this time down to 5-6 seconds.
However if change the cache for my OS disk (or a data disk) to be ReadOnly and then remount with barrier=0, the performance does not change at all and stays at 50 seconds.

This leads to the impression that "mount -o remount,nobarrier /" has no impact and "nobarrier" mount option should be present from the boot.

lhotari:
@maxim-lobanov PTAL , thank you

dsame (approved):

al-cheb:
Hello, @lhotari 

Thank you for your contribution.  
1. Could you please clarify how many changes are related to SSD as we use Standard HDD disk profile  for all ubuntu vms ? 
2. Please remove `eatmydata` lib, because customers can easily install it in runtime if it's necessary.

lhotari:
Thanks for the review @al-cheb . I have addressed your review comments. Please review the changes.

> * Could you please clarify how many changes are related to SSD as we use Standard HDD disk profile  for all ubuntu vms ?

Ok. I guess it's only the "elevator=noop" setting. I removed the "elevator=noop" setting. The commits in this PR have been rebased and revisited.

The changes disable discard / fstrim / unmap. I don't think that is SSD specific in Azure. This is explained in https://docs.microsoft.com/en-us/azure/virtual-machines/linux/attach-disk-portal#trimunmap-support-for-linux-in-azure:
_"Some Linux kernels support TRIM/UNMAP operations to discard unused blocks on the disk. This feature is primarily useful in standard storage to inform Azure that deleted pages are no longer valid and can be discarded, and can save money if you create large files and then delete them."_

> * Please remove `eatmydata` lib, because customers can easily install it in runtime if it's necessary.

I have removed the commit related to eatmydata.

chkimes:
> Could you please clarify how many changes are related to SSD as we use Standard HDD disk profile for all ubuntu vms ?

Ubuntu VMs run on Ephemeral VM Disks, meaning they use the local VM disk cache that is most definitely an SSD.

lhotari:
> Ubuntu VMs run on Ephemeral VM Disks, meaning they use the local VM disk cache that is most definitely an SSD.

@chkimes Thanks for confirming. In that case, do you recommend applying "elevator=noop" boot command line option? How about the changes in this PR in general? Would you be able to help reviewing this PR?

chkimes:
I'm not incredibly experienced with these Linux settings :), but I'd be happy to confirm any architecture questions that these may impact.

I would ask that for the settings that do _not_ require a reboot, could you queue some Actions runs to do before and after testing with these changes? For example, pick an open source repo like nodejs and a workflow file with mostly consistent runtime, then fork and modify the workflow file to set these settings and see how it changes? Or perhaps perform a few runs of a benchmarking tool?

lhotari:
> I'm not incredibly experienced with these Linux settings :), but I'd be happy to confirm any architecture questions that these may impact.
> 
> I would ask that for the settings that do _not_ require a reboot, could you queue some Actions runs to do before and after testing with these changes? For example, pick an open source repo like nodejs and a workflow file with mostly consistent runtime, then fork and modify the workflow file to set these settings and see how it changes? Or perhaps perform a few runs of a benchmarking tool?


 The settings with most impact (barrier=0 and data=writeback) cannot be changed without a reboot. I already tried doing that with `mount -o remount(+ options)` . The impact of disabling discard / fs-trim might be hard to reproduce since the slowdown seems to vary and it happens when the fs-trim process kicks in with certain conditions.

Would it be possible to get some help from the maintainers of this project to benchmark the changes?
I hope a Linux performance expert could provide more feedback on the changes so that we could get this PR to a resolution asap.

Btw. There is an experience report in https://github.com/MicrosoftDocs/azure-docs/issues/5607#issuecomment-444283811 about the impact of barrier=0 on Azure.

maxim-lobanov:
@lhotari , we can take a look and benchmark these changes from our side but I am not sure when we have capacity to handle it. Let's keep it in opened state until we have a chance to validate it.

lhotari:
> we can take a look and benchmark these changes from our side but I am not sure when we have capacity to handle it. Let's keep it in opened state until we have a chance to validate it.

thank you @maxim-lobanov . I'd assume that it makes sense to prioritize the handling of this PR. Builds in general do a lot of disk I/O. It might result in the best ever performance gains that any single PR has achieved. :)

lhotari:
The Azure performance optimization documentation recommends the usage of `nobarrier/barrier=0` filesystem mount option. https://docs.microsoft.com/en-us/azure/virtual-machines/linux/optimization#adding-disks-for-size-and-performance-targets

There is also a broader analysis of the `nobarrier` filesystem mount option in this article:
https://www.cockroachlabs.com/blog/2018_cloud_report/?s=09
> Why nobarrier Matters
>
> As a refresher, nobarrier is a method of writing directly to disk without guaranteeing that writes will be persisted in the same order in which they were performed. In many cases, nobarrier can offer superior performance, but at the risk of data loss. In the event of a power failure, data on disk can be corrupted causing you to lose meaningful data. In traditional deployments, nobarrier is used with battery-backed write caches. In cloud environments it is difficult to tell exactly what hardware is used, but it's also unnecessary: local SSDs do not survive a reboot of the host machine, so it is safe to set nobarrier on a cloud local SSD.
>
> We conduct most of our performance tests using nobarrier to demonstrate the best possible performance but understand that not all use cases can support this option. As a result, we tested I/O with and without nobarrier on GCP and AWS. Note, nobarrier has no impact on CPU or network testing and was not used to conduct those experiments.

@maxim-lobanov @chkimes I hope this helps validate the PR. Please let me know if reducing the scope of the PR or splitting it into multiple parts would be helpful.

miketimofeev:
@lhotari do you have a case that directly affects you? We could test at least your case until we've found something more robust.

lhotari:
> do you have a case that directly affects you? We could test at least your case until we've found something more robust.

@miketimofeev The case that directly affects me is reported as #1939 . However I don't think that's a very good test case because the build in Apache Pulsar isn't optimal at all. 

My assumption has been that it's not possible to change the barrier setting at runtime. However I'm not sure about this. I tried `mount -o remount,barrier=0 /` on my local linux env and it succeeded.
If a reboot isn't required, I guess I could try what [@chkimes suggested](https://github.com/actions/virtual-environments/pull/1942#issuecomment-733139274) to demonstrate the impact with the nodejs repository.

chkimes:
> In many cases, nobarrier can offer superior performance, but at the risk of data loss.

This is normally scary, but we force our machines to reimage if they reboot so this seems safe to me.

chkimes:
> tune vm.swappiness to a reasonable value (10). Currently using the default 60, which isn't optimal for CI environments where more RAM might be needed for running the application and dependent services.

I read a bit on the `swappiness` setting, and found a number of articles that either contradicted each other or said that this setting was tricky and not easy to understand. I'll pose the question here along with my concerns to make sure we're moving in the right direction:

If we set this swappiness value, do we expect the amount of swap usage to increase or decrease?

Cloud providers do not have the best disk transfer rates and will throttle on both IOPS and thoughput, but IOPS is the most common throttling mechanism. Getting into swap space means increased disk transfer and therefore poor performance. Anything we can do to decrease the usage of swap would be welcomed.

mont1985:
``

MaksimZhukov:
I'm closing this PR according to [this comment](https://github.com/actions/virtual-environments/issues/1939#issuecomment-751231625).

lhotari:
@LeonidLapshin I'm sorry for the late feedback on the benchmark results that were posted as [a comment to another issue](https://github.com/actions/virtual-environments/issues/1939#issuecomment-751231625). As discussed, those benchmark results were fio results for a VM running with the changes of this PR.

Changing the `barrier` mount option doesn't impact raw file i/o performance and therefore it's expected that there isn't an improvement when measured with fio. The main impact is on file metadata operations and therefore it has an impact on workloads that involve a lot of small files. This is a typical workload of many builds.
https://docs.gluster.org/en/latest/Administrator-Guide/Performance-Testing/#testing-tools lists a few tools for measuring filesystems performance. It seems that a tool called "smallfile", https://github.com/distributed-system-analysis/smallfile would be able to measure filesystem metadata operations performance and the impact of `nobarrier`/`barrier=0` mount option.

**Would you be able to measure the performance again with [smallfile](https://github.com/distributed-system-analysis/smallfile)?**

Another change that is hard to measure is the impact of the SSD TRIM operations (and the impact of `nodiscard` and disabling the scheduled fstrim.timer job), unless there is a special setup to measure it. I'd assume that it requires that a lot of files get deleted and that the fstrim.timer job kicks in. This is what might be happening sometimes with the Pulsar build when I discovered the slowness and reported #1939 . The Pulsar Github Actions workflow [deletes a lot of files to make diskspace available for the build](https://github.com/apache/pulsar/blob/9335c491f3c75cb39dd612a8fc688bbacd1f6785/.github/workflows/ci-integration-process.yaml#L69-L72). My assumption is that the delays reported in #1939 would be resolved by using `nodiscard` and disabling the `fstrim.timer` and `fstrim.service`.

**Would you be willing to accept a PR that has only the parts that disable the SSD TRIM operations (`nodiscard` mount option) and disables the `fstrim.timer`/`fstrim.service` jobs?** Do you have some concerns about making such change?

kadirselcuk (approved):

LeonidLapshin:
Hey, @lhotari!
I have tested barrier=0 and fstrim disabling/nodiscard settings with smallfiles python tool.

Each test was done 10+ times before applying PR changes and 10+ times after ("delete" test was done after each "create" test, so it was done 20+). I noticed that the disk I/O continued after test ends, so I have used a 180-second timeout between tests.

Test that i have made:
```
python /smallfiles/smallfile/smallfile_cli.py --operation create --threads 8 --file-size 128 --files 16192 --top /smallfiles/files 
```

files/sec ~1900

before:
files/sec = 1936.024344
files/sec = 1993.074399
files/sec = 1877.045824
files/sec = 1875.297552
...

after:
files/sec = 1908.713306
files/sec = 1914.975375
files/sec = 1881.322022
files/sec = 1865.764996
...
```
python /smallfiles/smallfile/smallfile_cli.py --operation create --threads 8 --files 16192 --top /smallfiles/files 
```
files/sec ~4600

before:
files/sec = 4635.323451
files/sec = 4725.304882
files/sec = 4713.044051
files/sec = 4647.851370
...

after:
files/sec = 4645.023528
files/sec = 4710.502606
files/sec = 4615.764552
files/sec = 4608.581693
...
```
python /smallfiles/smallfile/smallfile_cli.py --operation delete --threads 8 --files 16192 --top /smallfiles/files 
```
files/sec ~74776

before:
files/sec = 70731.529525
files/sec = 85484.903228
files/sec = 70126.359227
files/sec = 89344.374980
...

after:
files/sec = 72129.940228
files/sec = 86741.825683
files/sec = 73937.798046
files/sec = 87283.406204
...

As far as I can see, these results are almost identical. Could you please provide some other test samples to check?

lhotari:
> As far as I can see, these results are almost identical. Could you please provide some other test samples to check?

Thanks for testing @LeonidLapshin . I guess building some open source project would be a good test. Building the Linux kernel used to be a thing in the 90s. :) 

btw. Just interested, are there instructions how to do such experiments? 
Did you build a VM and boot it in Azure to test?
Let's say I'd like to try out different settings on a similar Azure VM that GitHub Actions uses. Are there public docs and tools/script for this?

LeonidLapshin:
Great idea @lhotari , I'll try to do some tests with open-source builds. There is actually one thing to mention about Ubuntu VMs - you should use ephemeral disks, I made these tests on Standard D4s v3 with ephemeral disks enabled, they are quite fast btw. I'll make tests and return with results here, thank you!

LeonidLapshin:
Hey, @lhotari ! 
I have run some Linux builds and do not see any significant build time changes with these settings
```
before:
real    39m55.073s
user    139m47.907s
sys     16m39.710s
---
after:
real    38m39.736s
user    135m48.327s
sys     16m33.099s
```
Sad, but as far as I can see, these changes doesn't speed up this build too.

maxim-lobanov:
As Leonid mentioned, these params don't help too.
We would like to move this thread to Discussions format. if you have any other proposals that makes sense to try in your understanding, please create [discussion](https://github.com/actions/virtual-environments/discussions) and we will try to validate it from our side.
Thank you for your proposals

PR #2021, opened 2020-11-10
Limit apt sources to the actual architectures

dsame:
# Description
Improvement

The Ubuntu mirrors http://azure.archive.ubuntu.com and http://security.ubuntu.com do not contains packages for arm architectures and this causes the error if an user attempts to apt-install packages compiled for ARM

This PR solves the problem by explicit definition the repos include only amd64 and i386 architectures and must be excluded from the search if ARM compiled packages are requested.

The result of the PR user become able to add the repos with ported packages like  this
```
deb http://ports.ubuntu.com/ubuntu-ports bionimain
deb http://ports.ubuntu.com/ubuntu-ports bioniuniverse
deb http://ports.ubuntu.com/ubuntu-ports bionimultiverse
deb http://ports.ubuntu.com/ubuntu-ports bionirestricted
```

and issue the command like `apt install -y libusb-1.0:armhf`

Without the PR the command fails because there's no `libusb-1.0:armhf` package in http://azure.archive.ubuntu.com and http://security.ubuntu.com 

## Check list


images/linux/scripts/base/apt-sources-list-arch.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash -e
+################################################################################
+##  File:  apt-sources-list-arch.sh
+##  Desc:  Limit apt sources to the actual architectures
+################################################################################
+
+sudo sed -i'' -E 's/^deb http:\/\/(azure.archive|security).ubuntu.com/deb [arch=amd64,i386] http:\/\/\1.ubuntu.com/' /etc/apt/sources.list
```
maxim-lobanov, line 7: Do we really need the separate script for one line?
I suggest to put it to [apt.sh](https://github.com/actions/virtual-environments/blob/main/images/linux/scripts/base/apt.sh) or [repos.sh](https://github.com/actions/virtual-environments/blob/main/images/linux/scripts/base/repos.sh)

images/linux/scripts/base/apt.sh
```diff
@@ -20,7 +20,15 @@ echo "APT::Get::Assume-Yes \"true\";" > /etc/apt/apt.conf.d/90assumeyes
 # Install aria2 and jq
 apt-get install aria2
 apt-get install jq
+
 # Use apt-fast for parallel downloads
 add-apt-repository -y ppa:apt-fast/stable
+
+# Limit apt sources to the actual architectures
```
maxim-lobanov, line 27: Could you please leave comment with issue link in code?
```
# Need to limit arch for default apt repos due to 
# https://github.com/actions/virtual-environments/issues/1961
```

maxim-lobanov (approved):

al-cheb (approved):

AlenaSviridenko (approved):

maxim-lobanov (approved):

PR #2043, opened 2020-11-12
Add Boost env variable to Linux docs

varunagrawal:
# Description

Add Boost environment varibale to Linux docs so it is consistent with Windows docs.

## Check list

### DIFF
--- a/images/linux/scripts/SoftwareReport/SoftwareReport.CachedTools.psm1
+++ b/images/linux/scripts/SoftwareReport/SoftwareReport.CachedTools.psm1
@@ -34,17 +34,20 @@ function Get-ToolcacheBoostVersions {
         return @()
     }
 
-    $ToolInstances = Get-CachedToolInstances -Name $Name
-    foreach ($Instance in $ToolInstances)
-    {
-        $VersionEnvVar = $Instance.Version.replace(".", "_")
-        $Instance."Environment Variable" = "BOOST_ROOT_${VersionEnvVar}"
-    }
+    $boostVersions = Get-ChildItem $toolcachePath -Name | Sort-Object { [Version]$_ }
+    $toolInstances = $boostVersions | ForEach-Object {
+        $VersionEnvVar = $_.replace(".", "_")
+        return @{
+            Version = $_
+            Architecture = "x64"
+            "Environment Variable" = "BOOST_ROOT_${VersionEnvVar}"
 
+        }
+    }
     $Content = $ToolInstances | New-MDTable -Columns ([ordered]@{
-        Version = "left";
-        Architecture = "left";
-        "Environment Variable" = "left"
+    Version = "left";
+    Architecture = "left";
+    "Environment Variable" = "left"
     })
 
     return Build-MarkdownElement -Head $Name -Content $Content
--- a/images/linux/scripts/SoftwareReport/SoftwareReport.CachedTools.psm1
+++ b/images/linux/scripts/SoftwareReport/SoftwareReport.CachedTools.psm1
@@ -46,13 +46,17 @@ function Get-ToolcacheBoostVersions {
         }
     }
     $Content = $ToolInstances | New-MDTable -Columns ([ordered]@{
-    Version = "left";
-    Architecture = "left";
-    "Environment Variable" = "left"
+        Version = "left";
+        Architecture = "left";
+        "Environment Variable" = "left"
     })
 
-    $markdown = New-MDHeader $Name -Level 4
-    $markdown += New-MDParagraph -Lines $Content
+    $markdown = ""
+
+    if ($Content.Count -gt 0) {
+        $markdown += New-MDHeader $Name -Level 4
+        $markdown += New-MDParagraph -Lines $Content
+    }
 
     return $markdown
 }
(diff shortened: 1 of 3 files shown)


AlenaSviridenko:
Hi @varunagrawal, Linux docs are auto-generated, please, do not update markdown directly.
This is how Boost section is generated for Windows https://github.com/actions/virtual-environments/blob/main/images/win/scripts/SoftwareReport/SoftwareReport.CachedTools.psm1#L1. Could you please adjust your PR to use the similar logic?

varunagrawal:
Ah I see. Sure thing.

varunagrawal:
@AlenaSviridenko my powershell is pretty bad so I expect quite a few iterations before we get this right. :slightly_smiling_face:

images/linux/scripts/SoftwareReport/SoftwareReport.CachedTools.psm1
```diff
@@ -33,7 +33,21 @@ function Get-ToolcacheBoostVersions {
     if (-not (Test-Path $toolcachePath)) {
         return @()
     }
-    return Get-ChildItem $toolcachePath -Name | Sort-Object { [Version]$_ }
+
+    $ToolInstances = Get-CachedToolInstances -Name $Name
```
maxim-lobanov, line 37: `Get-CachedToolInstances` doesn't exist on Linux.

I think we should do something like below
```pwsh
$boostVersions = Get-ChildItem $toolcachePath -Name | Sort-Object { [Version]$_ }
$toolInstances = $boostVersions | ForEach-Object {
  $VersionEnvVar = $_.replace(".", "_")
  return @{
    Version = $_
    Architecture = "x64"
    "Environment Variable" = "BOOST_ROOT_${VersionEnvVar}"
    
  }
}
$Content = $ToolInstances | New-MDTable -Columns ([ordered]@{
  Version = "left";
  Architecture = "left";
  "Environment Variable" = "left"
})

return Build-MarkdownElement -Head $Name -Content $Content
```

varunagrawal: Thanks @maxim-lobanov. I use a Linux machine so it is hard for me to test this and I am not good at Powershell scripting.

AlenaSviridenko: `Build-MarkdownElement` doesn't exist as a helper function for Linux report generation. It could be replaced with
```
$markdown = New-MDHeader $Name -Level 4
$markdown += New-MDParagraph -Lines $Content

return $markdown
```

maxim-lobanov: @varunagrawal, feel free to create an issue and we will take care about this change

AlenaSviridenko: @varunagrawal , you can create an issue in our repo to implement this change in docs, and our team will do that if it is not convenient for you to test these changes and spend time on Powershell scripting.

varunagrawal: I spent some time understanding the basic ideas behind Powershell scripting and I think I've gotten the hang of it.

Perhaps a better issue would be to refactor the code base to allow for the helper functions to be accessible across the scripts? Lots of reinventing-the-wheel happening here.

AlenaSviridenko: We investigated the possibility to create a common helpers code base for all images but didn't implement that due to a one reason:
as you can see, we are running validation with "azp run <image>" if the image-related code changed. If we used common helpers - we'd have to run all images validation on each common code update and it would take a significant amount of time. 

So here we will consider adding the same helpers functions to Linux code base as we have for Windows to make development easier.

images/linux/scripts/SoftwareReport/SoftwareReport.CachedTools.psm1
```diff
@@ -29,11 +29,32 @@ function Get-ToolcacheGoVersions {
...
+        }
+    }
+    $Content = $ToolInstances | New-MDTable -Columns ([ordered]@{
+    Version = "left";
+    Architecture = "left";
+    "Environment Variable" = "left"
+    })
+
+    $markdown = New-MDHeader $Name -Level 4
+    $markdown += New-MDParagraph -Lines $Content
```
maxim-lobanov, line 55: Almost done.
I think we don't these two lines here. Otherwise, we will have duplicate of header. Let's just `return $Content`

Currently, result looks like
image

maxim-lobanov:
@varunagrawal , feel free to create an issue in repo and we will finalize this change from our side.

images/linux/scripts/SoftwareReport/SoftwareReport.CachedTools.psm1
varunagrawal: Shouldn't we keep line 55 as `$markdown = New-MDParagraph -Lines $Content` so that the table actually renders?

varunagrawal: I figured out why this was happening, updated.

varunagrawal:
@maxim-lobanov sorry about the delay. This next change should do it.

maxim-lobanov:
Changes work as expected. I will run remain builds

al-cheb (approved):

AlenaSviridenko (approved):

PR #2047, opened 2020-11-12
[Windows] Replace npm feed to pre-cache official Ruby binaries 

al-cheb:
# Description
In scope of this PR , we  pre-cache official Ruby binaries on Windows images the same way as it done by [ruby/setup-ruby](https://github.com/ruby/setup-ruby), so [ruby/setup-ruby](https://github.com/ruby/setup-ruby) task could use them instead of downloading binaries every time.

#### Related issue:
https://github.com/actions/virtual-environments-internal/issues/1438

## Check list


al-cheb:
@MSP-Greg, @eregon, Hi, folks
Could you please review?

miketimofeev (approved):

MSP-Greg:
@al-cheb

I haven't looked thru all the code, but it looks good, and there isn't much to go wrong.  The Ruby 7z files include html documentation, which in ruby/setup-ruby we ignore.  It's ~45 M of disk space, and about 2.1k files.  The 7zip arguments we use (js) are:
```
exec.exec('7z', ['x', downloadPath, `-xr!${base}\\share\\doc`, `-o${parentDir}`], { silent: true }))
```

Thanks for helping with Ruby...

AlenaSviridenko (approved):

images/win/scripts/Installers/Install-Ruby.ps1
```diff
@@ -0,0 +1,144 @@
...
+            $sortVersions = $group | Sort-Object {[Version]$_.name.Replace("rubyinstaller-", "").Replace("-$Arch.$Extension","").Replace("-",".")}
+            $latestVersion = $sortVersions | Select-Object -Last 1
+            $versionLists[$majorMinorGroup.Name] = $latestVersion.browser_download_url
+
+        }
+        return $versionLists
+    }
+    catch
+    {
+        Write-Host "Enable to send request to the '$uri'. Error: '$_'"
```
eregon, line 32: Enable=>Unable

images/win/scripts/Installers/Install-Ruby.ps1
```diff
@@ -0,0 +1,144 @@
...
+        exit 1
+    }
+}
+
+# Most of this logic is from
+# https://github.com/MSP-Greg/actions-ruby/blob/master/lib/main.js
```
eregon, line 38: This comment is a bit old and the URL doesn't work anymore, so I'd suggest linking to https://github.com/ruby/setup-ruby/blob/master/windows.js instead

images/win/scripts/Installers/Install-Ruby.ps1
```diff
@@ -0,0 +1,144 @@
...
+        New-Item -ItemType Directory -Path $rubyVersionPath -Force | Out-Null
+
+        Write-Host "Moving Ruby '${rubyVersion}' files to '${rubyArchPath}'"
+        Move-Item -Path $tempFolder -Destination $rubyArchPath | Out-Null
+
+        Write-Host "Removing Ruby '${rubyVersion}' documentation '${rubyArchPath}\share\doc' folder"
+        Remove-Item -Path "${rubyArchPath}\share\doc" -Force -Recurse -ErrorAction Ignore
+
+        Write-Host "Installing Bundler"
+        cmd.exe /c "$rubyArchPath\bin\gem.cmd install bundler -v ""~> 2"" --no-document"
```
eregon, line 77: If ruby/setup-ruby reuses these installations in the future, `Bundler` should not be pre-installed, and the choice should be left to the user. The Bundler version used for `bundle ...` on the command line will be the latest version if there is no `Gemfile.lock` (otherwise, it will try to pick the same major version as in the Gemfile.lock).

Also Ruby 2.6 & 2.7 already come with their own Bundler as a default gem, and having Bundler 2.1.4 installed twice can lead to complications.

So, unless it's strictly required, I would leave installing Bundler to the user, which might want a specific version.

images/win/scripts/Installers/Install-Ruby.ps1
```diff
@@ -0,0 +1,144 @@
...
+
+    $rubyPath = Join-Path $env:AGENT_TOOLSDIRECTORY "/Ruby/${Version}*/${Arch}/bin"
+    $rubyDir = (Get-Item -Path $rubyPath).FullName
+
+    Write-Host "Use Ruby ${Version} as a system Ruby"
+    Add-MachinePathItem -PathItem $rubyDir | Out-Null
+    $env:Path = Get-MachinePath
+
+    # Update ruby gem to latest version
+    Invoke-Expression "gem update --system --no-document"
```
eregon, line 111: What's the motivation to update RubyGems? That might have various side effects, and make it harder for users to reproduce locally, as they typically have the version of RubyGems shipped with Ruby.

Also, `gem update --system` installs Bundler under `site_ruby`, which is rather confusing.

eregon (commented):
I don't know much PowerShell, but I commented on the extra steps.

I think installing Bundler & updating RubyGems should be avoided, if those builds are to be reused by ruby/setup-ruby. So to match the default RubyGems version that comes with Ruby (used by most users) and let users choose the Bundler version as needed.

images/win/scripts/Installers/Install-Ruby.ps1
al-cheb: Thank you. Done.

images/win/scripts/Installers/Install-Ruby.ps1
al-cheb: Thank you. Done.

images/win/scripts/Installers/Install-Ruby.ps1
al-cheb: Thank you. Removed.

images/win/scripts/Installers/Install-Ruby.ps1
al-cheb: Thank you. Removed.

eregon (approved):
Looks good to me

MSP-Greg:
> I think installing Bundler & updating RubyGems should be avoided

Apologies for missing the that.  @eregon is correct, and updating them creates a situation where the Ruby installs will change.  Once a Ruby version is included in Actions toolcache, it should remain the same.  `ruby/setup-ruby` has code to allow users to easily change/update them if they need to.

PR #2057, opened 2020-11-13
[Image.CI] Pick up a random datastore by condition

al-cheb:
There is a chance, we can face with the logical bug on condition that `$buildDatastore` value is not changed, but we already have had two or more tags , in that case we will never pick up a datastore.


images.CI/macos/select-datastore.ps1
```diff
@@ -61,7 +61,7 @@ function Select-DataStore {
     $buildDatastore = $allDatastores | Where-Object { $_.FreeSpaceGB -ge $thresholdInGb } | Where-Object {
         $vmOnDatastore = @((Get-ChildItem -Path $_.DatastoreBrowserPath).Name -notmatch "^\.").Count
         $vmOnDatastore -lt $vmCount
-    } | Select-Object -ExpandProperty Name -First 1
+    } | Get-Random | Select-Object -ExpandProperty Name
```
maxim-lobanov, line 64: 🤣

maxim-lobanov (approved):

miketimofeev (approved):

AlenaSviridenko (approved):

PR #2091, opened 2020-11-18
fix sporadic build fails

dsame:
# Description
Improvement

The `apt-get update` sporadically fails with the error output like 
```
==> azure-arm: E: Could not open file /var/lib/apt/lists/archive.ubuntu.com_ubuntu_dists_focal-backports_universe_i18n_Translation-en - open (2: No such file or directory)
==> azure-arm: E: Could not open file /var/lib/apt/lists/archive.ubuntu.com_ubuntu_dists_focal-backports_universe_binary-amd64_Packages - open (2: No such file or directory)
==> azure-arm: E: Could not open file /var/lib/apt/lists/archive.ubuntu.com_ubuntu_dists_focal-
``` 

As a workaround the apt is mocked with the script to check the output for the specific error and rerun the command after a delay

The `apt get` sporadically fails with the error output like 
```
gpg: can't connect to the agent: IPC connect call failed
```

Currently it is solved with the retry. 

## Check list

### DIFF
--- a/images/linux/scripts/base/apt-mock.sh
+++ b/images/linux/scripts/base/apt-mock.sh
@@ -17,17 +17,34 @@ while [ \$i -le 30 ];do
   result=\$?
   cat \$err >&2
   if [ \$result -eq  0 ];then
+    # no errors, continue
     rm \$err
     break
   fi
+  # some error exists, check for apt db locked
   grep -q 'Could not get lock' \$err
-  held=\$?
-  if [ \$held -ne  0 ];then
+  result=\$?
+  if [ \$result -ne  0 ];then
+    # some error exists, check for apt update is not complete
     grep -q 'Could not open file /var/lib/apt/lists' \$err
-    held=\$?
-    if [ \$held -ne  0 ];then
-      rm \$err
-      break
+    result=\$?
+    if [ \$result -ne  0 ];then
+      # some error exists, check for gpg-agent is busy or not run yet
+      grep -q 'IPC connect call failed' \$err
+      result=\$?
+      if [ \$result -ne  0 ];then
+        # some unhandled error, continue
+        rm \$err
+        break
+      # # TODO:
+      # # the delay should help with gpg-agent not ready
+      # # if it is not uncomment the folloing lines in order to
+      # # force restart gpg-agent
+      # else
+      #  pkill -9 gpg-agent
+      #  sleep 1
+      #  source <(gpg-agent --daemon)
+      fi
     fi
   fi
   rm \$err
--- a/images/linux/scripts/base/apt-mock.sh
+++ b/images/linux/scripts/base/apt-mock.sh
@@ -20,7 +20,11 @@ while [ \$i -le 30 ];do
   grep -q 'Could not get lock' \$err
   held=\$?
   if [ \$held -ne  0 ];then
-    break
+    grep -q 'Could not open file /var/lib/apt/lists' \$err
+    held=\$?
+    if [ \$held -ne  0 ];then
+      break
+    fi
   fi
   cat \$err >&2
   sleep 5
@@ -30,3 +34,4 @@ done
 EOT
   chmod +x $prefix/$tool
 done
+
--- a/images/linux/scripts/base/apt-mock.sh
+++ b/images/linux/scripts/base/apt-mock.sh
@@ -1,6 +1,5 @@
 #!/bin/bash -e
 
-set -x
 # A temporary workaround for https://github.com/Azure/azure-linux-extensions/issues/1238
 
 prefix=/usr/local/bin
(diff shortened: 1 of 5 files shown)


images/linux/scripts/base/apt-mock.sh
```diff
@@ -14,19 +16,44 @@ while [ \$i -le 30 ];do
...
+    rm \$err
     break
   fi
+  # some error exists, check for apt db locked
   grep -q 'Could not get lock' \$err
-  held=\$?
-  if [ \$held -ne  0 ];then
-    break
+  result=\$?
+  if [ \$result -ne  0 ];then
```
maxim-lobanov, line 28: It think this code should be refactored before merging. 
Currently, It contains 4 nested if to check 4 grep and it is absolutely unreadable structure already. What if we add more commands in future?

I suggest using one of the following options:
```bash
grep -q 'Could not get lock' \$err && \
grep -q 'Could not open file /var/lib/apt/lists' \$err && \
grep -q 'IPC connect call failed' \$err && \
held=\$?
if [ \$held -ne  0 ];then
  break
fi
```

Or even single grep with multipattern option:
```bash
optionA="Could not get lock"
optionB="Could not open file /var/lib/apt/lists"
optionC="IPC connect call failed"
grep -q '$optionA\|$optionB\|$optionC' \$err
held=\$?
if [ \$held -ne  0 ];then
  break
fi
```

dsame: yeah, i refactored the code, but in a bit different way. The code your suggested assumes there is the same handler 
for every error case that is not generally true. For example, currently there's no evidence the IPC problem could be solved with a delay.

images/linux/scripts/base/apt-mock.sh
```diff
@@ -14,15 +16,35 @@ while [ \$i -le 30 ];do
...
+    # apt update is not completed, needs retry
+    retry=true
+  elif grep -q 'IPC connect call failed' \$err;then
+    ## TODO:
+    ## the delay should help with gpg-agent not ready
+    ## if it is not then uncomment the folloing lines in order to
+    ## force restart gpg-agent
+    #  pkill -9 gpg-agent
+    #  sleep 1
+    #  source <(gpg-agent --daemon)
```
AlenaSviridenko, line 39: can we remove this comment or is it still waiting to be tested?

images/linux/scripts/base/apt-mock.sh
```diff
@@ -1,10 +1,12 @@
 #!/bin/bash -e
 
+set -x
```
maxim-lobanov, line 3: Why do we want to invoke this command?
It prints all script to logs, do we really want it?

images/linux/scripts/base/apt-mock.sh
```diff
@@ -14,15 +16,35 @@ while [ \$i -le 30 ];do
   err=\$(mktemp)
   $real_tool "\$@" 2>\$err
   result=\$?
-  if [ \$result -eq  0 ];then
-    break
+  cat \$err >&2
+
+  # no errors, continue
+  test \$result -eq  0 && continue
```
maxim-lobanov, line 22: Why it is not `break`?
Looks like with `continue` script will retry command even if it is finished with exit code 0

images/linux/scripts/base/apt-mock.sh
dsame: we should wait for the real life - it is not possible to be sure the simple delay solves the problem because it is not 100% reproducible

images/linux/scripts/base/apt-mock.sh
maxim-lobanov: The same for `which $tool || continue`. Let's get rid of additional output

images/linux/scripts/base/apt-mock.sh
maxim-lobanov: Also typo double space: `\$result -eq  0`

images/linux/scripts/base/apt-mock.sh
AlenaSviridenko: I'd prefer to remove it because if your fix will work - these comments will live in the code for an uncertain amount of time. 
If the fix doesn't work - we can add it in additional PR.

images/linux/scripts/base/apt-mock.sh
maxim-lobanov: Thank you, I like the your approach!

images/linux/scripts/base/apt-mock.sh
```diff
@@ -4,7 +4,8 @@
 
 prefix=/usr/local/bin
 
-for tool in apt apt-get apt-fast;do
+for tool in apt apt-get apt-fast apt-key;do
+  which $tool || continue
```
maxim-lobanov, line 8: Since we remove `set +x`, I think this like doesn't make sense too?

images/linux/scripts/base/apt-mock.sh
```diff
@@ -14,15 +15,35 @@ while [ \$i -le 30 ];do
   err=\$(mktemp)
   $real_tool "\$@" 2>\$err
   result=\$?
-  if [ \$result -eq  0 ];then
-    break
+  cat \$err >&2
+
+  # no errors, continue
+  test \$result -eq  0 && break
```
maxim-lobanov, line 21: nitpick: extra space: `-eq..0`

maxim-lobanov (approved):

AlenaSviridenko (approved):

PR #2126, opened 2020-11-24
[Ubuntu] Update CMake to 3.19.1

EwoutH:
Update CMake from 3.17.0 to 3.19.1 on the Linux/Ubuntu images. Release notes: CMake [3.18](https://cmake.org/cmake/help/latest/release/3.18.html), [3,19](https://cmake.org/cmake/help/latest/release/3.19.html).

Windows and macOS images are updated automatically.

PR #2131, opened 2020-11-25
[Ubuntu] Always install latest CMake version

### LINKED ISSUE #2129: [Ubuntu] Improvement - CMake - Install the latest version always
**Tool information**
- Tool name: CMake

Currently, version of Cmake for Ubuntu is hardcoded and we have to update it manually like https://github.com/actions/virtual-environments/pull/2126
We need to rework it to install the latest version always like we do on Windows and macOS.

**Question, Bug, or Feature?**:
Improvement

**Virtual environments affected**
- [x] Ubuntu 16.04
- [x] Ubuntu 18.04
- [x] Ubuntu 20.04
- [ ] macOS 10.15
- [ ] macOS 11.0
- [ ] Windows Server 2016 R2
- [ ] Windows Server 2019

**Can this tool be installed during the build?**

**Tool installation time in runtime**

**Are you willing to submit a PR?**

EwoutH:
Should fix #2129 and succeeds #2126 .

Thanks to @unfor19 for [this solution](https://gist.github.com/steinwaywhw/a4cd19cda655b8249d908261a62687f8#gistcomment-3459923).

Please test extensively.


images/linux/scripts/installers/cmake.sh
```diff
@@ -9,7 +9,8 @@ echo "Checking to see if the installer script has already been run"
 if command -v cmake; then
     echo "cmake is already installed"
 else
-	curl -sL https://cmake.org/files/v3.17/cmake-3.17.0-Linux-x86_64.sh -o cmakeinstall.sh \
+	VERSION=$(curl -s https://api.github.com/repos/Kitware/CMake/releases/latest | grep "tag_name" | cut -d'v' -f2 | cut -d'"' -f1)
```
maxim-lobanov, line 12: I suggest to use `jq` instead of grep here. It is more reliable way: `jq -r '.tag_name | ltrimstr("v")'`
```bash
VERSION=$(curl -s https://api.github.com/repos/Kitware/CMake/releases/latest | jq -r '.tag_name | ltrimstr("v")')
```

maxim-lobanov:
@EwoutH , Thank you for contribution!
It looks really great, I just left minor comment

images/linux/scripts/installers/cmake.sh
EwoutH: Good suggestion, thanks!

unfor19:
@EwoutH glad I could help.
Regarding the pipeline failure - 

You used **jq** and I'm not sure that it's installed on the VM/container which is running the pipeline. This is why I used grep and cut in my original solution :)

- Original Solution:
  ```
  VERSION=$(curl -s https://api.github.com/repos/${ORG_NAME}/${REPO_NAME}/releases/latest | grep "tag_name" | cut -d'v' -f2 | cut -d'"' -f1)
  ```

- Current Solution (pipeline fails):
  ```
  VERSION=$(curl -s https://api.github.com/repos/Kitware/CMake/releases/latest | jq -r '.tag_name | ltrimstr("v")')
  ```

- Adjusted Solution (pipeline should pass):
  ```
  VERSION=$(curl -s https://api.github.com/repos/Kitware/CMake/releases/latest | grep "tag_name" | cut -d'v' -f2 | cut -d'"' -f1)
  ```

maxim-lobanov:
Those failures are not related to this change. Looks like networking issues. I will rerun

unfor19:
Either way, it's better to avoid **jq** in a CI process, this way you're only dependant on **curl**.

I thought the issue is **jq** because I saw the logs of Azure Pipelines, they pointed to this specific line - https://github.com/actions/virtual-environments/pull/2131/files#annotation_629882609

EwoutH:
@maxim-lobanov It looks like jq is failing, shall I revert back to wget?

Maybe you could both expand a bit about the advantages and disadvantages of wget and jq.

maxim-lobanov:
> @maxim-lobanov It looks like jq is failing, shall I revert back to wget?
> 
> Maybe you could both expand a bit about the advantages and disadvantages of wget and jq.

As you can see in [build logs](https://github.visualstudio.com/virtual-environments/_build/results?buildId=90032&view=logs&j=011e1ec8-6569-5e69-4f06-baf193d1351e&t=486c0909-f67d-5748-80b3-bb1d531e7b43), it fails on installation of homebrew. It is not related to your changes. Nightly builds failed with the same. We will take a look and prepare the fix. 

JQ related logic works as expected (you can see in logs that cmake.sh script passes successfully and installed correct version) and we don't need to rollback it.
Solution with `grep` works only because GitHub API returns good formatted JSON response. It won't work for compressed json for example. `jq` is JSON parser that will work well with any valid json. 
We use jq [everywhere in image generation](https://github.com/actions/virtual-environments/search?q=jq) so we prefer to keep this way.

P.S. I have merged your previous PR https://github.com/actions/virtual-environments/pull/2126 just to make sure that new cmake will be included to next week deployment. This PR will be merged later and will improve installation logic.

images/linux/scripts/installers/cmake.sh
```diff
@@ -9,7 +9,8 @@ echo "Checking to see if the installer script has already been run"
 if command -v cmake; then
     echo "cmake is already installed"
 else
-	curl -sL https://cmake.org/files/v3.17/cmake-3.17.0-Linux-x86_64.sh -o cmakeinstall.sh \
+	VERSION=$(curl -s https://api.github.com/repos/Kitware/CMake/releases/latest | jq -r '.tag_name | ltrimstr("v")')
```
al-cheb, line 12: I don't think we should manually create a url link to download a script file. 
`url=$(curl -s https://api.github.com/repos/Kitware/CMake/releases/latest | jq -r '.assets[].browser_download_url | select(contains("Linux-x86_64.sh"))')`

Output:
```
$ curl -s https://api.github.com/repos/Kitware/CMake/releases/latest | jq -r '.assets[].browser_download_url | select(contains("Linux-x86_64.s
h"))'
https://github.com/Kitware/CMake/releases/download/v3.19.1/cmake-3.19.1-Linux-x86_64.sh
```

images/linux/scripts/installers/cmake.sh
```diff
@@ -9,7 +9,8 @@ echo "Checking to see if the installer script has already been run"
 if command -v cmake; then
```
al-cheb, line 9: Can we rid of the if else statement if we would like to install always the latest verison?

maxim-lobanov:
Issue with homebrew was fixed. Please resolve conflicts and @al-cheb comments and we will rerun validation

images/linux/scripts/installers/cmake.sh
EwoutH: Looks good, done!

images/linux/scripts/installers/cmake.sh
EwoutH: Good point. If an old CMake version is installed, we need to remove it first I think.

@maxim-lobanov Could you give some context about how likely it it we encounter a preinstalled CMake version? Are the images build from scratch or are the existing images updated (overwritten)?

EwoutH:
Resolved the conflicts and implemented the URL suggestion from al-cheb!

images/linux/scripts/installers/cmake.sh
al-cheb: We always build an image from scratch.

EwoutH: Is it harmful to just leave it there? Or maybe throw an error/warning instead of just a log message?

maxim-lobanov (approved):

al-cheb (approved):

miketimofeev (approved):

AlenaSviridenko (approved):

maxim-lobanov:
@EwoutH , I have merged PR. Thank you for contribution

PR #2135, opened 2020-11-26
[Ubuntu] Temporary workaround for Homebrew installation

miketimofeev:
# Description
Homebrew installation fails on Ubuntu after these changes https://github.com/Homebrew/install/pull/341
This is a temporary workaround until the initial issue is fixed in the https://github.com/Homebrew/install/pull/355

#### Related issue:
https://github.com/actions/virtual-environments-internal/issues/1520

## Check list

### DIFF
--- a/images/linux/scripts/installers/homebrew.sh
+++ b/images/linux/scripts/installers/homebrew.sh
@@ -9,7 +9,8 @@
 source $HELPER_SCRIPTS/etc-environment.sh
 
 # Install the Homebrew on Linux
-/bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install.sh)"
+# Install script is broken for CI in this PR https://github.com/Homebrew/install/pull/341, temporary use install.sh from the previous commit
+/bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/32bd2d42c6700f5497802b5989af4c23bfb5f2c9/install.sh)"
 eval $(/home/linuxbrew/.linuxbrew/bin/brew shellenv)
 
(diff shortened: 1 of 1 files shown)


AlenaSviridenko (approved):

al-cheb (approved):

maxim-lobanov (approved):

MikeMcQuaid:
> Homebrew installation fails on Ubuntu after these changes [Homebrew/install#341](https://github.com/Homebrew/install/pull/341)

I'm reverting this currently in https://github.com/Homebrew/install/pull/356.

MikeMcQuaid:
Merged. This PR should be unnecessary now.

andreineculau:
@miketimofeev (& anyone else reading this actually) could I ask how you are thinking about dynamic dependencies?
In the project that I maintain, that also depends on the homebrew installer, I basically did the same - locked homebrew/install to a commit SHA1, and will keep it even after the revert. I did similar things to the homebrew/brew and taps (e.g. homebrew/homebrew-core etc).

Short explanation: I'm totally over the edge of the spectrum that prefers running latest-and-greatest for (almost) stable dependencies like Homebrew 👏, but this aligns more with development environments (where there's usually a trigger to `brew update`) and less with automated (CI) environments where determinism trump IMO lagging behind fixes and features. Ideally I'd like all dependencies to have a stable and not-stable channels, where stable changes would lag a week/month/etc so that devs can use the beta/alpha/nightly/etc channel but automations would use the stable channel (so that fixes still reach them, as opposed to strict locking and manual labour to update). Anyways, I'm interested in hearing your thoughts if you find some minutes 🙄 . Thanks! 👋

PR #2146, opened 2020-11-27
Handle DNS error and errors in the child processes

dsame:
# Description
Bug fixing and Improvement

Аdd fix for build error
```
==> azure-arm: Provisioning with shell script: C:\agent\_work\9\s\images\linux/scripts/base/apt.sh
==> azure-arm: sudo: unable to resolve host pkrvmjd4gkmsxeh: Temporary failure in name resolution
==> azure-arm: Removed /etc/systemd/system/timers.target.wants/apt-daily.timer.
==> azure-arm: Removed /etc/systemd/system/timers.target.wants/apt-daily-upgrade.timer.
==> azure-arm: sudo: unable to resolve host pkrvmjd4gkmsxeh: Temporary failure in name resolution
```

Handle the case error has happened in the child process by removing the check for exit code and always 
grep err log

Handle the case the err log has not been created at all

## Check list

### DIFF
--- a/images/linux/scripts/base/apt-mock.sh
+++ b/images/linux/scripts/base/apt-mock.sh
@@ -13,11 +13,10 @@ i=1
 while [ \$i -le 30 ];do
   err=\$(mktemp)
   $real_tool "\$@" 2>\$err
-  result=\$?
-  cat \$err >&2
 
-  # no errors, continue
-  test \$result -eq 0 && break
+  # no errors, break the loop and continue normal flow
+  test -f \$err || break
+  cat \$err >&2
 
   retry=false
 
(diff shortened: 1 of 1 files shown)


images/linux/scripts/base/apt-mock.sh
```diff
@@ -13,11 +13,10 @@ i=1
 while [ \$i -le 30 ];do
   err=\$(mktemp)
   $real_tool "\$@" 2>\$err
-  result=\$?
-  cat \$err >&2
 
-  # no errors, continue
-  test \$result -eq 0 && break
+  # no errors, break the loop and continue normal flow
+  test -f \$err || break
```
maxim-lobanov, line 18: Is it safe to not look at $result?
The following cases are possible? :
- $result == 0 && $err != ""
- $result != 0 && $err == ""

cc: @al-cheb @miketimofeev

dsame: if even $result != 0 && $err == "" we do not handle the case (all grep-s ignored) anyway

maxim-lobanov (approved):

AlenaSviridenko (approved):

PR #1566, opened 2020-09-09
MSBuild/CL.exe fails to launch when called from a nodeJS script via Yarn 

asklar:
# Description
New tool, Bug fixing, or Improvement?   Bug

Some tools like Yarn will spawn nodeJS via CreateProcess and pass an environment block that contains duplicated environment variables, e.g:
image

Windows doesn't do any checking of environment variables being duplicated in CreateProcess so the node.exe process gets created with the two names. Then in our case, MSBuild gets launched from within node.exe, which later launches CL.exe and other build tools.

The Azure DevOps CI pipeline sets NPM_CONFIG_CACHE (all caps), while yarn will add the lowercase npm_config_cache to the environment block when a process is launched via child_process.exec()

As a result of this, we are hitting an error in our CI because MultiTaskTool is probably putting variables in a case-insensitive dictionary and doesn't expect to find the same variable twice:
C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\MSBuild\Microsoft\VC\v160\Microsoft.CppCommon.targets(375,5): error MSB6001: Invalid command line switch for "CL.exe". System.ArgumentException: Item has already been added. Key in dictionary: 'NPM_CONFIG_CACHE' Key being added: 'npm_config_cache' [D:\a\1\s\vnext\Common\Common.vcxproj]

See example run:  https://dev.azure.com/ms/react-native-windows/_build/results?buildId=107982&view=logs&j=5435db8c-d05a-5d7b-87ca-9953c4461652&t=397ce80a-0141-5d29-7c56-c592461a0308&l=4585

#### Related issue:
https://github.com/dotnet/msbuild/issues/5726

## Check list
- [ ] Related issue / work item is attached
- [ ] Tests are written (if applicable)
- [ ] Documentation is updated (if applicable)
- [ ] Changes are tested and related VM images are successfully generated

PR #2153, opened 2020-11-29
[Windows] use `npm config` instead of `npm_config_cache`

dr-js:
# Description

This commit change the way to set npm cache path from env to config file on Windows runners.

The `npm_config_cache` env, if set, will have highest priority, and harder to change.

One way to un-set this is add a global workflow env, like:
```
env:
  npm_config_cache: ''
```

This commit change the cache config to use `npm config set cache $CachePath --global`,
which should save the path under the global npmrc at: `C:\npm\prefix\etc\npmrc`,
and allow easier later reset with user/repo level `.npmrc` files.

For the record, my usage is to unify all platform's npm cache to `~/.npm/`,
then use the same cache action config on all platform to cache the folder.

## Check list

### DIFF
--- a/images/win/scripts/Installers/Install-NodeLts.ps1
+++ b/images/win/scripts/Installers/Install-NodeLts.ps1
@@ -18,9 +18,7 @@ $env:Path = Get-MachinePath
 setx npm_config_prefix $PrefixPath /M
 $env:npm_config_prefix = $PrefixPath
 
-setx npm_config_cache $CachePath /M
-$env:npm_config_cache = $CachePath
-
+npm config set cache $CachePath --global
 npm config set registry http://registry.npmjs.org/
 
 npm install -g cordova


asklar:
Please make sure this doesn't regress #1566

dr-js:
@asklar I do not think this will make #1566 worse, since the env is removed.
But I also don't know if this will affect `yarn` in your case.

dr-js:
Also on why preset env is better removed:
The windows runner will run each [step](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstepsrun) from the same job under the same process, for example compare the log of steps from this run:
https://github.com/dr-js/dr-dev/runs/1469095654?check_suite_focus=true#step:4:224
https://github.com/dr-js/dr-dev/runs/1469095654?check_suite_focus=true#step:5:198
```
    5412 |    │  │  ├─ C:\actions\runner-provisioner-Windows\provisioner.exe --agentdirectory c:...\settings\.cache (+30)
    1416 |    │  │  │  ├─ "c:\runners\2.274.2\bin\Runner.Listener.exe" run
    3292 |    │  │  │  │  └─ "c:\runners\2.274.2\bin\Runner.Worker.exe" spawnclient 1864 1876
    2080 |    │  │  │  │     ├─ \??\C:\windows\system32\conhost.exe 0x4
    4656 |    │  │  │  │     └─ "C:\Program Files\PowerShell\7\pwsh.EXE" -command ". 'D:\a\_temp...dfc0606bba.ps1'" (+27)
    7152 |    │  │  │  │        └─ C:\windows\system32\cmd.exe /c ""C:\hostedtoolcache\windows\n...dev.7 --ps tree" (+43)
    6448 |    │  │  │  │           └─ "C:\hostedtoolcache\windows\node\14.15.0\x64\\node.exe"  "...-dev.7 --ps tree (+95)
    5788 |    │  │  │  │              └─ WMIC PROCESS get Commandline,ParentProcessId,Processid
```
You'll see the same worker with pid `3292` will spawn command for each step.

So the problem with changing the env is:
- changing the global env won't work, since a "snapshot" of env is hold by the worker `3292`, it won't pick up the change
- changing the process env in the inner runner won't worker either, since it'll only last for current step and exit

To change the "snapshot" env hold by the worker (for all step of all job) can only be done with the [workflow env config](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#env),
but that won't work well for dynamically calculated env, or platform specific ones.

And with current workflow syntax, I don't see how to set persisting dynamically calculated env in one step, and be able to use that in follow up steps.

So I'd suggest use conf files instead of env for persisting global settings.

dr-js:
That's some long builds!

Not sure how to test the build result, but I think a few check should pass:
- no `npm_config_cache` set in env
- the user can change cache path with user level .npmrc or repo level .npmrc, and `npm config get cache` in corresponding path should print the most relevant setting
- there's a global npmrc file at: `C:\npm\prefix\etc\npmrc` with content `cache=C:\npm\cache` (the `$CachePath`)
- the `npm config ls` should pick up and list the content of the global npmrc and log something like:
```
; "builtin" config from C:\Program Files\nodejs\node_modules\npm\npmrc

; prefix = "C:\\Users\\Dr\\AppData\\Roaming\\npm" ; overridden by env

; "global" config from C:\npm\prefix\etc\npmrc

cache = "c:\\npm\\cache" ; NOTE: the cache path from install script

; "user" config from C:\Users\...\.npmrc

; "env" config from environment

prefix = "C:\\npm\\prefix\\" ; NOTE: the prefix path from ENV

; "cli" config from command line options

omit = []
user-agent = "npm/7.0.8 node/v15.1.0 win32 x64"

; node bin location = C:\...\node.exe
; cwd = C:\...
; HOME = C:\Users\...
; Run `npm config ls -l` to show all defaults.
```

--- --- --- ---

NOTE: the command `npm config set registry http://registry.npmjs.org/` in the install script is strange in some ways, and maybe don't set it is better:
- the [default registry](https://docs.npmjs.com/cli/v6/using-npm/config#registry) for `npm@6` and `npm@7` is HTTPS not HTTP
- this set is saved to the install user's .npmrc, but the install user may not be the runner user (`runneradmin`, from my test), so the change may not be pick up

kadirselcuk (approved):

kadirselcuk (approved):

maxim-lobanov (approved):

MaksimZhukov (approved):

miketimofeev (approved):

PR #2162, opened 2020-11-30
[Ubuntu] Add PHP 8.0

shivammathur:
# Description
- This PR adds `PHP 8.0` to the Ubuntu environments.
- PHP extension `xmlrpc` has been dropped from `PHP 8.0`, so installing it only on `PHP 7.4` and lower. 
   Reference: https://wiki.php.net/rfc/unbundle_xmlprc
- PHP extension `json` is now always enabled and does not have to be installed separately, hence has no package for PHP 8.0.
   Reference: https://wiki.php.net/rfc/always_enable_json
- Bumped PHPUnit version to PHPUnit 8.

## Check list

### DIFF
--- a/images/linux/scripts/installers/php.sh
+++ b/images/linux/scripts/installers/php.sh
@@ -14,15 +14,15 @@ apt-get update
 
 # Install PHP
 if isUbuntu16 ; then
-    php_versions="5.6 7.0 7.1 7.2 7.3 7.4"
+    php_versions="5.6 7.0 7.1 7.2 7.3 7.4 8.0"
 fi
 
 if isUbuntu18 ; then
-    php_versions="7.1 7.2 7.3 7.4"
+    php_versions="7.1 7.2 7.3 7.4 8.0"
 fi
 
 if isUbuntu20 ; then
-    php_versions="7.4"
+    php_versions="7.4 8.0"
 fi
 
 for version in $php_versions; do
@@ -44,7 +44,6 @@ for version in $php_versions; do
         php$version-imap \
         php$version-interbase \
         php$version-intl \
-        php$version-json \
         php$version-ldap \
         php$version-mbstring \
         php$version-mysql \
@@ -60,7 +59,6 @@ for version in $php_versions; do
         php$version-sybase \
         php$version-tidy \
         php$version-xml \
-        php$version-xmlrpc \
         php$version-xsl \
         php$version-zip
 
@@ -70,6 +68,10 @@ for version in $php_versions; do
 
     if [[ $version == "7.2" || $version == "7.3" ]]; then
         apt-fast install -y --no-install-recommends php$version-recode
+    fi
+
+    if [[ $version != "8.0" ]]; then
+        apt-fast install -y --no-install-recommends php$version-xmlrpc php$version-json
     fi
 done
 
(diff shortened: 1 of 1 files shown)


al-cheb (approved):

miketimofeev (approved):

AlenaSviridenko (approved):

maxim-lobanov (approved):

shivammathur:
`json` extension is always enabled on PHP 8.0 and thus,  does not have its own package. I have updated the PR. Please run the builds again.

shivammathur:
LGTM from my end. Changes related to this PR are successful.

shivammathur:
@miketimofeev Done

miketimofeev:
@shivammathur PHPUnit is missed from all Ubuntus'
image
Do we need to change something here?
https://github.com/actions/virtual-environments/blob/c45b5f5b6f8c7aadcfbe9b40fecbd80349c585d5/images/linux/scripts/installers/php.sh#L108

shivammathur:
@miketimofeev Thanks for noticing this, `PHPUnit` should be upgraded to PHPunit 8 to work with PHP 8. Should I update the PR bumping the version?

miketimofeev:
@shivammathur update this PR, please

shivammathur:
@miketimofeev Updated

maxim-lobanov:
@shivammathur , thank you for contribution!
Please pay attention that we have deployment freeze next week so these changes will be delivered only in 2 weeks

mfn:
Is it possible that the image landed today / recently?

I've started to see breakages of simply runs of php-cs-fixer which did not use anything explicit to require a specific PHP version, relying on the default. I've also seen other software I use (e.g. phpstan) start to break:
image

Just a heads up, this will cause a small ruckus for simple workflows not explicitly pinning their PHP version down.

mfn:
Given that these images use ppa:ondrej/php the fix is easy for the simple cases not requiring a full setup-php: explicitly specify `phpX.Y`, e.h. `php7.4 ./composer.phar …` in my case.

aszenz:
> Is it possible that the image landed today / recently?
> 
> I've started to see breakages of simply runs of php-cs-fixer which did not use anything explicit to require a specific PHP version, relying on the default. I've also seen other software I use (e.g. phpstan) start to break:
> image
> 
> Just a heads up, this will cause a small ruckus for simple workflows not explicitly pinning their PHP version down.

Thanks, just found the cause for our ci failures. php 8.0, guess we should have specified the version no

shivammathur:
@mfn, I would suggest using `setup-php`. even in cases where PHP is pre-installed, all `setup-php` does by default is switch the version and update composer. It would only take 1 to 2 seconds to finish.

mfn:
@shivammathur I understand, and thank your for you great work on setup-php 🙏 , but for simple cases it's not necessary and an explicit `php7.4` in front is sufficient

maxim-lobanov:
> Is it possible that the image landed today / recently?

@mfn , Yes, it is correct. 8.0 was added to images side by side with php 7.x and these changes were deployed recently.
`php` symlink was switched to point to latest php 8.0 but PHP 7.4 is still available as `php7.4`.

It is recommended to always specify version as `phpX.Y`. Even if you need PHP 8.0 for now and `php` works for you, it is better to use `php8.0` to make sure that you won't be broken by future updates like 8.1, 8.2, 9.0 and etc.

[setup-php](https://github.com/shivammathur/setup-php) is good alternative that allow you to select specific PHP and additional tools + guarantee that your builds won't be broken even if some old version will be deprecated from images, it will just install it on-flight. If version is pre-installed like 7.4, it will just take 1 second to switch it.

PR #2175, opened 2020-12-01
[Ubuntu] Additional verification for container's tools

Darleev:
# Description
Improvement
Currently, there are no checks in containers.sh file
https://github.com/actions/virtual-environments/blob/main/images/linux/scripts/installers/containers.sh
Thus we have to wait for 3 hours for these tools to be checked.

#### Related issue:
https://github.com/actions/virtual-environments-internal/issues/1539

## Check list

### DIFF
--- a/images/linux/scripts/installers/containers.sh
+++ b/images/linux/scripts/installers/containers.sh
@@ -15,17 +15,11 @@ mkdir -p /etc/containers
 echo -e "[registries.search]\nregistries = ['docker.io', 'quay.io']" | tee /etc/containers/registries.conf
 
 # Check that tools are installed properly
-if ! podman --version; then
-    echo "Podman was not installed"
-    exit 1
-fi
-
-if ! buildah --version; then
-    echo "Buildah was not installed"
-    exit 1
-fi
-
-if ! skopeo --version; then
-    echo "Skopeo was not installed"
-    exit 1
-fi
+cmd_packages=(podman buildah skopeo)
+echo "Testing to make sure that script performed as expected, and basic scenarios work" 
+for cmd in ${cmd_packages[@]}; do 
+    if ! command -v $cmd; then 
+        echo "$cmd was not installed" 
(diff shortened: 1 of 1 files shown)


images/linux/scripts/installers/containers.sh
```diff
@@ -12,3 +13,8 @@ apt-get update -qq
 apt-get -qq -y install podman buildah skopeo
 mkdir -p /etc/containers
 echo -e "[registries.search]\nregistries = ['docker.io', 'quay.io']" | tee /etc/containers/registries.conf
+
+# Check that tools are installed properly
+podman --version
+buildah --version
+skopeo --version
```
miketimofeev, line 20: I suggest changing it to another approach, which is implemented in many other scripts
```
if ! podman --version; then
    echo "Podman was not installed"
    exit 1
fi
```

Darleev: Thank you for the suggestion, I've changed these lines.

images/linux/scripts/installers/containers.sh
```diff
@@ -15,6 +15,17 @@ mkdir -p /etc/containers
...
+
+if ! buildah --version; then
+    echo "Buildah was not installed"
+    exit 1
+fi
+
+if ! skopeo --version; then
+    echo "Skopeo was not installed"
+    exit 1
+fi
```
miketimofeev, line 31: It's even better to do it through for loop like it's done in the basics.sh 
`https://github.com/actions/virtual-environments/blob/906e2ab86558ca8bbc0087057fa9baa0fd8691f6/images/linux/scripts/installers/basic.sh#L15-L22

Darleev: Thank you, done.

al-cheb (approved):

AlenaSviridenko (approved):

AlenaSviridenko (approved):

PR #2217, opened 2020-12-08
[macOS] Install Python2 from pkg installer

miketimofeev:
# Description
Due to maintenance concerns, we are going to stop installing Python 2.7 from the Homebrew formula, which was deprecated at the beginning of 2020, and instead, use the official macOS pkg installer available by the following link
https://www.python.org/ftp/python/2.7.18/python-2.7.18-macosx10.9.pkg
This package installs pip along with the python, but the version is outdated so the update required. Also, the GUI installation process requires to run the install certificates command after successful installation:

image

The corresponding step added to the installation script.

#### Related issue:
https://github.com/actions/virtual-environments-internal/issues/1543

## Check list

### DIFF
--- a/images/macos/provision/core/python.sh
+++ b/images/macos/provision/core/python.sh
@@ -7,11 +7,14 @@ echo "Installing Python Tooling"
 echo "Brew Installing Python 3"
 brew install python@3.9
 
-echo "Brew Installing Python 2"
-# Create local tap with formula due to python2 formula depreciation
-brew tap-new --no-git local/python2
-FORMULA_PATH=$(brew extract python@2 local/python2 | grep "Homebrew/Library/Taps")
-brew install $FORMULA_PATH
+echo "Install latest Python 2"
+Python2Url="https://www.python.org/ftp/python/2.7.18/python-2.7.18-macosx10.9.pkg"
+download_with_retries $Python2Url "/tmp" "python2.pkg"
+sudo installer -pkg /tmp/python2.pkg -target /
+pip install --upgrade pip
+
+echo "Install Python2 certificates"
+bash -c "/Applications/Python\ 2.7/Install\ Certificates.command"
 
 echo "Installing pipx"
 export PIPX_BIN_DIR=/usr/local/opt/pipx_bin


images/macos/provision/core/python.sh
```diff
@@ -7,11 +7,14 @@ echo "Installing Python Tooling"
...
-FORMULA_PATH=$(brew extract python@2 local/python2 | grep "Homebrew/Library/Taps")
-brew install $FORMULA_PATH
+echo "Install latest Python 2"
+Python2Url="https://www.python.org/ftp/python/2.7.18/python-2.7.18-macosx10.9.pkg"
+download_with_retries $Python2Url "/tmp" "python2.pkg"
+sudo installer -pkg /tmp/python2.pkg -target /
+pip install --upgrade pip
+
+echo "Install Python2 certificates"
+bash -c "/Applications/Python\ 2.7/Install\ Certificates.command"
```
miketimofeev, line 17: @maxim-lobanov @al-cheb I'm a bit worried about this step even though it's recommended

al-cheb (approved):

images/macos/provision/core/python.sh
miketimofeev: Looks good
```
>>> import requests
>>> 
>>> urls = ["https://www.google.com","https://www.ebay.com/broken-link",
... "https://notawebsite_broken"]
>>> 
>>> for u in urls:
...     try:
...         r = requests.get(u)
...         print "{} {}".format(u,r.status_code)
...     except Exception,e:
...         print "{} {}".format(u,e)
... 
https://www.google.com 200
https://www.ebay.com/broken-link 404
https://notawebsite_broken HTTPSConnectionPool(host='notawebsite_broken', port=443): Max retries exceeded with url: / (Caused by NewConnectionError('<urllib3.connection.HTTPSConnection object at 0x7f9355883cd0>: Failed to establish a new connection: [Errno 8] nodename nor servname provided, or not known',))
```

AlenaSviridenko (approved):

Issue #2228, opened 2020-12-10
Update/Add CMake 3.18.1 in Android SDK


atsushieno:
**Tool information**
- Tool name: CMake 3.18.1 in Android SDK components
- Tool license: Apache2
- Add or update?: maybe it is an "update" ?
- Desired version: it is fixed to 3.18.1 (provided by Android SDK)
- Approximate size: 50MB-ish in uncompressed state

**Area for Triage**: Android

**Question, Bug, or Feature?**: Feature

**Virtual environments affected**
- [x] Ubuntu 16.04
- [x] Ubuntu 18.04
- [x] Ubuntu 20.04
- [ ] macOS 10.15
- [ ] macOS 11.0
- [x] Windows Server 2016 R2
- [x] Windows Server 2019

**Can this tool be installed during the build?**

No on Linux. Maybe no on Windows. Maybe yes on macOS (CMakes are not installed there on existing images either).

**Tool installation time in runtime**

no right idea.

**Are you willing to submit a PR?**

I come up with a PR.

atsushieno:
Additional Information: CMake 3.18 is actually REQUIERD at the moment wherever the Android project builds native libs and it does not specify CMake version.

The related Android issue I reported: https://issuetracker.google.com/issues/175241013

atsushieno:
I wrote "no" to "can be installed during the build" question, but I'm not sure about what this actually means - if we write some tricky setup scripts then it is maybe doable. It cannot be automatically done in standard manner, hence my general answer.

maxim-lobanov:
Closing related issue according to https://github.com/actions/virtual-environments/pull/2229#issuecomment-743186160

ggfan:
CMake 3.18.1 is in stable channel, may you kindly add it into the virtual environment?

The installation command seems to work:
   `echo y | $ANDROID_HOME/cmdline-tools/latest/bin/sdkmanager "cmake;3.18.1"`

miketimofeev:
@ggfan could you please open a new issue?

ggfan:
done: https://github.com/actions/virtual-environments/issues/3321
Thank you!

PR #2229, opened 2020-12-10
Add CMake 3.18 in Android SDK components to toolsets (win/linux).

atsushieno:
# Description

This adds CMake 3.18.1 from Android SDK. The relevant information is at issue #2228.

#### Related issue:

https://github.com/actions/virtual-environments/issues/2228

### DIFF
--- a/images/linux/toolsets/toolset-1604.json
+++ b/images/linux/toolsets/toolset-1604.json
@@ -91,6 +91,7 @@
         "additional_tools": [
             "cmake;3.6.4111459",
             "cmake;3.10.2.4988404",
+            "cmake;3.18.1",
             "patcher;v4",
             "ndk-bundle",
             "platform-tools"
--- a/images/linux/toolsets/toolset-1804.json
+++ b/images/linux/toolsets/toolset-1804.json
@@ -87,6 +87,7 @@
         "additional_tools": [
             "cmake;3.6.4111459",
             "cmake;3.10.2.4988404",
+            "cmake;3.18.1",
             "patcher;v4",
             "ndk-bundle",
             "platform-tools"
--- a/images/linux/toolsets/toolset-2004.json
+++ b/images/linux/toolsets/toolset-2004.json
@@ -70,6 +70,7 @@
         ],
         "additional_tools": [
             "cmake;3.10.2.4988404",
+            "cmake;3.18.1",
             "patcher;v4",
             "ndk-bundle",
             "platform-tools"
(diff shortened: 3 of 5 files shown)


maxim-lobanov (approved):

al-cheb:
@atsushieno, All images failed with the exception -  ` Warning: Failed to find package cmake;3.18.1. Could you please check a package name?

atsushieno:
According to view-source:file:///tmp/mozilla_atsushi0/repository2-1.xml the version ID is correct, but since it requires android-sdk-preview-license, it is likely failing your image builder internals.

Since the build failure message has almost no information, I have no further clue at the moment. Is there any way I can try to run the image builders locally?

al-cheb:
> According to view-source:file:///tmp/mozilla_atsushi0/repository2-1.xml the version ID is correct, but since it requires android-sdk-preview-license, it is likely failing your image builder internals.
> 
> Since the build failure message has almost no information, I have no further clue at the moment. Is there any way I can try to run the image builders locally?

I think you can test installation via GitHub Actions.

atsushieno:
Okay, now I read through the docs. Looks like I need my own azure account subscription to build images either from local console or actions anyways...

atsushieno:
Sorry I dump the idea, I don't want to expense by building potentially huge ARM image just to investigate the problem (I have almost zero knowledge on them). That should not be my business. I stick to things that I can contribute without paying.

miketimofeev:
@atsushieno it looks like the package is available as a preview only. We only install packages from the stable channel.

atsushieno:
> We only install packages from the stable channel.

Fair enough.

maxim-lobanov:
I am closing this PR for now since as @miketimofeev mentioned we use only stable channel on images. Please consider installing cmake 3.18 from preview channel in runtime.
As soon as it will be added to stable channel, we will be happy to accept this contribution.

atsushieno:
I think you are misunderstanding the actual consequence that this problem leads to.

> Please consider installing cmake 3.18 from preview channel in runtime.

Unfortunately it does not work, as GitHub Actions setup (that it installs Android SDK under /usr/local) leads to no writable permission error. It is just that GitHub Actions won't be usable for any Android applications that depends on native code and uses Android Gradle Plugin 7.0.0. There is another workaround for the issue though.

maxim-lobanov:
I have double checked and currently `cmake 3.18` is available only in canary channel (`sdkmanager --list --channel=3`).
I don't have much experience with Android development but not sure that it is expected that some software / libraries / plugins depend on non-stable sdk tools.
We don't bake tools from "canary" channel into images.

Also, I have checked and cmake 3.18 is installed without any problems in runtime:
```
steps:
      - run: echo "yes" | $ANDROID_HOME/tools/bin/sdkmanager "cmake;3.18.1" --channel=3
```
See results in build: https://github.com/maxim-lobanov/test-repo/runs/1537169598?check_suite_focus=true

PR #2147, opened 2020-11-27
[macOS] Clarify clang versions

LeonidLapshin:
# Description
Currently only the brew installation of clang is documented. This PR adds the clang version from Xcode installation to the appropriate docs section.

## Check list
- [ ] Related issue / work item is attached
- [ ] Tests are written (if applicable)
- [ ] Documentation is updated (if applicable)
- [ ] Changes are tested and related VM images are successfully generated

PR #2254, opened 2020-12-12
Looks like I get Clang 12 on macos-10.15

danger89:
Looks like I get Clang v12.0.0 on the macosx-10.15 image, right?

Regards,

Melroy

### DIFF
--- a/images/macos/macos-10.15-Readme.md
+++ b/images/macos/macos-10.15-Readme.md
@@ -13,7 +13,7 @@
 
 ## Installed Software
 ### Language and Runtime
-- Clang/LLVM 11.0.0
+- Clang/LLVM 12.0.0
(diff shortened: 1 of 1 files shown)


maxim-lobanov:
Hello @danger89 , this documentation file is auto-generated. No need to update it manually.
The issue that you have described is miss-understanding between two instances of clang installed on machine: default clang and brew clang.
We have already fixed it in https://github.com/actions/virtual-environments/pull/2147 and next week new correct docs should be deployed.
For information, it should look the following:
```
- Clang/LLVM 12.0.0 is default
- Clang/LLVM 11.0.0 is available on `$(brew --prefix llvm)/bin/clang`
```

danger89:
Ah oké, thanks for your explanation.

In that case, I'll close this pr.

PR #2256, opened 2020-12-14
Azpagent 20201207

jamesbrewster-whs:
# Description

## Check list


maxim-lobanov:
@jamesbrewster-whs , Azure Pipelines agent must not be installed as a part of image-generation.

Also, I see that your PR is mix from different changes (related to image-generation, image-generation scripts, docs and etc). We won't be able to accept it in this view.

Please create issues for every type of changes for additional discussion before submitting PR.

PR #2261, opened 2020-12-14
[macOS] change Invoke-Pester function

Darleev:
# Description
Improvement
We need to cover all installation scripts with Pester tests approach.

#### Related issue:
https://github.com/actions/virtual-environments-internal/issues/1600
## Check list


images/macos/provision/core/powershell.sh
```diff
@@ -35,5 +35,4 @@ sudo ln -s /usr/local/bin/pwsh /usr/local/bin/powershell
 # fix ~/.azure directory permissions
 sudo chown -R ${USER}: $HOME/.azure
 
-# Import Pester Module
-Import-Module Pester
\ No newline at end of file
+pwsh -c "Invoke-PesterTests -TestFile \"Powershell\""
```
maxim-lobanov, line 38: I don't think it will work. `Invoke-PesterTests` function should be imported before using.
Could you please also wrap it to some bash helper function?

images/macos/helpers/Tests.Helpers.psm1
```diff
@@ -1,3 +1,5 @@
+Import-Module "$PSScriptRoot/Common.Helpers.psm1"
+
 # Invokes command and validate that the exit code is 0
 function Validate-ZeroExitCode($command) {
     $result = Get-CommandResult $command
```
maxim-lobanov, line 5: Could you please remove this function? We should use `ShouldReturnZeroExitCode` instead of it and having two function could confuse

images/macos/helpers/Tests.Helpers.psm1
```diff
@@ -1,3 +1,5 @@
+Import-Module "$PSScriptRoot/Common.Helpers.psm1"
```
maxim-lobanov, line 1: I think you should also update YAML template and remove step for publishing test results: https://github.com/actions/virtual-environments/blob/main/images.CI/macos/azure-pipelines/image-generation.yml#L124.

Currently, your builds have warning due to it: 
image

maxim-lobanov (approved):

al-cheb (approved):

images/macos/provision/core/android-toolsets.sh
```diff
@@ -1,5 +1,6 @@
 #!/bin/bash -e -o pipefail
 source ~/utils/utils.sh
+source ~/utils/invoke-tests.sh
```
vsafonkin, line 3: There is one small function in this file, I believe it can be moved to `utils.sh`

dibir-magomedsaygitov (approved):

images/macos/provision/core/android-toolsets.sh
maxim-lobanov: We have a plan to do this trick: https://github.com/actions/virtual-environments/pull/2270/files#r544082173
In this case, it won't be needed to import it everywhere

vsafonkin: ok, thanks!

vsafonkin (approved):

images/macos/helpers/Tests.Helpers.psm1
```diff
@@ -84,4 +80,40 @@ function ShouldReturnZeroExitCode {
...
+
+    $configuration = [PesterConfiguration] @{
+        Run = @{ Path = $testPath; PassThru = $true }
+        Output = @{ Verbosity = "Detailed" }
+    }
+    if ($TestName) {
+        $configuration.Filter.FullName = $TestName
+    }
+
+    # Switch ErrorActionPreference to Stop temporary to make sure that tests will on silent errors too
```
AlenaSviridenko, line 109: typo: will **fail** (missed `fail`)

AlenaSviridenko (approved):

dmitry-shibanov (approved):

PR #2312, opened 2020-12-20
[macOS] add pester tests to install scripts part-2

Darleev:
# Description
Improvement
In order to avoid delays in macOS build, it is required to test all tools right after install.

#### Related issue:
1601 

## Check list


images/macos/provision/core/openssl.sh
```diff
@@ -10,3 +10,6 @@ brew install openssl@1.1
 
 # Symlink brew openssl@1.1 to `/usr/local/bin` as Homebrew refuses
 ln -sf $(brew --prefix openssl@1.1)/bin/openssl /usr/local/bin/openssl
+
+invoke_tests "Common" "OpenSSL"
+
```
miketimofeev, line 15: nitpick: extraline

AlenaSviridenko (approved):

maxim-lobanov (approved):

images/macos/provision/core/openssl.sh
Darleev: ty, resolved

PR #2314, opened 2020-12-21
[Ubuntu] Migrate tests for Haskell to Pester

vsafonkin:
Added Pester tests for Haskell on Ubuntu.

## Check list

### DIFF
--- a/images/linux/scripts/tests/Common.Tests.ps1
+++ b/images/linux/scripts/tests/Common.Tests.ps1
@@ -1,0 +1,13 @@
+Describe "Haskell" {
+    It "GHC" {
+        "ghc --version" | Should -ReturnZeroExitCode
+    }
+
+    It "Cabal" {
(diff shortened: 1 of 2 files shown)


images/linux/scripts/tests/Common.Tests.ps1
```diff
@@ -0,0 +1,13 @@
+Describe "Haskell" {
+    It "GHC" {
+        "ghc --version" | Should -ReturnZeroExitCode
```
maxim-lobanov, line 3: We install 3 versions of GHC, can we also validate that exactly 3 versions are installed?

vsafonkin: Done

AlenaSviridenko (approved):

maxim-lobanov (approved):

PR #2326, opened 2020-12-21
[macOS] Fix the issue with Python 2to3 symlink

miketimofeev:
# Description
2to3 is a Python program that reads Python 2.x source code and applies a series of fixers to transform it into valid Python 3.x code. Python version 2.7.18 includes this utility as well as python 3.x that comes from the brew, however, python3 installation can't overwrite the existing symlink for some reason so that we need to explicitly run `brew link --overwrite`.

#### Related issue:
https://github.com/actions/virtual-environments/issues/2322

## Check list

### DIFF
--- a/images/macos/provision/core/python.sh
+++ b/images/macos/provision/core/python.sh
@@ -4,9 +4,6 @@ source ~/utils/utils.sh
 
 echo "Installing Python Tooling"
 
-echo "Brew Installing Python 3"
-brew install python@3.9
-
 echo "Install latest Python 2"
 Python2Url="https://www.python.org/ftp/python/2.7.18/python-2.7.18-macosx10.9.pkg"
 download_with_retries $Python2Url "/tmp" "python2.pkg"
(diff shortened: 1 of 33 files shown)


images/macos/provision/core/python.sh
```diff
@@ -16,6 +13,13 @@ pip install --upgrade pip
 echo "Install Python2 certificates"
 bash -c "/Applications/Python\ 2.7/Install\ Certificates.command"
 
+# Remove /usr/local/bin/2to3 symlink created by Python2 since it conflicts with symlink from Python3
+# https://github.com/actions/virtual-environments/issues/2322
+rm -rf /usr/local/bin/2to3
```
maxim-lobanov, line 18: We shouldn't use this solution. It is unreliable and will bring more problems and maintenance in future.
What if brew updates "python@3.9" formula and add one more symlink? You will have to remove it too and it will break customers again.

maxim-lobanov: I think the correct solution will be:
- Install Python2 from package
- Install `python@3.9` from brew (with ignoring errors)
- run `brew link python@3.9`

This approach will override `2to3` symlink and any other symlinks that can be added in future.

maxim-lobanov: Something like `brew install <formula> || brew link --overwrite <formula>`

miketimofeev: @maxim-lobanov agree, that makes sense

maxim-lobanov (approved):

al-cheb (approved):

AlenaSviridenko (approved):

PR #2337, opened 2020-12-23
[Ubuntu] Migrate tests for common tools to Pester (part 3)

miketimofeev:
# Description
We should migrate tests from files to Pester:

 - git.sh
 - heroku.sh
 - hhvm.sh
 - homebrew.sh and homebrew-validate.sh (not needed anymore)
 - julia.sh
 - kind.sh (Installation moved to kubernetes-tools.sh)
 - kubernetes-tools.sh
 - leiningen.sh
 - mercurial.sh
 - miniconda.sh
 - netlify.sh
 - packer.sh
 - pollinate.sh
 - pulumi.sh
 - phantomjs.sh
 - rndgenerator.sh
 
#### Related issue:
https://github.com/actions/virtual-environments-internal/issues/1582

## Check list


images/linux/scripts/installers/pollinate.sh
```diff
@@ -4,12 +4,9 @@
 ##  Desc:  Installs Pollinate
 ################################################################################
 
+source $HELPER_SCRIPTS/invoke-tests.sh
+
 # Install Pollinate
 apt-get install -y --no-install-recommends pollinate
```
AlenaSviridenko, line 10: Maybe move it to toolsets? So it will be installed along with other apt packages in [basic.sh](https://github.com/actions/virtual-environments/blob/main/images/linux/scripts/installers/basic.sh)
And the same for `haveged`

miketimofeev: In that case, we have a few more candidates for such migration:
1. https://github.com/actions/virtual-environments/blob/7403f3318073a9a0d8ec04eca0cc11e16a3e8196/images/linux/scripts/installers/build-essential.sh#L9-L15
2. https://github.com/actions/virtual-environments/blob/7403f3318073a9a0d8ec04eca0cc11e16a3e8196/images/linux/scripts/installers/image-magick.sh#L8-L12
3. https://github.com/actions/virtual-environments/blob/7403f3318073a9a0d8ec04eca0cc11e16a3e8196/images/linux/scripts/installers/subversion.sh#L8

AlenaSviridenko: then we probably need an extra work item for moving these tools to toolsets

al-cheb (approved):

images/linux/scripts/installers/pollinate.sh
maxim-lobanov: Agree with both.
Let's log low-priority item to get rid of simple scripts where we don't need additional logic except "apt install"

images/linux/scripts/tests/Tools.Tests.ps1
```diff
@@ -119,3 +119,121 @@ Describe "gfortran" {
...
+    }
+}
+
+Describe "Julia" {
+    It "julia" {
+        "julia --version" | Should -ReturnZeroExitCode
+    }
+}
+
+Describe "Kubernetes tools" {
```
maxim-lobanov, line 165: nitpick: avoid spaces `KubernetesTools`

maxim-lobanov (approved):

images/linux/scripts/tests/Tools.Tests.ps1
miketimofeev: But why? We have a lot of such Describes with witespaces even for ubuntu:
https://github.com/actions/virtual-environments/blob/6245d40db722e18bc79096e52ae02f29838e0cf5/images/linux/scripts/tests/Tools.Tests.ps1#L43
https://github.com/actions/virtual-environments/blob/6245d40db722e18bc79096e52ae02f29838e0cf5/images/linux/scripts/tests/CLI.Tools.Tests.ps1#L2
https://github.com/actions/virtual-environments/blob/6245d40db722e18bc79096e52ae02f29838e0cf5/images/linux/scripts/tests/CLI.Tools.Tests.ps1#L14
and so on

images/linux/scripts/installers/pollinate.sh
miketimofeev: Done:
https://github.com/actions/virtual-environments-internal/issues/1636

PR #2339, opened 2020-12-24
[Ubuntu] Migrate Toolset tests to Pester

dibir-magomedsaygitov:
# Description
In scope of this PR we are migrating tests for Toolset tools to Pester.

## Check list


al-cheb (approved):

images/linux/scripts/installers/Install-Toolset.ps1
```diff
@@ -55,4 +57,6 @@ foreach ($tool in $tools) {
 
 chown -R "$($env:SUDO_USER):$($env:SUDO_USER)" /opt/hostedtoolcache/Python
 chown -R "$($env:SUDO_USER):$($env:SUDO_USER)" /opt/hostedtoolcache/node
-chown -R "$($env:SUDO_USER):$($env:SUDO_USER)" /opt/hostedtoolcache/go
\ No newline at end of file
+chown -R "$($env:SUDO_USER):$($env:SUDO_USER)" /opt/hostedtoolcache/go
+
+Invoke-PesterTests -TestFile "Toolset" -TestName "Toolset"
```
vsafonkin, line 62: I would suggest to run tests in `Configure-Toolset.ps1`

AlenaSviridenko: And no need to specify TestName here, it is the only test suite in the file

images/linux/scripts/tests/Toolset.Tests.ps1
```diff
@@ -0,0 +1,69 @@
+Describe "Toolset" {
+    $tools = (Get-ToolsetContent).toolcache
```
AlenaSviridenko, line 2: so in fact we are testing toolcache here. Maybe makes sense to rename file and test suite to Toolcache?

dmitry-shibanov (approved):

images/linux/scripts/installers/Install-Toolset.ps1
dibir-magomedsaygitov: Thank you for your suggestions. I would prefer to leave it as it is. We install tools in this script, so it make sense to invoke tests from here as well.

images/linux/scripts/tests/Toolset.Tests.ps1
dibir-magomedsaygitov: Thank you for your point, @AlenaSviridenko. I'd prefer to leave current naming as it is for now. Just to be consistent with other OS's.
cc: @maxim-lobanov

images/linux/scripts/installers/Install-Toolset.ps1
vsafonkin: But earlier we had actions: `install` -> `configure` -> `validate`. Now, this chain will be changed to `install` -> `validate (run tests)`-> `configure`.

dibir-magomedsaygitov: Good point, @vsafonkin, thank you! Moved test invocation to `Configure-Toolset.ps1`.

vsafonkin (approved):

images/linux/scripts/tests/Toolset.Tests.ps1
```diff
@@ -0,0 +1,69 @@
...
+                        }
+                    }
+                }
+
+                $foundVersionName = $foundVersion.name
+                if ($tool.name -eq 'PyPy')
+                {
+                    $pypyVersion = & "$foundVersionPath/bin/python" -c "import sys;print(sys.version.split('\n')[1])"
+                    $foundVersionName = "{0} {1}" -f $foundVersionName, $pypyVersion
+                }
```
maxim-lobanov, line 65: Let's remove these lines. Seems redundant

images/linux/scripts/tests/Toolset.Tests.ps1
```diff
@@ -0,0 +1,69 @@
...
+                    | Sort-Object -Property {[version]$_.name} -Descending `
+                    | Select-Object -First 1
+                $foundVersionPath = Join-Path $foundVersion $tool.arch
+
+                if($toolExecs) {
+                    foreach ($executable in $toolExecs["tools"]) {
+                        $executablePath = Join-Path $foundVersionPath $executable   
+    
+                        It "Validate $executable" -TestCases @{ExecutablePath = $executablePath} {
+                            $ExecutablePath | Should -Exist
```
maxim-lobanov, line 55: nitpick: spaces

maxim-lobanov (approved):

PR #2343, opened 2020-12-24
Change Android SDK root

dsame:
# Description
 Improvement

Create a hardlink `C:\Android\android-sdk` pointing to `C:\Program Files (x86)\Android\android-sdk` and set `ANDROID_SDK_ROOT` to `C:\Android\android-sdk`

Problem to solve:

ANDROID_NDK_PATH / HOME should not contain spaces. Otherwise, the script C:\Program Files (x86)\Android\android-sdk\ndk-bundle\ndk-build.cmd gives an error

```
 'C:\Program' is not recognized as an internal or external command,
operable program or batch file.
```
changing just ndk-build.cmd is not (nearly) enough. We depend on all the compiler wrapper scripts in ./toolchains/llvm/prebuilt/windows-x86_64/bin/<architecture>-clang*.cmd being capable of handling spaces, and have not gone beyond checking what else might break.

## Check list


images/win/scripts/Installers/Update-AndroidSDK.ps1
```diff
@@ -74,7 +76,7 @@ Install-AndroidSDKPackages -AndroidSDKManagerPath $sdkManager `
                           -AndroidPackages $androidToolset.additional_tools
 
 # Android NDK root path.
-$ndkRoot = "C:\Program Files (x86)\Android\android-sdk\ndk-bundle"
```
miketimofeev, line 77: Do we need to preserve this path for backward compatibility?

dsame: No, we must not. 
This path containing the spaces was the cause of the whole store and to remove it is the purpose of the PR.

In the issue there's a thread i resisted to make the change and tried to manage out the problem with fixing the google .bat files, but the client has convinced me there's no way but to move the SDK/NDK to the path without the spaces.

The those clients who hardcoded the path will still be able to use it because we have a symbolic link

AlenaSviridenko (approved):

al-cheb (approved):

miketimofeev (approved):

images/win/scripts/Installers/Update-AndroidSDK.ps1
```diff
@@ -19,15 +19,17 @@
...
 $content = [System.Convert]::FromBase64String($base64Content)
 Set-Content -Path .\android-sdk-licenses.zip -Value $content -Encoding Byte
-Expand-Archive -Path .\android-sdk-licenses.zip -DestinationPath 'C:\Program Files (x86)\Android\android-sdk' -Force
+$sdkInstallRoot = "C:\Program Files (x86)\Android\android-sdk"
+$sdkRoot = "C:\Android\android-sdk"
+Expand-Archive -Path .\android-sdk-licenses.zip -DestinationPath $sdkInstallRoot -Force
+New-Item -Path "C:\Android" -ItemType Directory
+New-Item -Path "$sdkRoot" -ItemType SymbolicLink -Value "$sdkInstallRoot"
```
MarijnS95, line 26: @dsame You mention `hardlink` in the PR description, but this is clearly a `SymbolicLink`. Shouldn't it be `HardLink` or `Junction`?

I don't know if Windows has `realpath`-alike functionality like UNIX but those resolve symbolic links, resulting in the path with spaces.

PR #2344, opened 2020-12-24
[macOS] Pester tests fail on error.

Darleev:
# Description
Improvement
Currently, pester tests do not fail on error. Need to change the logic.

#### Related issue:
-


maxim-lobanov (approved):

AlenaSviridenko (approved):

al-cheb (approved):

images/macos/tests/Xamarin.Tests.ps1
```diff
@@ -180,6 +186,25 @@ Describe "Xamarin.Android" {
                 Join-Path $xamarinLibPath "xbuild-frameworks" | Should -Exist
             }   
         }
+    }
+
+        Context "NDK toolchains" -Skip:($os.IsBigSur) {
```
maxim-lobanov, line 191: I think it should be separate describe or even separate file because Xamarin NDKs are installed in the separate test file: `xamarin-android-ndk.sh` that is run separately from `xamarin.sh`

images/macos/provision/core/openjdk.sh
```diff
@@ -26,6 +26,7 @@ createEnvironmentVariable() {
         JAVA_VERSION="1.${JAVA_VERSION}"
     fi
     local JAVA_PATH=$(/usr/libexec/java_home -v${JAVA_VERSION})
+    export ${VARIABLE_NAME}=${JAVA_PATH}
```
maxim-lobanov, line 29: Why do we need this line? Please remove it

Darleev: it is required since without it tests will fail on "JAVA_PATH" verification

images/macos/tests/Xamarin.Tests.ps1
Darleev: Thank for the good point, moved it to separate file.

images/macos/provision/core/openjdk.sh
maxim-lobanov: I think we should add `source $HOME/.bashrc` line as a part of `invoke_tests` function to reload the whole environment instead of setting specific variables.

akkim31 (approved):

PR #2270, opened 2020-12-15
[Ubuntu] Implement first Pester tests

dibir-magomedsaygitov:
# Description
In scope of this PR we implementing and adding all necessary helper logic, updating templates to run Pester tests and moving two simple tests to new approach: `7-zip.sh, azcopy.sh`

## Check list
- [x] Related issue / work item is attached
- [ ] Tests are written (if applicable)
- [ ] Documentation is updated (if applicable)
- [ ] Changes are tested and related VM images are successfully generated

PR #2349, opened 2020-12-24
[Ubuntu] Create symlink for Pester test running

vsafonkin:
Implementation of @al-cheb [suggestion](https://github.com/actions/virtual-environments/pull/2270#discussion_r544082173), added logic to create symlink in `/usr/local/bin` for test running. Symlink is removed after image generation.

## Check list


dibir-magomedsaygitov (approved):

AlenaSviridenko (approved):

images/linux/scripts/installers/configure-environment.sh
```diff
@@ -25,3 +25,7 @@ chmod -R 777 $AGENT_TOOLSDIRECTORY
 # https://www.elastic.co/guide/en/elasticsearch/reference/current/vm-max-map-count.html
 # https://www.suse.com/support/kb/doc/?id=000016692
 echo 'vm.max_map_count=262144' | tee -a /etc/sysctl.conf
+
+# Create symlink for tests running
+chmod -R 777 $HELPER_SCRIPTS/invoke-tests.sh
+ln -s $HELPER_SCRIPTS/invoke-tests.sh /usr/local/bin
```
maxim-lobanov, line 31: may be name it without `sh`? just `invoke-tests`. In this case, it will look like `invoke_tests` that is more beatiful

images/linux/scripts/installers/configure-environment.sh
```diff
@@ -25,3 +25,7 @@ chmod -R 777 $AGENT_TOOLSDIRECTORY
 # https://www.elastic.co/guide/en/elasticsearch/reference/current/vm-max-map-count.html
 # https://www.suse.com/support/kb/doc/?id=000016692
 echo 'vm.max_map_count=262144' | tee -a /etc/sysctl.conf
+
+# Create symlink for tests running
+chmod -R 777 $HELPER_SCRIPTS/invoke-tests.sh
```
maxim-lobanov, line 30: IMHO, `chmod +x` should be enough.
Also we don't need `-R` since we apply it to the single file.

images/linux/scripts/installers/configure-environment.sh
vsafonkin: Done

images/linux/scripts/installers/configure-environment.sh
vsafonkin: Fixed, thanks!

maxim-lobanov (approved):

PR #2354, opened 2020-12-25
[macOS] Fix the issue with brew install

MaksimZhukov:
# Description
Recently, installation of some tools using the `brew install` command started to fail with the following error:
```
==> vsphere-clone: Error: azure-cli: no bottle available!
==> vsphere-clone: You can try to install from source with e.g.
==> vsphere-clone:   brew install --build-from-source azure-cli
```
In scope of this PR we fixed it.

## Details
Previously, the `brew install` command downloaded and poured a package bottle first. If there were no bottles available, it fell back to building from source. This behavior was changed recently in scope of [this PR](https://github.com/Homebrew/brew/pull/9518). Currently, if no bottle is found, the `brew install` command will fail with the error we've got.

To fix this, we check bottle availability and use the `brew install --build-from-source <package>` if no bottle is found for a specific image.

## Check list


images/macos/provision/utils/utils.sh
```diff
@@ -108,4 +108,66 @@ brew_cask_install_ignoring_sha256() {
...
+        echo "$OSTYPE is unknown operating system"
+        exit 1
+    fi
+
+    local build_from_source=$(should_build_from_source "$tool_name" "$os_name")
+    if $build_from_source; then
+        echo "Bottle of the $tool_name for the $os_name was not found. Building $tool_name from source..."
+        brew install --build-from-source $tool_name
+    else
+        echo "Downloading $tool_name..."
```
maxim-lobanov, line 170: nitpick: `Downloading bottle for $tool_name...`

images/macos/provision/utils/utils.sh
```diff
@@ -108,4 +108,66 @@ brew_cask_install_ignoring_sha256() {
...
+should_build_from_source() {
+    local tool_name=$1
+    local os_name=$2
+    local tool_info=$(brew info --json=v1 $tool_name)
+    local bottle_disabled=$(echo "$tool_info" | jq ".[0].bottle_disabled")
+
+    # No need to build from source if a bottle is disabled
+    # Use the simple 'brew install' command to download a package
+    if $bottle_disabled; then
+        echo "false"
```
maxim-lobanov, line 136: Should we use `return` after echo? to avoid running function

images/macos/provision/utils/utils.sh
```diff
@@ -108,4 +108,66 @@ brew_cask_install_ignoring_sha256() {
...
+    fi
+
+    # The aws-sam-cli has the 'aws-sam-cli-*.sierra.bottle.tar.gz' bottle
+    # It can be installed on all macOS images by the simple 'brew install' command
+    if [[ "$tool_name" == "aws-sam-cli" ]]; then
```
maxim-lobanov, line 141: Discussed offline, let's try to get rid of additional condition here

vsafonkin (approved):

images/macos/provision/utils/utils.sh
```diff
@@ -108,4 +108,66 @@ brew_cask_install_ignoring_sha256() {
...
+        echo "true"
+    else
+        echo "false"
+    fi
+}
+
+# brew provides package bottles for different macOS versions
+# The 'brew install' command will fail if a package bottle does not exist
+# Use the '--build-from-source' option to build from source in this case
+brew_install() {
```
AlenaSviridenko, line 156: To me it looks strange to try to mimic the usual `brew install`, it looks unclear for customers until they dig into the code.
Maybe it worth renaming it to something like `brew_smart_install` so it will be obvious that we modified and extended classic `brew install`

images/macos/provision/utils/utils.sh
MaksimZhukov: This is not quite correct. Bottle is not downloaded if `bottle_disabled` is true

images/macos/provision/utils/utils.sh
MaksimZhukov: Good catch, thanks!

images/macos/provision/utils/utils.sh
MaksimZhukov: I reverted changes related to the `aws-sam-cli`

images/macos/provision/utils/utils.sh
MaksimZhukov: Good suggestion, thank you!

al-cheb (approved):

images/macos/provision/core/homebrew.sh
```diff
@@ -12,10 +14,10 @@ echo "Installing jq..."
 brew install jq
```
al-cheb, line 14: nitpick: brew_smart_install

maxim-lobanov (approved):

PR #2363, opened 2020-12-28
Add Android Command Line tools for MacOS images

dsame:
# Description
New tool 

size:  100Mb
time: 30 secs

Virtual environments currently installs Android "SDK Tools", which is deprecated: https://developer.android.com/studio/releases/sdk-tools

Google suggests using the new "cmdline-tools" package instead:
https://developer.android.com/studio/command-line

## Check list


AlenaSviridenko (approved):

maxim-lobanov (approved):

miketimofeev (changes requested):
@dsame the tool is currently installed only for macOS-11 judging by the software output

AlenaSviridenko (approved):

PR #2364, opened 2020-12-28
Add Android Command Line tools to Linux images

dsame:
# Description
New tool 

size:  100Mb
time: 30 secs

Virtual environments currently installs Android "SDK Tools", which is deprecated: https://developer.android.com/studio/releases/sdk-tools

Google suggests using the new "cmdline-tools" package instead:
https://developer.android.com/studio/command-line

## Check list

### DIFF
--- a/images/linux/scripts/SoftwareReport/SoftwareReport.Android.psm1
+++ b/images/linux/scripts/SoftwareReport/SoftwareReport.Android.psm1
@@ -49,6 +49,10 @@ function Build-AndroidTable {
             "Version" = Get-AndroidBuildToolVersions -PackageInfo $packageInfo
         },
         @{
+            "Package" = "Android Command Line Tools"
+            "Version" = Get-AndroidPackageVersions -PackageInfo $packageInfo -MatchedString "Android Command Line Tools"
+        },
+        @{
             "Package" = "Android emulator"
             "Version" = Get-AndroidPackageVersions -PackageInfo $packageInfo -MatchedString "Android Emulator"
         },
--- a/images/linux/scripts/SoftwareReport/SoftwareReport.Android.psm1
+++ b/images/linux/scripts/SoftwareReport/SoftwareReport.Android.psm1
@@ -50,7 +50,7 @@ function Build-AndroidTable {
         },
         @{
             "Package" = "Android Command Line Tools"
-            "Version" = Get-AndroidPackageVersions -PackageInfo $packageInfo -MatchedString "Android Command Line Tools"
+            "Version" = Get-AndroidPackageVersions -PackageInfo $packageInfo -MatchedString "Android SDK Command-line Tools"
         },
         @{
             "Package" = "Android emulator"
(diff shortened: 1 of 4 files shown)


AlenaSviridenko:
Could you please add tests for cmdline-tools?

maxim-lobanov (approved):

images/linux/scripts/installers/android.sh
```diff
@@ -73,6 +73,8 @@ availableBuildTools=$(echo ${allBuildTools[@]//*rc[0-9]/})
 filter_components_by_version $minimumPlatformVersion "${availablePlatforms[@]}"
 filter_components_by_version $minimumBuildToolVersion "${availableBuildTools[@]}"
 
+components+=" cmdline-tools;latest"
```
miketimofeev, line 76: Maybe add it to the toolset?

dsame: well, no. It was done from the very first commit, but after discussion we decided to do not pollute the image with the many obviously useless old versions and install only the latest one. As a result there's no need to have an extra setting in the toolset.json.  If i understood you right.

maxim-lobanov: I think we can add ` cmdline-tools;latest` to `additional_tools` section: https://github.com/actions/virtual-environments/blob/main/images/linux/toolsets/toolset-1804.json#L87

maxim-lobanov: it should work for some other images too

images/linux/toolsets/toolset-1604.json
```diff
@@ -93,7 +93,8 @@
             "cmake;3.10.2.4988404",
             "patcher;v4",
             "ndk-bundle",
-            "platform-tools"
+            "platform-tools",
+            "cmdline-tools;latest"
```
AlenaSviridenko, line 97: I think we don't need `latest` here, we don't have it for `ndk-bundle` but we still install the latest one, right?

maxim-lobanov: I think It is expected. In output of `sdkmanager --list`, It prints ndk-bundle as `ndk-bundle` and cmdline-tools as `cmdline-tools;latest`. Otherwise, it won't find it

dmitry-shibanov (approved):

images/linux/toolsets/toolset-1604.json
AlenaSviridenko: ok

AlenaSviridenko (approved):

images/linux/scripts/SoftwareReport/SoftwareReport.Android.psm1
```diff
@@ -48,6 +48,10 @@ function Build-AndroidTable {
             "Package" = "Android SDK Build-tools"
             "Version" = Get-AndroidBuildToolVersions -PackageInfo $packageInfo
         },
+        @{
+            "Package" = "Android Command Line Tools"
+            "Version" = Get-AndroidPackageVersions -PackageInfo $packageInfo -MatchedString "Android Command Line Tools"
```
miketimofeev, line 53: Should be 
`-MatchedString "Android SDK Command-line Tools (latest)"`

miketimofeev (changes requested):

PR #2374, opened 2020-12-30
[Ubuntu] Migrate missing tools tests to Pester

dibir-magomedsaygitov:
# Description
In scope of this PR we are migrating missing tools tests to Pester and removing test for powershellcore.
 
List of tools:
`containers.sh`
`nodejs.sh`
`nvm.sh`
`powershellcore.sh`
`python.sh`

## Check list

### DIFF
--- a/images/linux/scripts/installers/nodejs.sh
+++ b/images/linux/scripts/installers/nodejs.sh
@@ -20,11 +20,4 @@ apt-get update
 # Install yarn
 apt-get install -y --no-install-recommends yarn
 
-# Run tests to determine that the software installed as expected
-echo "Testing to make sure that script performed as expected, and basic scenarios work"
-for cmd in node grunt gulp webpack parcel yarn newman; do
-    if ! command -v $cmd; then
-        echo "$cmd was not installed"
-        exit 1
-    fi
-done
+invoke_tests "Tools" "NodeJS"
--- a/images/linux/scripts/installers/nodejs.sh
+++ b/images/linux/scripts/installers/nodejs.sh
@@ -20,4 +20,4 @@ apt-get update
 # Install yarn
 apt-get install -y --no-install-recommends yarn
 
-invoke_tests "Tools" "NodeJS"
+invoke_tests "Tools" "Node.js"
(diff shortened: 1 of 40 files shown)


maxim-lobanov (approved):

images/linux/scripts/installers/nodejs.sh
```diff
@@ -20,11 +20,4 @@ apt-get update
...
 
-# Run tests to determine that the software installed as expected
-echo "Testing to make sure that script performed as expected, and basic scenarios work"
-for cmd in node grunt gulp webpack parcel yarn newman; do
-    if ! command -v $cmd; then
-        echo "$cmd was not installed"
-        exit 1
-    fi
-done
+invoke_tests "Tools" "NodeJS"
```
AlenaSviridenko, line 23: correct name in `Node.js`

images/linux/scripts/installers/powershellcore.sh
```diff
@@ -6,10 +6,3 @@
 
 # Install Powershell
 apt-get install -y powershell
-
-# Run tests to determine that the software installed as expected
-echo "Testing to make sure that script performed as expected, and basic scenarios work"
-if ! command -v pwsh; then
```
AlenaSviridenko, line 12: Shouldn't we do `invoke_tests "Tools" "PowerShell Core"` here?

dibir-magomedsaygitov: @AlenaSviridenko we have decided to remove this test, because we use `powershellcore` in testing.

images/linux/scripts/installers/nodejs.sh
dibir-magomedsaygitov: thank you, changed to `Node.js`

images/linux/scripts/installers/powershellcore.sh
AlenaSviridenko: ok, but probably then we should remove test suite for PowerShell Core https://github.com/actions/virtual-environments/pull/2374/files#diff-566d9fbecb5003c69d8bf0a9de1f159b931a82e63c7d96acd1ab48c0f40c768cR86

because I don't see where it is called. At least in current `main`

dibir-magomedsaygitov: Thank you! Decided to bring back `pwsh` test.

miketimofeev (approved):

PR #2394, opened 2021-01-06
[Ubuntu] Fix `setEtcEnvironmentVariable` function

vsafonkin:
Function `setEtcEnvironmentVariable` adds new variable value with quotes. It cause incorrect installation for new packages via pipx on Ubuntu.

Example of installation a new package: https://buildcanary.visualstudio.com/PipelineCanary/_build/results?buildId=1314837&view=logs&j=2666b424-a65e-5d07-a841-d60507eedf54&t=30f8b51d-c83a-54b4-a539-070a66f8e560

Also, `pipx list` command not shows already installed packages.

## Check list

### DIFF
--- a/images/linux/scripts/installers/python.sh
+++ b/images/linux/scripts/installers/python.sh
@@ -29,8 +29,8 @@ if isUbuntu18 || isUbuntu20 ; then
     python3 -m pipx ensurepath
 
     # Update /etc/environment
-    setEtcEnvironmentVariable "PIPX_BIN_DIR" $PIPX_BIN_DIR
-    setEtcEnvironmentVariable "PIPX_HOME" $PIPX_HOME
+    echo "PIPX_BIN_DIR=$PIPX_BIN_DIR" | tee -a /etc/environment
+    echo "PIPX_HOME=$PIPX_HOME" | tee -a /etc/environment
     prependEtcEnvironmentPath $PIPX_BIN_DIR
 
     # Test pipx
--- a/images/linux/scripts/installers/python.sh
+++ b/images/linux/scripts/installers/python.sh
@@ -29,8 +29,8 @@ if isUbuntu18 || isUbuntu20 ; then
     python3 -m pipx ensurepath
 
     # Update /etc/environment
-    echo "PIPX_BIN_DIR=$PIPX_BIN_DIR" | tee -a /etc/environment
-    echo "PIPX_HOME=$PIPX_HOME" | tee -a /etc/environment
+    setEtcEnvironmentVariable "PIPX_BIN_DIR" $PIPX_BIN_DIR
+    setEtcEnvironmentVariable "PIPX_HOME" $PIPX_HOME
     prependEtcEnvironmentPath $PIPX_BIN_DIR
 
     # Test pipx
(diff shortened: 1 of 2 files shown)


images/linux/scripts/installers/python.sh
```diff
@@ -29,8 +29,8 @@ if isUbuntu18 || isUbuntu20 ; then
     python3 -m pipx ensurepath
 
     # Update /etc/environment
-    setEtcEnvironmentVariable "PIPX_BIN_DIR" $PIPX_BIN_DIR
-    setEtcEnvironmentVariable "PIPX_HOME" $PIPX_HOME
+    echo "PIPX_BIN_DIR=$PIPX_BIN_DIR" | tee -a /etc/environment
```
al-cheb, line 32: Could you please describe why we shouldn't use `setEtcEnvironmentVariable` function?

addEtcEnvironmentVariable $variable_name $variable_value -> echo "$variable_name=\"$variable_value\"" | sudo tee -a /etc/environment

vsafonkin: We have `PIPX_BIN_DIR="/opt/pipx_bin"` in variables instead `PIPX_BIN_DIR=/opt/pipx_bin` and when we will try to install new packages pipx shows next message:
`Note: '/home/vsts/work/1/s/"/opt/pipx_bin"' is not on your PATH environment variable.`

New package will be installed to another place, not `/opt/pipx_bin`.

maxim-lobanov: Can we just remove quotes here: https://github.com/actions/virtual-environments/blob/446d7493c14b2aaf08befb32a50ae11dcc15ed92/images/linux/scripts/helpers/etc-environment.sh#L33 ?

vsafonkin: @maxim-lobanov , I believe it will be better to fix here:
`echo "$variable_name=\"$variable_value\"" | sudo tee -a /etc/environment`

maxim-lobanov: @vsafonkin  Yep, let's try

vsafonkin: Fixed

al-cheb (approved):

images/linux/scripts/installers/python.sh
maxim-lobanov: cc: @dsame to review this PR

dmitry-shibanov (approved):

images/linux/scripts/installers/python.sh
dsame: @maxim-lobanov , i've just double checked, the quotes in `/etc/environment` does not play any important role - i've added them initially just for the convention - the original ubuntu `/etc/environment ` has the PATH variable quoted.

So the quotes should not leak to python installation IF `/etc/environment` read by PAM-aware shell. We did reboot in order to enforce PAM-aware `login` to re-read the environments. I guess the `reloadEtcEnvironment` used instead of reboot and that might cause the appearing of the quotes.

But my final verdict is "let's remove the quotes from `scripts/helpers/etc-environment.sh` "

AlenaSviridenko (approved):

images/linux/scripts/installers/python.sh
maxim-lobanov: @dsame thank you for review!

PR #2395, opened 2021-01-06
[Windows] Add shell symlinks

al-cheb:
# Description
Add shell symlinks for Windows:

| Name         | Target                            |
| ------------ | --------------------------------- |
| gitbash.exe  | C:\Program Files\Git\bin\bash.exe |
| msys2bash.cmd | C:\msys64\usr\bin\bash.exe        |
| wslbash.exe  | C:\windows\System32\bash.exe      |

#### Related issue:
https://github.com/actions/virtual-environments/issues/1525

## Check list


al-cheb:
@eine, Could you please review this PR?

Ms230491 (approved):

eine:
@al-cheb, I believe that `C:\msys64\usr\bin\bash.exe` and `C:\msys64\usr\bin\sh.exe` are, in fact, the same shell. I don't think it's necessary to add both of them. @lazka, wdyt?

At the same time, I would suggest adding a wrapper, instead of symlinking directly. Based on https://github.com/msys2/setup-msys2/blob/master/main.js#L165-L169, the wrapper might be `msys2bash.cmd`:

```
@echo off
setloca
IF NOT DEFINED MSYS2_PATH_TYPE set MSYS2_PATH_TYPE=strict
set CHERE_INVOKING=1
C:\msys64\usr\bin\bash.exe -leo pipefail %*
```

al-cheb:
> @al-cheb, I believe that `C:\msys64\usr\bin\bash.exe` and `C:\msys64\usr\bin\sh.exe` are, in fact, the same shell. I don't think it's necessary to add both of them. @lazka, wdyt?
> 
> At the same time, I would suggest adding a wrapper, instead of symlinking directly. Based on https://github.com/msys2/setup-msys2/blob/master/main.js#L165-L169, the wrapper might be `msys2bash.cmd`:
> 
> ```
> @echo off
> setlocal
> IF NOT DEFINED MSYS2_PATH_TYPE set MSYS2_PATH_TYPE=strict
> IF NOT DEFINED MSYSTEM set MSYSTEM=mingw64
> set CHERE_INVOKING=1
> C:\msys64\usr\bin\bash.exe -leo pipefail %*
> ```

Thank you. Remove the link for `sh` shell and add a wrapper for bash.

images/win/scripts/Tests/Shell.Tests.ps1
```diff
@@ -0,0 +1,21 @@
+Describe "Shell" {
+    $IsWin16 = Test-IsWin16
+
+    $shellTestCases = @(
```
maxim-lobanov, line 4: Any reason why we can't add all 3 shells to this array and test?
I think `wslbash` can be added withoout any problem

maxim-lobanov:
```
$shellTestCases = ()
if (Test-IsWin16) {
  $shellTestCases += @{}
}
```

al-cheb: Thank you. Fixed.

maxim-lobanov (approved):

dmitry-shibanov (approved):

PR #2404, opened 2021-01-08
Parallel validation for Xcode installation

dmitry-shibanov:
# Description
In scope of this pull request we add logic to parallelise validation for Xcode during its installation.

## Check list


images/macos/provision/core/xcode.ps1
```diff
@@ -14,20 +14,28 @@ $os = Get-OSVersion
 $xcodeVersions = Get-ToolsetValue "xcode.versions"
 $defaultXcode = Get-ToolsetValue "xcode.default"
 [Array]::Reverse($xcodeVersions)
+$threadCount = 8
```
maxim-lobanov, line 17: There is a function to return the number of processors, may be it is better to use it?
image

images/macos/helpers/Xcode.Installer.psm1
```diff
@@ -1,4 +1,4 @@
-Import-Module "$PSScriptRoot/Xcode.Helpers.psm1"
+Import-Module "~/image-generation/helpers/Xcode.Helpers.psm1"
```
maxim-lobanov, line 1: Let's revert this line if we don't need it

images/macos/provision/core/xcode.ps1
al-cheb: nitpick: `[Environment]::ProcessorCount` as a variant

maxim-lobanov: Yep, `[Environment]::ProcessorCount` is even better from readability side

PR #2435, opened 2021-01-13
[Ubuntu] Get toolset value function.

Darleev:
# Description
Improvement
get_toolset_value replaced jq in every scripts where we're trying to get toolset during the build.

#### Related issue:
https://github.com/actions/virtual-environments-internal/issues/1672
## Check list


images/linux/scripts/helpers/get-toolset-value.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash -e
+################################################################################
+##  File:  toolset.sh
+##  Desc:  Helper functions to retrive toolset data
+################################################################################
+
+toolset_path="/imagegeneration/installers/toolset.json"
```
maxim-lobanov, line 7: We shouldn't use this approach with file for every helper function.
Let's just add function to existing helper https://github.com/actions/virtual-environments/blob/main/images/linux/scripts/helpers/install.sh that is imported almost everywhere.

images/linux/scripts/installers/clang.sh
```diff
@@ -32,10 +32,8 @@ function SetDefaultClang {
 wget https://apt.llvm.org/llvm.sh
 chmod +x llvm.sh
 
-toolset="$INSTALLER_SCRIPT_FOLDER/toolset.json"
-
-versions=$(jq -r '.clang.versions[]' $toolset)
-default_clang_version=$(jq -r '.clang.default_version' $toolset)
+versions=($(get_toolset_value '.clang.versions[]'))
+default_clang_version=($(get_toolset_value '.clang.default_version'))
```
maxim-lobanov, line 36: Why do we need extra brackets here? `default_clang_version($())`
I think they are necessary only when getting array from toolset

images/linux/scripts/installers/dotnetcore-sdk.sh
```diff
@@ -9,9 +9,8 @@ source $HELPER_SCRIPTS/install.sh
 source $HELPER_SCRIPTS/os.sh
 
 # Ubuntu 20 doesn't support EOL versions
-toolset="$INSTALLER_SCRIPT_FOLDER/toolset.json"
-LATEST_DOTNET_PACKAGES=$(jq -r '.dotnet.aptPackages[]' $toolset)
-versions=$(jq -r '.dotnet.versions[]' $toolset)
+LATEST_DOTNET_PACKAGES=($(get_toolset_value '.dotnet.aptPackages[]'))
```
maxim-lobanov, line 12: nitpick: please use consistent case for variables. Could you please rename this one to be consistent with `versions`

images/linux/scripts/installers/clang.sh
maxim-lobanov: The same in some other places

AlenaSviridenko:
please, attach correct task link to the description

maxim-lobanov (approved):

dibir-magomedsaygitov (approved):

vsafonkin (approved):

AlenaSviridenko (approved):

PR #2444, opened 2021-01-14
openssl php extension fix

LeonidLapshin:
# Description
Openssl extension installation on Windows do not use correct string replacements. 
Currently no string that contains "php_openssl.dll" can be found at php.ini and also documentation says that 

> "The syntax used in previous PHP versions ('extension=<ext>.so' and 'extension='php_<ext>.dll') is supported for legacy reasons and may be deprecated in a future PHP major version. So, when it is possible, please move to the new ('extension=<ext>) syntax"

so `extension=openssl` should be used instead.

## Check list

### DIFF
--- a/images/win/scripts/Installers/Install-PHP.ps1
+++ b/images/win/scripts/Installers/Install-PHP.ps1
@@ -11,7 +11,7 @@ Choco-Install -PackageName php -ArgumentList "--force", "--params", "/InstallDir
 Choco-Install -PackageName composer -ArgumentList "--ia", "/DEV=$installDir /PHP=$installDir"
 
 # update path to extensions and enable curl and mbstring extensions, and enable php openssl extensions.
-((Get-Content -path $installDir\php.ini -Raw) -replace ';extension=curl','extension=curl' -replace ';extension=mbstring','extension=mbstring' -replace ';extension_dir = "ext"','extension_dir = "ext"' -replace 'extension=";php_openssl.dll"','extension_dir = "php_openssl.dll"') | Set-Content -Path $installDir\php.ini
+((Get-Content -path $installDir\php.ini -Raw) -replace ';extension=curl','extension=curl' -replace ';extension=mbstring','extension=mbstring' -replace ';extension_dir = "ext"','extension_dir = "ext"' -replace ';extension=openssl','extension=openssl') | Set-Content -Path $installDir\php.ini
 
 # Set the PHPROOT environment variable.
 setx PHPROOT $installDir /M


al-cheb (approved):

AlenaSviridenko (approved):

images/win/scripts/Installers/Install-PHP.ps1
```diff
@@ -11,7 +11,7 @@ Choco-Install -PackageName php -ArgumentList "--force", "--params", "/InstallDir
 Choco-Install -PackageName composer -ArgumentList "--ia", "/DEV=$installDir /PHP=$installDir"
 
 # update path to extensions and enable curl and mbstring extensions, and enable php openssl extensions.
-((Get-Content -path $installDir\php.ini -Raw) -replace ';extension=curl','extension=curl' -replace ';extension=mbstring','extension=mbstring' -replace ';extension_dir = "ext"','extension_dir = "ext"' -replace 'extension=";php_openssl.dll"','extension_dir = "php_openssl.dll"') | Set-Content -Path $installDir\php.ini
+((Get-Content -path $installDir\php.ini -Raw) -replace ';extension=curl','extension=curl' -replace ';extension=mbstring','extension=mbstring' -replace ';extension_dir = "ext"','extension_dir = "ext"' -replace ';extension=openssl','extension=openssl') | Set-Content -Path $installDir\php.ini
```
al-cheb, line 14: Maybe I miss something, but:

Get-Content c:\tools\php\php.ini

1. Windows Server 2016
image

2. Windows Server 2019
image

LeonidLapshin: yes, but this script now shows all php extensions (curl, mbstring, openssl) that are enabled even if default php.ini will be changed sometimes.

maxim-lobanov:
cc: @shivammathur for any thoughts

images/win/scripts/Installers/Install-PHP.ps1
miketimofeev: @al-cheb @LeonidLapshin Strangely on a clean VM with php installed by choco it's commented 🤔
image

shivammathur:
@maxim-lobanov  @al-cheb  @miketimofeev 
This rightly fixes the script as extension file extensions as detected automatically now, so there is no `php_openssl.dll` in `php.ini`. 
https://github.com/php/php-src/blob/PHP-8.0.1/php.ini-production#L939

`Openssl` extension is a requirement of `Composer` and it currently enabled in the images when `chocolatey` runs `Composer-Setup.exe`, so soft of a side effect.

So, this PR should be merged.

shivammathur (approved):

maxim-lobanov (approved):

Issue #2388, opened 2021-01-04
Add chromium-browser to Ubuntu


Krinkle:
**Description**

Most package managers on Linux do not provide Google Chrome, they provide Chromium. As such, when projects provide bindings to different browsers across platforms, it often assumed that Chrome is used on Windows or macOS and Chromium on Linux.

As the Ubuntu image appears to be missing Chromium, this is making it difficult for developers to manage their local dev environment and CI setup in a way that works for contributors across operating systems.

**Area for Triage**:  
Containers, JavaScript and Node.js, Packages, Testing and code coverage

**Question, Bug, or Feature?**:  
Feature

**Virtual environments affected**  
- [ ] Ubuntu 16.04
- [ ] Ubuntu 18.04
- [x] Ubuntu 20.04
- [ ] macOS 10.15
- [ ] macOS 11.0
- [ ] Windows Server 2016 R2
- [ ] Windows Server 2019

See also:
- Follows-up https://github.com/actions/virtual-environments/issues/126.
- Ref https://github.com/karma-runner/karma-chrome-launcher/issues/141.

CatTheHacker:
From what I remember, Google Chrome is already installed on Ubuntu runners (https://github.com/actions/virtual-environments/blob/main/images/linux/scripts/installers/google-chrome.sh). Is there any preferred difference in Chromium over Chrome?

Krinkle:
This is not about preference. I do prefer Chromium over Chrome, and I think CI is the last place I expected Google Chrome to be installed over plain Chromum. It is provided by Ubuntu and Debian package managers out of the box, and is much more aligned with the ethos of the ecosystem it runs in. The end-user convenience trade-off that a Google setup offers afaik doesn't really apply to a CI context.

But... this ticket isn't about preference. It's about interoperability. Developer tooling that expects to find Chromium on Linux currently fails on GitHub Actions because neither `chromium` nor `chromium-browser` are discoverable in the PATH.

CatTheHacker:
What could be done is to create `chromium`/`chromium-browser` symlinks that point to `chrome`/`chrome-browser` in the image but that might sabotage people who install Chromium by themselves during workflow.
Removing Google Chrome would have far-wide impact on many people that already use it and installing Chromium additionally would take space, time during build and can affect workflows which install Chromium.
[...]

PR #2463, opened 2021-01-17
Chromium browser has been added to Ubuntu 20 

andy-mishechkin:
# Description
New tool.
Chromium browser has been added to Ubuntu 20 image. 
Additional disk space: 164 Kb
Time of installation: about 8 sec.

#### Related issue:
https://github.com/actions/virtual-environments/issues/2388

## Check list
- [X ] Tests are written (if applicable)


dibir-magomedsaygitov (approved):

images/linux/scripts/installers/chromium-browser.sh
```diff
@@ -0,0 +1,17 @@
...
+
+# Source the helpers for use with the script
+source $HELPER_SCRIPTS/os.sh
+
+if ! isUbuntu20 ; then
+    exit 0
+fi
+
+#Install chromium browser
+apt install chromium-browser -y
```
miketimofeev, line 15: we've moved all such installations to the toolset here https://github.com/actions/virtual-environments/pull/2376

images/linux/scripts/installers/chromium-browser.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash -e
+################################################################################
+##  File:  chromium-browser.sh
+##  Desc:  Installs Chromium browser
+################################################################################
+
+# Source the helpers for use with the script
+source $HELPER_SCRIPTS/os.sh
+
+if ! isUbuntu20 ; then
```
vsafonkin, line 10: I believe it can be removed, because we invoke script only ubuntu 20.04 template.

images/linux/scripts/SoftwareReport/SoftwareReport.Generator.ps1
```diff
@@ -193,12 +193,16 @@ $markdown += New-MDList -Style Unordered -Lines @(
...
 $markdown += New-MDHeader "Browsers and Drivers" -Level 3
-$markdown += New-MDList -Style Unordered -Lines @(
+$browsersAndDriversList = @(
     (Get-ChromeVersion),
     (Get-ChromeDriverVersion),
     (Get-FirefoxVersion),
     (Get-GeckodriverVersion)
 )
+if (Test-IsUbuntu20) {
+    $browsersAndDriversList += @(Get-ChromiumVersion)
```
maxim-lobanov, line 203: I think we need to reorder it a bit:
1) ChromiumVersion
2) ChromeVersion
3) ChromeDriverVersion
4) FirefoxVersion
5) GeckodriverVersion

The reason for this order is that Chrome and Chromium are almost identical and use the same ChromeDriver.

images/linux/scripts/installers/chromium-browser.sh
andy-mishechkin: Installation of chromium-browser has been moved to toolset

images/linux/scripts/SoftwareReport/SoftwareReport.Generator.ps1
andy-mishechkin: Done

images/linux/toolsets/toolset-2004.json
```diff
@@ -106,6 +106,7 @@
     "apt": {
         "common_packages": [
             "build-essential",
+            "chromium-browser",
```
miketimofeev, line 109: I suggest moving it to cmd_packages because those are packages with executables

images/linux/scripts/tests/Browsers.Tests.ps1
```diff
@@ -16,4 +16,10 @@ Describe "Chrome" {
     It "Chrome Driver" {
         "chromedriver --version" | Should -ReturnZeroExitCode
     }
+}
+
+Describe "Chromium" -Skip:(-not (Test-IsUbuntu20)) {
```
miketimofeev, line 21: We don't need this test after moving the installation to apt, but, please pay attention that if the executable name differs from the package name then you need to add a condition like this one
https://github.com/actions/virtual-environments/blob/a0656ad9b456750dfb92253c17669b6d4d882b3f/images/linux/scripts/tests/Apt.Tests.ps1#L12-L15

andy-mishechkin: We've decided to leave the separate test for chromium browser.

images/linux/toolsets/toolset-2004.json
andy-mishechkin: We've decided to leave the separate test for chromium browser.

images/linux/scripts/SoftwareReport/SoftwareReport.Generator.ps1
```diff
@@ -199,12 +199,18 @@ $markdown += New-MDList -Style Unordered -Lines (@(
...
     (Get-ChromeVersion),
     (Get-ChromeDriverVersion),
     (Get-FirefoxVersion),
     (Get-GeckodriverVersion)
 )
+if (Test-IsUbuntu20) {
+    $browsersAndDriversList = @(Get-ChromiumVersion) + $browsersAndDriversList
+}
+
+$markdown += New-MDList -Style Unordered -Lines ($browsersAndDriversList | Sort-Object)
```
maxim-lobanov, line 213: Why do we need `Sort-Object` here? I guess it will break order

andy-mishechkin: Fixed

maxim-lobanov (approved):

miketimofeev (approved):

al-cheb (approved):

PR #2483, opened 2021-01-19
[Windows] Clear local NuGet resources list in cache

vsafonkin:
Sometimes Dotnet application' builds are failed with error:
```
error NU1101: Unable to find package <package name>. No packages exist with this id in source(s): Microsoft Visual Studio Offline Packages
```
Nuget only looking for locally cached packages rather than using nuget.org.
Added clean-up NuGet resources list in cache after Dotnet installation.

[Link to NuGet resources cache documentation.](https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-nuget-locals)

## Check list

### DIFF
--- a/images/win/scripts/Installers/Install-DotnetSDK.ps1
+++ b/images/win/scripts/Installers/Install-DotnetSDK.ps1
@@ -109,7 +109,7 @@ function RunPostInstallationSteps()
     Set-ItemProperty -Path "HKLM:\Software\Microsoft\Windows\CurrentVersion\Run" -Name "DOTNETUSERPATH" -Value $cmdDotNet
 
     # Clear list of local NuGet resources in temporary cache
-    Invoke-Expression "dotnet nuget locals all --clear"
+    dotnet nuget locals all --clear
 }
 
 InstallAllValidSdks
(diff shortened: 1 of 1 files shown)


images/win/scripts/Installers/Install-DotnetSDK.ps1
```diff
@@ -107,6 +107,9 @@ function RunPostInstallationSteps()
 
     # Update Run key to run a script at logon
     Set-ItemProperty -Path "HKLM:\Software\Microsoft\Windows\CurrentVersion\Run" -Name "DOTNETUSERPATH" -Value $cmdDotNet
+
+    # Clear list of local NuGet resources in temporary cache
+    Invoke-Expression "dotnet nuget locals all --clear"
```
al-cheb, line 112: nitpick: Why should we use Invoke-Expression? -> dotnet nuget locals all --clear

vsafonkin: Done

maxim-lobanov: I think these changes won't help.

If you run `dotnet nuget locals all --list`, it will print
```
http-cache: C:\Users\VssAdministrator\AppData\Local\NuGet\v3-cache
global-packages: C:\Users\VssAdministrator\.nuget\packages\
temp: C:\Users\VssAdministrator\AppData\Local\Temp\NuGetScratch
plugins-cache: C:\Users\VssAdministrator\AppData\Local\NuGet\plugins-cache
```
All these directories are located under current user folder. We always re-create user profile during deployment.
Also no one of these folders exist on clean Hosted VM

Issue #2455, opened 2021-01-15
Add /home/runner/.local/bin to PATH on Linux


abitrolly:
`/home/runner/.local/bin` is not in PATH on Linux, which makes Python CLI tools installed with `pip install <command>` unavailable. Adding this dir to PATH will make `pip` in build environment behave similar to how things work in `virtualenv` where installed commands are immediately available.

See also https://github.com/actions/setup-python/issues/99

- Add
- If this is an add request:
  - Provide a basic test case to validate the tool's functionality:
```
pip install coverage
coverage
echo $?
```

**Area for Triage**:
Python Scripting and command line Feature

**Question, Bug, or Feature?**:

**Virtual environments affected**
- [ ] Ubuntu 16.04
- [x] Ubuntu 18.04
- [x] Ubuntu 20.04
- [ ] macOS 10.15
- [ ] macOS 11.0
- [ ] Windows Server 2016 R2
- [ ] Windows Server 2019

**Are you willing to submit a PR?**
Yes, if you show me where.

Darleev:
Hello @abitrolly,
We've added the required changes to our templates. PATH will be updated after the next image rollout.
I will keep you informed.

Darleev:
Hello @abitrolly,
We've added the required changes to the code, but they don't work at this time, due to some unexpected behavior for agent machines. We will investigate the issue with this path internally. As soon as a fix is applied, the path will be added to the agents.
Currently, it is possible to set environment variables using this approachs:
[Github Actions](https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#environment-files)
[Azure DevOps](https://github.com/microsoft/azure-pipelines-tasks/blob/master/docs/authoring/commands.md)

In case of any questions, feel free to contact us.

PR #2491, opened 2021-01-19
[Ubuntu] Add /home/runner/.local/bin to PATH

Darleev:
# Description
Improvement
/home/runner/.local/bin is not in PATH on Linux, which makes Python CLI tools installed with pip install <command> unavailable. Adding this dir to PATH will make pip in build environment behave similar to how things work in virtualenv where installed commands are immediately availabl

#### Related issue:
https://github.com/actions/virtual-environments/issues/2455

## Check list


images/linux/scripts/installers/python.sh
```diff
@@ -40,4 +40,7 @@ if isUbuntu18 || isUbuntu20 ; then
     fi
 fi
 
+# Adding this dir to PATH will make installed pip commands are immediately available.
+prependEtcEnvironmentVariable "$HOME/.local/bin"
```
maxim-lobanov, line 44: This won't work. we have different users on ADO and GA images.

maxim-lobanov: This code will add the following line `/home/packer/.local/bin` to `/etc/environments`.
After deployment, we will have the following users: `/home/vsts` (ADO) and `/home/runner` (GA) so this line will do nothing.

Please reevaluate if we really need to add this folder to PATH. If it is necessary, we can:
1) move pip cache under non-user directory
2) use post-build scripts

Darleev: $HOME/.local/bin added to skel for the user. Thank you for the meaningful reply.

al-cheb (approved):

vsafonkin (approved):

dmitry-shibanov (approved):

AlenaSviridenko (approved):

MaksimZhukov (approved):

Issue #2094, opened 2020-11-18
You have reached your pull rate limit


hutson:
**Description**  

When building Ubuntu images (It may also effect Windows) using a fork of this repository, I get the following error: `toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit`

**Area for Triage**:  

**Question, Bug, or Feature?**:  

Bug

**Virtual environments affected**  
- [x] Ubuntu 16.04
- [x] Ubuntu 18.04
- [x] Ubuntu 20.04
- [ ] macOS 10.15
- [ ] macOS 11.0
- [ ] Windows Server 2016 R2
- [ ] Windows Server 2019

**Expected behavior**  

I imagine I can extend the provisioners for the Ubuntu images to setup the appropriate authentication credentials, or to point to a Docker Hub mirror (The latter is preferred), which would avoid the rate limiting imposed on unauthenticated requests.

I simply do not know the best way to apply that change without it becoming difficult to maintain in a fork.

**Actual behavior**  

When building Ubuntu images (It may also effect Windows) using a fork of this repository, I get the following error: `toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit`

```
    ami:  OSType: linux
    ami:  Architecture: x86_64
    ami:  CPUs: 2
    ami:  Total Memory: 7.59GiB
    ami:  Name: ip-172-31-197-208
    ami:  ID: Y7LB:2E63:E6EC:M3FG:HQPG:XFD2:E6XT:CKZL:KOLY:4BBJ:NFFU:A3FT
    ami:  Docker Root Dir: /var/lib/docker
    ami:  Debug Mode: false
    ami:  Registry: https://index.docker.io/v1/
    ami:  Labels:
    ami:  Experimental: false
    ami:  Insecure Registries:
    ami:   127.0.0.0/8
    ami:  Live Restore Enabled: false
    ami:
    ami: 3.7: Pulling from library/alpine
==> ami: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit
==> ami: Provisioning step had errors: Running the cleanup provisioner, if present...
==> ami: Terminating the source AWS instance...
==> ami: Cleaning up any extra volumes...
==> ami: No volumes to clean up, skipping
==> ami: Deleting temporary security group...
[...]

PR #2548, opened 2021-01-25
feat(docker): authenticate using DockerHub credentials

hutson:
Ubuntu builds, by default, pull images anonymously from the
official DockerHub. DockerHub rate limits requests from
anonymous and unpaid accounts:
- https://docs.docker.com/docker-hub/download-rate-limit/

When those rate limits are reached the Packer build will
fail with an error indicated the rate limit has been
reached.

Add support for providing credentials for a paid
account that can be used by the Docker setup steps to
authenticate when pulling images from DockerHub, thereby
avoiding the rate limit.

Related to #2094

### DIFF
--- a/images/linux/scripts/installers/docker-moby.sh
+++ b/images/linux/scripts/installers/docker-moby.sh
@@ -28,10 +28,21 @@ systemctl is-enabled --quiet docker.service || systemctl enable docker.service
 sleep 10
 docker info
 
+# If credentials are provided, attempt to log into Docker Hub
+# with a paid account to avoid Docker Hub's rate limit.
+if [ "${DOCKERHUB_LOGIN}" ] && [ "${DOCKERHUB_PASSWORD}" ]; then
+  docker login --username "${DOCKERHUB_LOGIN}" --password "${DOCKERHUB_PASSWORD}"
+fi
+
 # Pull images
 images=$(get_toolset_value '.docker.images[]')
 for image in $images; do
     docker pull "$image"
 done
+
+# Always attempt to logout so we do not leave our credentials on the built
+# image. Logout _should_ return a zero exit code even if no credentials were
+# stored from earlier.
+docker logout
 
 invoke_tests "Tools" "Docker"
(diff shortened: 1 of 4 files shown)


images/linux/ubuntu1604.json
```diff
@@ -166,6 +168,19 @@
             ],
             "execute_command": "sudo sh -c '{{ .Vars }} pwsh -f {{ .Path }}'"
         },
+        {
+            "type": "shell",
+            "scripts": [
+                "{{template_dir}}/scripts/installers/docker-compose.sh",
+                "{{template_dir}}/scripts/installers/docker-moby.sh"
```
hutson, line 175: I intentionally moved these scripts into their own provisioner block so that the DockerHub credentials were not exposed to any other unrelated setup scripts.

images/linux/scripts/installers/docker-moby.sh
```diff
@@ -28,10 +28,18 @@ systemctl is-enabled --quiet docker.service || systemctl enable docker.service
...
 # Pull images
 images=$(get_toolset_value '.docker.images[]')
 for image in $images; do
     docker pull "$image"
 done
 
-invoke_tests "Tools" "Docker"
\ No newline at end of file
+# Always attempt to logout so we do not leave our credentials on the built image.
+docker logout
```
hutson, line 43: In my testing `docker logout` will not return a non-zero exit code even if you never logged into DockerHub. Therefore I believe it's safe/safer to always call it in this script.

hutson:
This PR only adds support for Ubuntu. My team only works with Ubuntu at the moment, so we're only focused on getting authentication working for that use case. In the future we will look into ensuring this works for Windows builds.

AlenaSviridenko:
Hi @hutson!
Thank you for contriubution, we don't mind accepting these changes for Ubuntu and later for Windows. Please, pull latest `main` changes, we merged a fix for broken cargo installation here https://github.com/actions/virtual-environments/pull/2545, and I will run checks to validate your PR.

hutson:
Pull request has been rebased on top of the latest from `main`.

miketimofeev (approved):

AlenaSviridenko (approved):

al-cheb (approved):

PR #2573, opened 2021-01-27
[macOS] move Pester tests to separate files/contexts/describes

Darleev:
# Description
Improvement
Need to move tests according to other environments in separate folder and contexts

#### Related issue:
https://github.com/actions/virtual-environments-internal/issues/1710

## Check list


vsafonkin (approved):

images/macos/tests/Powershell.Tests.ps1
```diff
@@ -38,4 +40,10 @@ Describe "Powershell" {
             }
         }
     }
+    
+    Context "Azure CLI" {
```
al-cheb, line 44: Why is az cli related to PowerShell section?

images/linux/scripts/installers/basic.sh
```diff
@@ -12,4 +12,4 @@ for package in $common_packages $cmd_packages; do
     apt-get install -y --no-install-recommends $package
 done
 
-invoke_tests "Apt"
\ No newline at end of file
+invoke_tests "BasicTools"
```
al-cheb, line 15: Why do we change linux scripts?

images/macos/tests/Powershell.Tests.ps1
Darleev: Az CLI installs in powershell.sh script. I think it's the best option to run it together with powershell itself.

images/linux/scripts/installers/basic.sh
Darleev: Thank you, very good point. I've fixed it.

dmitry-shibanov (approved):

dibir-magomedsaygitov (approved):

AlenaSviridenko (approved):

al-cheb (approved):

Issue #2458, opened 2021-01-16
Update cabal-install


juhp:
**Tool information**
- Tool name: cabal-install
- Add or update?  Update
- Desired version: 3.2.0.0
- Approximate size: 

**Area for Triage**:

Haskell

**Question, Bug, or Feature?**:
Bug

**Virtual environments affected**
- [x] Ubuntu 16.04
- [x] Ubuntu 18.04
- [x] Ubuntu 20.04
- [ ] macOS 10.15
- [ ] macOS 11.0
- [ ] Windows Server 2016 R2
- [ ] Windows Server 2019

**Can this tool be installed during the build?**

**Tool installation time in runtime**

**Are you willing to submit a PR?**

Probably

**Details**:
The problem is that the latest version (3.4.0.0) of cabal-install in hvr's ppa is a _prerelease_ (RC4) but haskell-setup defaults latest to 3.2.0.0.  So with either `cabal-version: 3.2` or without one gets the warning:

```
  Warning: cabal 3.2.0.0 was not found in the cache. It will be downloaded.
  If this is unexpected, please check if version 3.2.0.0 is pre-installed.
  The list of pre-installed versions is available here: https://help.github.com/en/actions/reference/software-installed-on-github-hosted-runners
  The above list follows a common haskell convention that the two latest major releases of cabal are commonly supported.
  If the list is outdated, please file an issue here: https://github.com/actions/virtual-environments
  by using the appropriate tool request template: https://github.com/actions/virtual-environments/issues/new/choose
  Attempting to install cabal-install 3.2 using apt-get
```

Admittedly cabal-install-3.4 should be released RSN but the RC's have already been out for months: maybe they are waiting for ghc 9.0 to be released, I dunno?

My suggestion is to override cabal-install to install 3.2.0.0 in the image and wait for 3.4 to be released but changing it back.
I mean here: https://github.com/actions/virtual-environments/blob/main/images/linux/scripts/installers/haskell.sh#L29
It is lucky that the ppa does not have a ghc-9.0 pre-release build yet. ;-)
Actually even ghc-8.10.3 is rather on the new side: Stackage LTS is still on 8.8.4.

I could open a haskell/actions ticket too to confirm about cabal-install.

juhp:
To add to this: I don't feel the solution taken in https://github.com/actions/virtual-environments/blob/main/images/linux/scripts/installers/haskell.sh is good - it is not really safe to assume that the latest version of ghc or cabal-install is always stable.
[...]

PR #2636, opened 2021-02-05
[Ubuntu] Rework haskell.sh to install Cabal using GHCup 

nikita-bykov:
# Description
We have reworked haskell.sh to install Cabal using `GHCup` instead of `apt-get install` to avoid installing a pre-release Cabal.

#### Related issue:
#2458
## Check list


images/linux/scripts/SoftwareReport/SoftwareReport.Common.psm1
```diff
@@ -219,6 +219,12 @@ function Get-GHCVersion {
     return "GHC $ghcVersion"
 }
 
+function Get-GHCupVersion {
+    $(ghcup --version) -match "version v(?<version>\d+\.\d+\.\d+)" | Out-Null
```
miketimofeev, line 223: nitpick: if version can be like 8.3.4.5 than it's better to use `\d+(\.\d+){2,}`

images/linux/scripts/installers/haskell.sh
```diff
@@ -7,34 +7,29 @@
...
-ghcMajorMinorVersions=$(echo "$allGhcVersions" | cut -d "." -f 1,2 | sort --unique --version-sort | tail -3)
-
-for version in $ghcMajorMinorVersions; do
-    # Get latest patch version for given Major.Minor one (ex. 8.6.5 for 8.6) and install it
-    exactVersion=$(echo "$allGhcVersions" | grep $version | sort --unique --version-sort | tail -1)
-    apt-get install -y ghc-$exactVersion
-    ghcInstalledVersions+=("$exactVersion")
-    defaultGHCVersion=$exactVersion
+# Any nonzero value for noninteractive installation
+export BOOTSTRAP_HASKELL_NONINTERACTIVE=1
```
maxim-lobanov, line 11: I think we should add this variable to etc/environments.
Otherwise, customers won' be able to use ghcup

images/linux/scripts/SoftwareReport/SoftwareReport.Common.psm1
jared-w: GHC's versions are always `a.b.c`, however Chocolatey can (and does) use `a.b.c.d` for patch updates to chocolatey's installation itself as all versions are immutable in Chocolatey.

(It doesn't change the correctness of this in any way)

al-cheb (approved):

dibir-magomedsaygitov (approved):

vsafonkin (approved):

MaksimZhukov (approved):

maxim-lobanov (approved):

PR #2644, opened 2021-02-07
Add PSScriptAnalyzer Powershell module

bergmeister:
# Description

New tool: [PSScriptAnalyzer](https://github.com/PowerShell/PSscriptanalyzer), which is PowerShell's code analysis tool and linter.

Size after installation is 288MB.
Download size is 16 MB when being downloaded as a zip, therefore 'installation time' is mainly limited by Internet connection and the speed to extract, on my local machine with a 70 Mbit/s connection, it takes 14 seconds, but the majority of the time is spent on downloading, therefore I'd assume this to be reduced to only a few seconds in a build machine that typically has a much faster Internet connection.

## Check list


AlenaSviridenko:
Hi @bergmeister,
thank you for your contribution, but we are a bit concerned about installation time and tool size, especially for Ubuntu images where we have issues with available disk space.

Our suggestion is to create an issue in this repository and monitor "thumbs up" for this tool to understand its demand. We can take care of this issue, or you can create it by yourself. What do you think?

bergmeister:
Ok, would you be happier to have it only for Windows images first as a starter and then have an issue for the *nix images?

AlenaSviridenko:
We'd prefer to track popularity for all OSs for 2 reasons:
- consistency of tools we have on images
- tool installation time will be added to image generation time, which is already ~6h for Windows now. Image generation speed is important for image deployment and changes testing, so we'd like to be sure about the tools we are preinstalling on images.

bergmeister:
Ok, I opened issue #2659 
Feel free to close this PR for now or keep it open, whatever you prefer.

al-cheb (approved):

bergmeister:
I've removed the entry for Ubuntu 1604 and 1804 as per issue feedback here: https://github.com/actions/virtual-environments/issues/2659#issuecomment-775943371

maxim-lobanov (approved):

AlenaSviridenko (approved):

PR #2668, opened 2021-02-10
Remove GOROOT on linux images

dsame:
# Description
 Improvement

The GOROOT environment variable is mostly deprecated, and no setup requires it. 
The problem with hardcoding a GOROOT is that if any other go tool is in use (for example, because one compiled Go tip), the environment variable will override its autodetected GOROOT.

## Check list


AlenaSviridenko:
Please, also removed from Windows images

AlenaSviridenko (approved):

miketimofeev (approved):

PR #2687, opened 2021-02-11
[macOS] Remove nvm aliases and old node versions.

dmitry-shibanov:
# Description
In scope of this pull request we deprecate some old node versions from nvm and remove nvm node aliases.

## Check list

### DIFF
--- a/images/macos/provision/core/nvm.sh
+++ b/images/macos/provision/core/nvm.sh
@@ -1,6 +1,6 @@
 #!/bin/bash -e -o pipefail
 ###########################################################################
-# The script installs node version manager with node versions 6,8,10 and 12
+# The script installs node version manager with node versions 10,12 and 14
 #
 ###########################################################################
 source ~/utils/utils.sh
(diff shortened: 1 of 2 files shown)


maxim-lobanov (approved):

AlenaSviridenko (approved):

images/macos/provision/core/nvm.sh
```diff
@@ -11,18 +11,9 @@ curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/$VERSION/install.sh | bash
 if [ $? -eq 0 ]; then
```
MaksimZhukov, line 11: Please update the script description above

images/macos/provision/core/nvm.sh
```diff
@@ -11,18 +11,9 @@ curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/$VERSION/install.sh | bash
...
         . ~/.bashrc
         nvm --version
-        nvm install lts/boron # 6.x
-        nvm install lts/carbon # 8.x
         nvm install lts/dubnium # 10.x
         nvm install lts/erbium # 12.x
-        nvm install v13 # 13.x
         nvm install v14 # 14.x
```
MaksimZhukov, line 16: Does it make sense to use a version number instead of codename for Node 10 and 14 and put all Node versions to an array to avoid adding a new line for each version? Something like this:
```
nodeVersions=("v10" "v12" "v14")
for version in ${nodeVersions[@]}; do
  nvm install $version
done
```

MaksimZhukov (approved):

images/macos/provision/core/nvm.sh
dmitry-shibanov: Thank you. Done.

images/macos/provision/core/nvm.sh
dmitry-shibanov: Thank you. Done.

al-cheb (approved):

PR #2710, opened 2021-02-15
Remove GOROOT on Windows images

dsame:
Improvement

The GOROOT environment variable is mostly deprecated, and no setup requires it.
The problem with hardcoding a GOROOT is that if any other go tool is in use (for example, because one compiled Go tip), the environment variable will override its autodetected GOROOT.

## Check list


maxim-lobanov:
@dsame , this PR has exactly same name like https://github.com/actions/virtual-environments/pull/2668 . Could you please update both PRs to specify platform where we remove the variable?

AlenaSviridenko (approved):

miketimofeev (approved):

al-cheb (approved):

PR #2729, opened 2021-02-17
[Ubuntu] Do not import Az.Account module on Ubuntu 18.04

al-cheb:
# Description
After installing `dbus-user-session` package on Ubuntu 18.04 ,` Import-Module Az.Accounts` has returned `** Message: 17:43:58.624: Remote error from secret service: org.freedesktop.DBus.Error.ServiceUnknown: The name org.freedesktop.secrets was not provided by any .service` files which has broken our CI due to writing to the stderr.

#### Related issue:
https://github.com/actions/virtual-environments-internal/issues/1831

## Check list

### DIFF
--- a/images/linux/scripts/tests/PowerShellModules.Tests.ps1
+++ b/images/linux/scripts/tests/PowerShellModules.Tests.ps1
@@ -34,30 +34,16 @@ Describe "AzureModules" {
                 $modulePath = Join-Path -Path $modulesRootPath -ChildPath "${moduleName}_${version}"
                 $moduleInfo = @{ moduleName = $moduleName; modulePath = $modulePath; expectedVersion = $version }
                 It "<expectedVersion> exists in modules directory" -TestCases $moduleInfo {
-                    $sb = {
+                    $testJob = Start-Job -ScriptBlock {
                         param (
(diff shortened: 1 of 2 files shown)


images/linux/scripts/tests/PowerShellModules.Tests.ps1
```diff
@@ -44,9 +43,21 @@ Describe "AzureModules" {
                         $env:PSModulePath = "${modulePath}:${env:PSModulePath}"
                         Import-Module -Name $moduleName
                         (Get-Module -Name $moduleName).Version.ToString()
+                    }
+                    
+                    if (Test-IsUbuntu18) {
```
miketimofeev, line 48: maybe follow this approach for all Ubuntus'?

AlenaSviridenko: Agreed. If it doesn't change test reliability, then we can do this for all Ubuntu images and avoid conditions

al-cheb: Thx. Fixed.

maxim-lobanov (approved):

PR #2731, opened 2021-02-17
[Ubuntu] Updating release file pattern for Kitware/CMake

whywaita:
# Description
This change fixes to install cmake.

https://github.com/actions/virtual-environments/pull/2714 has added an install script for [Kitware/CMake](https://github.com/Kitware/CMake).
But this script failed in installing process.

Because Kitware/CMake changed the name of assets after 3.20.0-rc that is the latest release (not pre-release).
e.g) `cmake-3.19.5-Linux-x86_64.sh` to `cmake-3.20.0-rc1-linux-x86_64.sh`

So, this Pull Request changes the pattern of assets. 

## Check list
- [x] Related issue / work item is attached
- [ ] Tests are written (if applicable)
    - not applicable
- [ ] Documentation is updated (if applicable)
    - not applicable
- [x] Changes are tested and related VM images are successfully generated

PR #2732, opened 2021-02-17
updating cmake URL condition

agorti:
# Description
We are getting an empty URL due to case sensitivity in AWS.

## Check list


al-cheb:
@agorti , Duplicate - https://github.com/actions/virtual-environments/pull/2731

PR #2733, opened 2021-02-17
[Ubuntu] remove unnecessary quotes from skel path.

Darleev:
# Description
Bug fixing
Currently, we added some paths using skel .bashrc file, but it does not work since we use quotes in the PATH variable.
So, it is required to change such approach:
`echo 'export XDG_CONFIG_HOME="$HOME/.config"' | tee -a /etc/skel/.bashrc`
and unnecessary quotes:
`echo 'export XDG_CONFIG_HOME=$HOME/.config' | tee -a /etc/skel/.bashrc`

#### Related issue:
https://github.com/actions/virtual-environments/issues/2455

## Check list
- [-] Tests are written (if applicable)


catthehacker:
I'd recommend using interactive rebase when pulling latest changes from upstream to your branch instead of merge, then you can just `git push --force` your branch and it will not bloat commit history.
`git config pull.rebase interactive`
`git config pull.ff only`

If you would like to just fix your history, I think that rebasing branch on commit **before** first merge and using `fixup` option on those merge commits should do the trick.

Darleev:
> I'd recommend using interactive rebase when pulling latest changes from upstream to your branch instead of merge, then you can just `git push --force` your branch and it will not bloat commit history.
> `git config pull.rebase interactive`
> `git config pull.ff only`
> 
> If you would like to just fix your history, I think that rebasing branch on commit **before** first merge and using `fixup` option on those merge commits should do the trick.

Thank you @catthehacker, I'll definitely fix this

maxim-lobanov (approved):

AlenaSviridenko (approved):

miketimofeev (approved):

Darleev:
New fix has been provdied.
https://github.com/actions/virtual-environments/pull/2826

PR #2735, opened 2021-02-17
Az Modules Size Reduction On Image

AmrutaKawade:
# Description

Currently hosted agent image installs multiple versions of Az, Azure and Azure RM modules which is taking more disk space.

- Linux

   - [12](https://github.com/actions/virtual-environments/blob/f534484cccc12e73d3a8108329e90d3d93e61b05/images/linux/toolsets/toolset-1804.json#L115) Az PowerShell modules are prebaked.
   - All these modules consume ~1.8 GB of disk space.
   - In ubuntu/linux agents 16 GB disk space is available and the above amount is already consumed by these modules which reduces the available disk space for customer's use.

- Windows

   - 11 Az, 5 Azure and 6 AzureRM [modules](https://github.com/actions/virtual-environments/blob/f534484cccc12e73d3a8108329e90d3d93e61b05/images/win/toolsets/toolset-2019.json#L100) are prebaked
   - All these modules consume ~5 GB of disk space
   - In case of windows agents we don't have space problem as of now but we can't keep on adding modules indefinitely.

- Mac
  - It comes bundled with only 1 version of Az module and its usage is almost nil so there is no concern for it as of now.

AzurePowerShell task and action, AzureFileCopy and AzureCloudPowershell tasks are taking dependency on az, azure and azurerm ps modules

- Windows

  - 1 latest Az version will be prebaked as folder in C:\modules\az_<version>. Will be downloaded from PSGallery using Save-Module.
  - All [currently supported](https://github.com/actions/virtual-environments/blob/f534484cccc12e73d3a8108329e90d3d93e61b05/images/win/toolsets/toolset-2019.json#L125) versions will be made available as zip in C:\modules\az_<version>.zip. Will be downloaded from [GH Release](https://github.com/Azure/az-ps-module-versions).
- Linux

  - 1 latest Az version will be prebaked as folder in /usr/share/az_<version>. Will be downloaded from PSGallery using Save-Module.
  - Versions 3.1.0 and 4.4.0 will be prebaked as zip in /usr/share/az_<version> similar to above in case of windows. Since these are mostly used versions by users apart from latest
- Mac

  - 1 latest Az version will be prebaked as folder in /usr/share/az_<version>. Will be downloaded from PSGallery using Save-Module.
  - No version prebaked as zip.

**Image creation logic for Azure and AzureRM**
- Azure and AzureRM modules are only supported in Windows.
- Azure 5.3.0(latest) and 2.1.0(default) will be bundled as folder. Will be downloaded from PSGallery using Save-Module
- AzureRM 6.13.1(latest) and 2.1.0(default) will be bundled as folder. Will be downloaded from PSGallery using Save-Module
- All currently supported versions will be made available as zip in C:\modules\<modulename>_<version>.zip. Will be downloaded from PSGallery using Save-Module, zipped using `Compress-Archive -Path C:\modules\<modulename>_<version> -DestinationPath C:\modules\<modulename>_<version>.zip` and then delete the module folder.

## Check list


images/linux/toolsets/toolset-1604.json
```diff
@@ -120,19 +120,13 @@
...
             "versions": [
-                "1.0.0",
-                "1.6.0",
-                "2.3.2",
-                "2.6.0",
-                "2.8.0",
+                "5.5.0"
+            ],
+            "zipversions": [
                 "3.1.0",
```
al-cheb, line 128: Why don't we zip the same versions for all images?

U16 - 3.8.0, 4.8.0
U18 - 3.8.0, 4.8.0
U20 - 3.8.0, 4.8.0

images/win/toolsets/toolset-2016.json
```diff
@@ -100,28 +100,36 @@
...
-                "6.13.1"
+                "6.7.0"
             ],
             "default": "2.1.0"
         },
         {
             "name": "azure",
             "versions": [
                 "2.1.0",
+                "5.3.0"
```
al-cheb, line 117: Why don't the 5.5.0 as the default?

images/linux/toolsets/toolset-1604.json
AmrutaKawade: We checked usage for each module on each image. since 3.1.0 and 4.4.0 were never been made available on ubuntu20 its usage is zero. so we don't want to add those versions on U20. 

for U16 and U18 3.1.0 and 4.4.0 are mostly used versions. for other versions which are rarely used we are modifying task which use these modules to download it on the run.

This is to reduce disk space

I have updated description to get more context.

images/linux/scripts/installers/Install-AzureModules.ps1
```diff
@@ -17,6 +17,25 @@ foreach ($module in $modules)
...
         Save-Module -Path $modulePath -Name $moduleName -RequiredVersion $version -Force -Verbose
     }
+
+    $assets = Invoke-RestMethod $module.url
+
+    # Get github release asset for each version
+    foreach ($toolVersion in $module.zipversions) {
+        $asset = $assets | Where-Object version -like $toolVersion `
+        | Select-Object -ExpandProperty files `
+        | Where-Object { ($_.platform -eq $module.platform) -and ($_.platform_version -eq $module.platform_version)} `
```
maxim-lobanov, line 27: I think we don't need this condition because versions-manifest.json doesn't contain any platforms and platform_version values.

images/win/scripts/Installers/Install-AzureModules.ps1
```diff
@@ -36,6 +36,46 @@ foreach ($module in $modules)
...
+            } else {
+                Write-Host "Asset was not found in versions manifest"
+                exit 1
+            }
+        }
+        # Install modules from powershell gallery
+        else 
+        {
+            $modulePath = Join-Path -Path $installPSModulePath -ChildPath "${moduleName}_${version}"
+            Write-Host " - $version [$modulePath]"
```
maxim-lobanov, line 65: As I understand, this code branch is needed for Azure and AzureRM modules that you don't distribute in Azure/az-ps-module-versions, am I right?

Considering the fact that new versions of Azure and AzureRM won't be released, can we compress them once and upload somewhere? If you don't want to use `Azure/az-ps-module-versions`, we can upload them to https://vstsagenttools.blob.core.windows.net/tools
We are trying to reduce image generation time so compressing the identical modules in every build looks excessive, especially because they won't be changed in future

AmrutaKawade:
I have split this PR into 2

https://github.com/actions/virtual-environments/pull/2828
https://github.com/actions/virtual-environments/pull/2827

Closing this PR

Issue #2676, opened 2021-02-10
Update/Add fastlane


alwell-kevin:
**Tool information**
- Tool name: [fastlane](https://github.com/fastlane/fastlane)
- Tool license: MIT
- Add or update? Add
- Desired version: latest
- Approximate size: unknown
- If this is an add request:
  - Brief description of tool: fastlane is a tool for iOS and Android developers to automate tedious tasks like generating screenshots, dealing with provisioning profiles, and releasing applications.
  - URL for tool's homepage: https://docs.fastlane.tools/
  - Provide a basic test case to validate the tool's functionality:

```
Navigate your terminal to your project's directory and run

_fastlane init_
You'll be asked to confirm that you're ready to begin, and then for a few pieces of information. To get started quickly:

Provide the package name for your application when asked (e.g. io.fabric.yourapp)
Press enter when asked for the path to your json secret file
Answer 'n' when asked if you plan on uploading info to Google Play via fastlane (we can set this up later)
That's it! fastlane will automatically generate a configuration for you based on the information provided.

You can see the newly created ./fastlane directory, with the following files:

Appfile which defines configuration information that is global to your app
Fastfile which defines the "lanes" that drive the behavior of fastlane
```

**Area for Triage**:

**Question, Bug, or Feature?**:

**Virtual environments affected**
- [X] Ubuntu 16.04
- [X] Ubuntu 18.04
- [X] Ubuntu 20.04
- [ ] macOS 10.15
- [ ] macOS 11.0
- [ ] Windows Server 2016 R2
- [ ] Windows Server 2019

**Can this tool be installed during the build?**

**Tool installation time in runtime**

**Are you willing to submit a PR?**

maxim-lobanov:
Hello @alwell-kevin , we don't have objections to pre-install fastlane on Ubuntu.
Moved to the backlog

miketimofeev:
@alwell-kevin don't you mind if we install Fastlane only on Ubuntu 20 since we have a lack of free space on Ubuntu 18 and Ubuntu 16? Furthermore, Ubuntu 16 is not supported as Fastlane requires Ruby 2.4 and above, but we have 2.3 there.

alwell-kevin:
Hi @miketimofeev @maxim-lobanov let me confirm with the ENG I am working with who made the initial request, thank you!

Chuckv01:
[...]

PR #2751, opened 2021-02-18
[Ubuntu] Add Fastlane

miketimofeev:
# Description
Fastlane is a ruby gem and requires Ruby >= 2.4 thus we don't want to preinstall it everywhere, only on Ubuntu 20 with system Ruby 2.7. However, in this PR, I've added the RubyGems section to each toolset in case we would need to install more gems in the future.
 I also noticed that we don't have tests for system ruby and don't provide RubyGems version in the readme  — added them as well. 
Another point worth to mention — it's a bad idea to update ruby gems from ruby package with `--system` flag and, for example, it's even disabled for Debian so we should either rely on RubyGems from Ruby-full package or install Ruby via other methods.

#### Related issue:
https://github.com/actions/virtual-environments/issues/2676

## Check list


al-cheb (approved):

images/macos/software-report/SoftwareReport.Common.psm1
```diff
@@ -181,6 +181,11 @@ function Get-RubyVersion {
     return "Ruby $rubyVersion"
 }
 
+function Get-RubyGemsVersion {
+    $rubyGemsVersion = Run-Command "gem --version"
+    return "RubyGems $rubyGemsVersion"
+}
+
```
maxim-lobanov, line 188: Should this function be in Linux software report? Currently, it is in macOS

miketimofeev: Looks like I love macOS more than Linux, fixed

images/linux/scripts/installers/ruby.sh
```diff
@@ -1,13 +1,22 @@
...
-sudo apt-get install ruby-full
-sudo gem update --system
+apt-get install ruby-full
+gem update
+
+# Install ruby gems from toolset
+gemsToInstall=$(get_toolset_value ".rubygems[] .name")
+if [ -n "$gemsToInstall" ]; then
+    for gem in $gemsToInstall; do
+        echo "Installing rubygem $gem"
```
AlenaSviridenko, line 16: RubyGem is a tool name, the package itself is called gem, so I think the message should be "Installing gem"

miketimofeev: makes sense

AlenaSviridenko (approved):

dmitry-shibanov (approved):

PR #2763, opened 2021-02-20
Add lerna to linux images

dsame:
# Description
New tool

A tool for managing JavaScript projects with multiple packages => enabling monorepo projects.

Approximate size: 80mb

## Check list


images/linux/scripts/SoftwareReport/SoftwareReport.Tools.psm1
```diff
@@ -28,6 +28,11 @@ function Get-BazelVersion {
     return "Bazel $bazelVersion"
 }
 
+function Get-LernaVersion {
+    $lernaVersion = bazel --version
```
al-cheb, line 32: Why bazel?

images/linux/scripts/installers/lerna.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash -e
```
al-cheb, line 1: Use the nodejs.sh script to install lerna

AlenaSviridenko:
please, remove Lerna for Ubuntu 16 and 18, we should install it to Ubuntu 20 only

images/linux/scripts/installers/lerna.sh
miketimofeev: I suggest creating a new toolset section with Node.js modules as well

images/linux/toolsets/toolset-2004.json
```diff
@@ -259,5 +259,8 @@
     },
     "rubygems": [
         {"name": "fastlane"}
+    ],
+    "node_modules": [
```
miketimofeev, line 263: There are more modules in nodejs provisioner script

images/linux/scripts/tests/Node.Tests.ps1
```diff
@@ -0,0 +1,14 @@
+Describe "Node.js" {
```
AlenaSviridenko, line 1: please, remove Node.js tests from Tools.Tests.ps1

images/linux/scripts/installers/nodejs.sh
```diff
@@ -4,14 +4,27 @@
...
+# remove commenting out of this line as nmp migrated to v7
+# npm install -g $node_modules
 
-# Install the Netlify CLI using --unsafe-perm=true options to avoid permission issues
-npm install -g --unsafe-perm=true netlify-cli
+# TODO: workaround for Netlify CLI with npm6. Remove 19-28 and uncomment 17 if migration to npm7 compelted
+for module in $node_modules; do
+    echo "Installing node module $module"
+    if [ $module = "netlify-cli" ];then
+      npm install -g --unsafe-perm=true $module
```
AlenaSviridenko, line 23: please, restore comment for netlify installation flag `# Install the Netlify CLI using --unsafe-perm=true options to avoid permission issues`

images/linux/scripts/tests/Node.Tests.ps1
```diff
@@ -0,0 +1,14 @@
...
+    $binaries = ,"node"
```
AlenaSviridenko, line 2: I think `,` should not be here

images/linux/toolsets/toolset-2004.json
```diff
@@ -259,5 +259,51 @@
...
+		    "command": "webpack-cli"
+	    },
+	    {
+	            "name": "netlify-cli",
+		    "command": "netlify"
+	    },
+	    {
+	            "name": "lerna",
+		    "command": "lerna"
+	    }
```
maxim-lobanov, line 307: Could you please revisit all toolset files and replace tabs to spaces?
Looks like you have mixed them in toolset files and in nodejs.sh

images/linux/scripts/tests/Toolset.Tests.ps1
```diff
@@ -63,4 +63,4 @@ Describe "Toolset" {
             }
         }
     }
-}
\ No newline at end of file
+}
```
maxim-lobanov, line 66: Please revert empty file

images/linux/scripts/SoftwareReport/SoftwareReport.Tools.psm1
```diff
@@ -311,4 +311,4 @@ function Get-SphinxVersion {
 
 function Get-YamllintVersion {
     return "$(yamllint --version)"
-}
\ No newline at end of file
+}
```
maxim-lobanov, line 314: Please revert empty file

images/linux/scripts/tests/Node.Tests.ps1
dsame: It must. This creates array of one item

images/linux/scripts/installers/nodejs.sh
```diff
@@ -4,14 +4,28 @@
...
 # Install LTS Node.js and related build tools
 curl -sL https://raw.githubusercontent.com/mklement0/n-install/stable/bin/n-install | bash -s -- -ny -
 ~/n/bin/n lts
-npm install -g grunt gulp n parcel-bundler typescript newman vercel
-npm install -g --save-dev webpack webpack-cli
+# Install node modules
+node_modules=$(get_toolset_value '.node_modules[].name')
+
+# remove commenting out of this line as nmp migrated to v7
+# npm install -g $node_modules
```
AlenaSviridenko, line 17: these comments can be removed, right?

dsame: I think they should not. This command line will install all modules in time and speed up the the build a bit. We have to use the foreach loop while we live with npm6 and need a workaround for netlify

al-cheb (approved):

images/linux/scripts/installers/nodejs.sh
```diff
@@ -4,14 +4,28 @@
...
 
-# Install the Netlify CLI using --unsafe-perm=true options to avoid permission issues
-npm install -g --unsafe-perm=true netlify-cli
+# TODO: workaround for Netlify CLI with npm6. Remove 19-28 and uncomment 17 if migration to npm7 compelted
+for module in $node_modules; do
+    echo "Installing node module $module"
+    if [ $module = "netlify-cli" ];then
+      # Install the Netlify CLI using --unsafe-perm=true options to avoid permission issues
+      npm install -g --unsafe-perm=true $module
+    else
```
maxim-lobanov, line 25: nitpick: looks like you have mixed tabs with spaces:
image

maxim-lobanov (approved):

AlenaSviridenko (approved):

dmitry-shibanov (approved):

images/linux/scripts/tests/Node.Tests.ps1
miketimofeev: why not define array using `$binaries = @("node")`?

images/linux/scripts/installers/nodejs.sh
dsame: fixed and double checked with `grep -r $'\t' images/linux/`

images/linux/scripts/tests/Node.Tests.ps1
```diff
@@ -0,0 +1,14 @@
...
+    $binaries = @("node")
+    $module_commands = (Get-ToolsetContent).node_modules | ForEach-Object { $_.command }
+    $testCases = $binaries + $module_commands | ForEach-Object { @{NodeCommand = $_} }
+
+    It "<NodeCommand>" -TestCases $testCases {
+        param (
+            [string] $NodeCommand
+        )
+
+        "$NodeCommand --version" | Should -ReturnZeroExitCode
```
maxim-lobanov, line 11: As I understand, we have to add `command` to toolset because package name can be non-equal to command, right? 
if you have decided to put the command to toolset, let's also put the `--version` to toolset too to be more flexible.

maxim-lobanov (approved):

images/linux/scripts/tests/Node.Tests.ps1
AlenaSviridenko: @maxim-lobanov I think putting `--version` to the toolset is out of the toolset purpose, maybe it is better to rename toolset fields to avoid confusion, like 
```
{
packageName: "netlify-cli",
cmdName: "netlify"
}
```

dsame: You are correct, i do not think the semantic for "command" attribute should be "command to test" because, in the theory, the command can be used for other purposes. In fact you suggest to return to the "commandShouldRetunrZero" attribute i had in the very beginning :)
I think we should to live the the command so far and the third attribute like "testCommand" or something in case we would really need it

dsame: @AlenaSviridenko i do not think we need to rename the fields as well
1. introducing another word "package" in addition to "node module" leads to the extra ambiguity for my opinion
2. we have pretty self explainable notation node_modules[].name and  node_modules[].command

PR #2827, opened 2021-03-03
[Windows] Az Modules Size Reduction On Image

AmrutaKawade:
# Description
Improvement

Currently hosted agent image installs multiple versions of Az, Azure and Azure RM modules which is taking more disk space.
- Windows

   - 11 Az, 5 Azure and 6 AzureRM [modules](https://github.com/actions/virtual-environments/blob/f534484cccc12e73d3a8108329e90d3d93e61b05/images/win/toolsets/toolset-2019.json#L100) are prebaked
   - All these modules consume ~5 GB of disk space
   - In case of windows agents we don't have space problem as of now but we can't keep on adding modules indefinitely.

AzurePowerShell task and action, AzureFileCopy and AzureCloudPowershell tasks are taking dependency on az, azure and azurerm ps modules

- Windows

  - 1 latest Az version will be prebaked as folder in C:\modules\az_<version>. Will be downloaded from PSGallery using Save-Module.
  - All [currently supported](https://github.com/actions/virtual-environments/blob/f534484cccc12e73d3a8108329e90d3d93e61b05/images/win/toolsets/toolset-2019.json#L125) versions will be made available as zip in C:\modules\az_<version>.zip. Will be downloaded from [GH Release](https://github.com/Azure/az-ps-module-versions).

**Image creation logic for Azure and AzureRM**
- Azure and AzureRM modules are only supported in Windows.
- Azure 5.3.0(latest) and 2.1.0(default) will be bundled as folder. Will be downloaded from PSGallery using Save-Module
- AzureRM 6.13.1(latest) and 2.1.0(default) will be bundled as folder. Will be downloaded from PSGallery using Save-Module
- All currently supported versions will be made available as zip in C:\modules\<modulename>_<version>.zip. Will be downloaded from https://vstsagenttools.blob.core.windows.net/tools/azurepowershellmodules/

## Check list


images/win/scripts/Installers/Install-AzureModules.ps1
```diff
@@ -36,6 +36,46 @@ foreach ($module in $modules)
...
+                Compress-Archive -Path $modulePath -DestinationPath "${modulePath}.zip"
+                Remove-Item $modulePath -Recurse -Force
+            }
+            catch
+            {
+                Write-Host "Error: $_"
+                exit 1
+            }
+        }
+    }
```
maxim-lobanov, line 78: As I understand, this code branch is needed for Azure and AzureRM modules that you don't distribute in `Azure/az-ps-module-versions`, am I right?

Considering the fact that new versions of Azure and AzureRM won't be released, can we compress them once and upload somewhere? If you don't want to use `Azure/az-ps-module-versions`, we can upload them to `https://vstsagenttools.blob.core.windows.net/tools`
We are trying to reduce image generation time so compressing the identical modules in every build looks excessive, especially because they won't be changed in future

images/win/scripts/Installers/Install-AzureModules.ps1
```diff
@@ -36,6 +36,46 @@ foreach ($module in $modules)
         }
     }
 
+    if($null -ne $module.url)
+    {
+        $assets = Invoke-RestMethod $module.url
```
amit-avit, line 41: @maxim-lobanov don't we need to auth the request here with GH token to avoid rate-limiting as well as below while making use of `asset.download_url` while downloading?

images/win/toolsets/toolset-2016.json
```diff
@@ -102,28 +102,36 @@
...
             ],
             "default": "2.1.0"
         },
         {
             "name": "azure",
             "versions": [
                 "2.1.0",
+                "5.3.0"
+            ],
+            "zipversions": [
```
amit-avit, line 121: nit: camelCase t/o

images/win/scripts/Installers/Install-AzureModules.ps1
maxim-lobanov: Nope, it is not needed here because we run image-generation on self-hosted machines and they don't meet rate limits (at least, we interact with GH API a lot and never faced it yet :) )

images/win/scripts/Installers/Install-AzureModules.ps1
```diff
@@ -36,6 +36,46 @@ foreach ($module in $modules)
...
+    }
+
+    foreach ($version in $module.zipversions)
+    {
+        # Install modules from GH Release
+        if($null -ne $assets) 
+        {
+            $asset = $assets | Where-Object version -like $version `
```
amit-avit, line 49: why making use of like `version -like $version` shouldn't `-eq` check be more appropriate

images/win/toolsets/toolset-2016.json
AmrutaKawade: camel case is not followed in this json. I have added underscore to make it consistent with other keys

images/win/scripts/Installers/Install-AzureModules.ps1
AmrutaKawade: done

images/win/scripts/Installers/Install-AzureModules.ps1
AmrutaKawade: PR for uloading azure and azurerm modules to blob https://dev.azure.com/mseng/AzureDevOps/_git/CIPlat.Externals/pullrequest/601350

maxim-lobanov: Thank you!

amit-avit:
You need M1 approval to merge this PR.

images/win/scripts/Installers/Install-AzureModules.ps1
AmrutaKawade: code changes to download from blob are done. please review

al-cheb (approved):

PR #2828, opened 2021-03-03
[Linux] Az Modules Size Reduction On Image

AmrutaKawade:
# Description
Improvement

Currently hosted agent image installs multiple versions of Az modules which is taking more disk space.

- Linux

   - [12](https://github.com/actions/virtual-environments/blob/f534484cccc12e73d3a8108329e90d3d93e61b05/images/linux/toolsets/toolset-1804.json#L115) Az PowerShell modules are prebaked.
   - All these modules consume ~1.8 GB of disk space.
   - In ubuntu/linux agents 16 GB disk space is available and the above amount is already consumed by these modules which reduces the available disk space for customer's use.

AzurePowerShell task and action are taking dependency on az ps modules

- Linux

  - 1 latest Az version will be prebaked as folder in /usr/share/az_<version>. Will be downloaded from PSGallery using Save-Module.
  - Versions 3.1.0 and 4.4.0 will be prebaked as zip in /usr/share/az_<version>.zip Will be downloaded from [GH Release](https://github.com/Azure/az-ps-module-versions). Since these are mostly used versions by users apart from latest

## Check list


maxim-lobanov (approved):

images/linux/toolsets/toolset-1604.json
```diff
@@ -119,19 +119,13 @@
...
+            "url" : "https://raw.githubusercontent.com/Azure/az-ps-module-versions/main/versions-manifest.json",
             "versions": [
-                "1.0.0",
-                "1.6.0",
-                "2.3.2",
-                "2.6.0",
-                "2.8.0",
+                "5.5.0"
+            ],
+            "zipversions": [
```
amit-avit, line 126: nit: camelCase t/o

images/linux/scripts/installers/Install-AzureModules.ps1
```diff
@@ -19,6 +19,24 @@ foreach ($module in $modules)
     }
 }
 
+$assets = Invoke-RestMethod $module.url
+
+# Get github release asset for each version
+foreach ($toolVersion in $module.zipversions) {
+    $asset = $assets | Where-Object version -like $toolVersion `
```
amit-avit, line 26: Should we use `-eq` check instead to be more explicit

images/linux/toolsets/toolset-2004.json
```diff
@@ -103,7 +103,11 @@
     "azureModules": [
         {
             "name": "az",
+            "url" : "https://raw.githubusercontent.com/Azure/az-ps-module-versions/main/versions-manifest.json",
             "versions": [
+                "5.5.0"
+            ],
+            "zip_versions": [
                 "3.8.0",
```
amit-avit, line 111: Please check if these versions are needed. Most used version in ubuntu is 3.1.0, most likely that would be better to bundle instead of these versions.
@maxim-lobanov can u confirm if these versions are really needed in Ubuntu 2004

AmrutaKawade: ubuntu 2004 was always released with these versions. if we don't include this by default unnecessarily we are increasing time for existing pipelines who are using these versions. 

Also there is provision now to download any specific version who ever wants this version they can use it.

I don't think we should change existing behaviour

amit-avit: @AmrutaKawade the reason for my ask is because we have a plan that `ubuntu-latest` will start making use of 2004 instead of 1804, so many jobs will switch or would have already switched to start making use of 2004 from 1804. These jobs will get affected as the az versions would then get dynamically downloaded.
If you look at the telemetry data as of today:
- 3.8.0 usage in ubuntu is 54
- 4.8.0 usage in ubuntu is 1497
whereas
- 3.1.0 usage in ubuntu is 18746 out of which 15997 is happening through GHRelease

amit-avit: Below is the telemetry data sorted based on linuxJobCount:
image

al-cheb (approved):

images/linux/scripts/installers/Install-AzureModules.ps1
```diff
@@ -17,8 +17,27 @@ foreach ($module in $modules)
...
+    # Get github release asset for each version
+    foreach ($toolVersion in $module.zip_versions) {
+        $asset = $assets | Where-Object version -eq $toolVersion `
+        | Select-Object -ExpandProperty files `
+        | Select-Object -First 1
+
+        Write-Host "Installing $($module.name) $toolVersion ..."
+        if ($null -ne $asset) {
+            Write-Host "Download $($asset.filename)"
+            wget $asset.download_url -nv --retry-connrefused --tries=10 -P $installPSModulePath
```
al-cheb, line 32: nitpick: we have `download_with_retries` function

images/linux/scripts/installers/Install-AzureModules.ps1
```diff
@@ -17,8 +17,27 @@ foreach ($module in $modules)
...
+        } else {
+            Write-Host "Asset was not found in versions manifest"
+            exit 1
+        }
+    }
 }
 
+
 # If Az.Accounts > 1.0.0 unable to load module with error: Assembly with same name is already loaded
```
al-cheb, line 41: nitpick:  we can delete 41-48 line, because we don't have az 1.0 anymore.

Issue #2978, opened 2021-03-18
Breaking change due to missing '.dotnet/tools to path'


edumserrano:
**Description**  

Today I've noticed my workflow failing on the step that is trying to run the installed dotnet tool because it can't find it.

I've noticed that there is a difference in the last line of the logs produced by the `actions/setup-dotnet@v1` from when it was working to now when it is failing:

* output from previously working `actions/setup-dotnet@v1`:
```
/home/runner/work/_actions/actions/setup-dotnet/v1/externals/install-dotnet.sh --version 5.0.201
dotnet_install: Warning: Unable to locate zlib. Probable prerequisite missing; install zlib.
dotnet-install: Downloading link: https://dotnetcli.azureedge.net/dotnet/Sdk/5.0.201/dotnet-sdk-5.0.201-linux-x64.tar.gz
dotnet-install: Extracting zip from https://dotnetcli.azureedge.net/dotnet/Sdk/5.0.201/dotnet-sdk-5.0.201-linux-x64.tar.gz
dotnet-install: Adding to current process PATH: `/home/runner/.dotnet`. Note: This change will be visible only when sourcing script.
dotnet-install: Installation finished successfully.
/home/runner/.dotnet:/home/linuxbrew/.linuxbrew/bin:/home/linuxbrew/.linuxbrew/sbin:/opt/pipx_bin:/usr/share/rust/.cargo/bin:/home/runner/.config/composer/vendor/bin:/usr/local/.ghcup/bin:/home/runner/.dotnet/tools:/snap/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin
```

* output from now failing `actions/setup-dotnet@v1`:
```
/home/runner/work/_actions/actions/setup-dotnet/v1/externals/install-dotnet.sh --version 5.0.201
dotnet_install: Warning: Unable to locate zlib. Probable prerequisite missing; install zlib.
dotnet-install: Downloading link: https://dotnetcli.azureedge.net/dotnet/Sdk/5.0.201/dotnet-sdk-5.0.201-linux-x64.tar.gz
dotnet-install: Extracting zip from https://dotnetcli.azureedge.net/dotnet/Sdk/5.0.201/dotnet-sdk-5.0.201-linux-x64.tar.gz
dotnet-install: Adding to current process PATH: `/home/runner/.dotnet`. Note: This change will be visible only when sourcing script.
[...]

PR #2982, opened 2021-03-18
[Ubuntu] Revert $BASH_ENV variable

miketimofeev:
# Description
It turned out that $BASH_ENV causes some issues since updated PATH is not propagated to other shells like pwsh.
Reverts https://github.com/actions/virtual-environments/pull/2826

#### Related issue:
https://github.com/actions/virtual-environments/issues/2978

## Check list

### DIFF
--- a/images/linux/scripts/installers/configure-environment.sh
+++ b/images/linux/scripts/installers/configure-environment.sh
@@ -7,14 +7,9 @@ echo ImageOS=$IMAGE_OS | tee -a /etc/environment
 # Set the ACCEPT_EULA variable to Y value to confirm your acceptance of the End-User Licensing Agreement
 echo ACCEPT_EULA=Y | tee -a /etc/environment
 
-# Create a file to store user-related global environment variables
-touch /etc/profile.d/env_vars.sh
-# Set BASH_ENV variable pointed to the file with user-related global environment variables for non-interactive sessions
-echo "BASH_ENV=/etc/profile.d/env_vars.sh" | tee -a /etc/environment
-
 # This directory is supposed to be created in $HOME and owned by user(https://github.com/actions/virtual-environments/issues/491)
 mkdir -p /etc/skel/.config/configstore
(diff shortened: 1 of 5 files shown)


al-cheb (approved):

maxim-lobanov (approved):

catthehacker:
@miketimofeev
*I've rambled a bit about `/etc/environment` && `/etc/profile` && shells here: https://github.com/nektos/act/pull/570#issuecomment-801560100*

For `pwsh` that is invoked without `-l`/`-login`, equivalent would be to create `$profile.AllUsersCurrentHost` or `$profile.AllUsersAllHosts` and add those variables in that file.
```powershell
$profile | gm -MemberType NoteProperty | % { return [pscustomobject]@{ Type = $_.Name ; Location = $profile."$($_.Name)" }}
```

image

e.g.:
```powershell
$env:XDG_CONFIG_HOME = "${env:HOME}/.config"
# or
$env:PATH = "$($env:PATH)$([IO.Path]::PathSeparator)$($env:HOME)/.dotnet/tools"
# or
$env:PATH = ('{0}{1}{2}/.dotnet/tools' -f $env:PATH, [IO.Path]::PathSeparator, $env:HOME)
```

Unfortunately I do not know how `zsh` will behave and how to fix it since I don't use it but from http://zsh.sourceforge.net/Intro/intro_3.html it appears that `.zshenv` could be used if it was put in `/etc/skel`.

cognifloyd:
Aargh. This just broke things for me because now ~/.local/bin isn't on the PATH anymore.

cognifloyd:
https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstepsrun
> Commands run using non-login shells by default

And on my ubuntu jobs, it is running this shell: `bash --noprofile --norc -eo pipefail {0}`

So, .bashrc doesn't get loaded now (`--norc`) whereas the profile.d entries WERE getting loaded by default.

miketimofeev:
@cognifloyd we're going to get it back soon, but using another approach. Sorry for the inconvenience

phuchau1989 (approved):

Issue #3001, opened 2021-03-22
ubuntu-latest has broken pre-installed ansible tool (please remove)


ssbarnea:
**Description**  
The preinstalled ansible on GHA is broken because it cannot be called from within an activated python virtualenvs. The root cause is a broken DEB package but pre-installing it is doing more damage.

**Area for Triage**:  

**Question, Bug, or Feature?**:  

**Virtual environments affected**  
- [x] Ubuntu 16.04
- [x] Ubuntu 18.04
- [x] Ubuntu 20.04

**Image version**
all ubuntu images

**Expected behavior**  
* Either not having ansible installed at all and relying on user to install it, or
* Install a working version of ansible and assure regression does not happen.

**Actual behavior**  

```bash
ansible --version # works
vritualenv foo
source foo/bin/activate
ansible --version
(foo) root@buntu:~# ansible --version
Traceback (most recent call last):
  File "/usr/bin/ansible", line 34, in <module>
    from ansible import context
ModuleNotFoundError: No module named 'ansible'
```

Please note that installing ansible using `pip` or `pipx` at `--user` or even root level does install a working copy of ansible, one that works regardless if a virtualenv is active or not.

As a first step, please remove the broken copy.

This bug impacts the testing of Python based projects where 99% are using virtualenvs for testing, likely with tox.

webknjaz:
I bet this problem is not specific to the Ansible package — Debian derivatives are know to ignore the existence of virtualenvs: https://gist.github.com/tiran/2dec9e03c6f901814f6d1e8dad09528e#gistcomment-3674863

webknjaz:
Have you checked if Ubuntu's `head -n /usr/bin/ansible` has `#!/usr/bin/python` or `#!/usr/bin/env python`? I bet it's the latter.

ssbarnea:
Yes, I can confirm that the broken Ubuntu one, is using `#!/usr/bin/env python` shebang line, which will obviously break inside activated virtualenvs.

While the env based shebang may be seen as portable and prefered for standalone scripts, it is clearly not a good idea for tools that rely on external libraries, as virtualenv activation will likely break those tools as they will endup calling another python interpreter, one that does not have access to their dependencies.

al-cheb:
Hey, @ssbarnea 
I think we can replace the ansible pkg from apt repo and install it using pipx:

```
    steps:
      - run: sudo apt remove ansible
      - name: ansible
        run: |
          sudo pipx install ansible-base
[...]

PR #3015, opened 2021-03-24
[Ubuntu] use pipx to install ansible

al-cheb:
# Description
- use pipx to install ansible on Ubuntu 18.04/20.04
- remove ppa:ansible/ansible on Ubuntu 18.04

#### Related issue:
https://github.com/actions/virtual-environments/issues/3001

## Check list

### DIFF
--- a/images/linux/toolsets/toolset-1804.json
+++ b/images/linux/toolsets/toolset-1804.json
@@ -224,6 +224,10 @@
         {
             "package": "aws-sam-cli",
             "cmd": "sam"
+        },
+        {
+            "package": "ansible-base",
+            "cmd": "ansible"
         }
     ],
     "dotnet": {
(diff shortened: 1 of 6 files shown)


images/linux/toolsets/toolset-1804.json
```diff
@@ -224,6 +224,10 @@
         {
             "package": "aws-sam-cli",
             "cmd": "sam"
+        },
```
maxim-lobanov, line 227: What about Ubuntu 16.04? Can we use the same approach for it?
If not - let's put some comment to the `ansible.sh` that this script is used only on Ubuntu 16.04 to avoid confusing

maxim-lobanov (approved):

miketimofeev (approved):

images/linux/toolsets/toolset-1804.json
ssbarnea: While I would prefer using the same method for both I think it may not be possible due to the ancient versions of python from Ubuntu 16.04 (all EOL!). pipx requires 3.6+ and even if there are probably some older version I doubt they will work fine. 

Shotly, I support this dual approach.

ssbarnea (approved):

Issue #2967, opened 2021-03-17
Chromium fails to launch in Ubuntu 20 due to permissions


amclin:
**Description**  
In Ubuntu 20, the Chromium browser installed fails to launch due to permissions (18 didn't have Chromium provided). GitHub actions just [switched from the Ubuntu 18 to 20 image as the default](https://github.com/actions/virtual-environments/issues/1816).

As a result, in pipelines where available browsers are auto-detected, Chromium is now detected, but when it is launched it fails due to permission errors when attempting to create the profile directory.

```
17 03 2021 13:18:07.841:INFO [framework.detect-browsers]: which.sync(process.env[browser.ENV_CMD]):  /usr/bin/google-chrome
17 03 2021 13:18:07.844:INFO [framework.detect-browsers]: The following browsers will be used: [ 'Chrome', 'Chromium', 'Firefox' ]
17 03 2021 13:18:08.942:INFO [karma-server]: Karma v6.2.0 server started at http://localhost:9876/
17 03 2021 13:18:08.942:INFO [launcher]: Launching browsers Chrome, Chromium, Firefox with concurrency unlimited
17 03 2021 13:18:08.951:INFO [launcher]: Starting browser Chrome
17 03 2021 13:18:08.965:INFO [launcher]: Starting browser Chromium
17 03 2021 13:18:08.978:INFO [launcher]: Starting browser Firefox
17 03 2021 13:18:11.821:ERROR [launcher]: Cannot start Chromium
	/bin/bash: /etc/profile.d/env_vars.sh: Permission denied
mkdir: cannot create directory ‘/run/user/1001’: Permission denied
No protocol specified
[3065:3065:0317/131811.781769:ERROR:browser_main_loop.cc(1390)] Unable to open X display.
```

**Area for Triage**:  
['Area: JavaScript and Node.js']['Area: Testing and code coverage']

**Question, Bug, or Feature?**:  
Bug

**Virtual environments affected**  
- [ ] Ubuntu 16.04
- [ ] Ubuntu 18.04
- [x] Ubuntu 20.04
- [ ] macOS 10.15
- [ ] macOS 11.0
- [ ] Windows Server 2016 R2
- [ ] Windows Server 2019

**Image version**

`20210309.1`

**Expected behavior**  
Chromium launches consistent with Chrome and Firefox launching

**Actual behavior**  
Chromium fails to launch with permission errors for the profile paths

**Repro steps**  
[...]

PR #3029, opened 2021-03-25
[Ubuntu] Rework Сhromium installation

nikita-bykov:
# Description
Changing the Chromium installation from `apt-get` to manually, because even using `apt-get install` Chromium was installed as a `snap` package. It led to an issue like `mkdir: cannot create directory ‘/run/user/1001’: Permission denied` when we tried to use Chromium.

#### Related issue:
#2967
## Check list


miketimofeev:
@nikita-bykov @maxim-lobanov I wonder if we should take the version corresponding to the google chrome one as it is described here https://www.chromium.org/getting-involved/download-chromium
image

maxim-lobanov:
> @nikita-bykov @maxim-lobanov I wonder if we should take the version corresponding to the google chrome one as it is described here https://www.chromium.org/getting-involved/download-chromium
> image

@miketimofeev @nikita-bykov 
I think it makes sense to make sure that Chromium is compatible with ChromeDriver.
Looks like we just need to retrieve installed Chrome version and put it to the url: `https://omahaproxy.appspot.com/deps.json?version=44.0.2403.157`. This url will return `chromium_base_position` that we should use instead of LAST CHANGE content.

images/linux/scripts/installers/chromium.sh
```diff
@@ -1,11 +1,35 @@
...
+        if [ $revision -lt  $version ]; then
+            echo $RIGHT_REVISION
+            return
+        fi
+        RIGHT_REVISION=$version
+    done
+    echo $RIGHT_REVISION
+}
+
+CHROME_VERSION=$(google-chrome --version | cut -d " " -f 3)
```
miketimofeev, line 26: I would prefer to get rid of this dependency and get the latest chrome version from the internet

nikita-bykov: Changed it to get the version from `http://omahaproxy.appspot.com/`

maxim-lobanov: I think we should install Chromium based on installed Chrome version because we install Chrome Driver that is available with Chrome. So Chromium should be the same version to be compatible with Chrome Driver

maxim-lobanov: Why?
We install Chrome from apt. Then we find compatible version of ChromeDriver based on installed Chrome. Chromium should be the same version to be compatible.

If we retrieve Chrome version from the internet - we can catch the case where Chromium and ChromeDriver are not compatible.
Like apt still contains Chrome 89.x and we install ChromeDriver 89.x but `http://omahaproxy.appspot.com/` or other remote site already contains Chrome 90.x that is not uploaded to apt yet. In this case, we will install Chromium 90.x that is not compatible with ChromeDriver.

It is better to rely on Chrome version that is already installed on image because we 100% sure that ChromeDriver is compatible with it.

images/linux/scripts/installers/chromium.sh
```diff
@@ -0,0 +1,44 @@
...
+
+source $HELPER_SCRIPTS/install.sh
+
+function GetChromiumRevision {
+    CHROME_VERSION=$1
+
+    URL="https://omahaproxy.appspot.com/deps.json?version=${CHROME_VERSION}"
+    REVISION=$(curl -s $URL | jq -r '.chromium_base_position')
+    # Take the first part of revision variable for search
+    FIRST_PART_OF_REVISION=${REVISION:0:${#REVISION}/2}
```
maxim-lobanov, line 11: May be `FIRST_PART_OF_REVISION=$(echo $REVISION | cut -c1-3)` looks simpler

images/linux/scripts/installers/chromium.sh
miketimofeev: @maxim-lobanov we install chrome from the deb package now https://github.com/actions/virtual-environments/pull/3036
If we rely on chrome then we depend on the other provisioning script and we already have a few complaints about such unclear dependencies

maxim-lobanov: Yep, and I am not sure we can retrieve exact version from the link https://dl.google.com/linux/direct/google-chrome-stable_current_amd64.deb
Retrieving version from any other resource won't work for us because they can have different versions

In my opinion, it is better to depend on other provisioning script than having Chromium that is not compatible with ChromeDriver.

As an option, we can install chromium in the same script like Chrome and ChromeDriver but I am not sure that it will be better.

miketimofeev: Hmm, I like the idea to move chromium installation to the chrome provisioning script. In that case, everything will be in a single script

AlenaSviridenko:
Is it correct that we will now install Chromium to all Ubuntu versions and not on Ubuntu 20.04 only?

nikita-bykov:
> 
> 
> Is it correct that we will now install Chromium to all Ubuntu versions and not on Ubuntu 20.04 only?

We now have enough free space on all Ubuntu images, so maybe we can add Chromium not only on Ubuntu 20. If not, I'll add a condition to install only on Ubuntu 20.

images/linux/scripts/installers/google-chrome.sh
```diff
@@ -35,4 +56,21 @@ chmod +x $CHROMEDRIVER_BIN
...
+CHROMIUM_DIR="/usr/local/share/chromium"
+CHROMIUM_BIN="$CHROMIUM_DIR/chrome-linux/chrome"
+
+# Download and unzip Chromium archive
+download_with_retries $ZIP_URL "/tmp" $ZIP_FILE
+mkdir $CHROMIUM_DIR
+unzip /tmp/${ZIP_FILE} -d $CHROMIUM_DIR
+
+ln -s $CHROMIUM_BIN /usr/bin/chromium
+
```
al-cheb, line 75: Looks like `invoke_tests "Browsers" "Chromium"`  test is missing.

images/linux/scripts/installers/google-chrome.sh
```diff
@@ -35,4 +56,21 @@ chmod +x $CHROMEDRIVER_BIN
...
+ZIP_URL="https://www.googleapis.com/download/storage/v1/b/chromium-browser-snapshots/o/Linux_x64%2F${REVISION}%2Fchrome-linux.zip?alt=media"
+ZIP_FILE="${REVISION}-chromium-linux.zip"
+
+CHROMIUM_DIR="/usr/local/share/chromium"
+CHROMIUM_BIN="$CHROMIUM_DIR/chrome-linux/chrome"
+
+# Download and unzip Chromium archive
+download_with_retries $ZIP_URL "/tmp" $ZIP_FILE
+mkdir $CHROMIUM_DIR
+unzip /tmp/${ZIP_FILE} -d $CHROMIUM_DIR
```
al-cheb, line 72: Please use  -qq to suppress the output.

miketimofeev:
> > Is it correct that we will now install Chromium to all Ubuntu versions and not on Ubuntu 20.04 only?
> 
> We now have enough free space on all Ubuntu images, so maybe we can add Chromium not only on Ubuntu 20. If not, I'll add a condition to install only on Ubuntu 20.

Change the PR title, please

images/linux/scripts/installers/google-chrome.sh
nikita-bykov: Done

al-cheb (approved):

miketimofeev (approved):

miketimofeev:
@nikita-bykov @maxim-lobanov we need to change the condition in software report generator since we've decided to add chromium everywhere
https://github.com/actions/virtual-environments/blob/3195add3cdc31759b7f2999b84192c703f2fb620/images/linux/scripts/SoftwareReport/SoftwareReport.Generator.ps1#L220-L222

miketimofeev (changes requested):
Please change the condition in software generator
https://github.com/actions/virtual-environments/blob/3195add3cdc31759b7f2999b84192c703f2fb620/images/linux/scripts/SoftwareReport/SoftwareReport.Generator.ps1#L220-L222

maxim-lobanov:
> Please change the condition in software generator
> https://github.com/actions/virtual-environments/blob/3195add3cdc31759b7f2999b84192c703f2fb620/images/linux/scripts/SoftwareReport/SoftwareReport.Generator.ps1#L220-L222

The same for tests: https://github.com/actions/virtual-environments/blob/main/images/linux/scripts/tests/Browsers.Tests.ps1#L21

miketimofeev:
> > Please change the condition in software generator
> > https://github.com/actions/virtual-environments/blob/3195add3cdc31759b7f2999b84192c703f2fb620/images/linux/scripts/SoftwareReport/SoftwareReport.Generator.ps1#L220-L222
> 
> The same for tests: https://github.com/actions/virtual-environments/blob/main/images/linux/scripts/tests/Browsers.Tests.ps1#L21

it's already done here, don't know why the test moved to the chrome section, but anyway
https://github.com/actions/virtual-environments/pull/3029/files#diff-4c476e8348e741443009d80dc08c7f6b6913320238aae669042cd9da015fe952L21

maxim-lobanov (approved):

Issue #2954, opened 2021-03-17
Value of environment variable XDG_CONFIG_HOME is always "/home/runner/.config"


shibataka000:
**Description**  
I set some value to environment variable `XDG_CONFIG_HOME` by `jobs.<job_id>.env.XDG_CONFIG_HOME` . But value of environment variable `XDG_CONFIG_HOME` is always `/home/runner/.config` .

**Area for Triage**:  

Servers

**Question, Bug, or Feature?**:  

Bug

**Virtual environments affected**  
- [ ] Ubuntu 16.04
- [ ] Ubuntu 18.04
- [x] Ubuntu 20.04
- [ ] macOS 10.15
- [ ] macOS 11.0
- [ ] Windows Server 2016 R2
- [ ] Windows Server 2019

**Image version**

20210315.1

**Expected behavior**  
Value of environment variable `XDG_CONFIG_HOME` should be same as value of `jobs.<job_id>.env.XDG_CONFIG_HOME` .

**Actual behavior**  
Value of environment variable `XDG_CONFIG_HOME` is always `/home/runner/.config` .

**Repro steps**  
1. Set up workflows like follow.

```yaml
# .github/workflows/test.yaml
name: test
on: [push]
jobs:
  test:
    runs-on: ubuntu-latest
    env:
      XDG_CONFIG_HOME: /github/home/.config
    steps:
      - run: echo $XDG_CONFIG_HOME
```

2. See workflow output.

I created sample repository in https://github.com/shibataka000/github-actions-issue-2826 . You can see [workflow setting](https://github.com/shibataka000/github-actions-issue-2826/blob/main/.github/workflows/test.yaml) and [its result](https://github.com/shibataka000/github-actions-issue-2826/actions/runs/660438764).

shibataka000:
I guess this issue might be related to #2826 .

miketimofeev:
Hi @shibataka000!
Yes, you're right it's related. The variable didn't work at all before https://github.com/actions/virtual-environments/pull/2826
but now it looks like the variable from $BASH_ENV overrides the variable from the env block. Sorry for the inconvenience. Could you please try the following?
```
jobs:
  test:
    runs-on: ubuntu-latest
    steps:
      - run: echo "export XDG_CONFIG_HOME=/github/home/.config" | sudo tee -a $BASH_ENV
      - run: echo $XDG_CONFIG_HOME
```

shibataka000:
[...]

PR #3030, opened 2021-03-25
[Ubuntu] Set user-related environmental variables and PATH via post deployment script

miketimofeev:
# Description
It turned out that we can't use the `$BASH_ENV` variable because the values from it can't be overridden by workflow's `env` block.
This PR introduces another approach in which we set the appropriate variables during the machine provisioning process via a post-generation script.

#### Related issue:
https://github.com/actions/virtual-environments/issues/2954

## Check list


images/linux/post-generation/environment-variables.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+# Replace $HOME with the default user's home directory for environmental variables related to the default user home directory
+
+default_user=$(cut -d: -f1 /etc/passwd | tail -1)
```
al-cheb, line 5: We can request user home folder:
`homeDir=$(cut -d: -f6 /etc/passwd | tail -1)`

miketimofeev: Thanks!

maxim-lobanov (approved):

maxim-lobanov (approved):

EmFl:
What about changes from #2826 ? 
Right now on my system I get:
/home/ubuntu/$HOME/.config/configstore
/etc/environment contains XDG_CONFIG_HOME=$HOME/.config

Was the rollback intentional ?

miketimofeev:
@EmFl yes it was because $BASH_ENV causes some issues since updated PATH is not propagated to other shells like pwsh. 
Please consider running post-deployment scripts on your system — this one substitutes the values in /etc/environment with the actual username https://github.com/actions/virtual-environments/blob/main/images/linux/post-generation/environment-variables.sh

PR #3062, opened 2021-03-30
fix: Report correct Erlang version, fix Erlang tests

catthehacker:
# Description

SoftwareReport collects wrong version for Erlang, it reports Eshell version rather than Erlang itself.
reference: https://stackoverflow.com/questions/9560815/how-to-get-erlangs-release-version-number-from-a-shell
2nd answer

`erl` doesn't have `-v` flag, only `-version`
`erlc` `-v` flag is for verbose output
`rebar3` reports it's version with flag `-v`
Reverting back the erlang test logic to test it via `command -v` as it breaks during packer build.
reference: http://erlang.org/doc/man/erl.html

## Check list

### DIFF
--- a/images/linux/scripts/SoftwareReport/SoftwareReport.Common.psm1
+++ b/images/linux/scripts/SoftwareReport/SoftwareReport.Common.psm1
@@ -34,8 +34,9 @@ function Get-ClangVersions {
 }
 
 function Get-ErlangVersion {
-    $version = (erl -eval 'erlang:display(erlang:system_info(version)), halt().' -noshell).Trim('"')
-    return "Erlang $version"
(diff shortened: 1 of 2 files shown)


catthehacker:
Previous logic for `erl`, `erlc` and `rebar3` was defined here:

https://github.com/actions/virtual-environments/commit/a9a236f7e5125b296edf587f38ef82cb1db4096e#diff-a4df2483eb99eddfab8faca6e043daaa0052e2d244b47d4dcc73cb208f5f9184L20-L27

If `command -v` is not enough for test, `erl` can be tested with `-version` flag, `erlc` with `-help` and `rebar3` with `-v`

al-cheb:
I think the command `"erlc -help"` will return 1 not 0.

miketimofeev (approved):

maxim-lobanov (approved):

PR #3077, opened 2021-04-01
[Ubuntu] Remove apt source for podman, buildah and skopeo

dibir-magomedsaygitov:
# Description
In scope of this PR we are simplifying `podman`, `buildah `and `skopeo` installation on `ubuntu20`. As for `ubuntu16` and `ubuntu18` we have to use current installation approach.

## Check list

### DIFF
--- a/images/linux/scripts/installers/containers.sh
+++ b/images/linux/scripts/installers/containers.sh
@@ -7,20 +7,24 @@
 source $HELPER_SCRIPTS/os.sh
 
 install_packages=(podman buildah skopeo)
+REPO_URL="https://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable"
 
 # Install podman, buildah, scopeo container's tools (on Ubuntu20 these tools can be installed without adding new repository)
+source /etc/os-release
+sh -c "echo 'deb ${REPO_URL}/x${NAME}_${VERSION_ID}/ /' > /etc/apt/sources.list.d/devel:kubic:libcontainers:stable.list"
+wget -qnv https://download.opensuse.org/repositories/devel:kubic:libcontainers:stable/x${NAME}_${VERSION_ID}/Release.key -O Release.key
+apt-key add Release.key
+apt-get update -qq
+apt-get -qq -y install ${install_packages[@]}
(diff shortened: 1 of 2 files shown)


images/linux/scripts/installers/containers.sh
```diff
@@ -15,4 +17,8 @@ apt-get -qq -y install ${install_packages[@]}
 mkdir -p /etc/containers
 echo -e "[registries.search]\nregistries = ['docker.io', 'quay.io']" | tee /etc/containers/registries.conf
 
+if isUbuntu20 ; then
+    rm /etc/apt/sources.list.d/devel:kubic:libcontainers:stable.list
```
miketimofeev, line 21: we need to document it somehow, not just remove from the list

images/linux/scripts/installers/containers.sh
```diff
@@ -4,6 +4,8 @@
 ##  Desc:  Installs container tools: podman, buildah and skopeo onto the image
 ################################################################################
 
+source $HELPER_SCRIPTS/os.sh
```
maxim-lobanov, line 7: Looks like we even don't need any new repository for 20.04. See installation guide: https://podman.io/getting-started/installation

Let's get rid of it

images/linux/scripts/installers/containers.sh
maxim-lobanov: Agree, May be we can collect the list of repos during image generation to some separate file like `echo "repo1" >> /image-generation/repos.txt` and attach this file to readme later?

maxim-lobanov: Btw, why we remove repo only on Ubuntu 20?

images/linux/scripts/installers/containers.sh
dibir-magomedsaygitov: Sadly, but `Ubuntu20 `still requires source repo to install these tools

al-cheb (approved):

miketimofeev (approved):

dmitry-shibanov (approved):

maxim-lobanov (approved):

PR #3083, opened 2021-04-01
Add instructions of how to use the Postgres database on Linux

fabioperrella:
# Description

This PR adds instructions on how to set up the Postgres database that is installed in Linux images.

Maybe there is a simpler/better way to do it, but this is the best I could do! 

Another alternative would be changing `images/linux/scripts/installers/postgresql.sh` to create a user `root` with password `root`, similar to MySql. Please, let me know if you think this approach is better!

## Check list

### DIFF
--- a/images/linux/scripts/SoftwareReport/SoftwareReport.Databases.psm1
+++ b/images/linux/scripts/SoftwareReport/SoftwareReport.Databases.psm1
@@ -38,6 +38,27 @@ function Build-MySQLSection {
     return $output
 }
 
+function Build-PostgreSqlSection {
+    $output = ""
+
+    $output += New-MDHeader "PostgreSql" -Level 4
+    $output += New-MDList -Style Unordered -Lines @(
+        (Get-PostgreSqlVersion ),
+        "PostgreSql Server (user:root password:root)"
+    )
+    $output += New-MDCode -Lines @(@"
+        To create the root user and enable the service, use the following commands:
+        sudo systemctl start postgresql.service
+        echo "create user root password 'root';" > /tmp/create_user.sql
+        echo "ALTER ROLE root WITH CREATEDB;" >> /tmp/create_user.sql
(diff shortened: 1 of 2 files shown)


AlenaSviridenko:
Hi @fabioperrella,
thank you for your contribution, but unfortunately, we cannot accept it due to a couple of reasons:
1. Documenting tool usage is out of scope of our README files if it doesn't differ from the standard usage or configuration, because in that case it is described in the official tool docs. 
2. We also worried about adding a default user `root` because it can break existing customers' scenarios, and we cannot predict the impact of this change.

please, let me know if you have any questions

fabioperrella:
tks for your feedback @AlenaSviridenko !

What if we create a different new user, for example with the name `user` and password `user`?

Maybe you can suggest a better name 😄 .

Giving more context about it, when I used the MySQL database, it was pretty easy because it was explicit what user and password to use, but when I tried Postgres, I struggled a lot to find out what user/password I should use.

whogonestopme662901 (approved):

AlenaSviridenko:
@fabioperrella, we also cannot be sure about this user, what if someone creates it in the workflow. Basically, we cannot be sure about any user 🙂
I understand that we precreated a user for MySQL, but it was done a long time ago, and we had several issues regarding different somehow preconfigured services, so we decided that it is more safe to not configure services by ourselves and leave setup flexibility to customers.

fabioperrella:
> @fabioperrella, we also cannot be sure about this user, what if someone creates it in the workflow. Basically, we cannot be sure about any user 🙂
> I understand that we precreated a user for MySQL, but it was done a long time ago, and we had several issues regarding different somehow preconfigured services, so we decided that it is more safe to not configure services by ourselves and leave setup flexibility to customers.

yes, you're right!

So, I will close this PR and create a post in my blog about it, then I hope people can find it when they search for it (I couldn't 😄)

Thank you!

PR #3095, opened 2021-04-05
[Ubuntu] Rework Git installation to avoid using apt repo

dibir-magomedsaygitov:
# Description
In scope of this PR we are reworking `git` and `git-lfs` to avoid using apt repo.

## Check list

### DIFF
--- a/images/linux/scripts/installers/git.sh
+++ b/images/linux/scripts/installers/git.sh
@@ -28,8 +28,8 @@ add-apt-repository --remove $GIT_REPO
 rm /etc/apt/sources.list.d/github_git-lfs.list
 
 # Document apt source repo's
-echo "git $GIT_REPO" >> $HELPER_SCRIPTS/apt-sources.txt
-echo "lfs $GIT_LFS_REPO" >> $HELPER_SCRIPTS/apt-sources.txt
+echo "git-core $GIT_REPO" >> $HELPER_SCRIPTS/apt-sources.txt
+echo "git-lfs $GIT_LFS_REPO" >> $HELPER_SCRIPTS/apt-sources.txt
 
 #Install hub
 tmp_hub="/tmp/hub"
--- a/images/linux/scripts/installers/git.sh
+++ b/images/linux/scripts/installers/git.sh
@@ -13,7 +13,6 @@ apt-get install git -y
 git --version
 
 # Install git-lfs
-apt-get update -y
 apt-get install -y git-lfs
 
 # Install git-ftp
--- a/images/linux/scripts/installers/git.sh
+++ b/images/linux/scripts/installers/git.sh
@@ -8,14 +8,21 @@
 source $HELPER_SCRIPTS/install.sh
 
 ## Install git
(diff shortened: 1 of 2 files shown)


images/linux/scripts/installers/git.sh
```diff
@@ -8,14 +8,21 @@
...
+latest_git_version=$(curl -s -L "https://api.github.com/repos/git/git/tags" | jq -r '.[].name' | grep -ve "-.*" | head -1 | tr -d "v")
+download_with_retries https://github.com/git/git/archive/v${latest_git_version}.tar.gz "/tmp" "git.tar.gz"
+git_installation_path="/usr/local/git"
+git_tar_tmp="/tmp/git.tar.gz"
+mkdir -p "${git_installation_path}"
+tar -zxf "${git_tar_tmp}" -C "${git_installation_path}"
+rm -f "${git_tar_tmp}"
+cd "${git_installation_path}/git-${latest_git_version}"
+make prefix=/usr/local all
+make prefix=/usr/local install
```
maxim-lobanov, line 20: Do we really need to build git from source code during image-generation? I don't think that it is good solution.
I think both `git` and `git-lfs` should be available in default apt sources without adding additional repos and installation will be faster

dibir-magomedsaygitov: @maxim-lobanov thank you, simplified installation

images/linux/scripts/installers/git.sh
```diff
@@ -8,14 +8,13 @@
...
 ## Install git
-add-apt-repository ppa:git-core/ppa -y
 apt-get update
 apt-get install git -y
 git --version
 
 # Install git-lfs
-curl -s https://packagecloud.io/install/repositories/github/git-lfs/script.deb.sh | bash
-apt-get install -y --no-install-recommends git-lfs
+apt-get update -y
```
maxim-lobanov, line 16: I think we don't need to run apt-get update one more time. We have already run it on line 11

maxim-lobanov (approved):

images/linux/scripts/installers/git.sh
```diff
@@ -8,14 +8,12 @@
 source $HELPER_SCRIPTS/install.sh
 
 ## Install git
-add-apt-repository ppa:git-core/ppa -y
 apt-get update
 apt-get install git -y
```
miketimofeev, line 12: is the version in the repo fresh enough? I think it's pretty outdated at least for ubuntu 16 & 18

images/linux/scripts/installers/git.sh
```diff
@@ -8,14 +8,12 @@
...
 git --version
 
 # Install git-lfs
-curl -s https://packagecloud.io/install/repositories/github/git-lfs/script.deb.sh | bash
-apt-get install -y --no-install-recommends git-lfs
+apt-get install -y git-lfs
```
miketimofeev, line 16: same here

al-cheb (approved):

dmitry-shibanov (approved):

miketimofeev (approved):

Issue #11, opened 2019-10-03
Better vcpkg support


jherico:
The vcpkg binary exists on the target environments, but it would be _much_ more useful if the YAML could specify a specific vcpkg commit, and a set of packages required for a given build.  If the server side could recognize and cache this state, it could potentially shave an enormous amount of time off of individual builds.

**Tool information**
- Tool name: vcpkg
- Add or update? update (new functionality)

**Virtual environments affected**
- [X] macOS 10.14
- [X] Ubuntu 16.04 LTS
- [X] Ubuntu 18.04 LTS
- [X] Windows Server 2016 R2
- [X] Windows Server 2019

**Can this tool be installed during the build?**
Yes

**Are you willing to submit a PR?**
Once caching support exists in the GA syntax, yes, I would be willing to write a tool like this.

simonsan:
I support this opinion. We're also using **vcpkg** to build dependencies ourselves, this step actually takes 10 min and is not cachable because it's outside of **GITHUB_WORKSPACE**. 

We are using `dirent eigen3 fontconfig freetype harfbuzz libepoxy libogg libpng opus opusfile sdl2 sdl2-image` so nothing special, would be nice if there would be some more packages cached or if the **cache** action gets an option to cache also the **vcpkg-dir**.

stevencpp:
It would be helpful if someone could make an action that makes this easier, but I did it without one here: https://github.com/stevencpp/cpp_modules/blob/master/.github/workflows/build.yml . I can specify the hash of the vcpkg commit and the list of dependencies as environment variables. I can collect all of the information (compiler version, CXXFLAGS, the deps and the commit id) into a file that the cache action can use as part of the key. I used the preinstalled binary instead of bootstrapping it but chose to clone into a subfolder of the workspace which is then cached.

simonsan:
That's pretty neat, well done! The problem is also, that the individual cache size is max. **400 MiB**, I'm pretty sure, that we are way on top of that. Not sure, if that solution is working out. Will try, though. Thank you!

adam4813:
You just need to cache the `installed` directory under vcpkg for proper caching (saves space bot caching the buildtrees directory).

simonsan:
Yeah, this would be still a great enhancement to the GHA...

maxim-lobanov:
Hello, have you looked at https://github.com/actions/cache , I believe it could help you cache packages and reduce the time

alepauly:
[...]

PR #3193, opened 2021-04-15
Securing packer builds via allowed_inbound_ip_addresses

seqdan:
# Description
Improvement: As detailed in https://github.com/actions/virtual-environments/discussions/3192 this is an attempt to secure the packer build machine without virtual networks. This is achieved by using the `allowed_inbound_ip_addresses` packer parameter and passing the build agent's public IP address over as a value.

The feature got a switch in `GenerateResourcesAndImage.ps1` so that it is fully backwards-compatible (default = off) and can be switched on where needed.

Tests for the win2019 server image are in progress; we can probably skip tests for the other images if one passes?! At the time of writing this, the test has successfully passed the critical stage where the VM is created and secured.

## Check list
- Related issue / work item is attached
- Tests are being executed
- Documentation is updated (parameter list in powershell script)
- [ ... ] Changes are tested and related VM images are successfully generated (in progress)


seqdan:
Ah, it was ready, but after merging `main` into this, I can't run it anymore. Investigating...

seqdan:
image
image was successfully built

Rada8732 (approved):

SonjayaJetBrain (approved):

da-art85 (approved):

maxim-lobanov (approved):

helpers/GenerateResourcesAndImage.ps1
```diff
@@ -89,6 +89,9 @@ Function GenerateResourcesAndImage {
         .PARAMETER AzureTenantId
             Tenant needs to be provided for optional authentication via service principal. Example: "11111111-1111-1111-1111-111111111111"
 
+        .PARAMETER RestrictToAgentIpAddress
+            If set, access to the VM used by packer to generate the image is restricted to the public IP address this script is run from
```
miketimofeev, line 93: I think we should add a notice here that the parameter can't be used in combination with `virtual_network_name`, which also exists in our template
image

seqdan: Thanks, good point. I will pre-check the wording with my colleagues here and then push a suggestion.

seqdan: Done; let me know if the updated text suits!

helpers/GenerateResourcesAndImage.ps1
```diff
@@ -89,6 +89,9 @@ Function GenerateResourcesAndImage {
         .PARAMETER AzureTenantId
             Tenant needs to be provided for optional authentication via service principal. Example: "11111111-1111-1111-1111-111111111111"
 
+        .PARAMETER RestrictToAgentIpAddress
+            If set, access to the VM used by packer to generate the image is restricted to the public IP address this script is run from. This parameter cannot be used in combination with the virtual_network_name packer parameter.
```
miketimofeev, line 93:
```suggestion
            If set, access to the VM used by packer to generate the image is restricted to the public IP address this script is run from. 
            This parameter cannot be used in combination with the virtual_network_name packer parameter.
```
I would prefer to have it on the new line 😅

seqdan: Sure!

seqdan:
Is there anything I can do to make this PR more compliant or nicer?

nohhhon (approved):
NIAP approval in accordance with CNSSP #11. This was primarily due to the potential interconnection of separate networks designated as National Security Systems

PR #3366, opened 2021-05-11
[Ubuntu] Remove pip symlink creation

miketimofeev:
# Description
Ubuntu 20 build fails with:
`ln: failed to create symbolic link '/usr/bin/pip': File exists`
There is no need to create pip3 -> pip symlink anymore since pip executable is now included in the python3-pip package.

#### Related issue:
https://github.com/actions/virtual-environments-internal/issues/2157

## Check list

### DIFF
--- a/images/linux/scripts/installers/python.sh
+++ b/images/linux/scripts/installers/python.sh
@@ -11,13 +11,10 @@ source $HELPER_SCRIPTS/os.sh
 
 # Install Python, Python 3, pip, pip3
 if isUbuntu16 || isUbuntu18; then
-    apt-get install -y --no-install-recommends python python-dev python-pip python3 python3-dev python3-pip python3-venv
+    apt-get install -y --no-install-recommends python python-dev python-pip
 fi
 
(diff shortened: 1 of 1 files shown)


images/linux/scripts/installers/python.sh
```diff
@@ -11,13 +11,10 @@ source $HELPER_SCRIPTS/os.sh
 
 # Install Python, Python 3, pip, pip3
 if isUbuntu16 || isUbuntu18; then
-    apt-get install -y --no-install-recommends python python-dev python-pip python3 python3-dev python3-pip python3-venv
```
maxim-lobanov, line 14: Why do we remove Python 3 packages?

miketimofeev: @maxim-lobanov we don't remove them. I moved the installation to the line 17 since we don't need any condition for python3

maxim-lobanov (approved):

al-cheb (approved):

Issue #3432, opened 2021-05-21
Ubuntu-18.04 is too intruisive - forcing non compatible libstdc++ by default for c++


Milerius:
**Description**  

If you go start a fresh ubuntu 18.04 machine you should have the following:

```
GLIBCXX_3.4
GLIBCXX_3.4.1
GLIBCXX_3.4.2
GLIBCXX_3.4.3
GLIBCXX_3.4.4
GLIBCXX_3.4.5
GLIBCXX_3.4.6
GLIBCXX_3.4.7
GLIBCXX_3.4.8
GLIBCXX_3.4.9
GLIBCXX_3.4.10
GLIBCXX_3.4.11
GLIBCXX_3.4.12
GLIBCXX_3.4.13
GLIBCXX_3.4.14
GLIBCXX_3.4.15
GLIBCXX_3.4.16
GLIBCXX_3.4.17
GLIBCXX_3.4.18
GLIBCXX_3.4.19
GLIBCXX_3.4.20
GLIBCXX_3.4.21
GLIBCXX_3.4.22
GLIBCXX_3.4.23
GLIBCXX_3.4.24
GLIBCXX_3.4.25
```

However on your machine we got:

```
GLIBCXX_3.4
GLIBCXX_3.4.1
GLIBCXX_3.4.2
GLIBCXX_3.4.3
GLIBCXX_3.4.4
GLIBCXX_3.4.5
GLIBCXX_3.4.6
GLIBCXX_3.4.7
GLIBCXX_3.4.8
GLIBCXX_3.4.9
GLIBCXX_3.4.10
GLIBCXX_3.4.11
GLIBCXX_3.4.12
GLIBCXX_3.4.13
GLIBCXX_3.4.14
GLIBCXX_3.4.15
GLIBCXX_3.4.16
GLIBCXX_3.4.17
GLIBCXX_3.4.18
GLIBCXX_3.4.19
GLIBCXX_3.4.20
GLIBCXX_3.4.21
GLIBCXX_3.4.22
GLIBCXX_3.4.23
GLIBCXX_3.4.24
GLIBCXX_3.4.25
GLIBCXX_3.4.26
GLIBCXX_3.4.27
GLIBCXX_3.4.28
GLIBCXX_3.4.29
```

with `GLIBCXX_3.4.29` -> https://gcc.gnu.org/onlinedocs/libstdc++/manual/abi.html -> break abi forcing end user that download a app that have been build on github actions to have gcc 11.1.0 installed

[...]

Issue #3376, opened 2021-05-13
Ubuntu-18.04 20210510 broke Clang 3.9 and 9


sfan5:
**Description**  
Since image version `20210510` both clang-3.9 and clang-9 are broken, specifically compilation errors in the C++ stdlib and binaries that segfault.
With no changes to code or CI `20210504.1` does _not_ exhibit this issue, hence this must be a bug in this repo.

**Area for Triage**:  
C/C++

**Question, Bug, or Feature?**:  
Bug

**Virtual environments affected**  
- [x] Ubuntu 18.04

**Image version**
  Version: 20210510.0

**Expected behavior**  
Clang 3.9 and 9 compile C++ code without erroring and/or the resulting binaries segfaulting.
This is the case on `20210504.1` (https://github.com/minetest/minetest/runs/2535472812).

**Actual behavior**  
* Clang 3.9 throws the following warning on every C++ source file: (https://github.com/minetest/minetest/pull/11263/checks?check_run_id=2575483229)
<pre>
[...]
In file included from /usr/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/cstdint:38:
/usr/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/x86_64-linux-gnu/c++/11/bits/c++config.h:558:2: warning: "__STRICT_ANSI__ seems to have been undefined; this is not supported" [-W#warnings]
#warning "__STRICT_ANSI__ seems to have been undefined; this is not supported"
 ^
1 warning generated.
</pre>
...also leading to errors in some cases:
<pre>
In file included from /home/runner/work/minetest/minetest/src/serverenvironment.cpp:20:
In file included from /usr/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/algorithm:61:
In file included from /usr/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/bits/stl_algobase.h:63:
/usr/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/ext/numeric_traits.h:64:7: error: static_assert failed "invalid specialization"
      static_assert(__is_integer_nonstrict<_Value>::__value,
      ^             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[...]

PR #3291, opened 2021-04-30
Ubuntu Bionic/Focal: Add GCC 11

EwoutH:
Add GCC 11 to the Ubuntu 18.04 Bionic and 20.04 Focal toolsets.

Closes #3280.

Please run `/azp run ubuntu1804, ubuntu2004` to test the images.

We probably should also start talking about a date to make GCC 11 the default.

------
### Check list
- [x] Related issue / work item is attached
- [x] Tests are written: By CI
- [x] Documentation is updated: By CI
- [x] Changes are tested and related VM images are successfully generated

PR #3436, opened 2021-05-22
Revert "[ubuntu] Add GCC 11 (#3291)"

jesec:
This reverts commit 3ebf601284cae6a4406b7ea1ad08cd593a4d619b.

# Description

Bug fix

It is a not good idea to add experimental toolchain to the image, as the addition/modification of standard C/C++ headers and libraries can have wide-ranging (and often undesired) consequences across the whole system.

#### Related issue:

#3376

### DIFF
--- a/images/linux/toolsets/toolset-1804.json
+++ b/images/linux/toolsets/toolset-1804.json
@@ -249,15 +249,13 @@
     "gcc": {
         "versions": [
             "g++-9",
-            "g++-10",
-            "g++-11"
+            "g++-10"
         ]
     },
     "gfortran": {
         "versions": [
             "gfortran-9",
-            "gfortran-10",
-            "gfortran-11"
+            "gfortran-10"
         ]
     },
     "php": {
--- a/images/linux/toolsets/toolset-2004.json
+++ b/images/linux/toolsets/toolset-2004.json
@@ -251,15 +251,13 @@
     "gcc": {
         "versions": [
             "g++-9",
-            "g++-10",
-            "g++-11"
+            "g++-10"
         ]
     },
     "gfortran": {
         "versions": [
             "gfortran-9",
-            "gfortran-10",
-            "gfortran-11"
+            "gfortran-10"
         ]
     },
     "php": {


Milerius:
> This reverts commit [3ebf601](https://github.com/actions/virtual-environments/commit/3ebf601284cae6a4406b7ea1ad08cd593a4d619b).
> 
> # Description
> Bug fix
> 
> It is a not good idea to add experimental toolchain to the image, as the addition/modification of standard C/C++ headers and libraries can have wide-ranging (and often undesired) consequences across the whole system.
> 
> #### Related issue:
> #3376

Thanks for the pr, this also fix #3432

MehdiChinoune:
If we fellow your logic, then gcc-9 and gcc-10 should be also removed from Ubuntu 18.04
Both of gcc-9 and gcc-11 are installed from the same ppa.

MehdiChinoune:
If we fellow your logic, then gcc-9 and gcc-10 should be also removed from Ubuntu 18.04
Both of gcc-9 and gcc-11 are installed from the same ppa.

jesec:
> If we fellow your logic, then gcc-9 and gcc-10 should be also removed from Ubuntu 18.04
> Both of gcc-9 and gcc-11 are installed from the same ppa.

So be it. In my opinion, things as important as the toolchain should never be pulled from an unsupported and experimental PPA. Users who wish to use them at their own risk should pull them themselves. Strictly speaking, only the ones in the Ubuntu archive should be deemed ready for production. 

Put that aside, I understand that many people would like to have newer toolchain available. In this case, gcc-9 and gcc-10 are "ready for use", but gcc-11 is not. 

See my previous comment: https://github.com/actions/virtual-environments/discussions/3459#discussioncomment-787734. Likely the Ubuntu maintainers are still working on this new version. Meanwhile, it is inappropriate to sabotage the entire environment with this experimental (and "useless") toolchain.

AlenaSviridenko (approved):

MaksimZhukov (approved):

miketimofeev (approved):

maxim-lobanov (approved):

PR #3451, opened 2021-05-26
Wrap gcc++ with the script to setup libstdc++.so link

dsame:
# Description
Improvement?  

Installing gcc++ brings libstdc++ of 29 version that makes the binaries built with the actions non-executable on fresh
ubuntu images.

I suggest to backup  the Ubutu default libstdc++ and use them as default till g++-9, g++-10, g++-11 runs for very first time.

g++-9, g++-10, g++-11 are wrapped with a script to setup the link target to 29th version 

## Check list

### DIFF
--- a/images/linux/scripts/installers/gcc.sh
+++ b/images/linux/scripts/installers/gcc.sh
@@ -13,7 +13,35 @@ function InstallGcc {
 
     echo "Installing $version..."
     apt-get install $version -y
+
+   # replace the original binaries with the script
+   # to make link libstdc++.so to target libstdc++.so.6.0.29
+   # not sure it is required for gcc++-9 and gcc++-10
+    binary=`which $version`
+    mv "$binary" "$binary.orig"
+    cat > "$binary" <<EOT
+#!/bin/sh
+
+$target=$(basename $(readlink -f /usr/lib/x86_64-linux-gnu/libstdc++.so.6))
+if [ $target != libstdc++.so.6.0.29 ];then
+ echo "WARNING: your are using $version that requires libstdc++.so.6.0.29"
+ echo "         the binaries compiled against it may crash on fresh Ubuntu images"
+ ln -s /usr/lib/x86_64-linux-gnu/libstdc++.so.6 /usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.29
+ echo "         libstdc++.so is a link to libstdc++.so.6.0.29 now"
+fi
+$binary.orig $@
+EOT
+    chmod +x "$binary"
 }
+
+# apt install of gcc-11 removes the existing
+# libstdc++ and replaces it with libstdc++.so.6.0.29
+# lets backup the default ubuntu distribution libstdc++
+#
+# make a copy of libstdc++ on ubuntu 20.04
+test -f /usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.28 && cp /usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.28 /tmp
+# make a copy of libstdc++ on ubuntu 18.04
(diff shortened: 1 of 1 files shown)


Milerius (approved):
I think this is an excellent and less-intrusive approach, does this mean that for example if i compile on a fresh machine with clang-12 with libstdc++ i will end up with a binary that link the default libstdc++ version and not the 6.0.29 ? If it's the case i'm in ! 

Furthermore it's always good to have the gcc version stick to his libstdc++ such as presented in the ABI page from gcc.

Thanks for the pr, i appreciate it a lot.

densamoilov:
@dsame, will it solve this issue #3454?

jesec:
This is very hacky. And, unfortunately, it will not work at all. This change has several severe issues:

- It is g++ not gcc++...... 
- It is generally a horrible idea to "workaround" one of cornerstone of the OS.
- It is generally a horrible idea to replace the compiler with a bash script. 
- You may not touch anything in `/usr` without sudo. However, obviously normal people don't run compiler with sudo, and build systems may not allow you to do that. 
- Some build systems (i.e. Bazel) use sandboxing techniques (to ensure reproducibility, etc.). The input elements would be mounted as RO, and `/usr` can not be modified in any way, even with sudo.
- It is not just libs. There are headers as well. It would be a disaster to mix incorrect versions of libs and headers. In this case, arguably the most important dependency of any C++ program: C++ standard library. 

This approach simply cannot work. In fact, there is no point to work on any kind of "harmony". In the end, it would always be: choose one, a proper OS or a functional GCC 11 (or if you "solve" the problem in the wrong way, none).

Also, I would say that the current version of GCC 11 is utterly useless if it can't provide binary compatibility with the base operating system. Even if you somehow decides to use this GCC 11, do you really want to ask your customers: OK we use GCC 11 to compile binaries, so to run our binaries you have to install the experimental version of standard C++ library from an unsupported PPA, which may f your entire environment up in the process? The toolchain is not production ready, and we should not allow this experimental piece of software to render the entire environment unusable. 

Simply wait for the Ubuntu maintainers to deliver a proper production-ready toolchain that is intended to work with Ubuntu 20.04.

Milerius:
> This is very hacky. And, unfortunately, it will not work at all. This change has several severe issues:
> 
> * It is g++ not gcc++......
> * It is generally a horrible idea to "workaround" one of cornerstone of the OS.
> * It is generally a horrible idea to replace the compiler with a bash script.
> * You may not touch anything in `/usr` without sudo. However, obviously normal people don't run compiler with sudo, and build systems may not allow you to do that.
> * Some build systems (i.e. Bazel) use sandboxing techniques (to ensure reproducibility, etc.). The input elements would be mounted as RO, and `/usr` can not be modified in any way, even with sudo.
> * It is not just libs. There are headers as well. It would be a disaster to mix incorrect versions of libs and headers. In this case, arguably the most important dependency of any C++ program: C++ standard library.
> 
> This approach simply cannot work. In fact, there is no point to work on any kind of "harmony". In the end, it would always be: choose one, a proper OS or a functional GCC 11 (or if you "solve" the problem in the wrong way, none).
> 
> Also, I would say that the current version of GCC 11 is utterly useless if it can't provide binary compatibility with the base operating system. Even if you somehow decides to use this GCC 11, do you really want to ask your customers: OK we use GCC 11 to compile binaries, so to run our binaries you have to install the experimental version of standard C++ library from an unsupported PPA, which may f your entire environment up in the process? The toolchain is not production ready, and we should not allow this experimental piece of software to render the entire environment unusable.
> 
> Simply wait for the Ubuntu maintainers to deliver a proper production-ready toolchain that is intended to work with Ubuntu 20.04.

I absolutely agree on that and more specifically, simple don't install any additional compiler that are not in the basic distro, especially for c++, let the user install what they want - philosophy of Linux in general.

PR #3007, opened 2021-03-22
Remove Clang 11 dependencies as they conflict with Clang 10 dependencies

catthehacker:
Reverts actions/virtual-environments#2851 since it breaks already installed clang-10
Fixes #2995
Unblocks #2939

PR #3479, opened 2021-05-29
Add libc++ and libc++abi when installing Clang and LLVM.

chandlerc:
# Description
These are relatively small, and important parts of the Clang/LLVM C++
toolchain, but are not installed by the script.

## Check list

Not sure which of these I need to do -- just trying to help out by doing this here rather than requiring folks to do this in each individual action configuration.


catthehacker:
`libc++` and `libc++abi` are installed via toolkit JSON file.
Won't this PR break Clang like last time? ref: https://github.com/actions/virtual-environments/pull/3007

chandlerc:
Abandoning due to the issues mentioned here: https://github.com/actions/virtual-environments/issues/3494#issuecomment-853584513

Hopefully at least something like https://github.com/actions/virtual-environments/pull/3523 can land.

PR #3549, opened 2021-06-08
fix wrong yaml label

aryairani:
This bug keeps snagging me.


al-cheb:
Hi, @aryairani 
Please check https://github.com/actions/virtual-environments/pull/3427.

aryairani:
@al-cheb Sorry about that; but neither one of them works.  Is it case sensitive and the wrong case? `macOS-11.0` works for me in one of our builds, but `macos-11` and `macos-11.0` don't seem to.

al-cheb:
@aryairani, Please sign a form -  https://github.com/actions/virtual-environments/blob/main/docs/macos-11-onboarding.md

PR #3560, opened 2021-06-10
Remove config

Toby-Valtech:
# Description
This is an Improvement, where one Piece of Information/Configuration on what Software to Install was previously hidden in a Script. This was problematic because a Script is a bad place for configuration we want to change via Script, as breaking changes could easily be made. This has been fixed and the Configuration now resides in the Toolset.json for all Ubuntu Versions as discussed in [this](https://github.com/actions/virtual-environments/discussions/3535) discussion.
The new Version has been tested for Ubuntu-2004.

## Check list

### DIFF
--- a/images/linux/scripts/installers/Install-Toolset.ps1
+++ b/images/linux/scripts/installers/Install-Toolset.ps1
@@ -43,13 +43,13 @@ foreach ($tool in $tools) {
 
         Write-Host "Installing $($tool.name) $toolVersion $($tool.arch)..."
         if ($null -ne $asset) {
-            #Install-Asset -ReleaseAsset $asset
+            Install-Asset -ReleaseAsset $asset
         } else {
             Write-Host "Asset was not found in versions manifest"
             exit 1
         }
     }
-} 
+}
 if ($toolsToInstall -contains "Python"){chown -R "$($env:SUDO_USER):$($env:SUDO_USER)" /opt/hostedtoolcache/Python}
 if ($toolsToInstall -contains "node"){chown -R "$($env:SUDO_USER):$($env:SUDO_USER)" /opt/hostedtoolcache/node}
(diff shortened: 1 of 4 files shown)


images/linux/toolsets/toolset-1804.json
```diff
@@ -68,6 +68,11 @@
             ]
         }
     ],
+    "tools":[
```
miketimofeev, line 71: I don't think we need this section. We can take the tools from the `toolcache` block that have `"url"` property

images/linux/scripts/installers/Install-Toolset.ps1
```diff
@@ -52,7 +52,6 @@ foreach ($tool in $tools) {
         }
     }
 }
-
-chown -R "$($env:SUDO_USER):$($env:SUDO_USER)" /opt/hostedtoolcache/Python
-chown -R "$($env:SUDO_USER):$($env:SUDO_USER)" /opt/hostedtoolcache/node
-chown -R "$($env:SUDO_USER):$($env:SUDO_USER)" /opt/hostedtoolcache/go
\ No newline at end of file
+if ($toolsToInstall -contains "Python"){chown -R "$($env:SUDO_USER):$($env:SUDO_USER)" /opt/hostedtoolcache/Python}
```
miketimofeev, line 55: I think we can do foreach here:
```
foreach ($toolToInstall in  $toolsToInstall) {
    chown -R "$($env:SUDO_USER):$($env:SUDO_USER)" /opt/hostedtoolcache/$toolToInstall
}
```

images/linux/toolsets/toolset-1804.json
maxim-lobanov: Agree with Mikhail. We don't need the separate section.
We can just take the tools with `url` property. Or add separate flag property as an alternative.

Toby-Valtech:
For legal reasons I will close this PR, and reopen it from my private account, as well as implementing the suggested changes.

PR #3590, opened 2021-06-15
update minGW version

rath3t:
# Description
Helly everybody,
i want to suggest an update of the quite old minGW version in the windows VM.
The version is currently 8.1.0 which means a gcc version of 8.1.0. This version does not cotain a windows implementation of `std::filesystem`. Even GCC 9 does still has only >Incomplete support for the C++17 Filesystem library and the Filesystem TS on Windows. which is a quote from https://gcc.gnu.org/gcc-9/changes.html.

Therefore,  building code in the windows VM which uses `std::filesystem` does not compile.

I tried it this at https://github.com/alandefreitas/matplotplusplus/pull/161 where i witnesses this drawback.

### DIFF
--- a/images/win/scripts/Installers/Install-Mingw64.ps1
+++ b/images/win/scripts/Installers/Install-Mingw64.ps1
@@ -3,7 +3,7 @@
 ##  Desc:  Install GNU tools for Windows
 ################################################################################
 
-Choco-Install -PackageName mingw -ArgumentList "--version=8.1.0"
+Choco-Install -PackageName mingw -ArgumentList "--version=10.2.0"
 
 # Make a copy of mingw32-make.exe to make.exe, which is a more discoverable name
 # and so the same command line can be used on Windows as on macOS and Linux


miketimofeev:
@rath3t it seems that the new version brings lots of breaking changes such as lack of mingw32-make binary, which cause the build to fail.
https://community.chocolatey.org/packages/mingw#discussion
image

rath3t:
Ok thank you for starting the workflow. On my pc it always worked since i just downloaded minGW from https://winlibs.com/.
Unfortunately, i didn't realize that this fundamental thing was missing in the "pure" minGW  package.

In my opinion the options are:
-  Trying the solution with the chocolatey package "make" as suggested in the comment. But maybe this is already installed and it is still not working?
-  Figure out what other packages need to be installed "by hand" to get in working. 
-  Install instead of pure mingw, also the package [Winlibs](https://community.chocolatey.org/packages/winlibs#install) from chocolatey and use this instead
-  Staying at 8.1 and migrate to something else 

What way is allowed here and what should i try?

Thanks and sorry for the inconvenience.

miketimofeev:
@rath3t Unfortunately, it's hard to predict the possible impact of any of these options. Have you considered mingw packages from MSYS2, which are preinstalled on the image, but not added to the path?

rath3t:
ok i will try it with MSYS and will inform you later thanks!

rath3t:
I tried several stuff but i didn't get it working, see the discussion after https://github.com/alandefreitas/matplotplusplus/pull/161#issuecomment-862991417

Especially, if i mimic the VM on my system it works with MSYS. This i mentioned at

https://github.com/alandefreitas/matplotplusplus/pull/161#issuecomment-863077559

std::filesystem is found but somehow in the VM the testing of FindFileSystem.cmake all fail.

Do you have any tips? Thanks!

miketimofeev:
@rath3t unfortunately, we don't have much expertise in either MSYS or MINGW. Maybe @MSP-Greg has any thoughts?

MSP-Greg:
@rath3t 

I've never looked to see where cmake exists in the Windows builds, or where it is in `Path`.  There are a lot of builds tools installed, one needs to be careful about which set is used.  The code I work with adds the below to `Path` before starting the build:
```
C:\msys64\mingw64\bin;C:\msys64\usr\bin
```

It might be better if you can run `C:\msys64\mingw64\bin\cmake`?  The gcc version is currently 10.3.0.  Note that sometime soon the MSYS2 ucrt packages/build tools may be available...

rath3t:
> It might be better if you can run `C:\msys64\mingw64\bin\cmake`? The gcc version is currently 10.3.0. Note that sometime soon the MSYS2 ucrt packages/build tools may be available...
> 

I tried to explicitly use `C:\msys64\mingw64\bin\cmake` but this also didn't help. Fortunately, i came up with another solution. I feel a bit stupid now, but nevertheless it works in the most obvious way.


So to compile using MinGW and using std::filesystem there is a `FindFileSystem.cmake` module from https://github.com/vector-of-bool/CMakeCM.

This module searches for experimental or other implementation of the std::filesystem library. It export a cmake target `std::filesystem` which can be used to link it to the own code.

Using MSYS with mingw 10.3 with gcc 10.3 this script returns that if found a filesystem library but it still returns 

```bash
-- Looking for C++ include filesystem
-- Looking for C++ include filesystem - found
-- Performing Test CXX_FILESYSTEM_NO_LINK_NEEDED
-- Performing Test CXX_FILESYSTEM_NO_LINK_NEEDED - Failed
-- Performing Test CXX_FILESYSTEM_STDCPPFS_NEEDED
-- Performing Test CXX_FILESYSTEM_STDCPPFS_NEEDED - Failed
-- Performing Test CXX_FILESYSTEM_CPPFS_NEEDED
-- Performing Test CXX_FILESYSTEM_CPPFS_NEEDED - Failed
CMake Error at FindFilesystem.cmake:250 (message):
-- Configuring incomplete, errors occurred!
See also "D:/a/testMingWWorkflow/testMingWWorkflow/build/CMakeFiles/CMakeOutput.log".
  Cannot run a simple program using std::filesystem
See also "D:/a/testMingWWorkflow/testMingWWorkflow/build/CMakeFiles/CMakeError.log".

  You need to update your C++ compiler to use this library
```
Therefore, the compilation is not possible. I tried several compiler flags etc. but the outcome was never correct.

I tried hard that `FindFileSystem.cmake` exports the correct target. Instead we can just neglect the search for a compile target and remove all linking since gcc 10.3 should care about that itself . Apparently that just works. 

So to summarize: 
Since ``FindFileSystem.cmake`  does not support this case with MinGW we can just don't search and don't export that target.
And also don't link to that target.

rath3t:
Thanks for your help! This can be closed or shouldn't it?

miketimofeev:
@rath3t let's close it! We can revisit this later if we receive more requests for updated mingw.

Issue #3645, opened 2021-06-24
Add `yq`


Yann-J:
**Tool information**
- Tool name: `yq`
- Tool license: MIT
- Add or update? Add
- Desired version: 4.x
- Approximate size: ~6Mb
- If this is an add request:
  - Brief description of tool: `yq` is the YAML equivalent to `jq` and provides convenient YAML query and edits by command line
  - URL for tool's homepage: https://github.com/mikefarah/yq
  - Provide a basic test case to validate the tool's functionality: 

```
[ "$(yq -n eval '.somekey = "somevalue"')" = "somekey: somevalue" ];
```

**Area for Triage**:

**Question, Bug, or Feature?**:

**Virtual environments affected**
- [x] Ubuntu 16.04
- [x] Ubuntu 18.04
- [x] Ubuntu 20.04
- [ ] macOS 10.15
- [ ] macOS 11
- [ ] Windows Server 2016 R2
- [ ] Windows Server 2019

**Can this tool be installed during the build?**

```
brew install yq
```

**Tool installation time in runtime**

~10s?

**Are you willing to submit a PR?**

Yes

AlenaSviridenko:
Hi @Yann-J,
there is a popular action for setting up `yq` https://github.com/marketplace/actions/setup-yq, have you tried this? Using actions is the preferable way to interact with images https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#using-preinstalled-software.

Yann-J:
Yes, it's true, but actions are not available in Azure Devops (I think), which is what we use... Right now we install it at runtime, i.e. for all of our builds, which works OK but since this is a popular and useful tool I think it makes sense to have it as part of the base image...

Yann-J:
I gave a shot at a PR for it...

catthehacker:
I would recommend installing it via Homebrew (`brew install yq`) since it's already available on ADO and GHA (if you install it at runtime) or install it via toolset
https://github.com/actions/virtual-environments/blob/405a368b22cbac48f31ce7b35f415369d31a27b1/images/linux/toolsets/toolset-2004.json#L199-L201
https://github.com/actions/virtual-environments/blob/405a368b22cbac48f31ce7b35f415369d31a27b1/images/linux/toolsets/toolset-1804.json#L199-L201
[...]

Issue #3287, opened 2021-04-29
Ubuntu 16.04 environment will be removed on September 20, 2021


AlenaSviridenko:
Traditional 5-years support of Ubuntu 16.04 by Canonical [ends in April, 2021](https://ubuntu.com/16-04). To keep our environment updated and secured, we will remove Ubuntu 16.04 on September 20, 2021.

You will need to migrate `ubuntu-16.04` workflows to `ubuntu-18.04` or `ubuntu-latest` which will run on Ubuntu 20.04 LTS.

To make sure everyone is aware of this change, we scheduled 2 short "brownouts". Ubuntu 16.04 builds will fail during the brownout period. Therefore, it is recommended to migrate your workflows prior to the September 6, 2021.

The brownouts are scheduled for the following dates and times:

- September 6, 2021 5:00pm UTC – 10:00pm UTC
- September 14, 2021 5:00pm UTC – 10:00pm UTC

Please, use this announcement issue to provide any feedback or ask questions regarding the upcoming change.
Thank you.

Rada8732:
Approve all

tomato42:
> The brownouts are scheduled for the following dates and times:
> 
>     * September 6, 2021 5:00pm UTC – 10:00pm UTC
> 
>     * September 14, 2021 5:00pm UTC – 10:00pm UTC

So anybody that works a 9 to 5 job in EMEA or APAC has no chance of hitting them?

hugovk:
![image](https://user-images.githubusercontent.com/1324225/119547257-914a1a80-bd9d-11eb-9bb9-9474feac34d1.png)

https://www.timeanddate.com/worldclock/meetingtime.html?iso=20210906&p1=136&p2=195&p3=101&p4=176&p5=33&p6=248&p7=152

joe192:
@joe192

joe192:
Ok

sergeyklay:
>  To keep our environment updated and secured, we will remove Ubuntu 16.04 on September 20, 2021.

Removing the support of distro version does not make the environment updated, it makes the environment less rich in features. There are many cases where you need to make sure that your project sill works on a previous version of the OS/distro. For example, your project depends on libraries that can only be installed on the Ubuntu 16.04 or even 14.04. The less choice you give, the less wide the range of systems where a particular project has been tested will be. Ultimately, this is an affect on the end user.

Wohlstand:
[...]

PR #3646, opened 2021-06-24
Install yq

Yann-J:
# Description

This is a new tool, `yq`. It is a small, popular tool to query and manipulate YAML files - equivalent to `jq` on JSON.

#### Related issue:

#3645

## Check list

### DIFF
--- a/images/linux/scripts/installers/yq.sh
+++ b/images/linux/scripts/installers/yq.sh
@@ -3,27 +3,15 @@
 # Source the helpers for use with the script
 source $HELPER_SCRIPTS/install.sh
 
-# Test to see if the software in question is already installed, if not install it
-echo "Checking to see if the installer script has already been run"
-if [ -z $YQ_VERSION ]; then
-    YQ_VERSION=v4.9.6
-    YQ_BINARY=yq_linux_amd64
+YQ_VERSION=$(curl -s https://api.github.com/repos/mikefarah/yq/releases/latest | jq -r '.tag_name')
+YQ_BINARY=yq_linux_amd64
 
-    # As per https://github.com/mikefarah/yq#wget
-    YQ_URL="https://github.com/mikefarah/yq/releases/download/${YQ_VERSION}/${YQ_BINARY}.tar.gz"
-    
-    download_with_retries "$YQ_URL" "/tmp" "${YQ_BINARY}.tar.gz"
-    tar xzf "/tmp/${YQ_BINARY}.tar.gz"
-    mv ${YQ_BINARY} /usr/bin/yq
-else
-    echo "yq already installed with version $YQ_VERSION"
-fi
+# As per https://github.com/mikefarah/yq#wget
+echo "Downloading yq version ${YQ_VERSION}"
+YQ_URL="https://github.com/mikefarah/yq/releases/download/${YQ_VERSION}/${YQ_BINARY}.tar.gz"
 
-# Run tests to determine that the software installed as expected
-echo "Testing to make sure that script performed as expected, and basic scenarios work"
-if [ ! "$(yq -n eval '.somekey = "somevalue"')" = "somekey: somevalue" ]; then
-    echo "yq installation failed"
-    exit 1
-else
-    echo "yq successfully installed"
-fi
+download_with_retries "$YQ_URL" "/tmp" "${YQ_BINARY}.tar.gz"
+tar xzf "/tmp/${YQ_BINARY}.tar.gz" -C "/tmp"
+mv /tmp/${YQ_BINARY} /usr/local/bin/yq
+
+invoke_tests "Tools" "yq"
--- a/images/linux/scripts/installers/yq.sh
+++ b/images/linux/scripts/installers/yq.sh
@@ -3,12 +3,11 @@
 # Source the helpers for use with the script
 source $HELPER_SCRIPTS/install.sh
 
-YQ_VERSION=$(curl -s https://api.github.com/repos/mikefarah/yq/releases/latest | jq -r '.tag_name')
 YQ_BINARY=yq_linux_amd64
 
 # As per https://github.com/mikefarah/yq#wget
-echo "Downloading yq version ${YQ_VERSION}"
-YQ_URL="https://github.com/mikefarah/yq/releases/download/${YQ_VERSION}/${YQ_BINARY}.tar.gz"
+YQ_URL=$(curl -s https://api.github.com/repos/mikefarah/yq/releases/latest | jq -r ".assets[].browser_download_url | select(endswith(\"$YQ_BINARY.tar.gz\"))")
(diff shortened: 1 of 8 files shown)


AlenaSviridenko:
Hi @Yann-J,
thank you for contribution!
could you please, re-work installation script so it will require less manual maintenance and the latest version will always be installed automatically.
Also, some common notes:
1. Here is how we invoke tests for the tools https://github.com/actions/virtual-environments/blob/851f0dfc316c21f935e7c161c6663fd2a6bf21df/images/linux/scripts/installers/packer.sh#L13
Test are written with Pester testing framework on Powershell.
2. Latest release can be retrieved via `https://github.com/mikefarah/yq/releases/latest` so there is no need to hard-code the exact version.
3. Ubuntu 16.04 is currently under deprecation https://github.com/actions/virtual-environments/issues/3287, we don't update it, so there is no need to include installer script to its template

images/linux/scripts/installers/yq.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash -e
+
+# Source the helpers for use with the script
+source $HELPER_SCRIPTS/install.sh
+
+# Test to see if the software in question is already installed, if not install it
+echo "Checking to see if the installer script has already been run"
+if [ -z $YQ_VERSION ]; then
+    YQ_VERSION=v4.9.6
```
catthehacker, line 9:
```suggestion
    YQ_VERSION=$(curl -s https://api.github.com/repos/mikefarah/yq/releases/latest | jq -r '.tag_name')
```

from https://github.com/Yann-J/virtual-environments/blob/203224ed3d218f7e0b14af71cdd0f772308412d4/images/linux/scripts/installers/nvm.sh#L9

Yann-J:
Thanks for the comments! Working on it now!

Yann-J:
Let me know what you guys think @catthehacker @AlenaSviridenko

images/linux/scripts/installers/yq.sh
```diff
@@ -0,0 +1,17 @@
...
+
+# Source the helpers for use with the script
+source $HELPER_SCRIPTS/install.sh
+
+YQ_VERSION=$(curl -s https://api.github.com/repos/mikefarah/yq/releases/latest | jq -r '.tag_name')
+YQ_BINARY=yq_linux_amd64
+
+# As per https://github.com/mikefarah/yq#wget
+echo "Downloading yq version ${YQ_VERSION}"
+YQ_URL="https://github.com/mikefarah/yq/releases/download/${YQ_VERSION}/${YQ_BINARY}.tar.gz"
```
al-cheb, line 11: We could parse the direct link from: 
`curl -s https://api.github.com/repos/mikefarah/yq/releases/latest | jq -r '.assets[].browser_download_url | select(endswith("yq_linux_amd64.tar.gz"))'`

al-cheb:
You should update scripts to provide yq version in the output:
https://github.com/MarcusFelling/virtual-environments/blob/main/images/linux/scripts/SoftwareReport/SoftwareReport.Tools.psm1
https://github.com/MarcusFelling/virtual-environments/blob/main/images/linux/scripts/SoftwareReport/SoftwareReport.Generator.ps1

Yann-J:
Ah yes, done, thanks!

images/linux/scripts/installers/yq.sh
miketimofeev: @Yann-J could you please apply the suggested changes?

images/win/scripts/SoftwareReport/SoftwareReport.Tools.psm1
```diff
@@ -270,3 +270,8 @@ function Get-SwigVersion {
     $swigVersion = $Matches.Version
     return "Swig $swigVersion"
 }
+
+function Get-YqVersion {
```
AlenaSviridenko, line 274: that should be in the linux folder, not windows :)

Yann-J: Oh dammit, you're right :)
Fixed now - hopefully!

images/linux/scripts/installers/yq.sh
Yann-J: OK I'm not convinced that this is much more readable, but I guess that's fine too

maxim-lobanov (approved):

AlenaSviridenko (approved):

images/linux/scripts/installers/yq.sh
```diff
@@ -0,0 +1,16 @@
...
+
+YQ_BINARY=yq_linux_amd64
+
+# As per https://github.com/mikefarah/yq#wget
+YQ_URL=$(curl -s https://api.github.com/repos/mikefarah/yq/releases/latest | jq -r ".assets[].browser_download_url | select(endswith(\"$YQ_BINARY.tar.gz\"))")
+echo "Downloading latest yq from $YQ_URL"
+
+download_with_retries "$YQ_URL" "/tmp" "${YQ_BINARY}.tar.gz"
+tar xzf "/tmp/${YQ_BINARY}.tar.gz" -C "/tmp"
+mv /tmp/${YQ_BINARY} /usr/local/bin/yq
```
skonto, line 14: Assuming this is the ubuntu:latest? If someone installs yq previously under /usr/bin using some previous version then this would break things.

miketimofeev: Hi @skonto! 
No, it wasn't made intentionally, we were not aware of typical yq usage. Is it usually installed in /usr/bin rather than /usr/local/bin?

skonto: Hi thanks for the quick response! It depends on how users setup things. Afaik `/usr/bin` is for OS level installations, `/usr/local/bin` for user supplied binaries (allows overriding system binaries). I think this change (PR) was intended to be at the VM/OS level. Btw we have such [a setup](https://github.com/openshift-knative/serverless-operator/blob/034e1eb83e0170677f32d794fc95cf311ac007f3/.github/workflows/validate.yaml#L31). WDYTH?

miketimofeev: Well, this setup won't override yq installed in `/usr/local/bin` 😩
I'll prepare a PR with moving yq to /usr/bin, but the new image with the fix will be deployed next week. Could you please add a workaround until that? Something like ` if [[ $(which yq) =~ "/usr/local/bin" ]]; then  sudo rm $(which yq); fi`
Sorry for the inconvenience!

Yann-J: Hmm apologies for this, looking at some other tool installation scripts, `/usr/local/bin` seemed to be the standard practice, but I agree `/usr/bin` probably makes more sense here...

skonto: No problem, yeah no strong opinions thanks for the response @Yann-J :)

PR #3651, opened 2021-06-24
updates java version to 16

chrigiansen:
# Description
Updates the Java version to 16.0.1+9 for all linux and windows images. No changes were made to macos images due to the comment below. All defaults remain set to an LTS version.

The reason for this update, is to close a multi year gap between the available and the newest version of Java.

## Check list


maxim-lobanov:
Hello @chrigiansen , according to our guidelines, we install only LTS Java versions on Hosted images. Java 16 is not LTS. https://github.com/actions/virtual-environments/blob/main/docs/software-and-images-guidelines.md#software-and-images-support-policy

Java 13 was installed before accepting this policy so we can't remove it without impact.
Please use official action to setup any Java in runtime (it is pretty quick, actually): https://github.com/actions/setup-java#basic

chrigiansen:
Sorry to bother you, but there are three points that I deem user unfriendly or are unclear to me:

1. In the first link you provided, it is specified that you install all LTS versions for Java, there is no mention of of exclusivity.
2. The other software packages lack, or have a much more lenient restriction, either because some latest version is included or their update policy has a narrower timeframe. Keep in mind, that the Java LTS cycle is 3 years.
3. I am building my application on Azure. The only way on azure to install a custom Java version is to host it by oneself, making it bothersome for all people wanting to use a newer non-LTS version of java.

Right now it seems like Java is being penalised, in my opinion a restriction similar to .NET Core's would be more useful. 

Anyway, I wish you a nice day, and i hope you take my words into consideration.

maxim-lobanov:
@chrigiansen , thank you for sharing these points.

There is no exclusivity. There are only two versions of Java that are LTS: 8 and 11 (Java 17 will be LTS and will be released on September 2021) and both are pre-installed. Any other Java versions are not LTS according to https://www.oracle.com/java/technologies/java-se-support-roadmap.html
Currently, Hosted images contain some non LTS Java versions like Java 13. They were installed before adding clear policy about it so we can't remove it without impact. For any further installations we follow policy.

Yes, policy is different for other tools because there are a bunch of factors that we need to keep on mind (like tool size, installation time, usage). We are trying to minimize the software list on Hosted images to be focused on stability of current software. It is why we need to follow this policy.

I understand your concern about [JavaToolInstaller](https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/tool/java-tool-installer?view=azure-devops) (Azure DevOps task) and it is less powerful than [actions/setup-java](https://github.com/actions/setup-java) (GitHub Actions task) that can install versions on-flight.
As an option, Java 16 can be installed on Azure DevOps on-flight without necessary to host it somewhere via the following steps (actually, it is what actions/setup-java does under hood) :
```yml
steps:
  - bash: |
      url="https://github.com/AdoptOpenJDK/openjdk16-binaries/releases/download/jdk-16.0.1%2B9/OpenJDK16U-jdk_x64_linux_hotspot_16.0.1_9.tar.gz"
      wget -O "$(Agent.TempDirectory)/java16.tar.gz" $url
  - task: JavaToolInstaller@0
    inputs:
      versionSpec: '16'
      jdkArchitectureOption: 'x64'
      jdkSourceOption: 'LocalDirectory'
      jdkFile: '$(Agent.TempDirectory)/java16.tar.gz'
      cleanDestinationDirectory: true
```

sfragata:
I thought the toolset file define the toolset installed, if it is true in this file
https://github.com/actions/virtual-environments/blob/main/images/macos/toolsets/toolset-10.15.json#L153 we can see java 12 and 14 
```
"8", "11", "12", "13", "14"
```

Issue #3638, opened 2021-06-22
[all OS versions] Android SDK tools will be changed to Command line tools on July, 12


miketimofeev:
**Breaking changes**
All the Android tools will be installed using `sdkmanager` from the Command line tools package rather than `sdkmanager` from SDK tools.

**Target date**
Image deployment will start July, 12 and will take 3-4 days.

**The motivation for the changes**
SDK Tools package is deprecated and no longer receiving updates, google recommends using command-line tools instead.
https://developer.android.com/studio/releases/sdk-tools
![image](https://user-images.githubusercontent.com/48208649/122936829-23801700-d37a-11eb-84bf-c66074c9d918.png)

**Possible impact**
There should be no visible impact, but it is better to switch your pipeline to use command line tools' sdkmanager as well.

**Virtual environments affected**
- [ ] Ubuntu 16.04
- [x] Ubuntu 18.04
- [x] Ubuntu 20.04
- [ ] macOS 10.13
- [x] macOS 10.14
- [x] macOS 10.15
- [x] macOS 11
- [x] Windows Server 2016 R2
- [x] Windows Server 2019

**Mitigation ways**
Change the path to the sdkmanager in your code:
- Windows
`$env:ANDROID_SDK_ROOT/tools/bin/sdkmanager.bat` -> `$env:ANDROID_SDK_ROOT/cmdline-tools/latest/bin/sdkmanager.bat` 
- Linux & macOS
`$ANDROID_SDK_ROOT/tools/bin/sdkmanager` -> `$ANDROID_SDK_ROOT/cmdline-tools/latest/bin/sdkmanager
`

CaoMinhThien:
Ok

miketimofeev:
All the images were deployed

PR #3682, opened 2021-07-02
[Windows] Switch android tools installation to use cmdline-tools' sdkmanager

miketimofeev:
# Description
SDK Tools package is deprecated and no longer receiving updates. We need to switch to command-line tools instead
https://developer.android.com/studio/releases/sdk-tools
image

A few notes:
- The installation script was renamed from Update to Install as for the other OSs
- The licenses are different for the cmdline-tools sdkmanager so the archive and corresponding base64 string were regenerated
- New sdkmanager doesn't overwrite the outdated `platform-tools` package from visual studio, creates the additional folder instead, and produces a warning message. We need to remove the outdated installation first to install the new version without such issues
- Old `SDK-tools` will also be available on the image as they are installed as a part of `build-tools` packages.
- A test added to make sure both "old" and "new" sdkmanagers are available

#### Related issue:
https://github.com/actions/virtual-environments/issues/3638
https://github.com/actions/virtual-environments-internal/issues/2340

## Check list


dmitry-shibanov (dismissed):

al-cheb (approved):

images/win/scripts/Installers/Install-AndroidSDK.ps1
```diff
@@ -0,0 +1,147 @@
+################################################################################
+##  File:  Install-AndroidSDK.ps1
+##  Desc:  Install and update Android SDK and tools
+################################################################################
+
+# install command-line tools
+$cmdlineToolsUrl = "https://dl.google.com/android/repository/commandlinetools-win-7302050_latest.zip"
```
AlenaSviridenko, line 7: is it ok to hardcode the version? will we have to manually update it in future?

images/win/scripts/Installers/Install-AndroidSDK.ps1
```diff
@@ -0,0 +1,147 @@
...
+New-Item -Path "C:\Android" -ItemType Directory
+New-Item -Path "$sdkRoot" -ItemType SymbolicLink -Value "$sdkInstallRoot"
+$sdkManager = "$sdkRoot\cmdline-tools\latest\bin\sdkmanager.bat"
+
+# Install the standard Android SDK licenses. Currently, there isn't a better way to do this,
+# so we are base64-encoded a zip of the licenses directory from another installation.
+# To create this base64 string, create a zip file that contains nothing but a 'licenses' folder,
+# which folder contains the accepted license files found in 'C:\Program Files (x86)\Android\android-sdk\licenses'.
+# Then, run this in PowerShell:
+#     $LicensesZipFileName = 'C:\Program Files (x86)\Android\android-sdk\Licenses.zip'
```
AlenaSviridenko, line 26: Can it be done in this script instead of having that base64 string? it looks scary

images/win/scripts/Installers/Install-AndroidSDK.ps1
```diff
@@ -0,0 +1,147 @@
...
+# install command-line tools
+$cmdlineToolsUrl = "https://dl.google.com/android/repository/commandlinetools-win-7302050_latest.zip"
+$cmdlineToolsArchPath = Start-DownloadWithRetry -Url $cmdlineToolsUrl -Name "cmdline-tools.zip"
+$sdkInstallRoot = "C:\Program Files (x86)\Android\android-sdk"
+$sdkRoot = "C:\Android\android-sdk"
+Expand-Archive -Path $cmdlineToolsArchPath -DestinationPath "${sdkInstallRoot}\cmdline-tools" -Force
+# cmdline tools should be installed in ${sdkInstallRoot}\cmdline-tools\latest\bin, but archive contains ${sdkInstallRoot}\cmdline-tools\bin 
+# we need to create the proper folder structure
+Rename-Item "${sdkInstallRoot}\cmdline-tools\cmdline-tools" "latest"
+# ANDROID_NDK_PATH/HOME should not contain spaces. Otherwise, the script ndk-build.cmd gives an error https://github.com/actions/virtual-environments/issues/1122
```
AlenaSviridenko, line 15: could you please add a new line between comment and a previous command? the whole text looks more unreadable

images/win/scripts/Installers/Install-AndroidSDK.ps1
```diff
@@ -0,0 +1,147 @@
...
+                | Sort-Object -Unique `
+                | Select-Object -Last 1
+
+$ndkLatestPackageName = Get-AndroidPackagesByName -AndroidPackages $androidPackages `
+                -PrefixPackageName "ndk;$ndkLatestMajorVersion" `
+                | Sort-Object -Unique `
+                | Select-Object -Last 1
+
+$androidNDKs = @($ndkLTSPackageName, $ndkLatestPackageName)
+
```
AlenaSviridenko, line 111: extra line

images/win/scripts/Installers/Install-AndroidSDK.ps1
miketimofeev: We have never updated the version of SDK tools back in the days, so I'm not sure we need to change something here.

images/win/scripts/Installers/Install-AndroidSDK.ps1
miketimofeev: Nope, unfortunately this is the only workaround, which was in the previous iteration as well

AlenaSviridenko (dismissed):

AlenaSviridenko (approved):

PR #3716, opened 2021-07-12
Download with Retry: Added HTTP status code check.

Darleev:
# Description
According to the latest unstable builds, it makes sense to implement an HTTP status code check before downloading the package from the source.

#### Related issue:
https://github.com/actions/virtual-environments-internal/issues/2402

## Check list

### DIFF
--- a/images/macos/provision/utils/utils.sh
+++ b/images/macos/provision/utils/utils.sh
@@ -15,6 +15,9 @@ download_with_retries() {
         COMMAND="curl $URL -4 -sL -o '$DEST/$NAME'"
     fi
 
+    echo "Verifying $URL HTTP status code..."
+    verify_http_status_code $URL
+
     echo "Downloading $URL..."
     retries=20
     interval=30
@@ -79,6 +82,20 @@ is_Less_BigSur() {
         true
     else
         false
+    fi
+}
+
+verify_http_status_code()
+{
+    URL="$1"
+    http_code=$(curl -s -o out.html -w '%{http_code}'  http://www.google.com/linux;)
+
+    if [[ $http_code -eq 200 ]]; then
+        exit 0
+    else
+        cat out.html
+        echo "Bad HTTP status code: $http_code for the provided link: $URL."
+        exit 1
     fi
 }
 


images/macos/provision/utils/utils.sh
```diff
@@ -82,6 +84,20 @@ is_Less_BigSur() {
     fi
 }
 
+verify_http_status_code()
+{
+    URL="$1"
+    http_code=$(curl -s -o out.html -w '%{http_code}'  http://www.google.com/linux;)
```
al-cheb, line 90: Should we replace google to URL?

images/macos/provision/utils/utils.sh
```diff
@@ -82,6 +84,20 @@ is_Less_BigSur() {
...
+
+    if [[ $http_code -eq 200 ]]; then
+        exit 0
```
al-cheb, line 93: You should use return statement.

images/macos/provision/utils/utils.sh
Darleev: @al-cheb of course, it's just copied example, thank you for pointing to this.

images/macos/provision/utils/utils.sh
Darleev: thank you

images/macos/provision/utils/utils.sh
```diff
@@ -87,15 +89,11 @@ is_Less_BigSur() {
 verify_http_status_code()
 {
     URL="$1"
-    http_code=$(curl -s -o out.html -w '%{http_code}' $URL;)
+    http_code=$(curl -sL -o out.html -w '%{http_code}' $URL)
 
-    if [[ $http_code -ge 400 ]]; then
-        cat out.html
+    if [[ $http_code -eq 200 ]]; then
```
miketimofeev, line 94: it should be not equal
```suggestion
    if [[ $http_code -ne 200 ]]; then
```

miketimofeev (approved):

al-cheb (approved):

AlenaSviridenko (approved):

PR #3721, opened 2021-07-12
[Ubuntu] Download with Retry: Added HTTP status code check

Darleev:
# Description
According to the latest unstable builds, it makes sense to implement HTTP status code check along with downloading a package from a source to make sure we download the package and not the web page with some error on it.
Also, there is no point in checking the curl return code as the curl will return 0 if the entire web page with an error is downloaded.

#### Related issue:
https://github.com/actions/virtual-environments-internal/issues/2402

## Check list


images/linux/scripts/helpers/install.sh
```diff
@@ -50,6 +52,20 @@ verlte() {
     [  "$1" = "$sortedVersion" ]
 }
 
+verify_http_status_code()
+{
+    URL="$1"
+    http_code=$(curl -s -o out.html -w '%{http_code}' $URL;)
```
al-cheb, line 58: nitpick: remove ;

images/linux/scripts/helpers/install.sh
```diff
@@ -50,6 +52,20 @@ verlte() {
...
+
+    if [[ $http_code -ge 400 ]]; then
+        cat out.html
+        echo "Bad HTTP status code: $http_code for the provided link: $URL."
+        return 1
+    else
+        return 0
```
al-cheb, line 65: Do we need else statement?

images/linux/scripts/helpers/install.sh
```diff
@@ -50,6 +52,20 @@ verlte() {
...
 
+verify_http_status_code()
+{
+    URL="$1"
+    http_code=$(curl -s -o out.html -w '%{http_code}' $URL;)
+
+    if [[ $http_code -ge 400 ]]; then
+        cat out.html
+        echo "Bad HTTP status code: $http_code for the provided link: $URL."
+        return 1
```
al-cheb, line 63: If we don't check the result I think we can rid it of.

images/linux/scripts/helpers/install.sh
miketimofeev, line 58: Does it make sense to add `-L` flag and check code `200` only?
```
URL="https://github.com/vercel/now"
curl -s -o out.html -w '%{http_code}' $URL
301
curl -sL -o out.html -w '%{http_code}' $URL
200
```

images/linux/scripts/helpers/install.sh
```diff
@@ -23,6 +23,8 @@ download_with_retries() {
     i=20
     while [ $i -gt 0 ]; do
         ((i--))
+        echo "Verifying $URL HTTP status code..."
+        verify_http_status_code $URL
```
al-cheb, line 27: I think we should add an additional condition to start download only if we have a successful http_code. In other case we can download broken page like https://google.com/linux and keep going.

images/linux/scripts/helpers/install.sh
```diff
@@ -50,6 +52,20 @@ verlte() {
     [  "$1" = "$sortedVersion" ]
 }
 
+verify_http_status_code()
+{
+    URL="$1"
+    http_code=$(curl -s -o out.html -w '%{http_code}' $URL;)
+
+    if [[ $http_code -ge 400 ]]; then
+        cat out.html
```
al-cheb, line 61: I don't think it's necessary to have this output, it can be useless. E.g: https://bing.com/404

images/linux/scripts/helpers/install.sh
miketimofeev: And why do you need out.html? You can use `curl -o /dev/null`

images/linux/scripts/helpers/install.sh
```diff
@@ -50,6 +54,19 @@ verlte() {
     [  "$1" = "$sortedVersion" ]
 }
 
+verify_http_status_code()
+{
+    URL="$1"
+    http_code=$(curl -sL -o out.html -w '%{http_code}' $URL)
```
miketimofeev, line 60: why do you need out.html? This should work fine
```suggestion
    http_code=$(curl -sL -o /dev/null -s -w "%{http_code}\n" $URL)
```

images/linux/scripts/helpers/install.sh
```diff
@@ -50,6 +54,19 @@ verlte() {
...
+
+    if [[ $http_code -eq 200 ]]; then
+        return 0
+    else
+        echo "Bad HTTP status code: $http_code for the provided link: $URL."
+        return 1
+    fi
```
miketimofeev, line 67: Looks like this block can be simplified
```suggestion
    if [[ $http_code -ne 200 ]]; then
        echo "Bad HTTP status code: $http_code for the provided link: $URL."
        return 1
    fi
```

images/linux/scripts/helpers/install.sh
```diff
@@ -23,12 +23,16 @@ download_with_retries() {
     i=20
     while [ $i -gt 0 ]; do
         ((i--))
-        eval $COMMAND
+        echo "Verifying $URL HTTP status code..."
+        verify_http_status_code $URL
         if [ $? != 0 ]; then
```
miketimofeev, line 28: why the condition is for non-zero exit code? Do we want to download the file from broken URLs only? 😀

miketimofeev (approved):

dmitry-shibanov (approved):

AlenaSviridenko (approved):

images/linux/scripts/helpers/install.sh
```diff
@@ -14,22 +14,29 @@ download_with_retries() {
...
+        # Temporary disable exit on error to retry on non-zero exit code
+        set +e
+        http_code=$(eval $COMMAND)
+        exit_code=$?
+        if [ $http_code -eq 200 ] && [ $exit_code -eq 0 ]; then
             echo "Download completed"
             return 0
+        else
+            echo "Error — Either HTTP response code for '$URL' is wrong - '$http_code' or exit code is not 0 - '$exit_code'. Waiting $interval seconds before the next attempt, $retries attempts left"
+            sleep 30
```
RUBenGAMArrarodRiguEZ-ToMtOm, line 36: Here the variable that is used for the echo above is not used and the interval is unexpectedly always 30 regardless of the value of the variable.
```suggestion
            sleep "$interval"
```

RUBenGAMArrarodRiguEZ-ToMtOm: Fixed here https://github.com/actions/runner-images/pull/7207

PR #3739, opened 2021-07-15
Update to latest Go Patch releases 1.15.14 and 1.16.6

berndverst:
# Description

Go versions have been outdated and actions/setup-go is unable to pull the latest patch version if a minor release is specified. This updates cached Go versions to 1.15.14 and 1.16.6. These releases addressed a CVE.

## Check list

### DIFF
--- a/images/linux/Ubuntu1804-README.md
+++ b/images/linux/Ubuntu1804-README.md
@@ -180,7 +180,7 @@
 #### Go
 - 1.13.15
 - 1.14.15
-- 1.15.13
+- 1.15.14
 
 #### Node.js
 - 10.24.1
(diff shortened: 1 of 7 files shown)


berndverst:
@miketimofeev & @maxim-lobanov fellow Microsoft person here! Can you make sure the Go versions get updated? 5 days ago there was a new patch release addressing a CVE and unfortunately actions/setup-go still doesn't pick up on it because the old version is cached.

It doesn't look like your `github-actions` bot picked up on this Golang release either.

The curious thing is that the bot did pick up on the new versions and updated `actions/go-versions`, however the virtual environment was not update.

https://github.com/actions/go-versions/commit/562bf9fcb9c34209c8b6352825b57274b487ccc9

maxim-lobanov:
@berndverst these documentation files are auto-generated. No need to update it manually.

Usually, we generate images on Sunday, start deploy on Monday and finish by Thursday every week. These versions were released on Monday this week so obviously they were not included to generated image. We have to wait next week until these updates are delivered. Both updates will be included to the next week rollout.

If you need to have these updates asap, you can modify actions/setup-go task to specify exact version:
```yml
steps:
  - uses: actions/setup-go@v2
    with:
      go-version: '1.16.6'
  - run: go version
```
actions/setup-go can install versions directly from `go-versions` if they are not found on image.

berndverst:
@maxim-lobanov yes I'm aware of being able to pin a version in `actions/setup-go` -- but what we pin is a minor release and expect the platform to pick up patch releases in a timely manner. For the moment we have indeed pinned the patch release but this isn't ideal.

Given that patch releases do not introduce breaking changes and fix important security issues these should always be picked up automatically in a timely manner. Unfortunately the current process here means the platform may be up to 2 weeks (!) behind in automatically picking up the patch release. For minor releases this wouldn't be a big deal, but patch releases are important.

If `actions/setup-go` would allow me to skip the local cache to instead pull from `actions/go-versions` which is updated much sooner that would be fine. Unfortunately it does not.

PR #3808, opened 2021-07-28
[macOS] Fix xcode installation script if there is only one Xcode version to install

miketimofeev:
# Description
There is an exception when we try to reverse 1 element array:
`MethodException: Cannot find an overload for "Reverse" and the argument count: "1"`
This PR fixes it by reversing arrays with more than 1 element.

#### Related issue:
n\a

## Check list

### DIFF
--- a/images/macos/provision/core/xcode.ps1
+++ b/images/macos/provision/core/xcode.ps1
@@ -17,7 +17,10 @@ $env:SPACESHIP_SKIP_2FA_UPGRADE = 1
 $os = Get-OSVersion
 $xcodeVersions = Get-ToolsetValue "xcode.versions"
 $defaultXcode = Get-ToolsetValue "xcode.default"
-[Array]::Reverse($xcodeVersions)
+if ($xcodeVersions.Count -gt 1)
+{
+    [Array]::Reverse($xcodeVersions)
+}
 $threadCount = "5"
 
 Write-Host "Installing Xcode versions..."
@@ -36,8 +39,8 @@ if ($os.IsLessThanCatalina) {
     $latestXcodeVersion = $xcodeVersions | Select-Object -Last 1 -ExpandProperty link
     Install-XcodeAdditionalPackages -Version $latestXcodeVersion
 }
-$xcodeVersions | ForEach-Object { 
-    Invoke-XcodeRunFirstLaunch -Version $_.link 
+$xcodeVersions | ForEach-Object {
+    Invoke-XcodeRunFirstLaunch -Version $_.link
 }
 Invoke-XcodeRunFirstLaunch -Version $defaultXcode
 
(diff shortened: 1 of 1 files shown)


AlenaSviridenko (dismissed):

images/macos/provision/core/xcode.ps1
```diff
@@ -17,7 +17,10 @@ $env:SPACESHIP_SKIP_2FA_UPGRADE = 1
 $os = Get-OSVersion
 $xcodeVersions = Get-ToolsetValue "xcode.versions"
 $defaultXcode = Get-ToolsetValue "xcode.default"
-[Array]::Reverse($xcodeVersions)
+if ($xcodeVersions.version.Count -gt 1)
+{
+    [Array]::Reverse($xcodeVersions)
+}
```
maxim-lobanov, line 23: May be it will be easier to just change:
`$xcodeVersions = Get-ToolsetValue "xcode.versions"`
to
`[array]$xcodeVersions = Get-ToolsetValue "xcode.versions"` ?

In this case, `$xcodeVersions` will be always array and it will be easier to work with it.

images/macos/provision/core/xcode.ps1
```diff
@@ -17,7 +17,10 @@ $env:SPACESHIP_SKIP_2FA_UPGRADE = 1
 $os = Get-OSVersion
 $xcodeVersions = Get-ToolsetValue "xcode.versions"
 $defaultXcode = Get-ToolsetValue "xcode.default"
-[Array]::Reverse($xcodeVersions)
+if ($xcodeVersions.version.Count -gt 1)
```
al-cheb, line 20: nitpick: `$xcodeVersions.Count`

al-cheb (approved):

images/macos/provision/core/xcode.ps1
miketimofeev: I thought the purpose of the reversing is to install Xcode starting from the lowest version, isn't it?

al-cheb: In case than Count is 1:
```
PS > $xcodeVersions = Get-ToolsetValue "xcode.versions"
PS > $xcodeVersions.GetType()

IsPublic IsSerial Name                                     BaseType
-------- -------- ----                                     --------
True     False    PSCustomObject                           System.Object
```

maxim-lobanov: yea, Alex is right. The reason of failure is that array with 1 element === 1 element and you run `Reverse` on single element instead of array with 1 element

miketimofeev: Thank you, folks!

MaksimZhukov (approved):

maxim-lobanov (approved):

PR #3833, opened 2021-08-02
[Windows] Clean up and simplify Windows templates

maxim-lobanov:
# Description
This PR clean up templates by hardcoding some values.
It is easier to write `$env:TEMP "toolset.json"` in two places instead of having huge expression in every provisioner:
```json
"environment_vars": [
    "TOOLSET_JSON_PATH={{user `toolset_json_path`}}"
],
```

It is easy to forget passing values to some provisioner group for example.
In this case, hardcoding values simplify template code and reduce templates (this approach is already used in macOS).

Also, union some scripts to single provisioner group


images/win/scripts/ImageHelpers/InstallHelpers.ps1
```diff
@@ -318,7 +318,8 @@ function Get-VSExtensionVersion
 
 function Get-ToolsetContent
 {
-    $toolsetJson = Get-Content -Path $env:TOOLSET_JSON_PATH -Raw
+    $toolsetPath = Join-Path $env:TEMP "toolset.json"
+    $toolsetJson = Get-Content -Path $toolsetPath -Raw
```
catthehacker, line 322: Wouldn't this be better?
```suggestion
    $toolsetJson = Get-Content -Path (Join-Path $env:TEMP "toolset.json") -Raw
```

maxim-lobanov: Thank you for your comment, I would prefer to keep it as multiple line because in my opinion, it is better for readability.

al-cheb (approved):

miketimofeev (approved):

AlenaSviridenko (approved):

PR #3839, opened 2021-08-03
[Windows] Implement minor improvements, rev 2

maxim-lobanov:
# Description
This PR brings minor improvements to Windows image generation:
1) Make sure that `MSYS2` installation works correctly if no mingw, msys packages are specified in toolset
2) `Install-WDK.ps1` and `Install-Wix.ps1`  should fail on different platforms because they are not compatible with them
3) Don't run TLS tests in scope of `images/win/scripts/Installers/Update-DotnetTLS.ps1` because changes are not applied immediately
4) Exclude `Cached Docker images` section from docs if no containers are cached

## Check list

### DIFF
--- a/images/win/scripts/Installers/Install-Msys2.ps1
+++ b/images/win/scripts/Installers/Install-Msys2.ps1
@@ -70,7 +70,6 @@ function Install-MingwPackages($Packages) {
   }
 }
 
-
 Install-Msys2
 
 # Add msys2 bin tools folders to PATH temporary
--- a/images/win/scripts/Installers/Install-Msys2.ps1
+++ b/images/win/scripts/Installers/Install-Msys2.ps1
@@ -10,19 +10,68 @@
 $dash = "-" * 40
 $origPath = $env:PATH
 
-$msys2_release = "https://api.github.com/repos/msys2/msys2-installer/releases/latest"
-$msys2Uri = ((Invoke-RestMethod $msys2_release).assets | Where-Object {
-  $_.name -match "^msys2-x86_64" -and $_.name.EndsWith(".exe") }).browser_download_url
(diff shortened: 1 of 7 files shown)


al-cheb (dismissed):

MaksimZhukov (approved):

AlenaSviridenko (dismissed):

images/win/scripts/Installers/Install-Msys2.ps1
```diff
@@ -10,19 +10,68 @@
...
+
+  $pkgs = pacman.exe -Q
+
+  foreach ($arch in $archs)
+  {
+    Write-Host "`n$dash Installed $arch packages"
+    $pkgs | grep ^${arch}-
+  }
+}
+
```
miketimofeev, line 72: extra line

images/win/scripts/Installers/Install-Msys2.ps1
```diff
@@ -35,40 +84,10 @@ pacman.exe -Syuu  --noconfirm
 taskkill /f /fi "MODULES eq msys-2.0.dll"
```
miketimofeev, line 84: Do we want to do all these steps in case we don't install any packages?

maxim-lobanov: I am not sure about exact purpose of these lines so decided to keep them. Do you have any context?
May be @al-cheb knows

maxim-lobanov: Per discussion with @al-cheb , these lines should be run even if we don't install packages.

al-cheb (approved):

miketimofeev (approved):

PR #3864, opened 2021-08-06
[Linux] install.sh: make the IsPackageInstalled function definition consistent

mikhailkoliada:
# Description
According to the [bash manual](http://www.gnu.org/savannah-checkouts/gnu/bash/manual/bash.html#Shell-Functions) the `function` keyword is not really mandatory, other functions defined in the file have no this keyword as well.

## Check list
- [n/a ] Related issue / work item is attached
- [n/a ] Tests are written (if applicable)
- [n/a ] Documentation is updated (if applicable)
- [n/a ] Changes are tested and related VM images are successfully generated

### DIFF
--- a/images/linux/scripts/helpers/install.sh
+++ b/images/linux/scripts/helpers/install.sh
@@ -48,7 +48,7 @@ download_with_retries() {
 ## if ! IsPackageInstalled packageName; then
 ##     echo "packageName is not installed!"
 ## fi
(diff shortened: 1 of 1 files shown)


miketimofeev (approved):

Milujem (commented):
@mikhailkoliada

Milujem (commented):
All of these hotels

PR #3888, opened 2021-08-12
Created template for Macos 11 using Anka

LeonidLapshin:


images/macos/templates/macOS-11-anka.pkr.hcl
```diff
@@ -0,0 +1,220 @@
...
+    type = string
+    default = "bigsur_300g"
+  }
+  
+  variable "vm_name" {
+    type = string
+    default = "71"
+  }
+  
+  variable "vcpu_count" {
```
maxim-lobanov, line 16: Do we really want to define all these values as a variables?
May be define them directly to
```
  source "veertu-anka-vm-clone" "template" {
    vm_name = "${var.vm_name}"
    source_vm_name = "${var.source_vm_name}"
    source_vm_tag = "${var.source_vm_tag}"
    vcpu_count = "${var.vcpu_count}"
    ram_size = "${var.ram_size}"
    stop_vm =  "true"
  }
```
section?
Currently, It looks a bit massively.

images/macos/templates/macOS-11-anka.pkr.hcl
```diff
@@ -0,0 +1,220 @@
...
+      source      = "./tests"
+    }
+    provisioner "file" {
+      destination = "image-generation/"
+      source      = "./software-report"
+    }
+    provisioner "file" {
+      destination = "image-generation/"
+      source      = "./helpers"
+    }
```
maxim-lobanov, line 68: I think we can unite all these provisioners into the single one.
`file` provisioner supports `sources` argument: https://www.packer.io/docs/provisioners/file

PR #3916, opened 2021-08-17
[Ubuntu] fix: add aria2 to toolset

catthehacker:
# Description
`aria2` is already installed as part of https://github.com/actions/virtual-environments/blob/133db654885e1796cc158ab04f9217826294857a/images/linux/scripts/base/apt.sh#L29
This change will add `aria2` to list of apt packages in generated README the same way `jq` is already done 

## Check list

### DIFF
--- a/images/linux/scripts/tests/Apt.Tests.ps1
+++ b/images/linux/scripts/tests/Apt.Tests.ps1
@@ -14,6 +14,11 @@ Describe "Apt" {
             $toolName = "getfacl"
         }
 
+        if ($toolName -eq "aria2")
+        {
+            $toolName = "aria2c"
+        }
+
         if ($toolName -eq "p7zip-full")
         {
             $toolName = "p7zip"
--- a/images/linux/scripts/base/apt.sh
+++ b/images/linux/scripts/base/apt.sh
@@ -25,8 +25,8 @@ echo 'APT sources limited to the actual architectures'
 cat /etc/apt/sources.list
 
 apt-get update
-# Install aria2 , jq
-apt-get install aria2 jq
+# Install jq
+apt-get install jq
 
 # Install apt-fast using quick-install.sh
 # https://github.com/ilikenwf/apt-fast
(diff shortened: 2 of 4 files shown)


images/linux/scripts/tests/Apt.Tests.ps1
```diff
@@ -14,6 +14,11 @@ Describe "Apt" {
             $toolName = "getfacl"
         }
 
+        if ($toolName -eq "aria2")
+        {
+        	$toolName = "aria2c"
+        }
```
maxim-lobanov, line 20:
```suggestion
        if ($toolName -eq "aria2")
        {
            $toolName = "aria2c"
        }
```
nitpick: use spaces instead of tabs

catthehacker: woops, did a quick edit in terminal and didn't noticed whitespace change, pushed fix

miketimofeev:
@catthehacker let's remove it from the https://github.com/actions/virtual-environments/blob/133db654885e1796cc158ab04f9217826294857a/images/linux/scripts/base/apt.sh#L29 as well

catthehacker:
@miketimofeev but `apt-fast` depends on it
https://github.com/actions/virtual-environments/blob/133db654885e1796cc158ab04f9217826294857a/images/linux/scripts/base/apt.sh#L29-L33

catthehacker:
Currently there are few scripts that use `apt-fast` before the toolset is installed https://github.com/actions/virtual-environments/search?q=apt-fast

miketimofeev:
@catthehacker yeah, but apt-fast installation script installs aria2 anyway
https://raw.githubusercontent.com/ilikenwf/apt-fast/master/quick-install.sh
```
  if ! dpkg-query --show aria2 >/dev/null 2>&1; then
    sudo apt-get update
    sudo apt-get install -y aria2
  fi
```

al-cheb (approved):

miketimofeev (approved):

PR #3929, opened 2021-08-19
Add Windows Server 2022 image templates

maxim-lobanov:
# Description
Windows Server 2022 has recently become available in Azure so we are preparing virtual-environments source code to support it.
https://docs.microsoft.com/en-us/windows-server/get-started/windows-server-release-info


al-cheb (approved):

images/win/scripts/Installers/Configure-Shell.ps1
```diff
@@ -15,8 +15,8 @@ C:\msys64\usr\bin\bash.exe -leo pipefail %*
 # gitbash <--> C:\Program Files\Git\bin\bash.exe
 New-Item -ItemType SymbolicLink -Path "$shellPath\gitbash.exe" -Target "$env:ProgramFiles\Git\bin\bash.exe" | Out-Null
 
-# WSL is available on Windows Server 2019
-if (Test-IsWin19)
+# WSL is available on Windows Server 2019 and Windows Server 2022
+if ((Test-IsWin19) -or (Test-IsWin22))
```
MaksimZhukov, line 19: I think, we can simplify this condition as follows `if (!Test-IsWin16)`

dibir-magomedsaygitov (dismissed):

images/win/scripts/Tests/Tools.Tests.ps1
```diff
@@ -42,13 +42,16 @@ Describe "DACFx" {
...
+    }
+
+    It "SqlLocalDB" -Skip:(Test-IsWin22) {
+        $sqlLocalDBPath = 'C:\Program Files\Microsoft SQL Server\130\Tools\Binn\SqlLocalDB.exe'
         "${sqlLocalDBPath}" | Should -Exist
     }
 }
 
-Describe "DotnetTLS" {
+Describe "DotnetTLS" -Skip:(Test-IsWin22) {
```
MaksimZhukov, line 54: Why do we skip it for the Windows 2022? As I can see, we added the Update-DotnetTLS.ps1 script to the windows2022 template.

images/win/toolsets/toolset-2022.json
```diff
@@ -0,0 +1,247 @@
...
+            "name": "az",
+            "url" : "https://raw.githubusercontent.com/Azure/az-ps-module-versions/main/versions-manifest.json",
+            "versions": [
+                "6.1.0"
+            ],
+            "zip_versions": []
+        }
+    ],
+    "java": {
+        "default": "11",
```
MaksimZhukov, line 107: Should it be Java 8?

maxim-lobanov: Fixed

images/win/scripts/Tests/Tools.Tests.ps1
maxim-lobanov: I think that `Update-DotnetTLS.ps1` and `DotnetTLS` test are not connected directly.
ps1 script sets default TLS version for .NET Core (and it is still needed on Windows Server 2022) but `DotnetTLS` validates system TLS.

System TLS should be set to 1.2 on Windows Server 2016 and 2019. On Windows Server 2022 it is not needed because it uses TLS 1.3 by default (and this test doesn't work properly for it)

MaksimZhukov (approved):

dibir-magomedsaygitov (approved):

al-cheb (approved):

Issue #3641, opened 2021-06-22
Windows image: Add Windows 10 SDK 10.0.20348.0


DSMitten:
**Tool information**
- Tool name: Windows 10 SDK 10.0.20348.0
- Add or update? Add
- Desired version: 10.0.20348.0
- Approximate size: 
- If this is an add request:
  - Brief description of tool: The latest published version of the Windows 10 SDK
  - URL for tool's homepage: https://developer.microsoft.com/en-us/windows/downloads/windows-10-sdk/

**Area for Triage**:

**Question, Bug, or Feature?**:

**Virtual environments affected**
- [ ] Ubuntu 16.04
- [ ] Ubuntu 18.04
- [ ] Ubuntu 20.04
- [ ] macOS 10.15
- [ ] macOS 11
- [ ] Windows Server 2016 R2
- [X] Windows Server 2019

**Can this tool be installed during the build?**
Yes, just run Win10 SDK installer, but it takes a long time.

**Tool installation time in runtime**
5-15 minutes.

**Are you willing to submit a PR?**
no

AlenaSviridenko:
Hi @DSMitten,
we are installing Windows 10 SDK here to install WDK 
https://github.com/actions/virtual-environments/blob/405a368b22cbac48f31ce7b35f415369d31a27b1/images/win/scripts/Installers/Install-WDK.ps1#L9

Do I get it right that it doesn't fit your needs, and you are asking for `Windows 10 SDK version 2104 and Server development`?

DSMitten:
Yes, we're asking for Windows 10 SDK version 2014 and Server development, link https://go.microsoft.com/fwlink/?linkid=2164145

When we switched from C++17 to C++20, we had compile errors in Windows header files, and we had to upgrade to the most recent release in order to build successfully. We are currently running the SDK installer now, but installing the SDK takes as long as building our application, so it would be very helpful if it was part of the VM image.

urkle:
This would be nice as I'm dealing with all sorts of issues with 10.0.17763 SDK (as there are bugs in many headers that cause compilation errors when using /W4 or /Wall as one project I'm building does).. and everything works fine w/ 10.0.14393.0 on the windows-2016 image.

AlenaSviridenko:
Hi,
We are concerned about installing this new big tools to our Windows images due to its size, possible maintenance effort and tool popularity, so I would suggest keep installing it in runtime. 
I am closing this issue, please, feel free to contact us in case of any questions.

[...]

PR #3942, opened 2021-08-21
Update windows-2019 image to use 20348 SDK & WDK

jpark37:
# Description
Need more recent Windows SDK to compile against new GraphicsCaptureAccess class. Also updated WDK because of the comment that they need to be on the same version.

Verified correct version of SDK, WDK, and WDK VS extension were all installed to a generated VM. My application compiles successfully.

I more or less did the same things as the previous upgrade to 19041: #935

#### Related issue:
Someone opened issue #3641 a couple months ago, and it was closed, but I'm hoping this is fine since the PR is similar to the previous one.

## Check list

### DIFF
--- a/images/win/scripts/Installers/Install-WDK.ps1
+++ b/images/win/scripts/Installers/Install-WDK.ps1
@@ -6,8 +6,8 @@
 # Requires Windows SDK with the same version number as the WDK
 if (Test-IsWin19)
 {
-    $winSdkUrl = "https://go.microsoft.com/fwlink/p/?linkid=2120843"
-    $wdkUrl = "https://go.microsoft.com/fwlink/?linkid=2128854"
+    $winSdkUrl = "https://go.microsoft.com/fwlink/?linkid=2164145"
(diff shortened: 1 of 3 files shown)


MehdiChinoune:
This should be added to windows-2022

jpark37:
Are you saying it should be added instead, or in addition?

It looks like the windows-2022 image is in beta, and uses VS 2022, which is also in beta. This 20348 SDK is listed as one of the components for VS 2019, so it seems reasonable to have it in the windows-2019 image.
image

MehdiChinoune:
> Are you saying it should be added instead, or in addition?
> 
> It looks like the windows-2022 image is in beta, and uses VS 2022, which is also in beta. This 20348 SDK is listed as one of the components for VS 2019, so it seems reasonable to have it in the windows-2019 image.
> image

in addition.

jpark37:
It doesn't look like the 2022 image is calling the WDK installation script yet, so I set this PR to draft for now.

maxim-lobanov (approved):

miketimofeev:
@jpark37 thank you for your contribution!

Issue #3953, opened 2021-08-23
MacOS 11 images contain invalid link to release and software report


maxim-lobanov:
### Description

![image](https://user-images.githubusercontent.com/16715858/130491627-ca337aeb-b69e-459a-b61f-a012742af392.png)
Environment and links contain `macos-11.5` term. It is invalid and all links are broken due to that.

The correct environment should be `macos-11`.

Looks like issue come from our repo and [preimagedata.sh](https://github.com/actions/virtual-environments/blob/main/images/macos/provision/configuration/preimagedata.sh#L11) file.

`label_version=$(echo $os_version | cut -d. -f1,2)` works correctly for `10.14`, `10.15` but for Big Sur, we need to invoke `cut -d. -f1` only.

### Virtual environments affected

- [ ] Ubuntu 16.04
- [ ] Ubuntu 18.04
- [ ] Ubuntu 20.04
- [ ] macOS 10.15
- [X] macOS 11
- [ ] Windows Server 2016
- [ ] Windows Server 2019
- [ ] Windows Server 2022

### Image version and build link

https://github.com/maxim-lobanov/test-repo/runs/3403094283?check_suite_focus=true

### Is it regression?

No

### Repro steps

See above

Darleev:
Hello,
The issue has been fixed.

PR #3954, opened 2021-08-23
[macOS 11] fix for invalid link to release and software report

Darleev:
# Description
Environment and links contain macos-11.5 term. It is invalid and all links are broken due to that.
The correct environment should be macos-11.

#### Related issue:
https://github.com/actions/virtual-environments/issues/3953

## Check list

### DIFF
--- a/images/macos/provision/configuration/preimagedata.sh
+++ b/images/macos/provision/configuration/preimagedata.sh
@@ -8,6 +8,9 @@ os_name=$(sw_vers -productName)
 os_version=$(sw_vers -productVersion)
 os_build=$(sw_vers -buildVersion)
 label_version=$(echo $os_version | cut -d. -f1,2)
+if is_BigSur; then
+  label_version=$(echo $os_version | cut -d. -f1)
+fi
 image_label="macos-${label_version}"
 release_label="macOS-${label_version}"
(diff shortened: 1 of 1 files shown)


images/macos/provision/configuration/preimagedata.sh
```diff
@@ -8,6 +8,9 @@ os_name=$(sw_vers -productName)
 os_version=$(sw_vers -productVersion)
 os_build=$(sw_vers -buildVersion)
 label_version=$(echo $os_version | cut -d. -f1,2)
+if is_BigSur; then
```
maxim-lobanov, line 11: Two suggestions:
1) include both branches to `if-else` statement to simplify readability.
2) Swap condition branches
```
# preudocode
if is_Less_BigSur {
  label_version=$(echo $os_version | cut -d. -f1,2)
} else {
  label_version=$(echo $os_version | cut -d. -f1)
}
```
So we won't need to edit this condition when macOS 12 / 13 is added

Darleev: Thank you, resolved.

maxim-lobanov (approved):

al-cheb (approved):

MaksimZhukov (approved):

Issue #3855, opened 2021-08-04
Make it easier to output "Managed Images" or "Shared Gallery Images" from template


spoelstraethan:
### Description

Currently we are building our own version of the agents to run in a Azure VM Scale Set in our accounts, but the VHD type throws a deprecation warning from Packer so we looked into "Managed Images" instead, and it turned out to be pretty easy, but not very streamlined, but it could be partially automated by integrating with the current Powershell scripts.

```
DEBUG:  217+      >>>> Get-LatestCommit -ErrorAction SilentlyContinue
DEBUG:   51+     process  >>>> {
DEBUG:   52+          >>>> Write-Host "Latest commit:"
Latest commit:
DEBUG:   53+          >>>> git --no-pager log --pretty=format:"Date: %cd; Commit: %H - %s; Author: %an <%ae>" -1
Date: Wed Jul 14 10:54:38 2021 +0000; Commit: ad358c8eb8fde1961567f6e3751ee47d66dc774a - Updating readme file for macOS-11 version 20210712.4 (#3723); Author: github-actions[bot] <41898282+github-actions[bot]@users.noreply.github.com>DEBUG:   54+      >>>> }
DEBUG:  219+      >>>> $packerBinary = Get-Command "packer"
DEBUG:  220+     if ( >>>> -not ($packerBinary)) {
DEBUG:  224+     if( >>>> $RestrictToAgentIpAddress -eq $true) {
# ... snipped
vhd: output will be in this color.

==> vhd: Running builder ...
==> vhd: Getting tokens using client secret
==> vhd: Getting tokens using client secret
    vhd: Creating Azure Resource Manager (ARM) client ...
==> vhd: Warning: You are using Azure Packer Builder to create VHDs which is being deprecated, consider using Managed Images. Learn more https://www.packer.io/docs/builders/azure/arm#azure-arm-builder-specific-options
```

The code we are using to make the switch as part of our build pipeline.

```
            git clone https://github.com/actions/virtual-environments.git
            Set-Location ./virtual-environments
            $templatePath = "images\win\windows2019.json"
            $JSON = Get-Content -Path "$templatePath" -Raw | ConvertFrom-Json
            # Print the incoming JSON
            $JSON | ConvertTo-Json -Depth 4 | Out-Host
            # Remove the builder properties that are only for VHD builds
            # basically https://github.com/actions/virtual-environments/blob/d522cafafedd41e155754b78f292b14f3ff62870/images/win/windows2019.json#L45-L50
[...]

PR #3955, opened 2021-08-23
Add note about managed image or shared gallery image

espoelstra:
# Description
Adds notes and code snippet on outputting a managed image or to a shared image gallery instead of a VHD.

#### Related issue:
#3855 

## Check list

### DIFF
--- a/docs/create-image-and-azure-resources.md
+++ b/docs/create-image-and-azure-resources.md
@@ -69,6 +69,79 @@ The function automatically creates all required Azure resources and kicks off pa
 
 *Please, check synopsis of `GenerateResourcesAndImage` for details about non-mandatory parameters.*
 
+##### Alternate output types (Managed Imaged/Shared Image Gallery)
+
+If you would like to use a more modern output image type you can add a small snippet to your pipeline to output a Managed Image or Shared Gallery Image instead.
+
+```
+git clone https://github.com/actions/virtual-environments.git
+Set-Location ./virtual-environments
+$templatePath = "images\win\windows2019.json"
+$JSON = Get-Content -Path "$templatePath" -Raw | ConvertFrom-Json
+# Print the incoming JSON
+$JSON | ConvertTo-Json -Depth 8 | Out-Host
+# Remove the builder properties that are only for VHD builds
+# basically https://github.com/actions/virtual-environments/blob/d522cafafedd41e155754b78f292b14f3ff62870/images/win/windows2019.json#L45-L50
+$JSON.builders[0].psobject.properties.remove('resource_group_name')
+$JSON.builders[0].psobject.properties.remove('storage_account')
+$JSON.builders[0].psobject.properties.remove('capture_container_name')
+$JSON.builders[0].psobject.properties.remove('capture_name_prefix')
+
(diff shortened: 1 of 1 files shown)


docs/create-image-and-azure-resources.md
```diff
@@ -69,6 +69,79 @@ The function automatically creates all required Azure resources and kicks off pa
 
 *Please, check synopsis of `GenerateResourcesAndImage` for details about non-mandatory parameters.*
 
+##### Alternate output types (Managed Imaged/Shared Image Gallery)
+
+If you would like to use a more modern output image type you can add a small snippet to your pipeline to output a Managed Image or Shared Gallery Image instead.
+
+```
```
catthehacker, line 76:
```suggestion
```pwsh
```

docs/create-image-and-azure-resources.md
```diff
@@ -69,6 +69,79 @@ The function automatically creates all required Azure resources and kicks off pa
...
+
+# Print the resulting JSON
+$JSON | ConvertTo-Json -Depth 8 | Out-Host
+# Write the JSON to a file
+$JSON | ConvertTo-Json -Depth 8 | Out-File $templatePath -Encoding Ascii -Force
+```
+
+For a Shared Gallery Image you need to have created the gallery and the initial Image Definition beforehand (defining Windows or Linux as the osType as appropriate).
+
+```
```
catthehacker, line 106:
```suggestion
```pwsh
```

roryza:
I like the bit of code to change the packer command in the JSON. I think it would ideally just be integrated into the main GenerateResourcesAndImage script as an option, which could then apply to any builder JSON.

If you use {{env `ARM_SUBSCRIPTION_ID`}} (has to be the same) , {{env `ARM_RESOURCE_GROUP`}} (at least as a default this could be the same param) and {{env `ARM_RESOURCE_LOCATION`}} (more than likely the same) in the JSON there's less to "customise" (or add as paramters to GenerateResourcesAndImage) and you basically just need gallery name, image name, image version.

I also think it might read a bit confusing in the docs to have this new segment, and then immediately followed by the section on CreateAzureVMFromPackerTemplate which isn't relevant if you're using Image Gallery/Managed Images.

miketimofeev:
@espoelstra We are sorry, but we decided not to add this guide to our docs even though it can be useful for someone. The reason is the maintenance burden — if something changes in the Managed Images/Shared Image Gallery then the deployment process can be broken and we will have to invest time in investigating the root cause and fixing the docs. We might get back to provide support for such types of images when we switch the image generation to use packer hcl files because it will allow us to create multiple configuration files with different variables.
Thank you!

brgsstm:
Just want to add here that I've been using this solution with success, however now that Ubuntu2204 (and I assume other ImageType moving forward) is using hcl2, the soultion is no longer viable.  Modifying and re-outputting hcl2 appears to only be possible in Go, which is way more involved than modifying JSON.

brgsstm:
Just to add - the reason we need to make these changes are due to Azure Policy that MS support as best practice, in that assigning public IPs to VMs is a bad idea.  MS also push a move to Managed Disk's rather than using VHDs so it would be nice to see this tooling make those best practices available as options.

PR #3960, opened 2021-08-25
[Windows] win: add validation InstalledSoftware.md and testResults.xml files

al-cheb:
# Description
Currently, if  tests are not passed or InstalledSoftware.md is missing the result status is green:
- Add validation InstalledSoftware.md
- Add validation testResults.xml 
- Exclude KB5005030
- Fix nginx test

## Check list

### DIFF
--- a/images/win/scripts/ImageHelpers/TestsHelpers.ps1
+++ b/images/win/scripts/ImageHelpers/TestsHelpers.ps1
@@ -53,6 +53,10 @@ function Invoke-PesterTests {
     if ($TestName) {
         $configuration.Filter.FullName = $TestName
     }
+    if ($TestFile -eq "*") {
+        $configuration.TestResult.Enabled = $true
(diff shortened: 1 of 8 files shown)


images/win/scripts/ImageHelpers/TestsHelpers.ps1
```diff
@@ -53,6 +53,10 @@ function Invoke-PesterTests {
     if ($TestName) {
         $configuration.Filter.FullName = $TestName
     }
+    if ($TestFile -eq "*") {
+        $configuration.TestResult.Enabled = $true
+        $configuration.TestResult.OutputPath = "C:\image\Tests\testResults.xml"
+    }
```
maxim-lobanov, line 59: Could you please double check that Pester will print test output (test results) to console with this option?
I would prefer to see test results in build logs too

PR #3985, opened 2021-08-30
[Windows] Add adoptium to javatools installer

mikhailkoliada:
# Description

- sets JAVA_HOME_8_X64 to use Adoptium jdk
- sets JAVA_HOME_11_x64 to use Adoptium jdk
- sets JAVA_HOME_13_X64 to use Adopt jdk (Adoptium is not available here at all)

## Check list

### DIFF
--- a/images/win/scripts/Installers/Install-JavaTools.ps1
+++ b/images/win/scripts/Installers/Install-JavaTools.ps1
@@ -7,10 +7,11 @@ function Set-JavaPath {
     param (
         [string] $Version,
         [string] $Architecture = "x64",
-        [switch] $Default
+        [switch] $Default,
+        [string] $VendorName
     )
 
-    $javaPathPattern = Join-Path -Path $env:AGENT_TOOLSDIRECTORY -ChildPath "Java_Adopt_jdk/${Version}*/${Architecture}"
+    $javaPathPattern = Join-Path -Path $env:AGENT_TOOLSDIRECTORY -ChildPath "Java_${VendorName}_jdk/${Version}*/${Architecture}"
     $javaPath = (Get-Item -Path $javaPathPattern).FullName
 
     if ([string]::IsNullOrEmpty($javaPath)) {
@@ -23,6 +24,7 @@ function Set-JavaPath {
 
     if ($Default)
     {
+        # Clean up any other Java folders from PATH to make sure that they won't conflict with each other
         $currentPath = Get-MachinePath
 
         $pathSegments = $currentPath.Split(';')
@@ -47,14 +49,21 @@ function Set-JavaPath {
     }
 }
 
-function Install-JavaFromAdoptOpenJDK {
+function Install-JavaJDK {
     param(
         [string] $JDKVersion,
-        [string] $Architecture = "x64"
+        [string] $Architecture = "x64",
+        [string] $VendorName
     )
 
-    # Get Java version from adopt openjdk api
-    $assetUrl = Invoke-RestMethod -Uri "https://api.adoptopenjdk.net/v3/assets/latest/${JDKVersion}/hotspot"
+    # Get Java version from api
+    if ($VendorName -eq "Temurin-Hotspot") {
+        $assetUrl = Invoke-RestMethod -Uri "https://api.adoptium.net/v3/assets/latest/${JDKVersion}/hotspot"
+    } elseif ($VendorName -eq "Adopt") {
+        $assetUrl = Invoke-RestMethod -Uri "https://api.adoptopenjdk.net/v3/assets/latest/${JDKVersion}/hotspot"
+    } else {
+        throw "$VendorName is invalid vendor name. 'Adopt' and 'Temurin-Hotspot' are allowed values"
+    }
     $asset = $assetUrl | Where-Object {
         $_.binary.os -eq "windows" `
         -and $_.binary.architecture -eq $Architecture `
@@ -68,13 +77,13 @@ function Install-JavaFromAdoptOpenJDK {
     # We have to replace '+' sign in the version to '-' due to the issue with incorrect path in Android builds https://github.com/actions/virtual-environments/issues/3014
     $fullJavaVersion = $asset.version.semver -replace '\+', '-'
     # Create directories in toolcache path
-    $javaToolcachePath = Join-Path -Path $env:AGENT_TOOLSDIRECTORY -ChildPath "Java_Adopt_jdk"
+    $javaToolcachePath = Join-Path -Path $env:AGENT_TOOLSDIRECTORY -ChildPath "Java_${VendorName}_jdk"
     $javaVersionPath = Join-Path -Path $javaToolcachePath -ChildPath $fullJavaVersion
     $javaArchPath = Join-Path -Path $javaVersionPath -ChildPath $Architecture
 
     if (-not (Test-Path $javaToolcachePath))
     {
-        Write-Host "Creating Adopt openjdk toolcache folder"
+        Write-Host "Creating ${VendorName} toolcache folder"
         New-Item -ItemType Directory -Path $javaToolcachePath | Out-Null
     }
 
@@ -87,17 +96,36 @@ function Install-JavaFromAdoptOpenJDK {
     New-Item -ItemType File -Path $javaVersionPath -Name "$Architecture.complete" | Out-Null
 }
 
-$jdkVersions = (Get-ToolsetContent).java.versions
-$defaultVersion = (Get-ToolsetContent).java.default
+$toolsetJava = (Get-ToolsetContent).java
+$jdkVendors = $toolsetJava.vendors
+$defaultVendor = $toolsetJava.default_vendor
+$defaultVersion = $toolsetJava.default
 
-foreach ($jdkVersion in $jdkVersions) {
-    Install-JavaFromAdoptOpenJDK -JDKVersion $jdkVersion
+foreach ($jdkVendor in $jdkVendors) {
+    $jdkVendorName = $jdkVendor.name
+    $jdkVersionsToInstall = $jdkVendor.versions
+    
+    $isDefaultVendor = $jdkVendorName -eq $defaultVendor
 
-    if ($jdkVersion -eq $defaultVersion) {
-        Set-JavaPath -Version $jdkVersion -Default
-    } else {
-        Set-JavaPath -Version $jdkVersion
+    foreach ($jdkVersionToInstall in $jdkVersionsToInstall) {
+        $isDefaultVersion = $jdkVersionToInstall -eq $defaultVersion
+
+        Install-JavaJDK -VendorName $jdkVendorName -JDKVersion $jdkVersionToInstall
+
+        if ($isDefaultVendor) {
+            if ($isDefaultVersion) {
+                Set-JavaPath -Version $jdkVersionToInstall -VendorName $jdkVendorName -Default
+            } else {
+                Set-JavaPath -Version $jdkVersionToInstall -VendorName $jdkVendorName
+            }
+        }
     }
+}
+
+# Setup JAVA_HOME_13_X64 as this is the only Adopt version needed
+# There is no jdk 13 on Windows 2022
+if (-not (Test-IsWin22)) {
+    Set-JavaPath -Version "13" -VendorName "Adopt"
 }
 
 # Install Java tools
--- a/images/win/scripts/SoftwareReport/SoftwareReport.Java.psm1
+++ b/images/win/scripts/SoftwareReport/SoftwareReport.Java.psm1
@@ -13,10 +13,11 @@ function Get-JavaVersions {
         $versionInPath = (Split-Path $javaPath) -replace "\w:\\.*\\"
         $version = $versionInPath -replace '-', '+'
         $defaultPostfix = ($javaPath -eq $defaultJavaPath) ? " (default)" : ""
+        $VendorName = ($javaPath -like '*Java_Adopt_jdk*') ? "Adopt OpenJDK" :  "Eclipse Temurin"
 
         [PSCustomObject] @{
             "Version" = $version + $defaultPostfix
-            "Vendor" = "Adopt OpenJDK"
+            "Vendor" = $VendorName
             "Environment Variable" = $_.Name
         }
     }
--- a/images/win/scripts/Tests/Java.Tests.ps1
+++ b/images/win/scripts/Tests/Java.Tests.ps1
@@ -1,6 +1,11 @@
 Describe "Java" {
-    [array]$jdkVersions = (Get-ToolsetContent).java.versions | ForEach-Object { @{Version = $_} }
-    $defaultVersion = (Get-ToolsetContent).java.default
+    $toolsetJava = (Get-ToolsetContent).java
+    $defaultVendor = $toolsetJava.default_vendor
+    $javaVendors = $toolsetJava.vendors
+    $defaultVersion = $toolsetJava.default
+
+    [array]$jdkVersions = ($javaVendors | Where-Object {$_.name -eq $defaultVendor}).versions | ForEach-Object { @{Version = $_} }
+    [array]$adoptJdkVersions = ($javaVendors | Where-Object {$_.name -eq "Adopt"}).versions | ForEach-Object { @{Version = $_} }
 
     It "Java <DefaultJavaVersion> is default" -TestCases @(@{ DefaultJavaVersion = $defaultVersion }) {
         $actualJavaPath = Get-EnvironmentVariable "JAVA_HOME"
@@ -28,6 +33,18 @@ Describe "Java" {
         $result = Get-CommandResult "`"$javaPath`" -version"
         $result.ExitCode | Should -Be 0
 
+        if ($Version -eq 8) {
+            $Version = "1.${Version}"
+        }
+        $result.Output[0] | Should -Match ([regex]::Escape("openjdk version `"${Version}."))
+    }
+
+    It "Java Adopt Jdk <Version>" -TestCases $adoptJdkVersions {
+        $adoptPath = Join-Path (Get-ChildItem ${env:AGENT_TOOLSDIRECTORY}\Java_Adopt_jdk\${Version}*) "x64\bin\java"
+    
+        $result = Get-CommandResult "`"$adoptPath`" -version"
+        $result.ExitCode | Should -Be 0
+    
         if ($Version -eq 8) {
             $Version = "1.${Version}"
         }
--- a/images/win/toolsets/toolset-2022.json
+++ b/images/win/toolsets/toolset-2022.json
@@ -106,8 +106,16 @@
     ],
     "java": {
         "default": "8",
-        "versions": [
-            "8", "11"
+        "default_vendor": "Temurin-Hotspot",
+        "vendors": [
+            {
+                "name": "Temurin-Hotspot",
+                "versions": [ "8", "11" ]
+            },
+            {
+                "name": "Adopt",
+                "versions": [ "8", "11"]
+            }
         ]
     },
     "android": {
--- a/images/win/toolsets/toolset-2016.json
+++ b/images/win/toolsets/toolset-2016.json
@@ -142,8 +142,16 @@
     ],
     "java": {
         "default": "8",
-        "versions": [
-            "8", "11", "13"
+        "default_vendor": "Temurin-Hotspot",
+        "vendors": [
+            {
+                "name": "Temurin-Hotspot",
+                "versions": [ "8", "11" ]
+            },
+            {
+                "name": "Adopt",
+                "versions": [ "8", "11", "13" ]
+            }
         ]
     },
     "android": {
--- a/images/win/toolsets/toolset-2019.json
+++ b/images/win/toolsets/toolset-2019.json
@@ -142,8 +142,16 @@
     ],
     "java": {
         "default": "8",
-        "versions": [
-            "8", "11", "13"
+        "default_vendor": "Temurin-Hotspot",
+        "vendors": [
+            {
+                "name": "Temurin-Hotspot",
+                "versions": [ "8", "11" ]
+            },
+            {
+                "name": "Adopt",
+                "versions": [ "8", "11", "13" ]
+            }
         ]
     },
     "android": {


images/win/scripts/Installers/Install-JavaTools.ps1
```diff
@@ -7,10 +7,11 @@ function Set-JavaPath {
     param (
         [string] $Version,
         [string] $Architecture = "x64",
-        [switch] $Default
+        [string] $DefaultVersion,
```
al-cheb, line 10: Why do we change [switch] to [string] and never use this value?

images/win/scripts/Installers/Install-JavaTools.ps1
```diff
@@ -21,8 +22,10 @@ function Set-JavaPath {
     Write-Host "Set 'JAVA_HOME_${Version}_X64' environmental variable as $javaPath"
     setx JAVA_HOME_${Version}_X64 $javaPath /M
 
-    if ($Default)
+    if ($DefaultVersion)
     {
+        # Here we remove any jdk versions installed prior to the main java installation (say, by Choco)
+        # then we rebuild the %PATH% variable value back to the normal one
```
maxim-lobanov, line 28:
```suggestion
        # Clean up any other Java folders from PATH to make sure that they won't conflict to each other
```

Please try to avoid using `here` and `we` in comments.

images/win/scripts/Installers/Install-JavaTools.ps1
```diff
@@ -47,14 +50,20 @@ function Set-JavaPath {
...
+        [string] $Architecture = "x64",
+        [string] $FullVendorName
     )
 
-    # Get Java version from adopt openjdk api
-    $assetUrl = Invoke-RestMethod -Uri "https://api.adoptopenjdk.net/v3/assets/latest/${JDKVersion}/hotspot"
+    # Get Java version from api
+    if ($FullVendorName -eq "Termurin-Hotspot") {
+        $assetUrl = Invoke-RestMethod -Uri "https://api.adoptium.net/v3/assets/latest/${JDKVersion}/hotspot"
+    }
```
maxim-lobanov, line 63: Please use consistent styling.
If open bracket on the same line with else - let's put close bracket on the same line.

```
if () {
  
} else {

}
```

images/win/scripts/Installers/Install-JavaTools.ps1
```diff
@@ -7,10 +7,11 @@ function Set-JavaPath {
     param (
         [string] $Version,
         [string] $Architecture = "x64",
-        [switch] $Default
+        [string] $DefaultVersion,
+        [string] $FullVendorName
```
al-cheb, line 11: nitpick: FullVendorName -> VendorName

images/win/scripts/Installers/Install-JavaTools.ps1
```diff
@@ -47,14 +50,20 @@ function Set-JavaPath {
...
     )
 
-    # Get Java version from adopt openjdk api
-    $assetUrl = Invoke-RestMethod -Uri "https://api.adoptopenjdk.net/v3/assets/latest/${JDKVersion}/hotspot"
+    # Get Java version from api
+    if ($FullVendorName -eq "Termurin-Hotspot") {
+        $assetUrl = Invoke-RestMethod -Uri "https://api.adoptium.net/v3/assets/latest/${JDKVersion}/hotspot"
+    }
+    else {
```
maxim-lobanov, line 64: Could you please convert this branch to the elseif to check exact Adoptium name and throw exception if unknown vendor is passed?
I would prefer to fail the build if vendor is invalid instead of silently installation of Adopt

```
if ($FullVendorName -eq "Temurin-Hotspot") {

} elseif ($FullVendorName -eq "Adopt") {

} else {
  throw "Invalid vendor name. 'Adopt' and 'Temurin-Hotspot' are allowed values"
}
```

Alternative will be make powershell validation of function params: https://docs.microsoft.com/en-us/powershell/scripting/developer/cmdlet/validateset-attribute-declaration?view=powershell-7.1#parameters

images/win/scripts/Installers/Install-JavaTools.ps1
```diff
@@ -87,19 +96,37 @@ function Install-JavaFromAdoptOpenJDK {
...
-        Set-JavaPath -Version $jdkVersion
     }
+
 }
 
+# Here we setup JAVA_HOME_13_X64 as this is the only Adopt version we want
+# Adoptium has only 8 & 11 so we do not have to check vendor name explicily as we know
+# 13 belongs to Adopt only.
+
+Set-JavaPath -Version $uniqVersion -FullVendorName "Adopt"
```
maxim-lobanov, line 128: May be just hardcode Java 13 here?
```
Set-JavaPath -Version "13" -FullVendorName "Adopt"
```

I don't see much profit of calculating `uniqVersion` in runtime because this logic will fail anyway if we remove Java 13 or add new Java from Adopt

images/win/scripts/Installers/Install-JavaTools.ps1
```diff
@@ -87,19 +96,37 @@ function Install-JavaFromAdoptOpenJDK {
     New-Item -ItemType File -Path $javaVersionPath -Name "$Architecture.complete" | Out-Null
 }
 
-$jdkVersions = (Get-ToolsetContent).java.versions
+$jdkVendors = (Get-ToolsetContent).java.vendors.name
```
al-cheb, line 99: I think we should remove .name property and don't filter at line 104.

```
$jdkVendors = (Get-ToolsetContent).java.vendors
$jdkVersionsToInstall = $jdkVendor.versions
```

images/win/scripts/Tests/Java.Tests.ps1
```diff
@@ -1,5 +1,5 @@
 Describe "Java" {
-    [array]$jdkVersions = (Get-ToolsetContent).java.versions | ForEach-Object { @{Version = $_} }
+    [array]$jdkVersions = (Get-ToolsetContent).java.vendors.versions | Sort | Get-Unique | ForEach-Object { @{Version = $_} }
```
maxim-lobanov, line 2: I think we need only versions from default vendor here. Otherwise, `It "Java <Version>" -TestCases $jdkVersions {` test won't work properly.

images/win/scripts/SoftwareReport/SoftwareReport.Java.psm1
```diff
@@ -16,7 +16,7 @@ function Get-JavaVersions {
 
         [PSCustomObject] @{
             "Version" = $version + $defaultPostfix
-            "Vendor" = "Adopt OpenJDK"
+            "Vendor" = "Adoptium OpenJDK"
```
maxim-lobanov, line 19: What about Adopt? Java 13 will still come from Adopt.

images/win/scripts/Installers/Install-JavaTools.ps1
```diff
@@ -21,8 +22,10 @@ function Set-JavaPath {
     Write-Host "Set 'JAVA_HOME_${Version}_X64' environmental variable as $javaPath"
     setx JAVA_HOME_${Version}_X64 $javaPath /M
 
-    if ($Default)
+    if ($DefaultVersion)
```
maxim-lobanov, line 25: Why do we need to rename [switch]`$Default` parameter to [string]`$DefaultVersion`?
We don't use its value anywhere.

images/win/scripts/Installers/Install-JavaTools.ps1
```diff
@@ -87,19 +96,37 @@ function Install-JavaFromAdoptOpenJDK {
...
+foreach ($jdkVendor in $jdkVendors) {
+
+    $jdkVersionsToInstall = ((Get-ToolsetContent).java.vendors | Where-Object { $_.name -eq $jdkVendor }).versions
+    $isDefaultVendor = $jdkVendor -eq $defaultVendor
+
+    foreach ($jdkVersionToInstall in $jdkVersionsToInstall)
+    {
+        Install-JavaJDK -FullVendorName $jdkVendor -JDKVersion $jdkVersionToInstall
+
+        if ($isDefaultVendor) {
```
maxim-lobanov, line 115: I think we need one more condition here.
We should invoke `Set-JavaPath` with `-DefaultVersion` only when isDefaultVendor and also isDefaultVersion.
If it is not default version, we should invoke `Set-JavaPath` without `-DefaultVersion`

images/win/scripts/SoftwareReport/SoftwareReport.Java.psm1
al-cheb, line 19: I couldn't find `Adoptium OpenJDK`, https://adoptopenjdk.net/ - uses AdoptOpenJDK naming.

images/win/toolsets/toolset-2022.json
```diff
@@ -105,8 +105,16 @@
...
-            "8", "11"
+        "default_vendor": "Temurin-Hotspot",
+        "vendors": [
+            {
+                "name": "Temurin-Hotspot",
+                "versions": [ "8", "11" ]
+            },
+            {
+                "name": "Adopt",
+                "versions": [ "8", "11", "13" ]
```
maxim-lobanov, line 116: Why do we need to install Java 13 on Windows Server 2022? We don't have this version on this platform.

images/win/scripts/Installers/Install-JavaTools.ps1
```diff
@@ -87,19 +96,37 @@ function Install-JavaFromAdoptOpenJDK {
...
-    Install-JavaFromAdoptOpenJDK -JDKVersion $jdkVersion
+# We get the difference between Adopt and Adoptium and then assign it as Adopt's JAVA_HOME
+$uniqVersion = (Get-ToolsetContent).java.vendors.versions | Group-Object -noelement | Where-Object {$_.count -eq 1} | Select-Object -expand Name
+
+foreach ($jdkVendor in $jdkVendors) {
+
+    $jdkVersionsToInstall = ((Get-ToolsetContent).java.vendors | Where-Object { $_.name -eq $jdkVendor }).versions
+    $isDefaultVendor = $jdkVendor -eq $defaultVendor
+
+    foreach ($jdkVersionToInstall in $jdkVersionsToInstall)
```
al-cheb, line 111: Let's revert the previous logic and pass VendorName param when necessary.

images/win/scripts/SoftwareReport/SoftwareReport.Java.psm1
maxim-lobanov: Also, it should be `Eclipse Temurin` instead of `Adoptium`

images/win/scripts/Installers/Install-JavaTools.ps1
```diff
@@ -87,19 +96,37 @@ function Install-JavaFromAdoptOpenJDK {
...
+$jdkVendors = (Get-ToolsetContent).java.vendors.name
+$defaultVendor = (Get-ToolsetContent).java.default_vendor
 $defaultVersion = (Get-ToolsetContent).java.default
 
-foreach ($jdkVersion in $jdkVersions) {
-    Install-JavaFromAdoptOpenJDK -JDKVersion $jdkVersion
+# We get the difference between Adopt and Adoptium and then assign it as Adopt's JAVA_HOME
+$uniqVersion = (Get-ToolsetContent).java.vendors.versions | Group-Object -noelement | Where-Object {$_.count -eq 1} | Select-Object -expand Name
+
+foreach ($jdkVendor in $jdkVendors) {
```
maxim-lobanov, line 106: Just a proposal for discussion:
May be it doesn't make sense to iterate through the vendors here.
We can create two loops:
```
foreach ($jdkVersion in $temurinVersions) {
  Install-JavaJDK -Version $jdkVersion -FullVendorName "Temurin"
  Set-JavaPath -Version $jdkVersion
}

foreach ($jdkVersion in $adoptVersions) {
  Install-JavaJDK -Version $jdkVersion -FullVendorName "Adopt"
}
```

Your current approach is more flexible but we won't have more vendors in future. It is 100%. Having two vendors is temporary solution.

With two looks we can just get rid of logic with default_vendor and simplify a bit.

images/win/scripts/Installers/Install-JavaTools.ps1
mikhailkoliada: Agreed

images/win/scripts/Installers/Install-JavaTools.ps1
mikhailkoliada: Makes sense

images/win/scripts/Installers/Install-JavaTools.ps1
mikhailkoliada: Done

images/win/scripts/Installers/Install-JavaTools.ps1
mikhailkoliada: Done

images/win/toolsets/toolset-2022.json
mikhailkoliada: removed

images/win/scripts/Installers/Install-JavaTools.ps1
mikhailkoliada: Hardcoded

images/win/scripts/Tests/Java.Tests.ps1
mikhailkoliada: Done

images/win/scripts/Installers/Install-JavaTools.ps1
mikhailkoliada: Good catch, otherwise it might indeed fail (as I see per ci)

images/win/scripts/Installers/Install-JavaTools.ps1
mikhailkoliada: Yes, it is hardcoded now for Adopt so filtering is not needed anymore

images/win/scripts/Installers/Install-JavaTools.ps1
```diff
@@ -87,17 +96,38 @@ function Install-JavaFromAdoptOpenJDK {
...
+
+        if ($isDefaultVendor -and $isDefaultVersion) 
+        {
+            Set-JavaPath -Version $jdkVersionToInstall -VendorName $jdkVendor -Default
+
+        } elseif ($isDefaultVendor -and (-not($isDefaultVersion))) {
+            # Set default vendor ant non-default vendor's version
+            # тot to coflict with Adopt
+            Set-JavaPath -Version $jdkVersionToInstall -VendorName $jdkVendor
+        }
```
maxim-lobanov, line 121: I suggest to simplify this condition a bit:
```suggestion
        if ($isDefaultVendor) 
        {
            Set-JavaPath -Version $jdkVersionToInstall -VendorName $jdkVendor -Default:($isDefaultVersion)

        }
```

images/win/scripts/Installers/Install-JavaTools.ps1
```diff
@@ -87,17 +96,38 @@ function Install-JavaFromAdoptOpenJDK {
...
-$jdkVersions = (Get-ToolsetContent).java.versions
+$jdkVendors = (Get-ToolsetContent).java.vendors.name
+$defaultVendor = (Get-ToolsetContent).java.default_vendor
 $defaultVersion = (Get-ToolsetContent).java.default
 
-foreach ($jdkVersion in $jdkVersions) {
-    Install-JavaFromAdoptOpenJDK -JDKVersion $jdkVersion
+foreach ($jdkVendor in $jdkVendors) {
+
+    $jdkVersionsToInstall = ((Get-ToolsetContent).java.vendors | Where-Object { $_.name -eq $jdkVendor }).versions
```
maxim-lobanov, line 105: I agree with @al-cheb suggestion above.
Currently, you do:
```
$jdkVendors = (Get-ToolsetContent).java.vendors.name
foreach ($jdkVendor in $jdkVendors) {
  $jdkVersionsToInstall = ((Get-ToolsetContent).java.vendors | Where-Object { $_.name -eq $jdkVendor }).versions
...
```

I think we don't really need this filtering.
We can simplify logic to 
```
$jdkVendors = (Get-ToolsetContent).java.vendors
foreach ($jdkVendor in $jdkVendors) {
  $jdkVendor = $jdkVendor.name
  $jdkVersionsToInstall = $jdkVendor.versions
...
```

mikhailkoliada: @maxim-lobanov I have no really strong opinion on that suggestion, but I've glanced my drafts/backlogs and I remember that I was proposing something like you and @al-cheb do, but @miketimofeev suggested the current implementation as the more uniform one. Lets stick to the current one, after all it is just simple pipe-filtering with plain value substitution (and readability does not suffer, in my very opinion).

images/win/scripts/Installers/Install-JavaTools.ps1
mikhailkoliada: Stopped on the nested `if else` implementation as discussed privately.

images/win/scripts/Installers/Install-JavaTools.ps1
maxim-lobanov: @mikhailkoliada  I disagree to proceed with current implementation because it is unreasonably overcomplicated.

With your implementation, we iterate through array of objects and retrieve the name of all vendors. After that we iterate through array of names, trying to find object for every name and work with it. I don't really understand why we need to do it.
We should just iterate through the array of objects.

maxim-lobanov (approved):

images/win/scripts/Installers/Install-JavaTools.ps1
```diff
@@ -87,19 +96,40 @@ function Install-JavaFromAdoptOpenJDK {
     New-Item -ItemType File -Path $javaVersionPath -Name "$Architecture.complete" | Out-Null
 }
 
-$jdkVersions = (Get-ToolsetContent).java.versions
+$jdkVendors = (Get-ToolsetContent).java.vendors
```
al-cheb, line 99: nitpick: we can create an additional variable for  `(Get-ToolsetContent).java` part, like:
```
$toolsetJava = (Get-ToolsetContent).java
$jdkVendors = $toolsetJava.vendors
```

images/win/scripts/Tests/Java.Tests.ps1
```diff
@@ -1,5 +1,6 @@
 Describe "Java" {
-    [array]$jdkVersions = (Get-ToolsetContent).java.versions | ForEach-Object { @{Version = $_} }
+    $defaultVendor = (Get-ToolsetContent).java.default_vendor
+    [array]$jdkVersions = ((Get-ToolsetContent).java.vendors | Where-Object {$_.name -eq $defaultVendor}).versions | ForEach-Object { @{Version = $_} }
```
al-cheb, line 3: I see we get only "Temurin-Hotspot" version. Where about Adopt are we planning to test them?

mikhailkoliada: @al-cheb it was concerned above https://github.com/actions/virtual-environments/pull/3985#discussion_r698445503

al-cheb: @mikhailkoliada , I mean very simple test just to validate that these versions exist.

Caddelldev19 (changes requested):
@caddell

images/win/scripts/Installers/Install-JavaTools.ps1
```diff
@@ -23,6 +24,7 @@ function Set-JavaPath {
 
     if ($Default)
     {
+        # Clean up any other Java folders from PATH to make sure that they won't conflict to each other
```
miketimofeev, line 27: nitpick:
```suggestion
        # Clean up any other Java folders from PATH to make sure that they won't conflict with each other
```

images/win/scripts/Installers/Install-JavaTools.ps1
```diff
@@ -68,13 +77,13 @@ function Install-JavaFromAdoptOpenJDK {
...
     # Create directories in toolcache path
-    $javaToolcachePath = Join-Path -Path $env:AGENT_TOOLSDIRECTORY -ChildPath "Java_Adopt_jdk"
+    $javaToolcachePath = Join-Path -Path $env:AGENT_TOOLSDIRECTORY -ChildPath "Java_${VendorName}_jdk"
     $javaVersionPath = Join-Path -Path $javaToolcachePath -ChildPath $fullJavaVersion
     $javaArchPath = Join-Path -Path $javaVersionPath -ChildPath $Architecture
 
     if (-not (Test-Path $javaToolcachePath))
     {
-        Write-Host "Creating Adopt openjdk toolcache folder"
+        Write-Host "Creating openjdk toolcache folder"
```
miketimofeev, line 86:
```suggestion
        Write-Host "Creating $VendorName openjdk toolcache folder"
```

images/win/scripts/Installers/Install-JavaTools.ps1
```diff
@@ -87,19 +96,40 @@ function Install-JavaFromAdoptOpenJDK {
...
+            # Set default vendor and non-default vendor's version
+            # not to coflict with Adopt
+            Set-JavaPath -Version $jdkVersionToInstall -VendorName $jdkVendorName
+            }
+        }
     }
 }
 
+# Setup JAVA_HOME_13_X64 as this is the only Adopt version needed
+# There is no jdk 16 on Windows 2022
```
miketimofeev, line 128:
```suggestion
# There is no jdk 13 on Windows 2022
```

images/win/scripts/Installers/Install-JavaTools.ps1
```diff
@@ -87,19 +96,40 @@ function Install-JavaFromAdoptOpenJDK {
...
-        Set-JavaPath -Version $jdkVersion
+    foreach ($jdkVersionToInstall in $jdkVersionsToInstall) {
+        $isDefaultVersion = $jdkVersionToInstall -eq $defaultVersion
+
+        Install-JavaJDK -VendorName $jdkVendorName -JDKVersion $jdkVersionToInstall
+
+        if ($isDefaultVendor) {
+            if ($isDefaultVersion) {
+            Set-JavaPath -Version $jdkVersionToInstall -VendorName $jdkVendorName -Default
+
```
miketimofeev, line 117: extra line

images/win/scripts/Installers/Install-JavaTools.ps1
```diff
@@ -87,19 +96,40 @@ function Install-JavaFromAdoptOpenJDK {
...
+            } else {
+            # Set default vendor and non-default vendor's version
+            # not to coflict with Adopt
```
miketimofeev, line 120: This comment sounds a bit confusing to me. Maybe remove it as the logic is quite clear here?

images/win/scripts/Tests/Java.Tests.ps1
```diff
@@ -1,5 +1,6 @@
 Describe "Java" {
-    [array]$jdkVersions = (Get-ToolsetContent).java.versions | ForEach-Object { @{Version = $_} }
+    $defaultVendor = (Get-ToolsetContent).java.default_vendor
```
miketimofeev, line 2: let's implement the same approach as al-cheb suggested:
```
$toolsetJava = (Get-ToolsetContent).java
$defaultVendor = $toolsetJava.default_vendor
```
and so on

images/win/scripts/Tests/Java.Tests.ps1
miketimofeev: @mikhailkoliada @maxim-lobanov Do we need to test Java-13 anyway as it's in the PATH?

mikhailkoliada: @miketimofeev I think it got changed on this stage https://github.com/actions/virtual-environments/pull/3985#discussion_r698445503

miketimofeev (approved):

images/win/scripts/SoftwareReport/SoftwareReport.Java.psm1
```diff
@@ -13,10 +13,11 @@ function Get-JavaVersions {
         $versionInPath = (Split-Path $javaPath) -replace "\w:\\.*\\"
         $version = $versionInPath -replace '-', '+'
         $defaultPostfix = ($javaPath -eq $defaultJavaPath) ? " (default)" : ""
+        $VendorName = if($javaPath -like '*Java_Adopt_jdk*') { "Adopt OpenJDK" } else { "Eclipse Temurin"}
```
al-cheb, line 16: nitpick: use the same syntax like at line 15

images/win/scripts/Installers/Install-JavaTools.ps1
```diff
@@ -87,19 +96,38 @@ function Install-JavaFromAdoptOpenJDK {
...
-    } else {
-        Set-JavaPath -Version $jdkVersion
+    foreach ($jdkVersionToInstall in $jdkVersionsToInstall) {
+        $isDefaultVersion = $jdkVersionToInstall -eq $defaultVersion
+
+        Install-JavaJDK -VendorName $jdkVendorName -JDKVersion $jdkVersionToInstall
+
+        if ($isDefaultVendor) {
+            if ($isDefaultVersion) {
+            Set-JavaPath -Version $jdkVersionToInstall -VendorName $jdkVendorName -Default
```
al-cheb, line 117: nitpick: add indent

images/win/scripts/Installers/Install-JavaTools.ps1
```diff
@@ -87,19 +96,38 @@ function Install-JavaFromAdoptOpenJDK {
...
+            } else {
+            Set-JavaPath -Version $jdkVersionToInstall -VendorName $jdkVendorName
```
al-cheb, line 119: nitpick: add indent

images/win/scripts/Tests/Java.Tests.ps1
```diff
@@ -1,6 +1,11 @@
 Describe "Java" {
-    [array]$jdkVersions = (Get-ToolsetContent).java.versions | ForEach-Object { @{Version = $_} }
-    $defaultVersion = (Get-ToolsetContent).java.default
+    $toolsetJava = (Get-ToolsetContent).java
+    $defaultVendor = $toolsetJava.default_vendor
+    $javaVendors = $toolsetJava.vendors
+    $defaultVersion = $toolsetJava.default
+
+    [array]$jdkVersions = ($toolsetJava.vendors | Where-Object {$_.name -eq $defaultVendor}).versions | ForEach-Object { @{Version = $_} }
```
miketimofeev, line 7:
```suggestion
    [array]$jdkVersions = ($javaVendors | Where-Object {$_.name -eq $defaultVendor}).versions | ForEach-Object { @{Version = $_} }
```

mikhailkoliada: @miketimofeev yep, you are ahead of me ;)

al-cheb (approved):

miketimofeev (approved):

maxim-lobanov (approved):

MaksimZhukov (approved):

PR #4016, opened 2021-09-03
Improve haskell ghcup

hasufell:
# Description

Improve ghcup handling on linux and mac, freeing up disk space. Add ghcup to windows (yes, it's supported).

This isn't tested manually (I don't know how).

## Check list

### DIFF
--- a/images/win/scripts/Installers/Install-Haskell.ps1
+++ b/images/win/scripts/Installers/Install-Haskell.ps1
@@ -33,3 +33,7 @@ Write-Host 'Installing cabal...'
 Choco-Install -PackageName cabal
 
 Invoke-PesterTests -TestFile 'Haskell'
+
+# install minimal ghcup, utilizing pre-installed msys2 at C:\msys64
+Set-ExecutionPolicy Bypass -Scope Process -Force;[System.Net.ServicePointManager]::SecurityProtocol = [System.Net.ServicePointManager]::SecurityProtocol -bor 3072;Invoke-Command -ScriptBlock ([ScriptBlock]::Create((Invoke-WebRequest https://www.haskell.org/ghcup/sh/bootstrap-haskell.ps1 -UseBasicParsing))) -ArgumentList $false, $true, $true, $false, $false, $false, $false, C:\, "", C:\msys64, C:\cabal
+
--- a/images/linux/scripts/installers/haskell.sh
+++ b/images/linux/scripts/installers/haskell.sh
@@ -25,18 +25,24 @@ curl --proto '=https' --tlsv1.2 -sSf https://get-ghcup.haskell.org | sh > /dev/n
 export PATH="$ghcup_bin:$PATH"
 prependEtcEnvironmentPath $ghcup_bin
 
-# Get 2 latest Haskell Major.Minor versions
-allGhcVersions=$(apt-cache search "^ghc-" | grep -Po '(\d*\.){2}\d*' | sort --unique --version-sort)
-ghcMajorMinorVersions=$(echo "$allGhcVersions" | cut -d "." -f 1,2 | sort --unique --version-sort | tail -2)
+availableVersions=$(ghcup list -t ghc -r | grep -v "prerelease" | awk '{print $2}')
 
-# We are using apt-get to install ghc, not ghcup,
-# because ghc installed through ghcup takes up too much disk space (2GB versus 1GB through apt-get)
-for version in $ghcMajorMinorVersions; do
-    # Get latest patch version for given Major.Minor one (ex. 8.6.5 for 8.6) and install it
-    exactVersion=$(echo "$allGhcVersions" | grep $version | sort --unique --version-sort | tail -1)
-    apt-get install -y ghc-$exactVersion
-    ghcInstalledVersions+=("$exactVersion")
-    defaultGHCVersion=$exactVersion
+# Install 3 latest major versions(For instance 8.6.5, 8.8.4, 8.10.2)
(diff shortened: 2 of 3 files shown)


images/win/scripts/Installers/Install-Haskell.ps1
```diff
@@ -33,3 +33,7 @@ Write-Host 'Installing cabal...'
 Choco-Install -PackageName cabal
 
 Invoke-PesterTests -TestFile 'Haskell'
+
+# install minimal ghcup, utilizing pre-installed msys2 at C:\msys64
+Set-ExecutionPolicy Bypass -Scope Process -Force;[System.Net.ServicePointManager]::SecurityProtocol = [System.Net.ServicePointManager]::SecurityProtocol -bor 3072;Invoke-Command -ScriptBlock ([ScriptBlock]::Create((Invoke-WebRequest https://www.haskell.org/ghcup/sh/bootstrap-haskell.ps1 -UseBasicParsing))) -ArgumentList $false, $true, $true, $false, $false, $false, $false, C:\, "", C:\msys64, C:\cabal
```
hasufell, line 38: For the arguments of the script see https://gitlab.haskell.org/haskell/ghcup-hs/-/blob/d61981bc1b8552d4ee3e007d5e110326b2f4ec21/bootstrap-haskell.ps1#L15

miketimofeev:
Hi @hasufell!
Thanks for the proposed changes! A few questions\notes:
- Could you please create a separate PR for each OS? It will be much easier to test them and make changes.
- Are there any drawbacks in deleting docs and profiling libs?
- What are the benefits of switching Haskell installation to ghcup for windows and ubuntu? I'm asking because we recommend using the following action to setup Haskell, which can be broken when using different installation paths https://github.com/actions/setup-haskell

hasufell:
> Could you please create a separate PR for each OS? It will be much easier to test them and make changes.

Sure

> Are there any drawbacks in deleting docs and profiling libs?

For CI, I don't think so. In fact, you're already doing that implicitly when using the ubuntu ppa: 

https://github.com/actions/virtual-environments/blob/19717dd75837cd8c0a877f78cb270505af1b1702/images/linux/scripts/installers/haskell.sh#L37

> What are the benefits of switching Haskell installation to ghcup for windows and ubuntu?

Some of the reasons have been discussed here: https://github.com/haskell/actions/issues/70

The main reason is that you get a consistent set of GHC versions across all platforms. Using 3 different methods (ghcup, ubuntu ppa and chocolatey) means you rely on 3 different release schedules of maintainers. On linux, ghcup is very well tested. Windows support is recent and may be discussed separately.

Kleidukos:
@hasufell I seem to recall that there was a disk space penalty when coming from the PPA to ghcup, do you think this is something ghcup can optimise, or will we have to wait for non-profiled bindists from GHCHQ?

Kleidukos:
>     * What are the benefits of switching Haskell installation to ghcup for windows and ubuntu? I'm asking because we recommend using the following action to setup Haskell, which can be broken when using different installation paths https://github.com/actions/setup-haskell

@miketimofeev This repo is abandoned, nowadays everyone uses https://github.com/haskell/actions instead :)

hasufell:
> I seem to recall that there was a disk space penalty when coming from the PPA to ghcup, do you think this is something ghcup can optimise, or will we have to wait for non-profiled bindists from GHCHQ?

See the PR: https://github.com/actions/virtual-environments/pull/4016/files#diff-be47a081552c94d85c52955897f314f7f38270b6444a776f793397e440fa0c85R38

Kleidukos:
@hasufell ah beautiful, I had missed that part. :)

jared-w:
> Hi @hasufell!
> Thanks for the proposed changes! A few questions\notes:
> 
>     * Could you please create a separate PR for each OS? It will be much easier to test them and make changes.
> 
>     * Are there any drawbacks in deleting docs and profiling libs?
> 
>     * What are the benefits of switching Haskell installation to ghcup for windows and ubuntu? I'm asking because we recommend using the following action to setup Haskell, which can be broken when using different installation paths https://github.com/actions/setup-haskell



> > * What are the benefits of switching Haskell installation to ghcup for windows and ubuntu?
> >   I'm asking because we recommend using the following action to setup Haskell, which can be broken when using different installation paths https://github.com/actions/setup-haskell
> 
> @miketimofeev This repo is abandoned, nowadays everyone uses https://github.com/haskell/actions instead :)

In addition to everyone using haskell/actions instead, that repo is something that tries to always correctly follow whatever actions/virtual-environments does so that the github provided runners will always work and always correctly utilize pre-installed resources whenever possible. So don't worry about breaking things for the action; but a little heads-up before doing so is always nice. That said, the action is also intended to be robust to missing binaries, setup, etc., as much as reasonable so that it can be used on self hosted runners as well; so changing things in actions/virtual-environments _shouldn't_ break anything in the haskell/actions repo.

Bodigrim:
@miketimofeev 
https://github.com/actions/virtual-environments/blob/4394e4414f48c6acaf0272add56b97a802826591/images/linux/scripts/installers/haskell.sh#L17-L21
Note that this PPA is no longer updated and already misses three last releases of GHC (8.10.5, 8.10.6 and 8.10.7). Because of this I strongly support switching to `ghcup` as a source.

miketimofeev:
@Bodigrim thanks for the explanation!

miketimofeev:
@hasufell I'm going to close this PR, please create separate ones for Ubuntu and Windows and we will take care of macOS as, unfortunately, we don't accept macOS contributions at the moment.

PR #4046, opened 2021-09-08
[macOS] Move Ruby and Go versions to toolset

dibir-magomedsaygitov:
# Description
In scope of this PR we are moving default Ruby and Go versions to toolset.

## Check list


images/macos/toolsets/toolset-10.15.json
```diff
@@ -304,5 +304,11 @@
         "xcpretty",
         "bundler",
         "fastlane"
-    ]
+    ],
+    "default_ruby": {
```
al-cheb, line 308: I think it should be called like `ruby` and contains `rubygems` and `default` like `java` section

maxim-lobanov (dismissed):

al-cheb (dismissed):

images/macos/toolsets/toolset-11.json
```diff
@@ -233,12 +233,18 @@
...
+            "xcode-install",
+            "cocoapods",
+            "nomad-cli",
+            "xcpretty",
+            "bundler",
+            "fastlane"
+        ]
+    },
+    "go": {
+        "default": "1.15"
```
MaksimZhukov, line 248: I believe we should remove `go@1.15` from the `common_packages` section as we did for others toolsets

dibir-magomedsaygitov: thank you, fixed

MaksimZhukov (approved):

al-cheb (approved):

images/macos/provision/core/golang.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash -e -o pipefail
+source ~/utils/utils.sh
+
+DEFAULT_GO_VERSION=$(get_toolset_value '.go.default')
+echo "Installing Go..."
+brew_smart_install "go@${DEFAULT_GO_VERSION}"
+
+# Create symlinks for Go 1.15 to preserve backward compatibility
+ln -sf $(brew --prefix go@${DEFAULT_GO_VERSION})/bin/* /usr/local/bin/
```
miketimofeev, line 9: the problem is that we need to create a symlink not only for 1.15 but for any non-latest go version. We have to either change a comment here or create a condition to check whether the installed version is the same as `brew install go` or not

maxim-lobanov (dismissed):

MaksimZhukov (dismissed):

images/macos/provision/core/golang.sh
```diff
@@ -5,7 +5,7 @@ DEFAULT_GO_VERSION=$(get_toolset_value '.go.default')
 echo "Installing Go..."
 brew_smart_install "go@${DEFAULT_GO_VERSION}"
 
-# Create symlinks for Go 1.15 to preserve backward compatibility
+#Create symlinks to preserve backward compatibility. Symlinks are not created when non-latest go is being installed
```
miketimofeev, line 8:
```suggestion
# Create symlinks to preserve backward compatibility. Symlinks are not created when non-latest go is being installed
```

MaksimZhukov (approved):

PR #4052, opened 2021-09-09
[MacOS] fix python installation order

mikhailkoliada:
# Description

The `azure-cli` package pulls python3.9 as a dependency from within the `powershell.sh` installer, which is being called before the actual python installer.  In our `python.sh` script we are installing both python2 from the upstream `.pkg` file and python3 from brew. python2 installation overrides the previously created `/usr/local/bin/2to3` symlink to its own internal path -`/Library/Frameworks/Python.framework/Versions/2.7/bin/2to3`, in the meantime python3 installation attempt is being ignored (as well as the `brew link --overwrite python@3.` call) as it had already been installed by azure cli. To prevent this azure-cli installation has been moved from `powershell.sh` to `commonutils.sh` to fix the installation order.

Apart from that:

- A test has been written to make sure that `2to3` symlink points to the correct place
- `az -v` and the `~/.azure` directory owner fix have been removed (it was figured that they are not needed anymore)

Interesting side effect:

 - Whenever you call `az -v` and then `pwsh -command "& {Import-Module Az}"` the `~/.azure` directory will be created and files created after `Import-Module Az` will be placed there as well
 
This works both ways:

- Whenever you call `pwsh -command "& {Import-Module Az}"` the `~/.Azure` directory will be created and files created after `az -v` call will be placed there.

## Check list

### DIFF
--- a/images/macos/provision/core/powershell.sh
+++ b/images/macos/provision/core/powershell.sh
@@ -1,14 +1,8 @@
 #!/bin/bash -e -o pipefail
 source ~/utils/utils.sh
 
-echo Installing Azure CLI...
-brew_smart_install "azure-cli"
-
 echo Installing PowerShell...
 brew install --cask powershell
-
-# A dummy call of `az` to initialize ~/.azure directory before the modules are installed
-az -v
 
 # Install PowerShell modules
 psModules=$(get_toolset_value '.powershellModules[].name')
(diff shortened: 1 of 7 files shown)


images/macos/provision/core/powershell.sh
```diff
@@ -31,7 +25,4 @@ pwsh -command "& {Import-Module Az}"
 # powershell link was removed in powershell-6.0.0-beta9
 sudo ln -s /usr/local/bin/pwsh /usr/local/bin/powershell
 
-# fix ~/.azure directory permissions
-sudo chown -R ${USER}: $HOME/.azure
```
maxim-lobanov, line 35: Looks like we drop this line? Are you sure that it is safe and won't break users?

mikhailkoliada: @maxim-lobanov Yes, we are ok here, I tested it and the directory and the files in it are created with the standard user's permissions and ownership, might be the fix was needed before, but now it is ok per se.

miketimofeev (approved):

al-cheb (approved):

Issue #1637, opened 2020-09-21
Add test signing support


jamesmistry:
**Tool information**
- Tool name: Test signing support
- Tool license: N/A
- Add or update? Config update
- Desired version: N/A
- Approximate size: 0
- If this is an add request:
  - Brief description of tool: Test signing allows the loading of test signed kernel drivers. Test signing can't be enabled using MS-hosted agents currently because the configuration change needs a reboot to take effect.
  - URL for tool's homepage: N/A
  - Provide a basic test case to validate the tool's functionality: 1. Build and test sign WDK driver. 2. Load the driver using devcon, pnputil or equivalent Windows APIs. If test signing is correctly enabled, the driver will successfully load.

**Area for Triage**:
C/C++

**Question, Bug, or Feature?**:
Feature

**Virtual environments affected**
- [ ] macOS 10.15
- [ ] Ubuntu 16.04 LTS
- [ ] Ubuntu 18.04 LTS
- [ ] Ubuntu 20.04 LTS
- [X] Windows Server 2016 R2
- [ ] Windows Server 2019

Note: not tested on Server 2019

**Can this tool be installed during the build?**
No. This requires a reboot.

**Tool installation time in runtime**
Time of a reboot

**Are you willing to submit a PR?**
Yes

Darleev:
Hello @jamesmistry,

We had a similar request in the past where we have discussed all points related to the feature. Please check it [by the link.](https://github.com/actions/virtual-environments/issues/1497).

Unfortunately, due to unpredictable impact, we are not able to add the feature at this time, but we will keep this request on mind and let you know if we have any updates.

In case of any questions, do not hesitate to contact us.

nibanks:
I would like to request that this be reopened. Several MSFT Windows teams are open sourcing Windows drivers and are leveraging Azure Pipelines for the automation. Currently, with test signing disabled, this forces us to maintain custom, self-hosted agents/machines. With new restrictions on self-hosted agents being put in place we're even more reliant on the default pool machines.

To your comment above related to "unpredictable impact" would you care to elaborate? It's a sandbox environment, right? Anything that breaks is just thrown away anyways. IMO, this "unknown" should not be reason enough to block this feature.

miketimofeev:
@nibanks are there any drawbacks of the "test-mode" besides the ones provided here?
[...]

PR #4098, opened 2021-09-16
[Windows] win2022: enable use of test-signed code

al-cheb:
# Description
https://docs.microsoft.com/en-us/windows-hardware/drivers/install/the-testsigning-boot-configuration-option#behavior-of-windows-when-loading-test-signed-code-is-enabled

Behavior of Windows when loading test-signed code is enabled

- When loading test-signed code is enabled, Windows does the following:
- Displays a watermark with the text "Test Mode" in all four corners of the desktop, to remind users the system has test-signing enabled. Note  Starting with Windows 7, Windows displays this watermark only in the lower right-hand corner of the desktop.
- Displays a watermark with the text "Test Mode" in the lower-left corner of the desktop to remind users that the system has test-signing enabled.
- The operating system loader and the kernel load drivers that are signed by any certificate. The certificate validation is not required to chain up to a trusted root certification authority. However, each driver image file must have a digital signature.

image

#### Related issue:
https://github.com/actions/virtual-environments/issues/1637

## Check list

### DIFF
--- a/images/win/windows2022.json
+++ b/images/win/windows2022.json
@@ -112,6 +112,14 @@
         },
         {
             "type": "powershell",
+            "inline": [
+                "bcdedit.exe /set TESTSIGNING ON"
+            ],
+            "elevated_user": "{{user `install_user`}}",
+            "elevated_password": "{{user `install_password`}}"
+        },
+        {
+            "type": "powershell",
             "environment_vars": [
                 "IMAGE_VERSION={{user `image_version`}}",
                 "IMAGE_OS={{user `image_os`}}",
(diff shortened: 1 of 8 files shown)


miketimofeev (approved):

images/win/windows2022.json
```diff
@@ -143,7 +151,7 @@
         },
         {
             "type": "windows-restart",
-            "restart_timeout": "10m"
+            "restart_timeout": "30m"
```
nibanks, line 154: Just curious, why this change?

nibanks (approved):
Looks good to me. Just wondering why you had to increase a restart timeout.

images/win/windows2022.json
al-cheb: During the image generation several attempts were unsuccessful due to short reboot timeout: `vhd: Timeout waiting for machine to restart.`

nibanks: Ok. Thanks! Interesting that test-signing caused that. Oh well!

PR #4143, opened 2021-09-24
Fix Visual Studio Experimental Instance Initialization

josetr:
We need to wait for the /ResetSettings command to complete otherwise it's not having any effect as shown below:

https://github.com/josetr/VsixTesting/runs/3697544944?check_suite_focus=true#step:4:1
https://github.com/josetr/VsixTesting/runs/3697544919?check_suite_focus=true#step:4:1

### DIFF
--- a/images/win/scripts/Installers/Install-VS.ps1
+++ b/images/win/scripts/Installers/Install-VS.ps1
@@ -34,6 +34,7 @@ if ($instanceFolders -is [array])
 $vsInstallRoot = (Get-VisualStudioInstance).InstallationPath
 
 # Initialize Visual Studio Experimental Instance
+# The Out-Null cmdlet is required to ensure PowerShell waits until the '/ResetSettings' command fully completes.
 & "$vsInstallRoot\Common7\IDE\devenv.exe" /RootSuffix Exp /ResetSettings General.vssettings /Command File.Exit | Out-Null
 
 # Updating content of MachineState.json file to disable autoupdate of VSIX extensions
--- a/images/win/scripts/Installers/Install-VS.ps1
+++ b/images/win/scripts/Installers/Install-VS.ps1
@@ -34,7 +34,7 @@ if ($instanceFolders -is [array])
 $vsInstallRoot = (Get-VisualStudioInstance).InstallationPath
 
 # Initialize Visual Studio Experimental Instance
-& "$vsInstallRoot\Common7\IDE\devenv.exe" /RootSuffix Exp /ResetSettings General.vssettings /Command File.Exit
+& "$vsInstallRoot\Common7\IDE\devenv.exe" /RootSuffix Exp /ResetSettings General.vssettings /Command File.Exit | Out-Null
 
 # Updating content of MachineState.json file to disable autoupdate of VSIX extensions
 $newContent = '{"Extensions":[{"Key":"1e906ff5-9da8-4091-a299-5c253c55fdc9","Value":{"ShouldAutoUpdate":false}},{"Key":"Microsoft.VisualStudio.Web.AzureFunctions","Value":{"ShouldAutoUpdate":false}}],"ShouldAutoUpdate":false,"ShouldCheckForUpdates":false}'


miketimofeev:
Hi @josetr!
Could you please clarify why do we need this `Out-Null`?

josetr:
@miketimofeev The /ResetSettings command usually takes a few minutes to complete as shown in the links I've provided but without the 'Out-Null' switch the `devenv.exe / ResetSettings` command will finish right away (in less than a second) without actually finishing executing the command. 

What probably happens next is that something interferes with the devenv process that is still executing the /ResetSettings command in the background or something kills it and therefore the /ResetSettings command never completes.

Maybe @sharwell knows better.

miketimofeev:
@josetr thanks for the clarification! Another question, what is the impact of resetting\not resetting the settings?

josetr:
I think the impact is clearly seen in the links: it took 5 minutes to reset the settings and 2.5 minutes to execute the tests. In smaller projects, the initialization cost will be even more noticeable and painful. 

To be honest, It's already painful for me because I'm comparing it to AppVeyor which I'm still using because it supports Visual Studio 2015.

Here are the results for the exact same commit in both AppVeyor & GitHub Actions:

https://ci.appveyor.com/project/josetr/vsixtesting/builds/40895007
**4 minutes and 32 seconds** to execute in both Visual Studio 2015 & Visual Studio 2017 since AppVeyor comes with multiple Visual Studio versions in a single image.

https://github.com/josetr/VsixTesting/runs/3697544944?check_suite_focus=true
**9 minutes and 10 seconds** to execute in Visual Studio 2019 only.

It would take more than 16 minutes if Microsoft wasn't so generous with their free concurrent jobs.

The point is that eventually someone will have to reset the settings if they want to deploy an extension to the Experimental Instance or if they want to start it.

josetr:
This is a bit off-topic but should I create an issue to ask for Visual Studio 2015 to be installed in the oldest image like Windows 2016?

I'm just asking because I made a similar issue in the old microsoft/azure-pipelines-image-generation repo but it was a bit ignored.

sharwell:
@josetr as a side note, we finally published [microsoft/vs-extension-testing](https://github.com/microsoft/vs-extension-testing).

Here's the preparation step we run for CI:
https://github.com/microsoft/vs-extension-testing/blob/38397283a48fa7127d2ba8751441d6fd3bc3157b/azure-pipelines/dotnet.yml#L1-L5

It will normally complete quickly, but there is some component of Visual Studio which makes it run extremely slow. If someone managed to capture a PerfView trace of this call, I could finally get that team to fix it.

sharwell:
@josetr there are a bunch of little things I would be interested to discuss with you if you'd like to use microsoft/vs-extension-testing. I've been able to get some internal buy in for the library so you may be able to depend on it working with less maintenance over time (e.g. it should support the next VS release by the time it's made public, so you're always ready to go). It didn't feel appropriate to file an issue about this on your competing library, but if you are interested please feel free to reach out either here or by filing an issue on [microsoft/vs-extension-testing](https://github.com/microsoft/vs-extension-testing) asking about feature gaps (my answer to such a question would have the details for where we could most benefit from merging the separate projects).

miketimofeev:
@josetr Unfortunately we don't have plans to support multiple visual studio versions on the image. Moreover, windows-2016 image will probably be retired next year.
> This is a bit off-topic but should I create an issue to ask for Visual Studio 2015 to be installed in the oldest image like Windows 2016?

miketimofeev:
> I think the impact is clearly seen in the links: it took 5 minutes to reset the settings and 2.5 minutes to execute the tests. In smaller projects, the initialization cost will be even more noticeable and painful.

Now I got it, thanks. Could you please add a comment before the `& "$vsInstallRoot\Common7\IDE\devenv.exe" /RootSuffix Exp /ResetSettings General.vssettings /Command File.Exit | Out-Null` to clarify that we need this out-null to wait for `ResetSettings` completion

josetr:
@sharwell 

At this point, I see no compelling reason to use vs-extension-testing as VsixTesting already does what I need it to do, and keeping it up to date is relatively simple but if at some it gets too hard I may consider it.

We can talk more about this in gitter if you want.

josetr:
> Unfortunately we don't have plans to support multiple visual studio versions on the image. Moreover, windows-2016 image will probably be retired next year.

Ok, thanks. Now I at least know that I will always need AppVeyor.

> Now I got it, thanks. Could you please add a comment before the & "$vsInstallRoot\Common7\IDE\devenv.exe" /RootSuffix Exp /ResetSettings General.vssettings /Command File.Exit | Out-Null to clarify that we need this out-null to wait for ResetSettings completion

Done.

Arifan998 (commented):
review

Arifan998:
images/win/scripts/Installers/Install-VS.ps1

al-cheb (approved):

miketimofeev (approved):

miketimofeev:
@josetr thank you for your contribution! The new images with the changes will be deployed next week.

josetr:
@miketimofeev Thanks a lot for the quick review and merge.

zuiwi:
g

PR #4168, opened 2021-09-28
[Ubuntu] Downgrade git-lfs version to 2.13.3

dibir-magomedsaygitov:
# Description
There is an issue with newest git-lfs version 3.0.0. We need to downgrade to 2.13.3 version until issue is fixed.
Successful run: https://buildcanary.visualstudio.com/PipelineCanary/_build/results?buildId=1811517&view=results

## Check list


dmitry-shibanov (dismissed):

MaksimZhukov (dismissed):

al-cheb (dismissed):

catthehacker:
Maybe it would be better to use GitHub release instead of apt package? that way it will be one repo less to keep track of
https://github.com/git-lfs/git-lfs/releases/latest

images/linux/scripts/installers/git.sh
```diff
@@ -18,7 +18,7 @@ git --version
 
 # Install git-lfs
 curl -s $GIT_LFS_REPO/script.deb.sh | bash
-apt-get install -y git-lfs
+apt-get install -y git-lfs=2.13.3 --allow-downgrades
```
miketimofeev, line 21: Why do we need `--allow-downgrades` flag?

miketimofeev:
> Maybe it would be better to use GitHub release instead of apt package? that way it will be one repo less to keep track of [git-lfs/git-lfs/releases/latest](https://github.com/git-lfs/git-lfs/releases/latest)

I can't recall why we decided to preserve the installation from ppa, but we will definitely revisit this, thanks!

images/linux/scripts/installers/git.sh
dibir-magomedsaygitov: Thank you, removed unnecessary flag

miketimofeev (approved):

Issue #4163, opened 2021-09-28
docker-compose v2 release breaks Linux docker-compose installer


cfbao:
### Description

the [installer](https://github.com/actions/virtual-environments/blob/1e7e6749a5fa9ad8d332d8d93ae24785579e5fc9/images/linux/scripts/installers/docker-compose.sh#L8) looks for `docker-compose-Linux-x86_64` in the latest release, but docker-compose [just released v2](https://github.com/docker/compose/releases), and the asset name is changed to `docker-compose-linux-arm64`.

pin v1 or update the name pattern in jq?

### Virtual environments affected

- [X] Ubuntu 16.04
- [X] Ubuntu 18.04
- [X] Ubuntu 20.04
- [ ] macOS 10.15
- [ ] macOS 11
- [ ] Windows Server 2016
- [ ] Windows Server 2019
- [ ] Windows Server 2022

### Image version and build link

latest

### Expected behavior

docker-compose installed

### Actual behavior

docker-compose.sh fails with

> curl: no URL specified!

### Repro steps

build the Linux runner

spyder007:
It should be noted, also, that the new version seems to be installed as a docker plugin rather than it's own binary.  I fixed the jq select locally, but the unit tests were failing when trying to call docker-compose --version.  It looks like docker-compose is installed before docker, so the unit test is incorrect as well.

miketimofeev:
@cfbao @spyder007 it seems we may stick to version 1 for now and then switch to the moby-compose package when 2.0 is available there.

u605902:
we are also stuck here when trying to build docker image. any temporary solution?

spyder007:
@u605902 

I changed the docker-compose.sh locally to this:
`# Install latest docker-compose from releases
URL=$(curl -s https://api.github.com/repos/docker/compose/releases/latest | jq -r '.assets[].browser_download_url | select(endswith("docker-compose-linux-amd64"))')
curl -L $URL -o /usr/local/bin/docker-compose
chmod +x /usr/local/bin/docker-compose

#invoke_tests "Tools" "Docker-compose"`

I'm testing to see if this works, but I don't use docker-compose in my builds, so I won't be able to tell if compose is successfully installed.

As @miketimofeev mentioned, they may drop to version 1 and use moby compose for 2.0 when it's available.

catthehacker:
New compose version https://github.com/docker/compose/releases/tag/v2.0.1
[...]

PR #4169, opened 2021-09-28
[Ubuntu] Update docker-compose assets filter

al-cheb:
# Description
Update jq filter patter to allow download v2 version due to  `docker-compose-Linux-x86_64`  is changed to `docker-compose-linux-amd64`

#### Related issue:
https://github.com/actions/virtual-environments/issues/4163

## Check list

### DIFF
--- a/images/linux/scripts/installers/docker-compose.sh
+++ b/images/linux/scripts/installers/docker-compose.sh
@@ -5,7 +5,7 @@
 ################################################################################
 
 # Install latest docker-compose from releases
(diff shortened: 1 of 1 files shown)


miketimofeev (approved):

catthehacker:
`compose` v2 should be installed as docker plugin, see https://github.com/docker/compose/issues/8687#issuecomment-929271959

miketimofeev:
> `compose` v2 should be installed as docker plugin, see [docker/compose#8687 (comment)](https://github.com/docker/compose/issues/8687#issuecomment-929271959) (although it might be better if guys maintaining `moby-*` packages at [packages.microsoft.com](https://packages.microsoft.com) could add it to repository since `compose` is a plugin like `buildx`)

Thanks for the clarification! We will contact folks maintaining Microsoft moby packages

PR #4207, opened 2021-10-02
[Ubuntu] Node.js: update LTS versions

EdJoPaTo:
# Description
Update the versions of Node.js (and the related Docker images).

The [Software and image guidelines](https://github.com/actions/virtual-environments/blob/main/docs/software-and-images-guidelines.md) state "3 latest LTS versions" for Node.js and "Tools and versions will typically be removed 6 months after they are deprecated or have reached end-of-life".

Node.js 10 is end-of-life since 2021-04-30 and Node.js 16 LTS starts 2021-10-26. See https://github.com/nodejs/Release

As people already test against the upcoming Node.js 16 LTS it could be added ahead of time.
Feel free to merge now or after 2021-10-26.

## Check list


miketimofeev:
Hi @EdJoPaTo! Thanks for the PR, but, I'm afraid we need to 
- split the PR by OS (unfortunately, we don't accept macOS contributions yet)
- investigate whether it's possible to remove Node10 as azure-devops agent relies on it
- make a proper announcement about node 10 deprecation

If you don't mind I will close this PR and we will take care of nodejs 16 when the time has come.
Thank you!

PR #4242, opened 2021-10-08
[Windows] Install the latest available azure-cli msi package

miketimofeev:
# Description
The latest azure-cli is not available via msi package, it can be installed using python only https://github.com/Azure/azure-cli/issues/19811
This PR changes the way we get the version as the latest release does not always include msi package.

#### Related issue:
n\a

## Check list

### DIFF
--- a/images/win/scripts/Installers/Install-AzureCli.ps1
+++ b/images/win/scripts/Installers/Install-AzureCli.ps1
@@ -4,8 +4,8 @@
 ################################################################################
 
 Write-Host "Install the latest Azure CLI release"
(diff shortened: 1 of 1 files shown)


jiasli:
Hi, this is Azure CLI developer. We would recommend installing from https://aka.ms/installazurecliwindows as shown in https://docs.microsoft.com/en-us/cli/azure/install-azure-cli-windows?tabs=azure-cli. This is the most reliable way.

miketimofeev:
Hi @jiasli! Is the link updated automatically with the new version release? I vaguely remember we had some issues with it back in the days but I might be wrong

PR #4245, opened 2021-10-09
Windows 2022 MSYS2 - add base-devel & compression groups

MSP-Greg:
# Description

Windows 2022 image removed MSYS2 mingw build tools, and also removed 'shared' MSYS2 tools.  Adds base-devel & compression groups to the MSYS2 Install.  Previously, no testing was done on these packages, which do add several exe files.

## Check list

### DIFF
--- a/images/win/toolsets/toolset-2022.json
+++ b/images/win/toolsets/toolset-2022.json
@@ -139,7 +139,10 @@
         }
     },
     "MsysPackages": {
-        "msys2": [],
+        "msys2": [
+            "base-devel",
+            "compression"
+        ],
         "mingw": []
     },
     "windowsFeatures": [


MSP-Greg:
Please consider installing the base-devel and compression groups to the MSYS2 installation on Windows-2022.  Both were installed on previous images.

The base-devel group is shared by the separate build tool chains.  For example, mingw or ucrt gcc won't run unless MSYS2 base-devel is installed.  It contains many build tools that are platform/compiler agnostic.  Without compression, there's no tar.  Pretty common command for CI.  I haven't checked for a while, but at one time the Windows tar did not support all the formats that MSYS2/*nix tar does.

I'm involved with Ruby, and am involved with the custom actions used by most Ruby repos for CI.  Windows Ruby has been built using MSYS2 for several years (it previously used MSYS).  A few Ruby repos run bash scripts, and need a bash shell somewhere, but the majority do not need a bash shell for CI testing, they need the bash shell for build utilities.  Note that many Ruby repos may not require build tools for their specific code, but they may require builds tools for their direct dependencies or their CI testing dependencies.

With Windows-2019, gcc tools are available for mingw64 builds, and many repos use what's installed, since typically they are no more than 3 weeks old due to the update frequency of Actions images.  Other repos (typically the actual Ruby repos) update the build tools and required packages (openssl, etc) so they are immediatedly aware of issues with new package releases.

Now today.  Ruby is now being built with both mingw64 and ucrt64 tools.  It's suggested that caching MSYS2 files be used, but we've found it much more efficient to have a separate cron repo running that packages all the mingw64 and ucrt64 tools and packages.  This means it's stored in one repo (not every repo as with caching), and since the files are independent of the base MSYS2 install, there's no conflict with installing from a prebuilt zip file.  It takes about ten seconds to install.

But, all of the build tool chains require all or part of the base-devel group to be installed.  We have tried to create a prebuilt zip file, and it takes about 20 seconds to decompress.  Also, some of its packages have dependencies that are shared by the existing MSYS install, so there could be issues with updates to those dependencies.

Summing up, without the base-devel and compression groups, many Ruby CI jobs (many only take a minute or two) will be slowed considerably and may be at risk of failure due to the above.  I suspect that the same will be true of some of the other repos using MSYS2 build tools.

**Off-topic**

Obviously, most commercial Ruby applications run under *nix.  But, many people learn coding on Windows, and in many parts of the world, everyone isn't running Win10 or Win11.  Even then, jumping to WSL2/Ubuntu (or native *nix) is a big jump.  Hence, I support Windows Ruby for the new users.  I run Windows (and have done so since DOS days), but most of my coding is using WSL2/Ubuntu.

miketimofeev:
Given the fact that some users will experience a 10+ minutes penalty, we would rather avoid merging these changes.

MSP-Greg:
As mentioned above, using windows-2022 with Ruby repos that require gcc tools, the additional time needed is between 45 and 60 seconds when CI is busy.

We're testing pre-assembling msys2, mingw64, and ucrt64 tools/files three times a day into 7z files, and extracting them on each job.  They're stored in a repo release.  That seems quickest, and removes the traffic from the MSYS2 servers, which would normally happen for each job.

@eine

> Acording to GitHub's search feature, it might be around 300 000

Yes, you've previously pointed out your ability to use the GitHub search feature.  Unfortunately, many of those results are for CI other than GitHub Actions.  If one adds a constraint to the workflow folder, the number is ~ 54k.   Not a high number.

> in your own workflows

My workflows?  A quick check of five or six popular Ruby repos/gems that require build tools, all show 'used by' numbers between 1M & 2.5M.  Many of those 'used by' repos may not be running CI on Windows, but that's still a lot of repos...

eine:
@MSP-Greg 

> As mentioned above, using windows-2022 with Ruby repos that require gcc tools, the additional time needed is between 45 and 60 seconds when CI is busy.

Precisely, it is at least 10 times less than 10+ minutes.

> We're testing pre-assembling msys2, mingw64, and ucrt64 tools/files three times a day into 7z files, and extracting them on each job. They're stored in a repo release. That seems quickest, and removes the traffic from the MSYS2 servers, which would normally happen for each job.

Should you succeed, please propose enhancements to [msys2/setup-msys2](https://github.com/msys2/setup-msys2) or [msys2.org/docs/ci](https://www.msys2.org/docs/ci/) (through [msys2/msys2.github.io](https://github.com/msys2/msys2.github.io)).

As you may know, there is a built-in feature in setup-msys2 for caching the whole installation: https://github.com/msys2/setup-msys2/blob/d07afaf11ae7963aa30c77d8e1f45a1be5f19954/main.js#L15-L16

It is currently disabled due to reliability issues. As explained in msys2/setup-msys2#61, GitHub's cache storage might be buggy.
However, that issue was created more than a year ago. Maybe GitHub/Microsoft improved the infrastructure since then. Feel free to fork and test.

> If one adds a constraint to the workflow folder, the number is ~ 54k. Not a high number.

Fair enough. Still, 54k is an order of magnitude below 300k and four orders of magnitude above "five or six".
I'm not the one bringing the quantitative arguments with regard to the user bases; but just reflecting non-solid arguments.

> My workflows? A quick check of five or six popular Ruby repos/gems that require build tools, all show 'used by' numbers between 1M & 2.5M. Many of those 'used by' repos may not be running CI on Windows, but that's still a lot of repos...

As explained several times already, if you want to bring data into the discussion, please do it properly.
Apart from GitHub's search feature, you can use the API for getting lots of data.
That would be very benefitial to stop circling around.
There is a Ruby toolkit: [octokit/octokit.rb](https://github.com/octokit/octokit.rb).
Unfortunately, vague and scope-limited claims are not very useful.
Particularly, as I guess you know, the "Used by" feature is unrelated to the usage of CI. You might want to constraint it to the ones using CI *and* Windows *and* the default installation of MSYS2; since that is the scope of this PR.

PR #4375, opened 2021-10-29
[Windows] Rework NodeJS install

nikolai-frolov:
## Description
This PR changes way of Node.js install to use default version from toolset instead of LTS. 

## Related issue: 
https://github.com/actions/virtual-environments-internal/issues/2890

## Test builds:
- [Windows2022](https://github.visualstudio.com/virtual-environments/_build/results?buildId=118470&view=results)

## Check list

### DIFF
--- a/images/win/scripts/Installers/Install-NodeLts.ps1
+++ b/images/win/scripts/Installers/Install-NodeLts.ps1
@@ -10,7 +10,8 @@ $CachePath = 'C:\npm\cache'
 New-Item -Path $PrefixPath -Force -ItemType Directory
 New-Item -Path $CachePath -Force -ItemType Directory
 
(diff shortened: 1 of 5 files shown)


MaksimZhukov (dismissed):

miketimofeev (approved):

gep13:
> @miketimofeev said...
could you please clarify why $ODataQuery = '$filter=(Title eq ''ghc'')' works but $ODataQuery = '$filter=(Title eq ''nodejs'')' doesn't? I thought it is universal for all the packages

Apologies, just seeing this now.

This is due to the fact that you are using `Title` and not `Id`.  The title of the nodejs package is `Node JS` which is why it isn't being returned.

A better, more reliable query would be:

```
$filter=(tolower(Id) eq 'nodejs')
```

PR #4406, opened 2021-11-02
[Windows] Rework NodeJS install to use latest version which corresponds to the default

nikolai-frolov:
## Description
This PR changes way of Node.js install to search the latest released version which corresponds to the default version from the toolset. 

## Related issue: 
https://github.com/actions/virtual-environments-internal/issues/2890

## Check list


images/win/scripts/Installers/Install-NodeLts.ps1
```diff
@@ -11,7 +11,16 @@ New-Item -Path $PrefixPath -Force -ItemType Directory
 New-Item -Path $CachePath -Force -ItemType Directory
 
 $defaultVersion = (Get-ToolsetContent).node.default
-Choco-Install -PackageName nodejs -ArgumentList "--version=$defaultVersion"
+$ODataQuery = "`$filter=(Id eq `'nodejs`') and (IsPrerelease eq false) and (Version ge `'$defaultVersion`')"
```
al-cheb, line 14: use the same name conversion, all variables should be in lowercase.

images/win/scripts/Installers/Install-NodeLts.ps1
```diff
@@ -11,7 +11,16 @@ New-Item -Path $PrefixPath -Force -ItemType Directory
...
+$Url = "https://community.chocolatey.org/api/v2/Packages()?$ODataQuery"
+$ChocoVersionsOutput = (Invoke-RestMethod -Uri $Url).properties.Version
+$highestVersion = [Version[]]($ChocoVersionsOutput | Where-Object {$_ -Like "$defaultVersion.*"}) `
+    | Sort-Object -Descending `
```
al-cheb, line 18: Let's remove -Descending and use Last

images/win/scripts/Installers/Install-NodeLts.ps1
```diff
@@ -11,7 +11,16 @@ New-Item -Path $PrefixPath -Force -ItemType Directory
 New-Item -Path $CachePath -Force -ItemType Directory
 
 $defaultVersion = (Get-ToolsetContent).node.default
-Choco-Install -PackageName nodejs -ArgumentList "--version=$defaultVersion"
+$ODataQuery = "`$filter=(Id eq `'nodejs`') and (IsPrerelease eq false) and (Version ge `'$defaultVersion`')"
+$Url = "https://community.chocolatey.org/api/v2/Packages()?$ODataQuery"
+$ChocoVersionsOutput = (Invoke-RestMethod -Uri $Url).properties.Version
+$highestVersion = [Version[]]($ChocoVersionsOutput | Where-Object {$_ -Like "$defaultVersion.*"}) `
```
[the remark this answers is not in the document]
nikolai-frolov: It's required to use $ChocoVersionsOutput as string first. It should be converted to the Version only after filtering corresponded to default versions.

images/win/scripts/Installers/Install-NodeLts.ps1
```diff
@@ -11,7 +11,16 @@ New-Item -Path $PrefixPath -Force -ItemType Directory
...
+    | Sort-Object -Descending `
+    | Select-Object -First 1
+# For case when $defaultVersion is the full version number
+if ( !$highestVersion ) { $highestVersion = $defaultVersion }
```
al-cheb, line 21: Why do we need this condition if we have `Version ge '$defaultVersion'` filter?

images/win/scripts/Installers/Install-NodeLts.ps1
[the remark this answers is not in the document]
al-cheb: Why can't we use Major property?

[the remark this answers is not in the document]
nikolai-frolov: Current implementation allows use not only major version as wildcard. We can use for example "14", "14.18", "14.18.1" as default and script returns correct latest version corresponded for them.
Conversion to Versions is required only to sort versions correctly. We can use -Like operator for Version as you suggest, but in this case it converts Versions back to Strings, but it's an additional time.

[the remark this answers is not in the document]
al-cheb: In that case we can move cast in Sort-Object and replace $highestVersion.ToString() -> $highestVersion:
`$chocoVersions | Where-Object { $_ -like "$defaultVersion.*" } | Sort-Object { [version]$_ } | Select-Object -Last 1`

images/win/scripts/Installers/Install-NodeLts.ps1
al-cheb: You don't need to escape single quotes.

vsafonkin: Also, maybe use more clear name, for example `queryFilter`

images/win/scripts/Installers/Install-NodeLts.ps1
nikolai-frolov: Correct, but in case $defaultVersion is the full version like "14.18.1" it will be filtered out by "| Where-Object { $_ -Like "$defaultVersion.*" }" so $highestVersion will be null here. Without "." it works incorrectly for versions like "14.1"

al-cheb: We can add this check to `Where-Object { $_ -Like "$defaultVersion.*" } `

images/win/scripts/Installers/Install-NodeLts.ps1
```diff
@@ -11,7 +11,15 @@ New-Item -Path $PrefixPath -Force -ItemType Directory
 New-Item -Path $CachePath -Force -ItemType Directory
 
 $defaultVersion = (Get-ToolsetContent).node.default
-Choco-Install -PackageName nodejs -ArgumentList "--version=$defaultVersion"
+$queryFilter = "`$filter=(Id eq 'nodejs') and (IsPrerelease eq false) and (Version ge '$defaultVersion')"
+$url = "https://community.chocolatey.org/api/v2/Packages()?$queryFilter"
+$chocoVersionsOutput = (Invoke-RestMethod -Uri $url).properties.Version
+$highestVersion = $chocoVersionsOutput `
```
al-cheb, line 17: Let's remove ` and swap to |

images/win/scripts/Installers/Install-NodeLts.ps1
```diff
@@ -15,11 +15,9 @@ $queryFilter = "`$filter=(Id eq 'nodejs') and (IsPrerelease eq false) and (Versi
 $url = "https://community.chocolatey.org/api/v2/Packages()?$queryFilter"
 $chocoVersionsOutput = (Invoke-RestMethod -Uri $url).properties.Version
 $highestVersion = $chocoVersionsOutput `
-    | Where-Object { $_ -Like "$defaultVersion.*" } `
+    | Where-Object { $_ -Like "$defaultVersion.*" -or $_ -eq $defaultVersion } `
```
miketimofeev:
```suggestion
    | Where-Object { $_ -match "^$defaultVersion" } `
```

nikolai-frolov: This works incorrectly for defaultVersion like "14.1" and returns "14.18.1" instead of "14.1.0"

miketimofeev: Do we ever need to install the `Major.Minor` version? In toolset we only have the Major one

miketimofeev: And even there such a case you can use match
`Where-Object { $_ -match "^$defaultVersion\." } `
image

nikolai-frolov: `Where-Object { $_ -match "^$defaultVersion\." }` works incorrectly for the full version number like `14.18.1`. I would like to have flexible function which can return correct version for any default version, not only for Major.

miketimofeev: why do we need to set the full version in the toolset? The toolset is not intended for that

nikolai-frolov: Can you guarantee that at some moment we don't have to stick to a specific version due to compatibility issues for example? I guess that is a good option to be safe in future.

images/win/scripts/Installers/Install-NodeLts.ps1
```diff
@@ -11,7 +11,15 @@ New-Item -Path $PrefixPath -Force -ItemType Directory
 New-Item -Path $CachePath -Force -ItemType Directory
 
 $defaultVersion = (Get-ToolsetContent).node.default
-Choco-Install -PackageName nodejs -ArgumentList "--version=$defaultVersion"
+$queryFilter = "`$filter=(Id eq 'nodejs') and (IsPrerelease eq false) and (Version ge '$defaultVersion')"
```
gep13, line 14: Would it make sense to extend this to include an upper limit?

```
`$filter=(Id eq 'nodejs') and (IsPrerelease eq false) and (Version ge '$defaultVersion') and (Version lt '15.0')
```

nikolai-frolov: This solution will differ from our regular approach which we use for versions in toolset file and probably can confuse someone.

miketimofeev (approved):

MaksimZhukov (approved):

miketimofeev (approved):

PR #4429, opened 2021-11-07
Update swift.sh

### LINKED ISSUE #4427: Swift installer script no longer works
### Description

```
    azure-arm: Downloading 'https://swift.org/builds/swift--release/ubuntu1804/swift--RELEASE/swift--RELEASE-ubuntu18.04.tar.gz' to '/tmp/swift--RELEASE-ubuntu18.04.tar.gz'...
    azure-arm: Download completed
==> azure-arm:
==> azure-arm: gzip: stdin: not in gzip format
==> azure-arm: tar: Child returned status 1
==> azure-arm: tar: Error is not recoverable: exiting now
==> azure-arm: mv: cannot stat 'swift--RELEASE-ubuntu18.04': No such file or directory
```

https://github.com/actions/virtual-environments/blob/main/images/linux/scripts/installers/swift.sh#L12 no longer returns swift version.

### Virtual environments affected

- [X] Ubuntu 18.04
- [X] Ubuntu 20.04
- [ ] macOS 10.15
- [ ] macOS 11
- [ ] Windows Server 2016
- [ ] Windows Server 2019
- [ ] Windows Server 2022

https://github.visualstudio.com/virtual-environments/_build/results?buildId=118803&view=logs&j=011e1ec8-6569-5e69-4f06-baf193d1351e&t=5431112d-2b61-5a5f-7042-ef698f761043&l=4851

### Is it regression?

_No response_

### Expected behavior

_No response_

### Actual behavior

_No response_

### Repro steps

N/A

JayDoubleu:
# Description
Fixes #4427 

## Check list

### DIFF
--- a/images/linux/scripts/installers/swift.sh
+++ b/images/linux/scripts/installers/swift.sh
@@ -9,7 +9,7 @@ source $HELPER_SCRIPTS/install.sh
 
 # Install
 image_label="$(lsb_release -rs)"
-swift_version=$(curl -s "https://api.github.com/repos/apple/swift/releases/latest" | grep -Po '"tag_name": "swift-\K[0-9.]+')
+swift_version=$(curl -s "https://api.github.com/repos/apple/swift/releases/latest" | jq -r '.tag_name | match("[0-9.]+").string')
 
 swift_tar_name="swift-$swift_version-RELEASE-ubuntu$image_label.tar.gz"
 swift_tar_url="https://swift.org/builds/swift-$swift_version-release/ubuntu${image_label//./}/swift-$swift_version-RELEASE/$swift_tar_name"
--- a/images/linux/scripts/installers/swift.sh
+++ b/images/linux/scripts/installers/swift.sh
@@ -9,7 +9,7 @@ source $HELPER_SCRIPTS/install.sh
 
 # Install
 image_label="$(lsb_release -rs)"
-swift_version=$(curl -s -L -N https://swift.org/download|awk -F"[ <]" '/id="swift-/ {print $4; exit}')
+swift_version=$(curl -s "https://api.github.com/repos/apple/swift/releases/latest" | grep -Po '"tag_name": "swift-\K[0-9.]+')
 
 swift_tar_name="swift-$swift_version-RELEASE-ubuntu$image_label.tar.gz"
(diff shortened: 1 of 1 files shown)


images/linux/scripts/installers/swift.sh
```diff
@@ -9,7 +9,7 @@ source $HELPER_SCRIPTS/install.sh
 
 # Install
 image_label="$(lsb_release -rs)"
-swift_version=$(curl -s -L -N https://swift.org/download|awk -F"[ <]" '/id="swift-/ {print $4; exit}')
+swift_version=$(curl -s "https://api.github.com/repos/apple/swift/releases/latest" | grep -Po '"tag_name": "swift-\K[0-9.]+')
```
mikhailkoliada, line 12: Is there a point of using github rather than simply pass the `--compressed` use flag to curl?

JayDoubleu: Yes, the point being that Github release syntax is not likely to change while current way of getting the release can break every time swift.org makes a change. This being one of examples of it.

AKA if I would have to choose one of these options on my projects I would go with github release API and forget about it. While if to leave it in current form there is always risk of it breaking.

Which even way you guys go about it i'm fine with it as long as it's fixed on each release tag :)

JayDoubleu: This wouldn't be such a big deal if Microsoft would publish your release images to the public so they can be used in private agent pools in ADO.
Right now we have to hook to your releases and build our own private agents and if it breaks here, it breaks in private projects which then translates to someone having to go through failed build and implement temporary fixes.
I hope you understand.

JayDoubleu: Actually I started discussion about it here : https://github.com/actions/virtual-environments/discussions/4430
I'm not sure if this was discussed already.

mikhailkoliada: @JayDoubleu I have nothing against github api in particular, just curious :)

grep is not the best tool to parse json in general, I'd use jq with something like

```
curl -s "https://api.github.com/repos/apple/swift/releases/latest" | jq -r '.tag_name | match("[0-9.]+").string'
```

JayDoubleu: Actually I thought about using JQ but wasn't sure if it's included in the image by default or installed before swift script run.

miketimofeev: Jq is installed before, so it's worth using it here

miketimofeev (approved):

PR #4447, opened 2021-11-09
[Ubuntu] Pin MongoDB to the specific version defined in the toolset

shilovmaksim:
# Description
We need to move mongo along with some other tools to the toolset on all OSes to make versions control more granular.

## Check list

### DIFF
--- a/images/linux/scripts/installers/mongodb.sh
+++ b/images/linux/scripts/installers/mongodb.sh
@@ -6,17 +6,21 @@
 
 # Source the helpers
 source $HELPER_SCRIPTS/os.sh
+source $HELPER_SCRIPTS/install.sh
 
 REPO_URL="https://repo.mongodb.org/apt/ubuntu"
+osLabel=$(getOSVersionLabel)
+toolsetVersion=$(get_toolset_value '.mongodb.default')
+latestVersion=$(curl $REPO_URL/dists/$osLabel/mongodb-org/ | grep $toolsetVersion\.. | awk -F'>|<' '{print $3}' | tail -1)
 
 #  Install Mongo DB
-wget -qO - https://www.mongodb.org/static/pgp/server-5.0.asc | sudo apt-key add -
-version=$(getOSVersionLabel)
-echo "deb [ arch=amd64,arm64 ] $REPO_URL $version/mongodb-org/5.0 multiverse" | sudo tee /etc/apt/sources.list.d/mongodb-org-5.0.list
+wget -qO - https://www.mongodb.org/static/pgp/server-$latestVersion.asc | sudo apt-key add -
+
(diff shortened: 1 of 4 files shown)


images/linux/toolsets/toolset-1804.json
```diff
@@ -329,5 +329,8 @@
             "name": "yarn",
             "command": "yarn"
         }
-    ]
+    ],
+    "mongodb": {
+        "default": "5"
```
miketimofeev, line 334: maybe version instead of the default? As we don't have multiple versions anyway 😃

shilovmaksim: Sure we can. I just used existing naming =)

miketimofeev: I should have left the same comment for nodejs, but forgot

images/linux/scripts/installers/mongodb.sh
```diff
@@ -6,17 +6,21 @@
 
 # Source the helpers
 source $HELPER_SCRIPTS/os.sh
+source $HELPER_SCRIPTS/install.sh
 
 REPO_URL="https://repo.mongodb.org/apt/ubuntu"
+osLabel=$(getOSVersionLabel)
+toolsetVersion=$(get_toolset_value '.mongodb.default')
+latestVersion=$(curl $REPO_URL/dists/$osLabel/mongodb-org/ | grep $toolsetVersion\.. | awk -F'>|<' '{print $3}' | tail -1)
```
miketimofeev, line 14: maybe slightly change to this to get rid of `\..`?
```suggestion
latestVersion=$(curl $REPO_URL/dists/$osLabel/mongodb-org/ | awk -F'>|<' '{print $3}' | grep "^$toolsetVersion" | tail -1)
```

miketimofeev (dismissed):

images/linux/toolsets/toolset-1804.json
shilovmaksim: Updated

images/linux/scripts/installers/mongodb.sh
shilovmaksim: Updated

images/linux/scripts/tests/Databases.Tests.ps1
```diff
@@ -3,7 +3,8 @@ Describe "MongoDB" {
         @{ ToolName = "mongo" }
         @{ ToolName = "mongod" }
     ) {
-        "$ToolName --version" | Should -ReturnZeroExitCode
```
miketimofeev, line 6: I think it makes sense to leave this test as well

miketimofeev (approved):

images/linux/scripts/tests/Databases.Tests.ps1
shilovmaksim: As discussed keeping this is redundant

al-cheb (approved):

PR #4527, opened 2021-11-16
CI: Run linter only on changes of JSON and MD files

EwoutH:
This PR updates the CI linter workflow to only run if a JSON of MD file is changed. This saves CI time on PRs that only have changes to other file types.

Also updates [github/super-linter](https://github.com/github/super-linter) to version 4.

### DIFF
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -21,7 +21,7 @@ jobs:
           fetch-depth: 0
 
       - name: Lint Code Base
-        uses: github/super-linter@v4
+        uses: github/super-linter/slim@v4
         env:
           VALIDATE_ALL_CODEBASE: false
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


EwoutH:
Noticed some additional things in the linter job, feel free to squash when merging.

.github/workflows/linter.yml
```diff
@@ -5,6 +5,9 @@ name: Linter
 on:
   pull_request:
     branches: [ main ]
+    paths:
+      - '**.json'
+      - '**.md'
```
catthehacker, line 10: This prevents running linter when only shell file have changes

EwoutH: Does that matter if only the JSON and Markdown files are linted?

If so, is there a certain file extension that needs to be included?

catthehacker: https://github.com/actions/virtual-environments/blob/499a9c967ffc320d970b964f57230065d08aa1f5/.github/workflows/linter.yml#L32-L34

EwoutH: It now also runs on changes to `.sh` files, which the shebang linter checks. Thanks!

miketimofeev (approved):

miketimofeev:
@EwoutH looks good, thanks!
@catthehacker thanks for the review!

PR #4551, opened 2021-11-21
[Ubuntu] install libpcre2-8-0 for PHP compat

mikhailkoliada:
# Description

Looks like php version in the PPA repo got updated and now we are lacking proper libcre2 version which blocks our builds, I guess it gonna be fixed eventually but lets workaround it for now.

#### Related issue:
https://github.com/actions/virtual-environments/pull/4551

## Check list

### DIFF
--- a/images/linux/scripts/installers/php.sh
+++ b/images/linux/scripts/installers/php.sh
@@ -13,8 +13,10 @@ source $HELPER_SCRIPTS/install.sh
 apt-add-repository ppa:ondrej/php -y
 apt-get update
 
-# install specific pcre-lib otherwise latest ppa-hosted php fails
-apt-get install -y libpcre2-8-0
+# install specific version of libpcre2 otherwise latest ppa-hosted php fails
+if isUbuntu20 ; then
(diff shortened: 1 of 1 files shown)


miketimofeev (approved):

Danack:
I'm not sure that updating individual packages by hand is an appropriate fix.

[Ondřej Surý wrote](https://github.com/oerdnj/deb.sury.org/issues/1682#issuecomment-974802224) "This is a bare minimum. You should be doing apt-get dist-upgrade otherwise you are running random mix of updated and outdated packages."

miketimofeev:
> I'm not sure that updating individual packages by hand is an appropriate fix.
> 
> [Ondřej Surý wrote](https://github.com/oerdnj/deb.sury.org/issues/1682#issuecomment-974802224) "This is a bare minimum. You should be doing apt-get dist-upgrade otherwise you are running random mix of updated and outdated packages."

On the one hand — yes, on the other hand, we will have an unpredicted set of updated packages that came from the PPA.

PR #4575, opened 2021-11-23
[Ubuntu] get rid of post-gen rust script

mikhailkoliada:
# Description

This pull request moves rust installation into the `/etc/skel` directory so it could be picked up later on user creation and owned as `runner:docker` (and therefore we could get rid of the post-generation script).  After everything is set up it is possible to unset both the `CARGO_HOME` and  `RUSTUP_HOME` env vars as by default rust (ab)uses `$HOME` and all the users runtime installations will be placed there.

CC: @maxim-lobanov 

## Check list


images/linux/scripts/installers/rust.sh
```diff
@@ -19,19 +19,13 @@ source $CARGO_HOME/env
...
-prependEtcEnvironmentPath "${CARGO_HOME}/bin"
+prependEtcEnvironmentPath "${HOME}/.cargo/bin"
 
-# Rust Symlinks are added to a default profile /etc/skel
-pushd /etc/skel
-ln -sf $RUSTUP_HOME .rustup
-ln -sf $CARGO_HOME .cargo
-popd
+# unset back to default
+unset RUSTUP_HOME CARGO_HOME
```
maxim-lobanov, line 29: No need to unset it. These variables will unset automatically after script execution.

images/linux/scripts/installers/rust.sh
```diff
@@ -19,19 +19,10 @@ source $CARGO_HOME/env
...
 
 # Update /etc/environemnt
-prependEtcEnvironmentPath "${CARGO_HOME}/bin"
-
-# Rust Symlinks are added to a default profile /etc/skel
-pushd /etc/skel
-ln -sf $RUSTUP_HOME .rustup
-ln -sf $CARGO_HOME .cargo
-popd
+prependEtcEnvironmentPath '$HOME/.cargo/bin'
```
al-cheb, line 26: I don't see any benefits in this approach, because you can't avoid invoking post-deployments scripts due to unknown $HOME variable. In fact, .rustup&.cargo are quite huge folders to multiple them at least two or three times.

miketimofeev: @maxim-lobanov don't mind invoking the script replacing $HOME afaik 😃

maxim-lobanov: Yea, replacing $HOME in `/etc/environment` is the part that we can't get rid of on self-hosted / premium runners. And it is not something that we can rework somehow.
So we replicate this script for Premium Runners. It is not big deal and takes little time so we can live with it.
We definitely can't maintain any big scripts on this stage (like having separate script for every tool) but replicating $HOME is some kind of core here.

P.S. We don't run https://github.com/actions/virtual-environments/blob/main/images/linux/post-generation/environment-variables.sh exactly but do pretty similar logic and this PR should work for us.

images/linux/scripts/SoftwareReport/SoftwareReport.Rust.psm1
```diff
@@ -1,6 +1,5 @@
 function Initialize-RustEnvironment {
-    ln -sf "/usr/share/rust/.rustup" $HOME/.rustup
-    ln -sf "/usr/share/rust/.cargo" $HOME/.cargo
+    $env:PATH = "/etc/skel/.cargo/bin:/etc/skel/.rustup/bin:$($env:PATH)"
```
maxim-lobanov, line 2:
```suggestion
    $env:PATH = "/etc/skel/.cargo/bin:/etc/skel/.rustup/bin:$($env:PATH)"
    $env:RUSTUP_HOME = "/etc/skel/.rustup"
    $env:CARGO_HOME = "/etc/skel/.cargo"
```

images/linux/scripts/tests/Tools.Tests.ps1
```diff
@@ -15,6 +15,7 @@ Describe "Bicep" {
 }
 
 Describe "Rust" {
+    $env:PATH = "/etc/skel/.cargo/bin:/etc/skel/.rustup/bin:$($env:PATH)"
```
maxim-lobanov, line 18:
```suggestion
    BeforeAll {
        $env:PATH = "/etc/skel/.cargo/bin:/etc/skel/.rustup/bin:$($env:PATH)"
    }
    
   
```

images/linux/scripts/tests/Tools.Tests.ps1
```diff
@@ -15,6 +15,11 @@ Describe "Bicep" {
 }
 
 Describe "Rust" {
+    BeforeAll {
+        $env:PATH = "/etc/skel/.cargo/bin:/etc/skel/.rustup/bin:$($env:PATH)"
```
maxim-lobanov, line 19:
```suggestion
        $env:PATH = "/etc/skel/.cargo/bin:/etc/skel/.rustup/bin:$($env:PATH)"
        $env:RUSTUP_HOME = "/etc/skel/.rustup"
        $env:CARGO_HOME = "/etc/skel/.cargo"
```

al-cheb (approved):

maxim-lobanov (approved):

MaksimZhukov (approved):

jgraettinger:
Hi, we're seeing a build break on the ubuntu20.04 runner that _may_ be linked to this PR (still trying to figure that out):

A run 8 days ago, which passed on the previous release of the runner:
https://github.com/estuary/flow/actions/runs/1495572974/attempts/1
A run today of this same commit, which fails consistently:
https://github.com/estuary/flow/actions/runs/1495572974/attempts/2

The specific break is within the `prost-build` [crate attempting to find and invoke `protoc`](https://github.com/estuary/flow/runs/4386823343?check_suite_focus=true). For background, the crate ships with a bundled copy of the protobuf compiler which is used by default (including in this case). It's (handwavy) not able to find this bundled copy of `protoc` for some reason, which may be related to these changes ?

jgraettinger:
Okay, a more specific failure mode theory: the failed builds happened to be re-using cached cargo states:
https://github.com/estuary/flow/blob/e8da2ce/.github/workflows/main.yml#L52-L62

Explicitly busting the cache appears to have it building again. So presumably it's an interaction of this changed and a cargo cache.

PR #4598, opened 2021-11-26
[Windows] Invoke Move-Item and Rename-Item with retry after Extract-7Zip

al-cheb:
# Description
Sometimes customers have an issue with `Access is denied`  during image generation after invoking Extract-7Zip, as a workaround is using a function with retries.

#### Related issue:
https://github.com/actions/virtual-environments/issues/4277

## Check list


mikhailkoliada (approved):

shilovmaksim:
Is "Invoke-SBWithRetry" needed for files/folders operations? Will it resume/overwrite files/folders if it encounter an error. Maybe we should use approach with "robocopy" used?

al-cheb:
> Is "Invoke-SBWithRetry" needed for files/folders operations? Will it resume/overwrite files/folders if it encounter an error? Maybe we should use approach with "robocopy" used?

This function is general to run any cmdlets with retries. It will depend on the logic `command` to rewrite/resume/etc. I don't see any benefits to use robocopy in this tricky case.

dmitry-shibanov (approved):

MaksimZhukov (approved):

PR #4628, opened 2021-11-29
[macOS] overwrite php symlink

mikhailkoliada:
# Description

Composer is depending on the latest version of php, which is now 8.1. In the toolset we keep the latest brew slot which is 8.0, as a result once composer is installed the `/usr/local/bin/php` symlink points to the latest installed php (even if php-8.0 is installed after 8.1). We need to override the symlink to make 8.0 available  by default (there is no the `php@8.1` slot so we can't simply update our toolset and be fine).

## Check list

### DIFF
--- a/images/macos/provision/core/php.sh
+++ b/images/macos/provision/core/php.sh
@@ -8,4 +8,6 @@ brew_smart_install "php@${phpVersionToolset}"
 echo Installing composer
 brew_smart_install "composer"
 
+brew link --overwrite php@${phpVersionToolset}
+
 invoke_tests "PHP"


al-cheb:
@shivammathur, Could you please review?

dibir-magomedsaygitov (approved):

shivammathur:
@al-cheb 
I would recommend upgrading the default version to PHP 8.1 instead of relinking 8.0. If you want to keep PHP 8.0 installed on macOS that is fine, but the default should now be 8.1.

mikhailkoliada:
ugh, my bad, I now see brew reflects @8.1, lets rather upgrade the toolset.

PR #4674, opened 2021-12-07
[Ubuntu] Downgrade mysql to 8.0.26

mikhailkoliada:
# Description

As has been discussed with @MaksimZhukov and @miketimofeev lets downgrade mysql on ubuntu to 8.0.26 as 8.0.27 is buggy. As there is no an older version in Oracle's PPA repo there was a need to install all the dependencies step by step utilizing the `dpkg -i` command (the order in the `mysql_debs` array matters). No toolset and other flexible solutions were considered yet as the future of this solution is uncertain and will likely be changed.

## Check list

### DIFF
--- a/images/linux/scripts/SoftwareReport/SoftwareReport.Databases.psm1
+++ b/images/linux/scripts/SoftwareReport/SoftwareReport.Databases.psm1
@@ -16,7 +16,7 @@ function Get-SqliteVersion {
 }
 
 function Get-MySQLVersion {
-    $mySQLVersion = mysqld --version | Take-OutputPart -Part 2 | Take-OutputPart -Part 0 -Delimiter "-"
+    $mySQLVersion = mysqld --version | Take-OutputPart -Part 2
     return "MySQL $mySQLVersion"
 }
 
--- a/images/linux/scripts/installers/mysql.sh
+++ b/images/linux/scripts/installers/mysql.sh
@@ -4,22 +4,41 @@
 ##  Desc:  Installs MySQL Client
 ################################################################################
 
-export ACCEPT_EULA=Y
+source /etc/os-release
 
 # Mysql setting up root password
 MYSQL_ROOT_PASSWORD=root
 echo "mysql-server mysql-server/root_password password $MYSQL_ROOT_PASSWORD" | debconf-set-selections
 echo "mysql-server mysql-server/root_password_again password $MYSQL_ROOT_PASSWORD" | debconf-set-selections
 
-# Install MySQL Client
-apt-get install mysql-client -y
+mkdir -p /tmp/mysql
+pushd /tmp
 
-# Install MySQL Server
-apt-get install -y mysql-server
+# Mandatory dependency
+apt-get install -y libmecab2 libaio1
 
-#Install MySQL Dev tools
-apt install libmysqlclient-dev -y
+wget -q https://downloads.mysql.com/archives/get/p/23/file/mysql-server_8.0.26-1ubuntu${VERSION_ID}_amd64.deb-bundle.tar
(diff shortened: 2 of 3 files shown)


miketimofeev (approved):

mattlord:
Hi @miketimofeev!

I apologize, cross posting my comment from the issue that led me here: https://github.com/actions/virtual-environments/issues/4732#issuecomment-992933533

Please let me know what you would recommend.

Thanks!

mattlord:
Hi @miketimofeev and @mikhailkoliada!

Just FYI, this change may cause more problems than it solved. It broke any of our actions that use `ubuntu-latest` and simply try to install the `mysql-{client,server}` packages via `apt-get`. Installing either one now exits with:
```
The following packages have unmet dependencies:
 mysql-client : Depends: mysql-client-8.0 but it is not going to be installed
 mysql-server : Depends: mysql-server-8.0 but it is not going to be installed
```

For an example that's failing:
  - source: https://github.com/vitessio/vitess/blob/main/.github/workflows/check_make_parser.yml#L31
  - run: https://github.com/vitessio/vitess/runs/4510817416?check_suite_focus=true

Should I open a new ticket for this? I'm curious which way forward you might recommend as we figure something out on our side.

Thank you!

wojsmol:
@mattlord Looking at this PR https://github.com/actions/virtual-environments/pull/4674/files#diff-7e1f5b360352a2ec74ab295b4fa1bfd65b9126094484bf56dbf6108b6f5478d4R28-R37 and your workflow definition changing 
```yml
sudo apt-get install -y mysql-server mysql-client make unzip g++ etcd curl git wget
```
to
```yml
sudo apt-get install -y make unzip g++ etcd curl git wget
```
should temporally fix the issue as `mysql-server` and `mysql-client` are pre-installed. AFAIK.

mattlord:
@wojsmol thank you for responding! We started with that idea too, but that would be a temporary fix with potential downsides in the future so instead we ended up opting to manually/explicitly install and configure the Oracle MySQL APT repo and then explicitly install MySQL from there. That should never hurt and it shields/protects us from any upstream changes where that repo may not be available or configured as expected so seemed like a good idea overall. For example:
```
    - name: Get dependencies
      run: |
        # Setup MySQL 8.0
        wget -c https://dev.mysql.com/get/mysql-apt-config_0.8.20-1_all.deb
        echo mysql-apt-config mysql-apt-config/select-server select mysql-8.0 | sudo debconf-set-selections
        sudo DEBIAN_FRONTEND="noninteractive" dpkg -i mysql-apt-config*
        sudo apt-get update
        # Install everything else we need, and configure
        sudo apt-get install -y mysql-server mysql-client make unzip g++ etcd curl git wget eatmydata
        sudo ln -s /etc/apparmor.d/usr.sbin.mysqld /etc/apparmor.d/disable/
        sudo apparmor_parser -R /etc/apparmor.d/usr.sbin.mysqld
```

In case it's helpful to anyone else or anyone has a better idea. 🙂

mikhailkoliada:
@mattlord Hello!

Problems in regards with apt were expected as apt is not aware of the manual installation method, we are likely to get back to the "normal" installation process as soon as Oracle/Canonical have the fix released, otherwise 8.0.27 is unusable due to the connectivity issues [bug](https://bugs.mysql.com/bug.php?id=105288).

Sorry for inconvenience caused!

mattlord:
Hi @miketimofeev!

Ah, OK. After having looked at https://bugs.mysql.com/bug.php?id=105288 I can completely understand why you did this. 😬 

We ended up doing something we probably should have been doing all along, so all good on our side. 

Thank you for fixing this for all of your MySQL users then and for the helpful response here!

Best Regards

PR #4715, opened 2021-12-10
[Windows] Add installer projects to VS on Windows-2022

V-Zabayrachny:
# Description
This PR adds  extension "Installer projects" to Visual Studio 2022 on Windows-2022

## Check list

### DIFF
--- a/images/win/scripts/ImageHelpers/InstallHelpers.ps1
+++ b/images/win/scripts/ImageHelpers/InstallHelpers.ps1
@@ -252,7 +252,6 @@ function Install-VsixExtension
         [string] $Url,
         [Parameter(Mandatory = $true)]
         [string] $Name,
-        [string] $InstallPath = ${env:ProgramFiles(x86)},
         [string] $FilePath,
         [Parameter(Mandatory = $true)]
         [string] $VSversion,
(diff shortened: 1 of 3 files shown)


mikhailkoliada (dismissed):

images/win/scripts/ImageHelpers/InstallHelpers.ps1
```diff
@@ -270,10 +271,15 @@ function Install-VsixExtension
     $vsEdition = (Get-ToolsetContent).visualStudio.edition
     try
     {
+        if (Test-IsWin22)
```
al-cheb, line 274: Why do we need a condition if we have InstallPath param?

V-Zabayrachny: In Windows 22 , the location `VSIXInstaller.exe` differs from previous versions of windows, in this regard, a condition was added to override the `InstallPath` variable.

al-cheb: That's why param InstallPath is useless if we always rewrite it.

V-Zabayrachny: Do you think it should just be moved to a variable?

al-cheb: Yep, it should be a variable.

V-Zabayrachny: I fixed it. Thanks.

dmitry-shibanov (approved):

miketimofeev (approved):

al-cheb (approved):

PR #4766, opened 2021-12-17
[macOS] Update Openssl up to 3* version

dibir-magomedsaygitov:
# Description
Currently we install `OpenSSL 1.1`, but in scope of `powershell.sh` script `OpenSSL`  is getting updated automatically up to 3* version. Let's update OpenSSL version up to 3* on our macOS images to avoid confusion.

## Check list

### DIFF
--- a/images/macos/provision/core/openssl.sh
+++ b/images/macos/provision/core/openssl.sh
@@ -1,13 +1,13 @@
 #!/bin/bash -e -o pipefail
 source ~/utils/utils.sh
 
-echo "Install openssl@1.1"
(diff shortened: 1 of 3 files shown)


mikhailkoliada (changes requested):
We must not update openssl version arbitrary, openssl 1.1.x and 3 are API and ABI incompatible, apart from that lots of software is built/patched to work with 1.1.x and have not yet been ported to 3 (and will not be ported in next 1-2 year I believe). Even though `openssl@3` is installed we are pointing to `openssl@1.1` on purpose.

PR #4784, opened 2021-12-21
[Ubuntu] Do not use apt-fast

mikhailkoliada:
# Description

apt-fast is kind of hack on top of the normal apt to speed up downloading by using either axel or aria2 downloading manager.
It is currently used by only `php.sh` and `java-tools.sh` installers and rationale behind its usage these days is unclear (we are using fast enough mirrors not to use an additional layer for packages management).
I have discussed it with @miketimofeev and we have come to conclusion it might be removed. 

EDIT: lets just stop using it but keep installed

## Check list

### DIFF
--- a/images/linux/scripts/installers/java-tools.sh
+++ b/images/linux/scripts/installers/java-tools.sh
@@ -53,7 +53,7 @@ echo "JAVA_HOME=/usr/lib/jvm/adoptopenjdk-${DEFAULT_JDK_VERSION}-hotspot-amd64" 
 # add extra permissions to be able execute command without sudo
 chmod -R 777 /usr/lib/jvm
 # Install Ant
-apt-fast install -y --no-install-recommends ant ant-optional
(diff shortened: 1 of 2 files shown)


images/linux/scripts/SoftwareReport/SoftwareReport.Generator.ps1
```diff
@@ -96,7 +96,6 @@ $markdown += New-MDList -Style Unordered -Lines ($projectManagementList | Sort-O
 $markdown += New-MDHeader "Tools" -Level 3
 $toolsList = @(
     (Get-AnsibleVersion),
-    (Get-AptFastVersion),
```
al-cheb, line 99: We shouldn't delete apt-fast, we should skip to use it. Please revert.

al-cheb (approved):

miketimofeev (approved):

MaksimZhukov (approved):

Issue #4770, opened 2021-12-19
Update VoiceOver Utility to allow VoiceOver to be controlled with AppleScript 


cmorten:
### Tool name

VoiceOver Utility

### Tool license

Copyright © 2005-2017 Apple Inc. / Unknown

### Add or update?

- [ ] Add
- [X] Update

### Desired version

Version 10 / All

### Brief description of tool

VoiceOver is a screen reader system application that provides auditory descriptions of elements help you easily navigate your screen with keyboard or gestures.

You can configure VoiceOver through the system application `VoiceOver Utility.app`. One such option is a checkbox `Allow VoiceOver to be controlled with AppleScript` which allows the user to use AppleScript scripts to automate VoiceOver tasks. (Reference: https://support.apple.com/en-gb/guide/voiceover/cpvougen/mac).

This feature request is to enable this option for the macos virtual environments.

![Screenshot 2021-12-19 at 16 03 54](https://user-images.githubusercontent.com/11313985/146690938-9a07cf42-9f52-46fa-b7a0-b763e3f34435.png)

### URL for tool's homepage

https://support.apple.com/en-gb/guide/voiceover/welcome/mac

### Provide a basic test case to validate the tool's functionality.

```bash
#!/bin/bash

test -f "/private/var/db/Accessibility/.VoiceOverAppleScriptEnabled" && echo "VoiceOver AppleScript Control Enabled"
```

### Virtual environments affected

- [ ] Ubuntu 18.04
- [ ] Ubuntu 20.04
- [X] macOS 10.15
- [X] macOS 11
- [ ] Windows Server 2016
- [ ] Windows Server 2019
- [ ] Windows Server 2022

### Can this tool be installed during the build?

I'm currently uncertain if you can set this option programatially...

From what I can tell, whether this is enabled or not is driven by the presence of a `/private/var/db/Accessibility/.VoiceOverAppleScriptEnabled` file (containing the single character `a` - as determined by digging into VoiceOver Utility / ScreenReaderCore.framework), but this directory has SIP so writing the file I'm assuming is a no-go (disabling SIP not an option).

It also appears trying to import a `.voprefs` file with the `SCREnableAppleScript` key set to true is simply ignored.

[...]

Issue #650, opened 2020-03-31
Disable SIP in Mac OS Catalina


alepauly:
We're planning on disabling SIP (System Integrity Protection) on Mac OS 10.15 images. The change is planned to take effect during the week of April 13th.

Please let us know of any concerns or questions you may have about this change.

**Virtual environments affected**
- [x] macOS 10.15
- [ ] Ubuntu 16.04 LTS
- [ ] Ubuntu 18.04 LTS
- [ ] Windows Server 2016 R2
- [ ] Windows Server 2019

jeroenleenarts:
Why disable SIP?

It creates a bigger distance between running build scripting on an end user system compared to the Action/Pipeline VMs. In practice I have never had the need to disable SIP, sometimes some scripting needs to be changes. But still...

alepauly:
thanks for the input @jeroenleenarts. At this point it's not set in stone that we'll disable it and we just wanted to get some feedback (rather quickly). The reason to disable it is because it allows for some workflows that don't with it enabled. 

We agree that there's a risk with making the CI environment even more different from where the artifacts will eventually be run but want to get a sense of how users feel about it as well. Thanks again.

ydnar:
@alepauly is this related to gaining programmatic access to microphone, location, camera?

alepauly:
> @alepauly is this related to gaining programmatic access to microphone, location, camera?
> 
> Could this be an argument on a per-job basis e.g. `disable_sip: true`?

I think different customers need it for different reasons including building drivers, getting more access to certain OS components, and perhaps access devices.

Unfortunately a per-job variable is not possible at this time but it would definitely be ideal.

alepauly:
At this time we're going to hold off on making this change based on feedback from customers in multiple channels. I'll reopen and pin this issue back up if we decide to go forward with it. Thanks all.

ydnar:
> I think different customers need it for different reasons including building drivers, getting more access to certain OS components, and perhaps access devices.
> 
> Unfortunately a per-job variable is not possible at this time but it would definitely be ideal.

Guessing this would be a wholly-separate macOS environment, e.g. `macos-latest-sip-disabled`?

Sorry to hear you’re abandoning this. It could be quite useful for testing iOS apps in a simulator.

matejc:
[...]

PR #4805, opened 2021-12-24
[macOS] Enable voiceover if SIP is disabled

miketimofeev:
# Description
VoiceOver is a screen reader system application that provides auditory descriptions of elements help you easily navigate your screen with keyboard or gestures.
This PR enables voiceover during the image generation by creating a special file (only works if SIP is disabled).

#### Related issue:
https://github.com/actions/virtual-environments/issues/4770

## Check list

### DIFF
--- a/images/macos/provision/configuration/configure-machine.sh
+++ b/images/macos/provision/configuration/configure-machine.sh
@@ -19,7 +19,7 @@ fi
 
 # Update VoiceOver Utility to allow VoiceOver to be controlled with AppleScript by creating a special file (SIP must be disabled)
 if csrutil status | grep -Eq  "System Integrity Protection status: (disabled|unknown)"; then
-    sudo echo -n "a" > /private/var/db/Accessibility/.VoiceOverAppleScriptEnabled
+    sudo bash -c 'echo -n "a" > /private/var/db/Accessibility/.VoiceOverAppleScriptEnabled'
 fi
 
 # https://developer.apple.com/support/expiration/
--- a/images/macos/provision/configuration/configure-machine.sh
+++ b/images/macos/provision/configuration/configure-machine.sh
@@ -19,7 +19,7 @@ fi
 
 # Update VoiceOver Utility to allow VoiceOver to be controlled with AppleScript by creating a special file (SIP must be disabled)
 if csrutil status | grep -Eq  "System Integrity Protection status: (disabled|unknown)"; then
-    sudo touch /private/var/db/Accessibility/.VoiceOverAppleScriptEnabled
+    sudo echo -n "a" > /private/var/db/Accessibility/.VoiceOverAppleScriptEnabled
 fi
 
 # https://developer.apple.com/support/expiration/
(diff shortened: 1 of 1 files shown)


images/macos/provision/configuration/configure-machine.sh
```diff
@@ -17,6 +17,11 @@ if [ -d "/Library/Application Support/VMware Tools" ]; then
     sudo "/Library/Application Support/VMware Tools/vmware-resolutionSet" 1176 885
 fi
 
+# Update VoiceOver Utility to allow VoiceOver to be controlled with AppleScript by creating a special file (SIP must be disabled)
+if csrutil status | grep -Eq  "System Integrity Protection status: (disabled|unknown)"; then
+    sudo touch /private/var/db/Accessibility/.VoiceOverAppleScriptEnabled
```
cmorten, line 22: Consider:

```suggestion
    sudo echo -n "a" > /private/var/db/Accessibility/.VoiceOverAppleScriptEnabled
```

To more closely mimic the system behaviour.

miketimofeev: Interestingly, when I enabled the setting via system menu this file didn't contain anything

cmorten: Curious! Perhaps a subtle difference between MacOS version? This is on Monterey:

Screenshot 2021-12-25 at 08 26 38

Screenshot 2021-12-25 at 08 28 18

Screenshot 2021-12-25 at 09 02 10

But not one to worry about today - Merry Christmas! ⛄

images/macos/provision/configuration/configure-machine.sh
```diff
@@ -17,6 +17,11 @@ if [ -d "/Library/Application Support/VMware Tools" ]; then
     sudo "/Library/Application Support/VMware Tools/vmware-resolutionSet" 1176 885
 fi
 
+# Update VoiceOver Utility to allow VoiceOver to be controlled with AppleScript by creating a special file (SIP must be disabled)
+if csrutil status | grep -Eq  "System Integrity Protection status: (disabled|unknown)"; then
```
cmorten, line 21: I was under the impression from #650 that SIP was disabled on agent images… so can this never evaluate to truthy and the step won’t ever run? Or is SIP actually/sometimes disabled during build?

mrcali213 (approved):

images/macos/provision/configuration/configure-machine.sh
miketimofeev: @cmorten correct, SIP is enabled and thus we enabled this setting on the image templates for now, but if someone would like to build their own VM based on our scripts this can be useful.

cmorten: Ok! Is there any guidance on building custom macOS images based on this repo, details for macOS are missing from https://github.com/actions/virtual-environments/blob/main/docs/create-image-and-azure-resources.md?

miketimofeev: Well, since macOS images can't be built in azure we haven't included them in this guide. We haven't created any special guide either as there were no requests for it.

images/macos/provision/configuration/configure-machine.sh
miketimofeev: Indeed, this is true for Big Sur as well! I must have overlooked it. Thanks for the help!

MaksimZhukov (approved):

PR #4814, opened 2021-12-28
[Ubuntu] restart snapd during images generation

mikhailkoliada:
# Description

We change snapd settings during the images generation process but there is no guarantee that snapd daemon (and its unix socket) is alive by the time the `snapd.sh` script is being run as we do not manage the snapd systemd unit explicitly (it is provided by the azure images themselves). Prior snapd restart must do the trick and bring the daemon/unix socket back before changes are made.

## Check list

### DIFF
--- a/images/linux/scripts/base/snap.sh
+++ b/images/linux/scripts/base/snap.sh
@@ -5,4 +5,10 @@
 # when they are rolling a new major update out.
 # Hold is calculated as today's date + 60 days
 
+# snapd is started automatically, but during image generation
+# a unix socket may die, restart snapd.service (and therefore snapd.socket)
+# to make sure the socket is alive.
+
+systemctl restart snapd.socket
+systemctl restart snapd
 snap set system refresh.hold="$(date --date='today+60 days' +%Y-%m-%dT%H:%M:%S%:z)"


images/linux/scripts/base/snap.sh
```diff
@@ -5,4 +5,10 @@
 # when they are rolling a new major update out.
 # Hold is calculated as today's date + 60 days
 
+# snapd is started automatically, but during image generation
+# a unix socket may die, restart snapd.service (and therefore dnapd.socket)
+# to make sure the socket is alive.
+
+systemctl restart snapd
+systemctl is-active --quiet snapd
```
al-cheb, line 13: Why should we check if a service active?

mikhailkoliada: @al-cheb I was actually to put a socket service restart there first (now done), so we indeed have no reason to check activity now.

images/linux/scripts/base/snap.sh
```diff
@@ -5,4 +5,10 @@
 # when they are rolling a new major update out.
 # Hold is calculated as today's date + 60 days
 
+# snapd is started automatically, but during image generation
+# a unix socket may die, restart snapd.service (and therefore snapd.socket)
+# to make sure the socket is alive.
+
+systemctl restart snapd.socket
```
al-cheb, line 12: Should we check if the socket exists before restart?

mikhailkoliada: @al-cheb this socket exists in anyway i.e. it just a different type of systemd unit provided by the snapd package. In short snapd.socket is not the unix socket I am referring to above it is just a service that polls snapd daemon from time to time

al-cheb (approved):

miketimofeev (approved):

PR #4852, opened 2022-01-10
[Ubuntu] remove apt-fast for good

monken:
# Description
Improvement

`apt-fast` is no longer used in any of the installers, so removing it from the system makes sense.

## Check list


miketimofeev:
@monken some users prefer to use apt-fast and we will definitely break lots of workflows with these changes. Let's leave it for now.

PR #4895, opened 2022-01-18
[Windows] Add WiX to VS on Windows-2022

V-Zabayrachny:
# Description
This PR adds extension "WiX" to Visual Studio 2022 on Windows-2022

## Check list


images/win/toolsets/toolset-2022.json
```diff
@@ -243,7 +243,8 @@
             "wasm.tools"
         ],
         "vsix": [
-            "VisualStudioClient.MicrosoftVisualStudio2022InstallerProjects"
+            "VisualStudioClient.MicrosoftVisualStudio2022InstallerProjects",
+            "WixToolset.WixToolsetVisualStudio2022Extension"
```
miketimofeev, line 247: Let's do the same for the VS2019 and VS2017 and get rid of these lines then
https://github.com/actions/virtual-environments/blob/37d8c26d314e6799cb74ec8e378d05cbbaaa3b1a/images/win/scripts/Installers/Install-Wix.ps1#L8-L25

miketimofeev: And these tests should also be removed
https://github.com/actions/virtual-environments/blob/37d8c26d314e6799cb74ec8e378d05cbbaaa3b1a/images/win/scripts/Tests/Wix.Tests.ps1#L12-L22

V-Zabayrachny: Ok, I made it.

miketimofeev (approved):

Issue #4913, opened 2022-01-19
Windows long paths support


JamieMagee:
### Description

When attempting to checkout https://github.com/Azure/azure-sdk-for-net into a subdirectory on a Windows 2019 based runner I receive many error messages like the following:

```
error: unable to create file sdk/cognitiveservices/Microsoft.Azure.Management.CognitiveServices/tests/SessionRecords/CognitiveServices.Tests.CognitiveServicesAccountTests/CognitiveServicesAccountCreateAllApisTest.json: Filename too long
```

Please enable long paths support when building Windows images

```
New-ItemProperty -Path "HKLM:\SYSTEM\CurrentControlSet\Control\FileSystem" -Name "LongPathsEnabled" -Value 1 -PropertyType DWORD -Force
```

### Virtual environments affected

- [ ] Ubuntu 18.04
- [ ] Ubuntu 20.04
- [ ] macOS 10.15
- [ ] macOS 11
- [ ] Windows Server 2016
- [X] Windows Server 2019
- [ ] Windows Server 2022

### Image version and build link

```
Environment: windows-2019
Version: 20220116.1
```

https://github.com/microsoft/component-detection/runs/4874590492?check_suite_focus=true

### Is it regression?

No

### Repro steps

1. Try and check out a repository to a path with more than 260 characters [^1]

[^1]: https://docs.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation

al-cheb:
Hey, @JamieMagee 
`LongPathsEnabled` is enabled - https://github.com/actions/virtual-environments/blob/a64db791f32ffe355d7397c1ac896d42686e9dfa/images/win/scripts/Installers/Initialize-VM.ps1#L70

JamieMagee:
Interesting! After some more searching, it might be a configuration issue with git[^1]. Is it possible to set a git config default for the image?

```
git config --system core.longpaths true
```

[^1]: https://stackoverflow.com/a/22575737

al-cheb:
> Interesting! After some more searching, it might be a configuration issue with git[1](#user-content-fn-1-b507e5cfdcfaa78f589133b5ab77bd53). Is it possible to set a git config default for the image?
> 
> ```
> git config --system core.longpaths true
> ```
> 
> ## Footnotes
> 1. https://stackoverflow.com/a/22575737 [↩](#user-content-fnref-1-b507e5cfdcfaa78f589133b5ab77bd53)

I will take a look at it. 

[...]

PR #4915, opened 2022-01-20
[Windows] Set core.longpaths option

al-cheb:
# Description
https://github.com/git-for-windows/git/blob/main/Documentation/config/core.txt
```
core.longpaths::
	Enable long path (> 260) support for builtin commands in Git for
	Windows. This is disabled by default, as long paths are not supported
	by Windows Explorer, cmd.exe and the Git for Windows tool chain
	(msys, bash, tcl, perl...). Only enable this if you know what you're
	doing and are prepared to live with a few quirks.
```

#### Related issue:
https://github.com/actions/virtual-environments/issues/4913

## Check list

### DIFF
--- a/images/win/scripts/Installers/Install-Git.ps1
+++ b/images/win/scripts/Installers/Install-Git.ps1
@@ -19,6 +19,10 @@ if (Test-IsWin16) {
     $env:Path += ";$env:ProgramFiles\Git\usr\bin\"
 }
 
+# Set core.longpaths option to true in the system-wide [path]/etc/gitconfig file
+# https://github.com/git-for-windows/git/blob/main/Documentation/config/core.txt
+git config --system core.longpaths true
+
 # Add well-known SSH host keys to ssh_known_hosts
 ssh-keyscan -t rsa github.com >> "C:\Program Files\Git\etc\ssh\ssh_known_hosts"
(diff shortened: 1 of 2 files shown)


al-cheb:
@dscho, Could you please review ?

miketimofeev (dismissed):

miketimofeev (approved):

images/win/scripts/Installers/Install-Git.ps1
```diff
@@ -14,11 +14,16 @@ Choco-Install -PackageName hub
...
 Add-MachinePathItem "C:\Program Files\Git\bin"
+$env:Path = Get-MachinePath
 
 if (Test-IsWin16) {
     $env:Path += ";$env:ProgramFiles\Git\usr\bin\"
 }
 
+# Set core.longpaths option to true in the system-wide [path]/etc/gitconfig file
+# https://github.com/git-for-windows/git/blob/main/Documentation/config/core.txt
+git config --system core.longpaths true
```
dscho, line 25: I _hope_ that this feature is stable enough to enable as widely as on the build agents... Be prepared to roll back on short notice, it might break users' scenarios in surprising and dramatic ways.

al-cheb: @dscho, It's better to be safe.

Issue #4424, opened 2021-11-04
Add .NET 6 SDK


akyamada-ms:
### Tool name

.NET 6 SDK

### Tool license

N/A

### Add or update?

- [X] Add
- [ ] Update

### Desired version

N/A

### Brief description of tool

I’m support engineer in Japan. One of my customers is using Ubuntu latest OS image for Microsoft-hosted agents in Azure DevOps Services. Does not have .NET 6 installed in Ubuntu latest OS image, so he adds the following definition to Pipeline and installs it every time.

steps:
- task: UseDotNet@2
displayName: 'Use .NET Core sdk'
inputs:
packageType: sdk
version: 6.0.100-rc.2.21505.57
installationPath: $(Agent.ToolsDirectory)/dotnet

I think you already have a plan, but I would like to request to include .NET 6 in the OS image.

Also, I would appreciate it if you could let me know when the OS image will include .NET 6 and if there is any publicly available information about future plans.

### URL for tool's homepage

https://github.com/dotnet/core/tree/main/release-notes/6.0

### Virtual environments affected

- [X] Ubuntu 18.04
- [X] Ubuntu 20.04
- [ ] macOS 10.15
- [ ] macOS 11
- [ ] Windows Server 2016
- [ ] Windows Server 2019
- [ ] Windows Server 2022

### Can this tool be installed during the build?

```bash
steps:
- task: UseDotNet@2
displayName: 'Use .NET Core sdk'
inputs:
packageType: sdk
version: 6.0.100-rc.2.21505.57
installationPath: $(Agent.ToolsDirectory)/dotnet
```

### Tool installation time in runtime

N/A

al-cheb:
Hey, @akyamada-ms 
.NET 6 will launch at .NET Conf November 9-11(https://devblogs.microsoft.com/dotnet/announcing-net-6-release-candidate-2/) and take 1 or 2 weeks to new rollout.

miketimofeev:
Hi @akyamada-ms!
How long does it take to install .Net 6 in runtime? We are thinking of changing the policy and do not pre-install .NET on the images due to maintenance and security concerns.

akyamada-ms:
Thank you, @al-cheb @miketimofeev 
I tried it five times with ubuntu-latest and it took an average of 8.6 seconds. It looks like there is no problem with the Pipeline execution time.

epvanhouten:
[...]

PR #4958, opened 2022-01-27
[Ubuntu] Add .NET SDK 6.0

nikolai-frolov:
## Description
This PR adds .NET SDK 6.0 to Ubuntu images.

## Related issue: 
https://github.com/actions/virtual-environments/issues/4424

## Check list

### DIFF
--- a/images/linux/scripts/installers/dotnetcore-sdk.sh
+++ b/images/linux/scripts/installers/dotnetcore-sdk.sh
@@ -32,8 +32,12 @@ for version in ${DOTNET_VERSIONS[@]}; do
     release_url="https://dotnetcli.blob.core.windows.net/dotnet/release-metadata/${version}/releases.json"
     download_with_retries "${release_url}" "." "${version}.json"
     releases=$(cat "./${version}.json")
-    sdks=("${sdks[@]}" $(echo "${releases}" | jq '.releases[]' | jq '.sdk.version'))
-    sdks=("${sdks[@]}" $(echo "${releases}" | jq '.releases[]' | jq '.sdks[]?' | jq '.version'))
+    if [[ $version == "6.0" ]]; then
+        sdks=("${sdks[@]}" $(echo "${releases}" | jq '.releases[]' | jq '.sdks[]?' | jq '.version' | grep -v preview | grep -v rc | grep -v display -m 1))
+    else
+        sdks=("${sdks[@]}" $(echo "${releases}" | jq '.releases[]' | jq '.sdk.version'))
+        sdks=("${sdks[@]}" $(echo "${releases}" | jq '.releases[]' | jq '.sdks[]?' | jq '.version'))
+    fi
     rm ./${version}.json
 done
 
--- a/images/linux/scripts/installers/dotnetcore-sdk.sh
+++ b/images/linux/scripts/installers/dotnetcore-sdk.sh
@@ -33,15 +33,15 @@ for version in ${DOTNET_VERSIONS[@]}; do
     download_with_retries "${release_url}" "." "${version}.json"
     releases=$(cat "./${version}.json")
     if [[ $version == "6.0" ]]; then
-        sdks=("${sdks[@]}" $(echo "${releases}" | jq '.releases[]' | jq '.sdks[]?' | jq '.version' | grep -v preview | grep -v rc | grep -v display -m 1))
+        sdks=("${sdks[@]}" $(echo "${releases}" | jq -r 'first(.releases[].sdks[]?.version | select(contains("preview") or contains("rc") | not))'))
     else
-        sdks=("${sdks[@]}" $(echo "${releases}" | jq '.releases[]' | jq '.sdk.version'))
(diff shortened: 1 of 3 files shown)


miketimofeev (changes requested):
@nikolai-frolov we should install only the latests available .Net 6 version

images/linux/scripts/installers/dotnetcore-sdk.sh
```diff
@@ -32,12 +32,15 @@ for version in ${DOTNET_VERSIONS[@]}; do
     release_url="https://dotnetcli.blob.core.windows.net/dotnet/release-metadata/${version}/releases.json"
     download_with_retries "${release_url}" "." "${version}.json"
     releases=$(cat "./${version}.json")
-    sdks=("${sdks[@]}" $(echo "${releases}" | jq '.releases[]' | jq '.sdk.version'))
-    sdks=("${sdks[@]}" $(echo "${releases}" | jq '.releases[]' | jq '.sdks[]?' | jq '.version'))
+    if [[ $version == "6.0" ]]; then
+        sdks=("${sdks[@]}" $(echo "${releases}" | jq '.releases[]' | jq '.sdk.version' | grep -v preview | grep -v rc | grep -v display)[0])
```
al-cheb, line 36: nitpick: grep has a parameter m to get only first match and get rid of [0]

images/linux/scripts/installers/dotnetcore-sdk.sh
```diff
@@ -32,8 +32,12 @@ for version in ${DOTNET_VERSIONS[@]}; do
     release_url="https://dotnetcli.blob.core.windows.net/dotnet/release-metadata/${version}/releases.json"
     download_with_retries "${release_url}" "." "${version}.json"
     releases=$(cat "./${version}.json")
-    sdks=("${sdks[@]}" $(echo "${releases}" | jq '.releases[]' | jq '.sdk.version'))
-    sdks=("${sdks[@]}" $(echo "${releases}" | jq '.releases[]' | jq '.sdks[]?' | jq '.version'))
+    if [[ $version == "6.0" ]]; then
+        sdks=("${sdks[@]}" $(echo "${releases}" | jq '.releases[]' | jq '.sdks[]?' | jq '.version' | grep -v preview | grep -v rc | grep -v display -m 1))
```
miketimofeev, line 36: Too many `greps` I believe 😃 Everything can be done using `jq` as it's a very powerful tool
```suggestion
        sdks=("${sdks[@]}" $(echo "${releases}" | jq -r 'first(.releases[].sdks[]?.version | select(contains("preview") or contains("rc") | not))'))
```

images/linux/scripts/installers/dotnetcore-sdk.sh
```diff
@@ -32,8 +32,12 @@ for version in ${DOTNET_VERSIONS[@]}; do
...
+    else
+        sdks=("${sdks[@]}" $(echo "${releases}" | jq '.releases[]' | jq '.sdk.version'))
+        sdks=("${sdks[@]}" $(echo "${releases}" | jq '.releases[]' | jq '.sdks[]?' | jq '.version'))
```
miketimofeev, line 39: What is the difference between those two?
Can they be combined into one using jq?

nikolai-frolov: I left it as it was, since more powerful constructions are harder to quickly understanding and both (.sdk.version and .sdks[].version) are required to keep current used .NET versions according to how .net releases are built. For example:
```
currently used versions: 2.1.302 2.1.403 2.1.526 2.1.617 2.1.701 2.1.818 3.1.120 3.1.202 3.1.302 3.1.416 5.0.104 5.0.210 5.0.303 5.0.404
without .sdk.version:                    2.1.526 2.1.617 2.1.701 2.1.818 3.1.120 3.1.202 3.1.302 3.1.416 5.0.104 5.0.210 5.0.303 5.0.404
without .sdks[].version: 2.1.302 2.1.403 2.1.507 2.1.607 2.1.701 2.1.818 3.1.101 3.1.201 3.1.302 3.1.416 5.0.102 5.0.202 5.0.302 5.0.404
```

images/linux/scripts/installers/dotnetcore-sdk.sh
nikolai-frolov: Completely agree there are too many `jp` and `grep`, but in my guess your suggestion is harder for understanding and it has the same length though `or contains("display")` is missed 😃 The same approach is used [a bit lower in this file](https://github.com/actions/virtual-environments/blob/6865a4ccfe6c5bcf5ac3a96c4eb84d035136903c/images/linux/scripts/installers/dotnetcore-sdk.sh#L40) so it keeps the consistency.

miketimofeev: In my opinion, `jq` string is much more readable that's why I suggest reworking the other strings as well and getting rid of so many grep pipes
Why do we need this `display`?

MaksimZhukov (approved):

miketimofeev (approved):

PR #5014, opened 2022-02-04
[macOS] install tcl-tk explicitly

mikhailkoliada:
# Description

While the workaround described in the attached issue works perfectly well, the actual symlinks creation happens much earlier than we have the Tcl/Tk package installed, as a result we are having broken symlinks created, to avoid this we need to install the tcl/tk package right in the place of the symlinks creation.

## Check list

### DIFF
--- a/images/macos/software-report/SoftwareReport.Common.psm1
+++ b/images/macos/software-report/SoftwareReport.Common.psm1
@@ -532,6 +532,11 @@ function Get-LibXextVersion {
     return "libXext $libXextVersion"
 }
 
+function Get-TclTkVersion {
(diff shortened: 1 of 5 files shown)


images/macos/software-report/SoftwareReport.Common.psm1
```diff
@@ -532,6 +532,11 @@ function Get-LibXextVersion {
     return "libXext $libXextVersion"
 }
 
+function Get-TclTkVersion {
+    $tcltkVersion = (brew info tcl-tk)[0] | Take-Part -Part
```
al-cheb, line 536: Missing Part argument

maldito187 (commented):
@maldito-dot

miketimofeev (approved):

Issue #4856, opened 2022-01-11
windows-latest workflows will use Windows Server 2022


miketimofeev:
### Breaking changes

Windows Server 2022 is ready to be the default version for the "windows-latest" label in GitHub Actions and Azure DevOps.

### Target date

This change will be rolled out over a period of several weeks beginning on January, 17. We plan to complete the migration by March, 6.

### The motivation for the changes

Starting from November 2021 Windows Server 2022 with Visual Studio 2022 image is generally available for all customers. We have monitored customer feedback to improve the Windows Server 2022 image stability and now we are ready to set it as the latest.

### Possible impact
~~⚠️  [CodeQL](https://github.com/github/codeql-action) does not currently support analysis of compiled languages on Windows Server 2022 ⚠️~~ Support for Windows Server 2022 has been added.

The Windows Server 2022 image has a different set of software than Windows Server 2019. The most significant changes are listed in the table below:
|Tool name|Windows-2019|Windows-2022|Notes|
|-|:-:|-|-|
|Go lang|1.15.*|1.16.*| Go lang version bumped to the more recent one |
|Python|3.7.*|3.9.*| Python version bumped to the more recent one |
|Ruby|2.5.*| 3.0.* | Ruby version bumped to the more recent one |
|Google Cloud SDK|latest| - | The software was removed due to maintenance concerns, consider using the official action https://github.com/google-github-actions/setup-gcloud |
|Innosetup|latest| - | The software was removed due to its low popularity and maintenance concerns |
|Parcel|latest| - | The software was removed due to its low popularity and maintenance concerns |
|Cloud Foundry CLI |latest| - | The software was removed due to its low popularity and maintenance concerns |
|Java|8(default)<br>11<br>13<br>17|8(default)<br>11<br>17| Due to our software guidelines we left only LTS versions on the image. If you need to use other versions of Java, consider using [setup-java](https://github.com/actions/setup-java) and [Java tool installer](https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/tool/java-tool-installer?view=azure-devops) tasks |
|Python toolcache|2.7<br>3.5<br>3.6<br>3.7<br>3.8<br>3.9<br>3.10|3.7<br>3.8<br>3.9<br>3.10| EOL python versions were removed |
|PyPy toolcache|2.7<br>3.6<br>3.7<br>3.8|2.7<br>3.7<br>3.8| Version 3.6 was deprecated as it won't receive any updates |
[...]

PR #5093, opened 2022-02-16
windows-latest is now windows-2022

alkino:
It seems that new windows-latest is now 2022, let's document it


miketimofeev:
Hi @alkino!
The transition is in progress — https://github.com/actions/virtual-environments/issues/4856
We will update all the docs once it is finished so I'm going to close this PR.
Thank you!

PR #5102, opened 2022-02-17
FIX: Write-Host Incorrect calls to variables.

jacob-blanc:
Error calling variables into last step of Azure VM creation. 

ResourceGroupName: The term 'ResourceGroupName' is not recognized as a name of a cmdlet, function, script file, or executable program.
Check the spelling of the name, or if a path was included, verify that the path is correct and try again.

# Description
Bug fixing

We get an error into the last step of creation of a VM Creation in Azure. Not retriving variables values because calls are incorrect. We should use $ResourceGroupName or ${ResourceGroupName} or $($ResourceGroupName) instead bad calling $(ResourceGroupName) 

## Check list

### DIFF
--- a/helpers/CreateAzureVMFromPackerTemplate.ps1
+++ b/helpers/CreateAzureVMFromPackerTemplate.ps1
@@ -72,5 +72,5 @@ Function CreateAzureVMFromPackerTemplate {
     Write-Host "`nCreating the VM"
     az group deployment create -g $ResourceGroupName -n $VirtualMachineName --subscription $subscriptionId --template-file $templateFilePath --parameters vmSize=$vmSize vmName=$VirtualMachineName adminUserName=$AdminUsername adminPassword=$AdminPassword networkInterfaceId=$networkId
     
-    Write-Host "`nCreated in $(ResourceGroupName):`n  vnet $(vnetName)`n  subnet $(subnetName)`n  nic $(nicName)`n  publicip $(publicIpName)`n  vm $(VirtualMachineName)"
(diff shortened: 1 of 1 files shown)


helpers/CreateAzureVMFromPackerTemplate.ps1
```diff
@@ -72,5 +72,5 @@ Function CreateAzureVMFromPackerTemplate {
     Write-Host "`nCreating the VM"
     az group deployment create -g $ResourceGroupName -n $VirtualMachineName --subscription $subscriptionId --template-file $templateFilePath --parameters vmSize=$vmSize vmName=$VirtualMachineName adminUserName=$AdminUsername adminPassword=$AdminPassword networkInterfaceId=$networkId
     
-    Write-Host "`nCreated in $(ResourceGroupName):`n  vnet $(vnetName)`n  subnet $(subnetName)`n  nic $(nicName)`n  publicip $(publicIpName)`n  vm $(VirtualMachineName)"
+    Write-Host "`nCreated in $($ResourceGroupName):`n  vnet $($vnetName)`n  subnet $($subnetName)`n  nic $($nicName)`n  publicip $($publicIpName)`n  vm $($VirtualMachineName)"
```
al-cheb, line 75: too many $(), please replace $(var) -> ${var}

jacob-blanc: Applied suggested changes. I also agree it is more elegant solution.

al-cheb (approved):

shilovmaksim (approved):

nikolai-frolov (approved):

miketimofeev (approved):

miketimofeev:
@jacob-blanc thank you for the contribution!

Issue #5231, opened 2022-03-15
Invalid kcpassword on macOS


t0rr3sp3dr0:
### Description

The /etc/kcpassword file on macOS Runners is invalid. It is a 54-byte file, but its size should be a multiple of 11, and trying to decode it doesn’t result on a valid password for user runner.

I believe this was introduced on #5153. If you run the old Python script and the new Bash one, you can check the output differ. The newer script produces an output similar to the expected, but adds or looses bytes depending on the input. More specifically, it looses bytes when it XORs a char with the key and it results in `\0` or `\n` and it adds bytes by padding the string too much.

### Example 1
Raw Input: `wVI=`
Base64 Output:
 - python: `Ct8bHtK83eqjuR8=` (11B decoded, 👍)
 - bash: `3xse0rzd6qO5H30=` (11B decoded, 👎 missing first byte + last byte is garbage)

### Example 2
Raw Input: `}_R#`
Base64 Output:
 - python: `ANYAANK83eqjuR8=` (11B decoded, 👍)
 - bash: `1tK83eqjuR99` (9B decoded, 👎 missing first, third, and fourth bytes + last byte is garbage)

### Example 3
Raw Input: `00000000000`
Base64 Output:
 - python: `BOcfHtK83eqjuR8=` (11B decoded, 👍)
 - bash: `BOcfHtK83eqjuR99` (12B decoded, 👎 last byte is garbage)

### Example 4
Raw Input: `111111111111`
Base64 Output:
 - python: `TLhjEuON7NuSiC5MiVIj0rzd6qO5Hw==` (22B decoded, 👍)
 - bash: `TLhjEuON7NuSiC5MiVIj0rzd6qO5H32J` (24B decoded, 👎 last 2 bytes are garbage)

Given the size of the file inside the runner, I think somehow the garbage bytes are being stripped from the file, but some byte is still being lost when the script is generates kcpassword.

### Virtual environments affected

- [ ] Ubuntu 18.04
- [ ] Ubuntu 20.04
- [X] macOS 10.15
- [X] macOS 11
- [ ] Windows Server 2016
- [ ] Windows Server 2019
- [ ] Windows Server 2022

### Image version and build link

20220307.1

### Is it regression?

Yes

### Expected behavior

kcpassword must be a file with size multiple of 11 and it should contain the password for the user runner when decoded.

### Actual behavior

kcpassword is not valid.

### Repro steps

Check kcpassword size:
```bash
ls -lha /etc/kcpassword
[...]

PR #5234, opened 2022-03-15
[macOS] Fix kcpassword compatibility issue with Python 3

maxim-lobanov:
# Description
In scope of https://github.com/actions/virtual-environments/pull/5115, we switched `kcpassword` script to use Python 3 instead of Python 2 because Python 2 was removed from macOS 12.
After that, kcpassword started to produce incorrect results. See https://github.com/actions/virtual-environments/issues/5231 for details.

### Root cause
`chr` function was changed in Python 3 and don't produce char by number anymore.
More details can be found in https://python-future.org/compatible_idioms.html#chr

### Fix
- I have removed converting to string completely and convert to byte array instead
- Writing byte array directly to file

### How it was tested
I have compared all test cases reported in https://github.com/actions/virtual-environments/issues/5231 and after my changes script product correct result for both Python 2 & 3

### DIFF
--- a/images/macos/provision/bootstrap-provisioner/kcpassword.py
+++ b/images/macos/provision/bootstrap-provisioner/kcpassword.py
@@ -27,13 +27,13 @@ def kcpassword(passwd):
             passwd[j] = passwd[j] ^ key[ki]
             ki += 1
 
-    passwd = [chr(x) for x in passwd]
-    return "".join(passwd)
+    return bytearray(passwd)
 
 if __name__ == "__main__":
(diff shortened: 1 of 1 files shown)


brcrista:
It seems like the compatibility issue isn't `chr()` itself, but that

```py
b'a' == 'a' # True in Python 2, False in Python 3
```

Anyway, using `bytestring` here seems appropriate.

I don't have context on the application domain.  What is `/etc/kcpassword` and how do we use it?

AlenaSviridenko (approved):

Issue #5285, opened 2022-03-25
Update LLVM to 14.0.0


mmuffins:
### Tool name

LLVM

### Tool license

Apache v2.0 

### Add or update?

- [ ] Add
- [X] Update

### Desired version

14.0.0

### Brief description of tool

LLVM 14.0 was released and is available on chocolatey, failing the builds for server 2016 and 2019 because their toolset manifest still check for version 13.

### Virtual environments affected

- [ ] Ubuntu 18.04
- [ ] Ubuntu 20.04
- [ ] macOS 10.15
- [ ] macOS 11
- [X] Windows Server 2016
- [X] Windows Server 2019
- [ ] Windows Server 2022

### Are you willing to submit a PR?

Yes

miketimofeev:
@mmuffins please take a look at the comment https://github.com/actions/virtual-environments/pull/5286#issuecomment-1079553563
We will update llvm a bit later and announce it first.

miketimofeev:
@FrancescElies hi! The announcement is right here https://github.com/actions/virtual-environments/issues/5835

PR #5286, opened 2022-03-25
Update LLVM to 14.0.0

mmuffins:
# Description
The latest version of llvm was released on chocolatey, breaking the builds for server 2016 and 2019 since they still validate for version 13. 2022 doesn't seem to have a version check for llvm, so no nothing needs to be updated there (besides the readme).
  
#### Related issue:
#5285 

## Check list

### DIFF
--- a/images/win/Windows2016-Readme.md
+++ b/images/win/Windows2016-Readme.md
@@ -13,7 +13,7 @@
 - Go 1.15.15
 - Julia 1.7.2
 - Kotlin 1.6.10
-- LLVM 14.0
+- LLVM 14.0.0
 - Node 16.14.0
 - Perl 5.32.1
 - PHP 8.1.3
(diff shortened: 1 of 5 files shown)


images/win/Windows2016-Readme.md
```diff
@@ -13,7 +13,7 @@
 - Go 1.15.15
 - Julia 1.7.2
 - Kotlin 1.6.10
-- LLVM 13.0.1
+- LLVM 14.0.0
```
panekj, line 16: Don't update README, it's generated automatically

mmuffins: Thanks for the info, didn't know that.

miketimofeev:
Hi @mmuffins! Thanks for your contribution but we wouldn't like to update llvm without an announcement so we will stick to version 13 for now. There is an error in the installation script, which is fixed here https://github.com/actions/virtual-environments/pull/5288

Issue #4799, opened 2021-12-23
[Ubuntu] Issue with libstdc++ cannot allocate memory in static TLS block


miketimofeev:
### Breaking changes

Due to a bug in MySql 8.0.27 (https://bugs.mysql.com/bug.php?id=105288) we had to revert the version to 8.0.26 yet it's not possible to use ubuntu upstream repo as it doesn't contain older versions. In this PR we switched the installation to install all the dependencies step by step utilizing the `dpkg -i `command https://github.com/actions/virtual-environments/pull/4674

This leads to unpredictable behavior in libstdc++ usage when calling from Python (there could be other cases we are not aware of) — https://github.com/actions/virtual-environments/issues/4733, https://github.com/actions/virtual-environments/issues/4790
Please use the provided workaround until we either got 8.0.26 in the upstream repo or the new version 8.0.28 is released.

### Possible impact

If your Python code uses some C++ libraries it can be broken.

### Virtual environments affected

- [X] Ubuntu 18.04
- [X] Ubuntu 20.04
- [ ] macOS 10.15
- [ ] macOS 11
- [ ] Windows Server 2016
- [ ] Windows Server 2019
- [ ] Windows Server 2022

### Mitigation ways

Reinstall MySQL version 8.0.27 using upstream repo:
```bash
sudo apt-get update
sudo apt-get remove mysql* && sudo apt-get install -y mysql-server
```

mikhailkoliada:
Announce with the workaround has been published. Sorry for inconvenience caused.

freddy123098:
Hi I saw this as open so may I help you? And what is the issue?

freddy123098:
Please provide repo-steps thank you.

freddy123098:
@miketimofeev pls provide repo steps thank you.

freddy123098:
Hi I saw this issue open.

Misshunter1:
to much violence do not want resilence

RW21:
To add to the mitigation method provided, if you are using `mysql-client`, you will need to install `libmysqlclient-dev`, otherwise you will get `OSError: mysql_config not found`. 
So:
```
    - name: GA fixes
      run: |
        sudo apt-get update
        sudo apt-get remove mysql* && sudo apt-get install -y mysql-server libmysqlclient-dev

```

anthillofoppies44:
jyz Ndoid havr linux 
idk

marcelgerber:
I think this issue can be fixed by now, right?
[...]

PR #5306, opened 2022-03-30
Remove Ubuntu20 mysql workaround

ddobranic:
# Description
Bug fix for #4799 - [Ubuntu] Issue with libstdc++ cannot allocate memory in static TLS block. Remove old workaround since it is not needed as new versions of MySQL are available

#### Related issue:
#4799
## Check list


al-cheb (approved):

mikhailkoliada (changes requested):
Remove the init.d script too, as the deb distribution was incomplete and we had to port it manually from ubuntu.

miketimofeev:
> Remove the init.d script too, as the deb distribution was incomplete and we had to port it manually from ubuntu.

@ddobranic this one https://github.com/actions/virtual-environments/blob/main/images/linux/scripts/helpers/mysql-service-helper.sh

miketimofeev (approved):

ball144love (approved):

Issue #5208, opened 2022-03-09
Update default go version to 1.16 or higher


g14com0:
### Tool name

go

### Tool license

MIT

### Add or update?

- [ ] Add
- [X] Update

### Desired version

1.16.*

### Brief description of tool

Upgrade default version of golang at least to v1.16 since golang 1.15 reached end of life and it's vulnerable.

### Virtual environments affected

- [X] Ubuntu 18.04
- [X] Ubuntu 20.04
- [X] macOS 10.15
- [X] macOS 11
- [x] Windows Server 2016
- [X] Windows Server 2019
- [X] Windows Server 2022

### Are you willing to submit a PR?

yes

g14com0:
Hello, [here](https://github.com/g14com0/virtual-environments/pull/1) you can find a PR in my forked repo, i didn't remove go 1.15 because i think you want first to announce this change also, go 1.16.15 and 1.17.8 are out and can be used once the pipeline in [go-version](https://github.com/actions/go-versions/actions/workflows/get-go-versions.yml) repo will be fixed as mentioned in #5202

maxim-lobanov:
Hello @g14com0 , thank you for bringing our attention to it.
Considering the fact that Go 1.15 reached EOL, we definitely should update default to 1.16 or 1.17 but I think we should some announcement regarding it.
I will double check and will return back.

g14com0:
v1.16 reached as well EOL as reported [here](https://endoflife.date/go), please consider only 1.17 and 1.18 as the default version

Issue #5280, opened 2022-03-25
[All OSs] Go versions less than 1.16 will be removed and the default will be set to 1.17 on April, 11


ddobranic:
### Breaking changes

All pre-cached Go versions less than 1.16 will be removed from all operating systems. Version 1.17 will be set as the new default version.

### Target date

April 11, 2022.

### The motivation for the changes

According to our [Software and image guidelines
](https://github.com/actions/virtual-environments/blob/main/docs/software-and-images-guidelines.md) we support [3 latest minor versions of Go],
(https://endoflife.date/go) which are 1.16, 1.17 & 1.18 at the moment.

Version 1.17 will be the new default as it is most popular according to telemetry data.

### Possible impact

If your project depends on a pre-installed Go version 1.15 it can be broken.

### Virtual environments affected

- [X] Ubuntu 18.04
- [X] Ubuntu 20.04
- [X] macOS 10.15
- [ ] macOS 11
- [X] Windows Server 2016
- [X] Windows Server 2019
- [ ] Windows Server 2022

### Mitigation ways

To install any versions in runtime use:
- https://github.com/actions/setup-go action for GH actions
- https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/tool/go-tool?view=azure-devops task for AzDO

nauil:
جميل جد @ #

- [ ] 
![Uploading 16502566231168533349594927354681.jpg…]()

PR #5336, opened 2022-04-04
Update default Go version to 1.16 or higher

ddobranic:
# Description
Update default go version to 1.17 since it is most popular
#### Related issue:
#5208
## Check list

### DIFF
--- a/images/linux/toolsets/toolset-1804.json
+++ b/images/linux/toolsets/toolset-1804.json
@@ -42,12 +42,11 @@
             "arch": "x64",
             "platform" : "linux",
             "versions": [
-                "1.15.*",
                 "1.16.*",
                 "1.17.*",
                 "1.18.*"
             ],
-            "default": "1.15.*"
+            "default": "1.17.*"
         },
         {
             "name": "Ruby",
--- a/images/linux/toolsets/toolset-2004.json
+++ b/images/linux/toolsets/toolset-2004.json
@@ -44,12 +44,11 @@
             "arch": "x64",
             "platform" : "linux",
             "versions": [
-                "1.15.*",
                 "1.16.*",
                 "1.17.*",
                 "1.18.*"
             ],
-            "default": "1.15.*"
+            "default": "1.17.*"
         },
         {
             "name": "Ruby",
--- a/images/macos/toolsets/toolset-10.15.json
+++ b/images/macos/toolsets/toolset-10.15.json
@@ -290,7 +290,6 @@
             "platform" : "darwin",
             "variable_template" : "GOROOT_{0}_{1}_X64",
             "versions": [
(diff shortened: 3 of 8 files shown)


images/linux/toolsets/toolset-1804.json
```diff
@@ -42,12 +42,11 @@
             "arch": "x64",
             "platform" : "linux",
             "versions": [
-                "1.15.*",
                 "1.16.*",
                 "1.17.*",
                 "1.18.*"
             ],
-            "default": "1.15.*"
+            "default": "1.17.*"
```
maxim-lobanov, line 49: I think we should update default to 1.18.
Updating from 1.15 to 1.17 will be painful anyway so it is worth to update to 1.18 as actual stable version right now.
1.17 will reach EOL on 2022-08-16

maxim-lobanov, line 49: Could you please confirm that we have an announcement for this update?
According to our docs we should publish announcement at least 2 weeks prior update default versions: https://github.com/actions/virtual-environments/blob/main/docs/software-and-images-guidelines.md#software-default-versions-update-policy-for-tools-with-multiple-versions-installed

ddobranic: We run Kusto queries and seen that 1.17 is the most used, currently. Also, we support three latest versions.
What do you think?

miketimofeev: Here it is https://github.com/actions/virtual-environments/issues/5280

miketimofeev: But maybe we should update the title as we mention the default version change only in the description

maxim-lobanov: Ah, makes sense. 
Actually, only default version really matters for customers because all other versions are "tool-cache". It means they can be used only via setup-tasks. But if customers are using setup-tasks - they won't be affected at all because tasks will download version on-flight. So real impact is only default version change.

maxim-lobanov: I think 1.17 is the most used because 1.18 was released recently and not all customers was able to switch to it yet.
That is probably up to you but keep on mind that if you update default to 1.17 for now, you will need to update to 1.18 one more time in 4-5 months.  It is a bit easier to bump it once.
I am not really sure how much breaking changes between 1.17 and 1.18 versions and how many customers will be affected. But I guess updating from 1.15 to 1.17 will be very painful anyway.

miketimofeev: we have announced 1.15 -> 1.17 switch already

azboyzranch (approved):

miketimofeev (approved):

PR #5373, opened 2022-04-08
Update README.md

ikem-krueger:
This patch makes it more clear, which architecture is used in the containers.

That's especially important for people, which expects a certain architecture like "armhf".

# Description

ARM devices will become more and more relevant. It's only a matter of time, when virtualization on ARM is becoming the norm.

I know, you should write your software architecture independent as possible, but that's not always possible.

Communicating clearly what is, and what is not supported within GitHub actions makes it easier to work with it.


mikhailkoliada:
hello @ikem-krueger as we do not provide arm-based vms it does not make much sense to distinguish anything as this is all about x86_64

PR #5390, opened 2022-04-12
[Ubuntu] remove php 7.1 from Bionic Beaver

mikhailkoliada:
# Description

According to the related issue php 7.1 must be removed from Ubuntu 18.04 on April 18th.

## Check list


miketimofeev:
@mikhailkoliada we have some conditions regarding the old versions like this
https://github.com/actions/virtual-environments/blob/main/images/linux/scripts/installers/php.sh#L67
please remove them

images/linux/scripts/installers/php.sh
```diff
@@ -64,18 +64,6 @@ for version in $php_versions; do
...
-
-    if [[ $version == "7.2" || $version == "7.3" ]]; then
-        apt-get install -y --no-install-recommends php$version-recode
-    fi
-
-    if [[ $version != "8.0" && $version != "8.1" ]]; then
-        apt-get install -y --no-install-recommends php$version-xmlrpc php$version-json
-    fi
-
     if [[ $version != "5.6" && $version != "7.0" ]]; then
```
al-cheb, line 67: Do we need this condition?

miketimofeev (approved):

PR #5396, opened 2022-04-12
Limit storageaccount name length to 24 characters

Tapanila:
# Description
If resource group name is long then the script will generate storage account name that's longer than 24 characters.
This change will limit the generated storageaccountname to 24 characters.

[Document about limits](https://docs.microsoft.com/en-us/azure/storage/common/storage-account-overview#:~:text=Storage%20account%20names%20must%20be,must%20be%20unique%20within%20Azure.)

### DIFF
--- a/helpers/GenerateResourcesAndImage.ps1
+++ b/helpers/GenerateResourcesAndImage.ps1
@@ -195,7 +195,9 @@ Function GenerateResourcesAndImage {
     
     
     # Storage Account Name can only be 24 characters long
-    $storageAccountName = $storageAccountName.Substring(0, 24)
+    if ($storageAccountName.Length -gt 24){
+        $storageAccountName = $storageAccountName.Substring(0, 24)
+    }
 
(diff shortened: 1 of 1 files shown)


helpers/GenerateResourcesAndImage.ps1
```diff
@@ -192,6 +192,10 @@ Function GenerateResourcesAndImage {
     # Resource group names may contain special characters, that are not allowed in the storage account name
     $storageAccountName = $storageAccountName.Replace("-", "").Replace("_", "").Replace("(", "").Replace(")", "").ToLower()
     $storageAccountName += "001"
+    
+    
+    # Storage Account Name can only be 24 characters long
+    $storageAccountName = $storageAccountName.Substring(0, 24)
```
al-cheb, line 198: You should check if the length of storageAccountName is more than 24:
```
PS > "ahbfff".Substring(0, 24)
MethodInvocationException: Exception calling "Substring" with "2" argument(s): "Index and length must refer to a location within the string. (Parameter 'length')"
```

Tapanila: Somehow missed this completely. Added a check for the length

al-cheb (approved):

dmitry-shibanov (approved):

miketimofeev (approved):

PR #5420, opened 2022-04-20
[Ubuntu] Add safe directory setting to git config

miketimofeev:
# Description
Due to the recent security changes in git actions\checkout task is experiencing issues https://github.com/actions/checkout/issues/760. The workaround is to set `git config --global --add safe.directory "*"` setting. This PR adds this setting in `/etc/skel` as the user is different between image generation VM and VM in production.

#### Related issue:

## Check list

### DIFF
--- a/images/linux/scripts/installers/git.sh
+++ b/images/linux/scripts/installers/git.sh
@@ -15,6 +15,11 @@ add-apt-repository $GIT_REPO -y
 apt-get update
 apt-get install git -y
 git --version
+# Git version 2.35.2 introduces security fix that breaks action\checkout https://github.com/actions/checkout/issues/760
+cat <<EOF >> /etc/skel/.gitconfig
+[safe]
+        directory = *
+EOF
 
 # Install git-lfs
 curl -s $GIT_LFS_REPO/script.deb.sh | bash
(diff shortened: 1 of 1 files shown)


miketimofeev:
@dscho could you please review?

al-cheb (dismissed):

ddobranic (dismissed):

vsafonkin (dismissed):

images/linux/scripts/installers/git.sh
```diff
@@ -15,6 +15,11 @@ add-apt-repository $GIT_REPO -y
 apt-get update
 apt-get install git -y
 git --version
+# Git version 2.35.2 introduces security fix that breaks action\checkout https://github.com/actions/checkout/issues/760
+cat <<EOF >> /etc/skel/.gitconfig
+[safe]
+        directory = *
+EOF
```
dscho, line 22: Would it not make more sense to put this into `/etc/gitconfig`? Or would that break usage in containers?

miketimofeev: @thboop @TingluoHuang do you have any thoughts regarding containers?

TingluoHuang: `/etc/gitconfig` should be fine since we just want to turn off the `safe.dir` enforcement in the hosted image.

ddobranic (approved):

mikhailkoliada (approved):

PR #5452, opened 2022-04-28
[MacOS]  Add safe directory setting to git config

AlexanderGitin:
# Description
Due to the recent security changes in git actions\checkout task is experiencing issues https://github.com/actions/checkout/issues/760. The workaround is to set git config --global --add safe.directory "*" setting. This PR adds this setting in /etc/gitconfig as it is used globally for every OS user.

#### Related issue:

## Check list

### DIFF
--- a/images/macos/provision/core/git.sh
+++ b/images/macos/provision/core/git.sh
@@ -8,6 +8,11 @@ brew extract --version=2.35.1 git local/git
 brew install git@2.35.1
 brew untap -f local/homebrew-git
 
+cat <<EOF >> /etc/gitconfig
+[safe]
+        directory = *
+EOF
+
 echo Installing Git LFS
 brew_smart_install "git-lfs"
 


images/macos/provision/core/git.sh
```diff
@@ -8,6 +8,11 @@ brew extract --version=2.35.1 git local/git
 brew install git@2.35.1
```
al-cheb, line 8: Should we remove this workaround?

miketimofeev: for sure, this was one of the goal @AlexanderGitin

AlexanderGitin: I removed temporary fix please review

al-cheb (dismissed):

images/macos/provision/core/git.sh
```diff
@@ -2,11 +2,12 @@
...
 
 echo Installing Git...
-# Temporary hardcode version 2.35.1 due to the issue with the actions\checkout https://github.com/actions/checkout/issues/76
-brew tap-new local/git
-brew extract --version=2.35.1 git local/git
-brew install git@2.35.1
-brew untap -f local/homebrew-git
+brew_smart_install "git"
+
+cat <<EOF >> /etc/gitconfig
```
miketimofeev, line 7: you need sudo permissions to write to this file

miketimofeev (approved):

PR #5453, opened 2022-04-28
[Windows] Add safe directory to gitconfig

AlexanderGitin:
# Description
Due to the recent security changes in git actions\checkout task is experiencing issues https://github.com/actions/checkout/issues/760. The workaround is to set git config --global --add safe.directory "*" setting. This PR adds this setting in /etc/gitconfig as it is used globally for every OS user.
#### Related issue:
## Check list


images/macos/provision/core/git.sh
```diff
@@ -2,11 +2,12 @@
 source ~/utils/utils.sh
```
al-cheb, line 3: Why do we have change for macOS?

images/win/scripts/Installers/Install-Git.ps1
```diff
@@ -4,8 +4,10 @@
 ################################################################################
 
 # Install git
-# Temporary hardcode version 2.35.1 due to the issue with the actions\checkout https://github.com/actions/checkout/issues/760
-Choco-Install -PackageName git -ArgumentList '--version', "2.35.1.2", '--installargs="/VERYSILENT /NORESTART /NOCANCEL /SP- /CLOSEAPPLICATIONS /RESTARTAPPLICATIONS /o:PathOption=CmdTools /o:BashTerminalOption=ConHost /o:EnableSymlinks=Enabled /COMPONENTS=gitlfs"'
+Choco-Install -PackageName git -ArgumentList '--installargs="/VERYSILENT /NORESTART /NOCANCEL /SP- /CLOSEAPPLICATIONS /RESTARTAPPLICATIONS /o:PathOption=CmdTools /o:BashTerminalOption=ConHost /o:EnableSymlinks=Enabled /COMPONENTS=gitlfs"'
+
+echo "[safe]
```
al-cheb, line 9: Why do we need echo?

images/win/scripts/Installers/Install-Git.ps1
```diff
@@ -4,8 +4,12 @@
 ################################################################################
 
 # Install git
-# Temporary hardcode version 2.35.1 due to the issue with the actions\checkout https://github.com/actions/checkout/issues/760
-Choco-Install -PackageName git -ArgumentList '--version', "2.35.1.2", '--installargs="/VERYSILENT /NORESTART /NOCANCEL /SP- /CLOSEAPPLICATIONS /RESTARTAPPLICATIONS /o:PathOption=CmdTools /o:BashTerminalOption=ConHost /o:EnableSymlinks=Enabled /COMPONENTS=gitlfs"'
+Choco-Install -PackageName git -ArgumentList '--installargs="/VERYSILENT /NORESTART /NOCANCEL /SP- /CLOSEAPPLICATIONS /RESTARTAPPLICATIONS /o:PathOption=CmdTools /o:BashTerminalOption=ConHost /o:EnableSymlinks=Enabled /COMPONENTS=gitlfs"'
+
+Add-Content -Path "C:\Program Files\Git\etc\gitconfig" -Value "
```
al-cheb, line 9: Why do not we use the same approach like we do for macOS?

`git config --system --add safe.directory "*"`

miketimofeev: @al-cheb on macOS the user remains the same after the image generation but on windows the setting will be erased as the user is different

al-cheb: @miketimofeev , agree, what's why I put `system` flag.

al-cheb (approved):

miketimofeev (approved):

PR #5572, opened 2022-05-19
[Ubuntu] Make software and release URLs using variables instead of hardcode

miketimofeev:
# Description
Previously we have to add new ubuntu versions manually, which is not convenient and produces extra code. This PR moves hardcoded parts to variables thus we don't need to take care of this file when a new Ubuntu version is released.

#### Related issue:
https://github.com/actions/virtual-environments-internal/issues/3760

## Check list

### DIFF
--- a/images/linux/scripts/installers/preimagedata.sh
+++ b/images/linux/scripts/installers/preimagedata.sh
@@ -5,8 +5,8 @@ image_version=$IMAGE_VERSION
 os_name=$(lsb_release -ds | sed "s/ /\\\n/g")
 os_version=$(lsb_release -rs)
 image_label="ubuntu-${os_version}"
-version_major=$(echo ${os_version} | cut -d "." -f 1)
-version_wo_dot=$(echo ${os_version} | tr -d ".")
+version_major=${os_version/.*/}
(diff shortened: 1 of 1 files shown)


ddobranic (approved):

images/linux/scripts/installers/preimagedata.sh
```diff
@@ -3,18 +3,14 @@
 imagedata_file=$IMAGEDATA_FILE
 image_version=$IMAGE_VERSION
 os_name=$(lsb_release -ds | sed "s/ /\\\n/g")
-image_label="ubuntu-$(lsb_release -rs)"
+os_version=$(lsb_release -rs)
+image_label="ubuntu-${os_version}"
+version_major=$(echo ${os_version} | cut -d "." -f 1)
+version_wo_dot=$(echo ${os_version} | tr -d ".")
```
al-cheb, line 9: nitpick: use substring `${os_version/./}`

dmitry-shibanov (approved):

al-cheb (approved):

ddobranic (approved):

PR #5587, opened 2022-05-23
[Ubuntu] Use new Hashicorp Releases API

smaeda-ks:
# Description

Changing the legacy API endpoint URL to the new endpoint that was announced recently:
ref. https://www.hashicorp.com/blog/announcing-the-hashicorp-releases-api

### legacy

```sh
curl -s https://checkpoint-api.hashicorp.com/v1/check/terraform | jq -r .current_version
```
### new

```sh
curl -s https://api.releases.hashicorp.com/v1/releases/terraform/latest | jq -r .version
```

## Check list

### DIFF
--- a/images/linux/scripts/installers/packer.sh
+++ b/images/linux/scripts/installers/packer.sh
@@ -4,10 +4,13 @@
 ##  Desc:  Installs packer
 ################################################################################
 
+source $HELPER_SCRIPTS/install.sh
+
 # Install Packer
 URL=$(curl -s https://api.releases.hashicorp.com/v1/releases/packer/latest | jq -r '.builds[] | select((.arch=="amd64") and (.os=="linux")).url')
-curl -L -o packer_linux_amd64.zip "${URL}"
-unzip -qq packer_linux_amd64.zip -d /usr/local/bin
-rm -f packer_linux_amd64.zip
+ZIP_NAME="packer_linux_amd64.zip"
+download_with_retries "${URL}" "/tmp" "${ZIP_NAME}"
+unzip -qq "/tmp/${ZIP_NAME}" -d /usr/local/bin
+rm -f "/tmp/${ZIP_NAME}"
 
 invoke_tests "Tools" "Packer"
--- a/images/linux/scripts/installers/packer.sh
+++ b/images/linux/scripts/installers/packer.sh
@@ -5,9 +5,9 @@
 ################################################################################
 
 # Install Packer
-PACKER_VERSION=$(curl -s https://api.releases.hashicorp.com/v1/releases/packer/latest | jq -r .version)
-curl -LO "https://releases.hashicorp.com/packer/${PACKER_VERSION}/packer_${PACKER_VERSION}_linux_amd64.zip"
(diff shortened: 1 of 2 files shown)


al-cheb (approved):

smaeda-ks:
Just realized that there's also one for Packer, so updating that script as well. Thanks for rpproving that @al-cheb

images/linux/scripts/installers/packer.sh
```diff
@@ -5,7 +5,7 @@
 ################################################################################
 
 # Install Packer
-PACKER_VERSION=$(curl -s https://checkpoint-api.hashicorp.com/v1/check/packer | jq -r .current_version)
+PACKER_VERSION=$(curl -s https://api.releases.hashicorp.com/v1/releases/packer/latest | jq -r .version)
```
al-cheb, line 8: Let's update the filter and get a link from json:
`'.builds[] | select((.arch=="amd64") and (.os=="linux")).url'`

images/linux/scripts/installers/terraform.sh
```diff
@@ -5,7 +5,7 @@
 ################################################################################
 
 # Install Terraform
-TERRAFORM_VERSION=$(curl -s https://checkpoint-api.hashicorp.com/v1/check/terraform | jq -r .current_version)
+TERRAFORM_VERSION=$(curl -s https://api.releases.hashicorp.com/v1/releases/terraform/latest | jq -r .version)
```
al-cheb, line 8: Let's update the filter and get a link from json:
`'.builds[] | select((.arch=="amd64") and (.os=="linux")).url'`

smaeda-ks:
Thanks for your review @al-cheb 
Made some more changes as per your suggestions.

images/linux/scripts/installers/packer.sh
```diff
@@ -5,7 +5,7 @@
...
 curl -LO "https://releases.hashicorp.com/packer/${PACKER_VERSION}/packer_${PACKER_VERSION}_linux_amd64.zip"
```
al-cheb: Replace curl to the helper function `download_with_retries` [source $HELPER_SCRIPTS/install.sh]

images/linux/scripts/installers/terraform.sh
```diff
@@ -5,7 +5,7 @@
...
 curl -LO "https://releases.hashicorp.com/terraform/${TERRAFORM_VERSION}/terraform_${TERRAFORM_VERSION}_linux_amd64.zip"
```
al-cheb: Replace curl to the helper function `download_with_retries` [source $HELPER_SCRIPTS/install.sh]

images/linux/scripts/installers/packer.sh
smaeda-ks: updated

images/linux/scripts/installers/terraform.sh
smaeda-ks: updated

ddobranic (approved):

miketimofeev (approved):

PR #5676, opened 2022-06-02
[Ubuntu] install Temurin gpg key manually on 22.04

mikhailkoliada:
# Description

As `apt-key` is deprecated, installing gpg keyrings manually is the best option here.

## Check list

### DIFF
--- a/images/linux/scripts/installers/java-tools.sh
+++ b/images/linux/scripts/installers/java-tools.sh
@@ -38,20 +38,24 @@ createJavaEnvironmentalVariable() {
 }
 
 enableRepositories() {
+
+osLabel=$(getOSVersionLabel)
+
     if isUbuntu18 || isUbuntu20; then
         # Add Adopt PPA
-        wget -qO - "https://adoptopenjdk.jfrog.io/adoptopenjdk/api/gpg/key/public" | apt-key add -
-        add-apt-repository --yes https://adoptopenjdk.jfrog.io/adoptopenjdk/deb/
+        wget -qO - https://adoptopenjdk.jfrog.io/adoptopenjdk/api/gpg/key/public | gpg --dearmor > /usr/share/keyrings/adopt.gpg
+        echo "deb [signed-by=/usr/share/keyrings/adopt.gpg] https://adoptopenjdk.jfrog.io/adoptopenjdk/deb/ $osLabel main" > /etc/apt/sources.list.d/adopt.list
     fi
 
     # Add Addoptium PPA
-    wget -qO - "https://packages.adoptium.net/artifactory/api/gpg/key/public" | apt-key add -
-    add-apt-repository --yes https://packages.adoptium.net/artifactory/deb/
+    # apt-key is deprecated, dearmor and add manually
+    wget -qO - https://packages.adoptium.net/artifactory/api/gpg/key/public | gpg --dearmor > /usr/share/keyrings/adoptium.gpg
+    echo "deb [signed-by=/usr/share/keyrings/adoptium.gpg] https://packages.adoptium.net/artifactory/deb/ $osLabel main" > /etc/apt/sources.list.d/adoptium.list
 
     if isUbuntu18 ; then
         # Install GPG Key for Azul Open JDK. See https://www.azul.com/downloads/azure-only/zulu/
-        apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 0xB1998361219BD9C9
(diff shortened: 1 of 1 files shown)


images/linux/scripts/installers/java-tools.sh
```diff
@@ -45,8 +45,14 @@ enableRepositories() {
...
 
     # Add Addoptium PPA
-    wget -qO - "https://packages.adoptium.net/artifactory/api/gpg/key/public" | apt-key add -
-    add-apt-repository --yes https://packages.adoptium.net/artifactory/deb/
+    if isUbuntu22; then
+        # apt-key is deprecated on Ubuntu22, dearmor and add manually
+        wget -qO - https://packages.adoptium.net/artifactory/api/gpg/key/public | gpg --dearmor > /usr/share/keyrings/adoptium.gpg
+        echo "deb [signed-by=/usr/share/keyrings/adoptium.gpg] https://packages.adoptium.net/artifactory/deb/ jammy main" > /etc/apt/sources.list.d/adoptium.list
+    else
+        wget -qO - "https://adoptopenjdk.jfrog.io/adoptopenjdk/api/gpg/key/public" | apt-key add -
```
al-cheb, line 53: Why do we duplicate line 41?

images/linux/scripts/installers/java-tools.sh
```diff
@@ -45,8 +45,14 @@ enableRepositories() {
     fi
 
     # Add Addoptium PPA
-    wget -qO - "https://packages.adoptium.net/artifactory/api/gpg/key/public" | apt-key add -
-    add-apt-repository --yes https://packages.adoptium.net/artifactory/deb/
+    if isUbuntu22; then
+        # apt-key is deprecated on Ubuntu22, dearmor and add manually
+        wget -qO - https://packages.adoptium.net/artifactory/api/gpg/key/public | gpg --dearmor > /usr/share/keyrings/adoptium.gpg
+        echo "deb [signed-by=/usr/share/keyrings/adoptium.gpg] https://packages.adoptium.net/artifactory/deb/ jammy main" > /etc/apt/sources.list.d/adoptium.list
```
al-cheb, line 51: We shouldn't hardcoded jammy.

images/linux/scripts/installers/java-tools.sh
```diff
@@ -45,8 +45,14 @@ enableRepositories() {
     fi
 
     # Add Addoptium PPA
-    wget -qO - "https://packages.adoptium.net/artifactory/api/gpg/key/public" | apt-key add -
-    add-apt-repository --yes https://packages.adoptium.net/artifactory/deb/
+    if isUbuntu22; then
```
al-cheb, line 48: Why do we need this condition if gpg --dearmor works for Ubuntu18+?

mikhailkoliada: @al-cheb what is the point to dearmor it on ubuntu18+ if apt-key does it here?

images/linux/scripts/installers/java-tools.sh
mikhailkoliada: well, it is already called only under ubuntu22 so the distro name is predictable in any single way

images/linux/scripts/installers/java-tools.sh
mikhailkoliada: @al-cheb or do you mean we must switch off apt-key generally? then it makes sense to me

images/linux/scripts/installers/java-tools.sh
```diff
@@ -45,8 +45,10 @@ enableRepositories() {
     fi
 
     # Add Addoptium PPA
-    wget -qO - "https://packages.adoptium.net/artifactory/api/gpg/key/public" | apt-key add -
-    add-apt-repository --yes https://packages.adoptium.net/artifactory/deb/
+    # apt-key is deprecated, dearmor and add manually
+    wget -qO - https://packages.adoptium.net/artifactory/api/gpg/key/public | gpg --dearmor > /usr/share/keyrings/adoptium.gpg
+    echo "deb [signed-by=/usr/share/keyrings/adoptium.gpg] https://packages.adoptium.net/artifactory/deb/ $(lsb_release -cs) main" > /etc/apt/sources.list.d/adoptium.list
+    rm /etc/apt/sources.list.d/adoptium.list
```
al-cheb, line 51: Are you sure that we need delete it right after?

miketimofeev (approved):

Issue #5682, opened 2022-06-02
Github Actions now break, since Visual Studio (v8.*) on Mac has been renamed to "Visual Studio 2019.app"


CartBlanche:
### Description

@al-cheb @miketimofeev Seem you fixed the macOS-12 image for VS2022, but now the macOS-11 image is broken for VS2019

/Users/runner/.nuget/packages/monodevelop.addins/0.4.7/build/MonoDevelop.Addins.targets(76,5): error : Could not find a global instance of Visual Studio for Mac or MonoDevelop. Try providing an explicit value with MDBinDir. [/Users/runner/work/VS_Mac_Meadow_Extension/VS_Mac_Meadow_Extension/main/VS4Mac_Meadow_Extension/Meadow.Sdks.IdeExtensions.Vs4Mac.csproj]

I see you've rename the VS2019 app to "Visual Studio 2019.app". So from a Github Action how do we specify that specific VS version for a particular build?

### Platforms affected

- [ ] Azure DevOps
- [ ] GitHub Actions

### Virtual environments affected

- [ ] Ubuntu 18.04
- [ ] Ubuntu 20.04
- [ ] Ubuntu 22.04
- [ ] macOS 10.15
- [X] macOS 11
- [ ] macOS 12
- [ ] Windows Server 2019
- [ ] Windows Server 2022

### Image version and build link

https://github.com/WildernessLabs/VS_Mac_Meadow_Extension/runs/6712367837?check_suite_focus=true

### Is it regression?

Yes, It was working last before the app rename happened

### Expected behavior

Well I wish our build didn't suddenly break.

### Actual behavior

It breaks our builds and I suspect anyone's build who used to use VS 8.* on Mac to build their projects.

### Repro steps

on macOS-11 run the following command on a C# project.

`msbuild MySpecial.csproj /t:Build /p:Configuration=Release /p:CreatePackage=true`

al-cheb:
Hey @CartBlanche,
You could pick up VS2019 using snippet below:

```
- name: Set VS2019 
  run: |
        mv "/Applications/Visual Studio.app" "/Applications/Visual Studio 2022.app"
        mv "/Applications/Visual Studio 2019.app" "/Applications/Visual Studio.app"
        /usr/libexec/PlistBuddy -c 'Print CFBundleVersion' '/Applications/Visual Studio.app/Contents/Info.plist'
```

[...]

PR #5701, opened 2022-06-06
[macOS] Add a note about VS 2019 usage

miketimofeev:
# Description
We have added VS for mac 2022 recently keeping VS 2019 on the images as well. However, it is not possible to use VS2019 directly without renaming the app. This PR adds a note on how to do that.

#### Related issue:
https://github.com/actions/virtual-environments/issues/5682#issuecomment-1147306237

## Check list


al-cheb (approved):

dmitry-shibanov (approved):

CartBlanche:
Nice! Thanks for making this happen so quickly. Greatly appreciated.

ddobranic (approved):

PR #5812, opened 2022-06-24
[Ubuntu] Enable systemd lingering

mikhailkoliada:
# Description

`loginctl enable-linger` allows us to enable user services on a machine boot and not on a user login, which in turn allows us having user sockets activated on every machine boot prior to further communication. In practice it sets up some user-specific variables (like the dbus ones)  needed for the `systemctl --user` calls.

I have tested it on:

* mpd
* dirmngr/gpg
* emacs
* snapd (this one is extremely essential as there is snapd in our images and some packages like firefox slowly migrate to the snap-only distribution approach).

More info: https://www.freedesktop.org/software/systemd/man/loginctl.html

## Check list

### DIFF
--- a/images/linux/post-generation/systemd-linger.sh
+++ b/images/linux/post-generation/systemd-linger.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+
+# Enable user session on boot, not on login
+UserId=$(cut -d: -f3 /etc/passwd | tail -1)
+loginctl enable-linger $UserId


images/linux/post-generation/systemd-linger.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+# Enable user session on boot, not on login
+loginctl enable-linger 1001
```
ddobranic, line 4: new line here?

al-cheb, line 4: We should get user id like we do - https://github.com/actions/virtual-environments/blob/main/images/linux/post-generation/environment-variables.sh

igorboskovic3 (dismissed):

ddobranic (dismissed):

al-cheb (approved):

dmitry-shibanov (approved):

ddobranic (approved):

miketimofeev (approved):

PR #5819, opened 2022-06-27
[macOS] Update Clang/LLVM test to use absolute path

al-cheb:
# Description
After https://github.com/Homebrew/homebrew-core/commit/34db45964ce662aba85cfd830c6bd5b11e517147 commit was merged there is no anymore the `/usr/local/opt/llvm` symlink.

Before:
```
lrwxr-xr-x  1 runner  admin  23 Jun 22 16:15 /usr/local/opt/llvm -> ../Cellar/llvm/13.0.1_1
lrwxr-xr-x  1 runner  admin  23 Jun 22 16:15 /usr/local/opt/llvm@13 -> ../Cellar/llvm/13.0.1_1
```

After:
```
lrwxr-xr-x  1 runner  admin  24 Jun 27 02:43 /usr/local/opt/llvm@13 -> ../Cellar/llvm@13/13.0.1

vsphere-clone: llvm@13 is keg-only, which means it was not symlinked into /usr/local,
vsphere-clone: because this is an alternate version of another formula.
```

#### Related issue:
https://github.com/actions/virtual-environments-internal/issues/3917

## Check list

### DIFF
--- a/images/macos/tests/LLVM.Tests.ps1
+++ b/images/macos/tests/LLVM.Tests.ps1
@@ -3,7 +3,7 @@ $os = Get-OSVersion
 Describe "Clang/LLVM" {
     It "Clang/LLVM is installed and version is correct" {
         $toolsetVersion = Get-ToolsetValue 'llvm.version'
-        $clangVersion = & "$(brew --prefix llvm)/bin/clang" --version
+        $clangVersion = & "$(brew --prefix llvm@$toolsetVersion)/bin/clang" --version
         $clangVersion[0] | Should -BeLike "*${toolsetVersion}*"
     }
 }


al-cheb:
cc @MikeMcQuaid, @danielnachun, Could you please review?

al-cheb:
If I try to link the formular I get:

```
Error: Could not symlink lib/libomp.dylib
Target /usr/local/lib/libomp.dylib
is a symlink belonging to libomp. You can unlink it:
  brew unlink libomp

To force the link and overwrite all conflicting files:
  brew link --overwrite llvm@13

To list all files that would be deleted:
  brew link --overwrite --dry-run llvm@13
Linking /usr/local/Cellar/llvm@13/13.0.1... 
```

ddobranic (approved):

danielnachun:
`/usr/local/opt/llvm` is always supposed to exist regardless of whether `llvm` is linked or not.  What was supposed to happen was that `/usr/local/opt/llvm` is should point to LLVM 14 now instead of LLVM 13.  How was `llvm` being installed here?  `brew install llvm` should be enough for that directory to exist - no linking is needed.

al-cheb:
> `/usr/local/opt/llvm` is always supposed to exist regardless of whether `llvm` is linked or not. What was supposed to happen was that `/usr/local/opt/llvm` is should point to LLVM 14 now instead of LLVM 13. How was `llvm` being installed here? `brew install llvm` should be enough for that directory to exist - no linking is needed.

@danielnachun, currently, we install LLVM@13.

danielnachun:
If you're explicitly installing `llvm@13`, then it will only be found through `/usr/local/opt/llvm@13`.  `/usr/local/opt/llvm` will only exist if you've installed `llvm`, and it will point to LLVM 14.  So if anything was previously hardcoded to `/usr/local/opt/llvm`, it will have to be changed to `/usr/local/opt/llvm@13` or LLVM 14 will have to be used instead.

miketimofeev (approved):

MikeMcQuaid (commented):
Makes sense to me 👍🏻

PR #5945, opened 2022-07-22
Add windows/server docker image for Server2022

LiuChangFreeman:
# Description
This pr adds **Windows Server** docker image for Server2022
[**Windows** - contains the full set of Windows APIs and system services (minus server roles).
**Windows Server** - contains the full set of Windows APIs and system services.
**Windows Server Core** - a smaller image that contains a subset of the Windows Server APIs–namely the full .NET framework. It also includes most but not all server roles (for example Fax Server is not included).
**Nano Server** - the smallest Windows Server image and includes support for the .NET Core APIs and some server roles.](https://docs.microsoft.com/en-us/virtualization/windowscontainers/manage-containers/container-base-images)

## Check list

### DIFF
--- a/images/win/toolsets/toolset-2022.json
+++ b/images/win/toolsets/toolset-2022.json
@@ -300,6 +300,7 @@
     },
     "docker": {
         "images": [
+            "mcr.microsoft.com/windows/server:ltsc2022",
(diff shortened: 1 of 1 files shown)


al-cheb:
@LiuChangFreeman, the image size is quite huge we decided not to add it on board.

PR #6140, opened 2022-08-24
[macOS] Install tvOS and watchOS runtime sims

mikhailkoliada:
# Description

We should install these components manually now.

## Check list

### DIFF
--- a/images/macos/provision/core/xcode.ps1
+++ b/images/macos/provision/core/xcode.ps1
@@ -47,6 +47,16 @@ $xcodeVersions | ForEach-Object {
     }
 }
 
+$xcodeVersions | ForEach-Object {
+    if ($_.link.StartsWith("14.")) {
+        Write-Host "Installing Simulator Runtimes..."
+
+        # tvOS and watchOS simulators are not included by default
+        $xcodebuildPath = Get-XcodeToolPath -Version $_.link -ToolName "xcodebuild"
+        Invoke-ValidateCommand "$xcodebuildPath -downloadAllPlatforms"
+    }
+}
+
 Write-Host "Rebuilding Launch Services database ..."
 $xcodeVersions | ForEach-Object {
     Rebuild-XcodeLaunchServicesDb -Version $_.link


images/macos/provision/core/xcode.ps1
```diff
@@ -52,6 +52,13 @@ $xcodeVersions | ForEach-Object {
     Rebuild-XcodeLaunchServicesDb -Version $_.link
 }
 
+# 14.0 and higher doesn't ship tvOS and watchOS, so installing it manually
+$xcodeVersions | Where-Object { $_.link -ge "14.0" } | ForEach-Object {
+    Write-Host "Installing Simulator Runtimes..."
+    Switch-Xcode -Version $_.link
```
maxim-lobanov, line 58: I think we don't really need to switch Xcode here.
`Switch-Xcode` changes `xcodebuild` symlink but you can invoke xcodebuild for specific Xcode using full path.
We already use this approach for some other invocations. See https://github.com/actions/runner-images/blob/6701203c1b625cf03ff525c8758afc0e9075639e/images/macos/helpers/Xcode.Helpers.psm1#L36

And since we don't need to switch Xcode, what do you think about moving `xcodebuild -downloadAllPlatforms` command to the first loop that we use for downloading and installing Xcode?
```powershell
Write-Host "Installing Xcode versions..."
$xcodeVersions | ForEach-Object -ThrottleLimit $threadCount -Parallel {
```

images/macos/provision/core/xcode.ps1
```diff
@@ -52,6 +52,13 @@ $xcodeVersions | ForEach-Object {
     Rebuild-XcodeLaunchServicesDb -Version $_.link
 }
 
+# 14.0 and higher doesn't ship tvOS and watchOS, so installing it manually
+$xcodeVersions | Where-Object { $_.link -ge "14.0" } | ForEach-Object {
```
maxim-lobanov, line 56: I think we shouldn't use `$_.link -ge "14.0" because this way we compare strings instead of versions.
May be `{ $_.link.StartsWith("14.") } like https://github.com/actions/runner-images/blob/7a7fafda99ef4ddb3539fc19596dd1cfc1ab0edc/images/macos/helpers/Xcode.Installer.psm1#L154

images/macos/provision/core/xcode.ps1
mikhailkoliada: @maxim-lobanov done

images/macos/provision/core/xcode.ps1
mikhailkoliada: why not

maxim-lobanov (approved):

al-cheb (approved):

miketimofeev (approved):

Issue #4241, opened 2021-10-08
Errors after using the latest version of images.


gdotdesign:
### Description

Since yesterday (2021-10-07) runs started fail for our project (private).

It uses Ruby and Ruby on Rails, and it fails when running migrations with several memory related errors:

```
malloc(): memory corruption
/home/runner/work/_temp/c04b1d3a-ffc0-43c0-95ea-73a168e17da5.sh: line 1: 16754 Aborted                 (core dumped) bundle exec rake db:create
```

```
free(): invalid pointer
/home/runner/work/_temp/3b451ec2-85a5-43bb-8763-e0baee22b273.sh: line 1:  2497 Aborted                 (core dumped) bundle exec rake db:create
```

We used the `ubuntu-latest` before, and now I tested with `ubuntu-18.04` but with both it fails.

The information for the last working run:

```
Current runner version: '2.283.2'
Operating System
  Ubuntu
  20.04.3
  LTS
Virtual Environment
  Environment: ubuntu-20.04
  Version: 20210929.1
  Included Software: actions/virtual-environments@ubuntu20/20210929.1/images/linux/Ubuntu2004-README.md
  Image Release: actions/virtual-environments@ubuntu20%2F20210929.1 (release)
Virtual Environment Provisioner
  1.0.0.0-master-20210922-1
GITHUB_TOKEN Permissions
  Actions: write
  Checks: write
  Contents: write
  Deployments: write
  Discussions: write
  Issues: write
  Metadata: read
  Packages: write
  Pages: write
  PullRequests: write
  RepositoryProjects: write
  SecurityEvents: write
  Statuses: write
Prepare workflow directory
Prepare all required actions
Getting action download info
Download action repository 'actions/checkout@v2' (SHA:5a4ac9002d0be2fb38bd78e4b4dbde5606d7042f)
Download action repository 'aws-actions/configure-aws-credentials@v1' (SHA:0d9a5be0dceea74e09396820e1e522ba4a110d2f)
Download action repository 'szenius/set-timezone@v1.0' (SHA:dd47655c84241eec2ffa0a855959c16c0920c3c4)
Download action repository 'ruby/setup-ruby@v1' (SHA:5e4f0a10bfc39c97cd5358121291e27e5d97e09b)
Download action repository 'actions/cache@v2' (SHA:c64c572235d810460d0d6876e9c705ad5002b353)
```

### Virtual environments affected

- [ ] Ubuntu 16.04
- [X] Ubuntu 18.04
[...]

PR #6192, opened 2022-09-07
[Windows] Add VCPKG_ROOT variable

erik-bershel:
# Description
Add VCPKG_ROOT variable in additional to VCPKG_INSTALLATION_ROOT + simple style fix

#### Related issue:
[#4241](https://github.com/actions/runner-images-internal/issues/4241)

## Check list

### DIFF
--- a/images/win/scripts/SoftwareReport/SoftwareReport.Common.psm1
+++ b/images/win/scripts/SoftwareReport/SoftwareReport.Common.psm1
@@ -349,6 +349,12 @@ function Build-PackageManagementEnvironmentTable {
             "Value" = $env:VCPKG_INSTALLATION_ROOT
         }
     )
+    $envVariables += @(
+        @{
+            "Name" = "VCPKG_ROOT"
+            "Value" = $env:VCPKG_ROOT
+        }
+    )
     if (Test-IsWin19) {
         $envVariables += @(
             @{
--- a/images/win/scripts/SoftwareReport/SoftwareReport.Common.psm1
+++ b/images/win/scripts/SoftwareReport/SoftwareReport.Common.psm1
@@ -347,9 +347,7 @@ function Build-PackageManagementEnvironmentTable {
         @{
             "Name" = "VCPKG_INSTALLATION_ROOT"
             "Value" = $env:VCPKG_INSTALLATION_ROOT
-        }
-    )
-    $envVariables += @(
+        },
         @{
             "Name" = "VCPKG_ROOT"
             "Value" = $env:VCPKG_ROOT
--- a/images/win/scripts/Installers/Install-Vcpkg.ps1
+++ b/images/win/scripts/Installers/Install-Vcpkg.ps1
@@ -17,5 +17,6 @@ Invoke-Expression "$InstallDir\$VcpkgExecPath integrate install"
 Add-MachinePathItem $InstallDir
 $env:Path = Get-MachinePath
 setx VCPKG_INSTALLATION_ROOT $InstallDir /M
+setx VCPKG_ROOT $InstallDir /M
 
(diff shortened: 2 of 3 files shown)


images/win/scripts/SoftwareReport/SoftwareReport.Common.psm1
```diff
@@ -349,6 +349,12 @@ function Build-PackageManagementEnvironmentTable {
             "Value" = $env:VCPKG_INSTALLATION_ROOT
         }
     )
+    $envVariables += @(
+        @{
+            "Name" = "VCPKG_ROOT"
+            "Value" = $env:VCPKG_ROOT
+        }
+    )
```
miketimofeev, line 357: we can add this ennv variable in the section above, no need to update the #envVariables

erik-bershel: In general, yes, it's redundant.

al-cheb (approved):

miketimofeev (approved):

nightlark:
This change broke several of our CI builds on GitHub Actions and Azure Pipelines (we use the presence of VCPKG_ROOT being set to indicate that a user wants to opt-in to building dependencies with vcpkg rather than ones installed on the system). Some form of announcement about this change would have been nice.

- Windows builds on Azure Pipelines started timing out: installing the Boost packages takes an excessively long time, given we only use the headers (last I checked there wasn't a vcpkg for grabbing just the headers and not compile anything), adding about 16 minutes to the overall build time
- MSYS2 builds for our MINGW package started failing (conflict between the vcpkg version of Boost and the mingw copy? not sure)

The solution has been to liberally add `unset VCPKG_ROOT` to our CI workflows and a new option for our CMake builds that forces it to not use vcpkg even if VCPKG_ROOT is set.

ddobranic:
@nightlark thank you for the feedback and workaround! we will investigate the issue.

miketimofeev:
> This change broke several of our CI builds on GitHub Actions and Azure Pipelines (we use the presence of VCPKG_ROOT being set to indicate that a user wants to opt-in to building dependencies with vcpkg rather than ones installed on the system). Some form of announcement about this change would have been nice.
> 
> * Windows builds on Azure Pipelines started timing out: installing the Boost packages takes an excessively long time, given we only use the headers (last I checked there wasn't a vcpkg for grabbing just the headers and not compile anything), adding about 16 minutes to the overall build time
> * MSYS2 builds for our MINGW package started failing (conflict between the vcpkg version of Boost and the mingw copy? not sure)
> 
> To get things working again our solution has been to liberally add `unset VCPKG_ROOT` to our CI workflows and a new option for our CMake builds that forces it to not use vcpkg even if VCPKG_ROOT is set.

@michaelbprice could you please comment on this? It seems it's not so safe to add that variable

michaelbprice:
I've left a more detailed comment over at https://github.com/actions/runner-images/pull/6196#issuecomment-1256468363.

We'll see how this shakes out, but it seems to me that if tools expect to use `VCPKG_ROOT` as an indication of where vcpkg is installed, it should not also serve the purpose of an opt-in, unless that opt-in is strictly predicated on the tool being installed in such an identifiable way.

My recommendation would be to change that opt-in behavior to depend on an environment variable more clearly within the control of the users' environment, by doing something like prepending an org identifier to the environment variable, such as `MY_AWESOME_PROJECT_USING_VCPKG_ROOT`.

michaelbprice:
> @michaelbprice could you please comment on this? It seems it's not so safe to add that variable

In the sense that it's never safe to add any environment variable, as you cannot predict if someone downstream might already be using it for some other purpose.

PR #6210, opened 2022-09-09
Add sccache

leighmcculloch:
# Description

Add the sccache tool to all images.

sccache is about 12MB once installed on my own system, but it depends on the operating system. It takes about 28 seconds to install.

#### Related issue:

Close https://github.com/actions/runner-images/issues/6209

## Check list

### DIFF
--- a/images/linux/Ubuntu1804-Readme.md
+++ b/images/linux/Ubuntu1804-Readme.md
@@ -149,6 +149,7 @@
 - Rust 1.62.1
 - Rustdoc 1.62.1
 - Rustup 1.25.1
+- Sccache 0.3.0
 
 #### Packages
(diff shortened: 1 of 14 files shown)


images/linux/Ubuntu1804-Readme.md
```diff
@@ -149,6 +149,7 @@
 - Rust 1.62.1
 - Rustdoc 1.62.1
 - Rustup 1.25.1
+- Sccache 0.3.0
```
leighmcculloch, line 152: I was not sure where sccache should be mentioned in the docs. It is popular to use in the Rust ecosystem as a rustc wrapper, but it also works with other compilers too so is relevant to more ecosystems than just Rust.

PR #6255, opened 2022-09-16
[macOS] Add Xcode 14.0.1 to macOS 12

miketimofeev:
# Description
Xcode 14.0.1 has been released https://developer.apple.com/news/releases/?id=09162022a

#### Related issue:
n/a

## Check list


ddobranic (approved):

EwoutH:
Should the old patch version (`14.0.0` in this case) be removed once a new patch version is available? That would keep the CI versions a bit smaller, and it looks like that's done with most other tools as well.

miketimofeev:
@EwoutH we usually deprecate such version in 3 months, thanks!

Issue #6216, opened 2022-09-11
Add colima


mvgijssel:
### Tool name

Colima

### Tool license

MIT license

### Add or update?

- [X] Add
- [ ] Update

### Desired version

0.4.4

### Approximate size

17mb

### Brief description of tool

Colima - container runtimes on macOS (and Linux) with minimal setup. Means you can run `docker` without having to install Docker for macOS and run into licensing issues (https://github.com/actions/runner-images/issues/2150). 

### URL for tool's homepage

https://github.com/abiosoft/colima

### Provide a basic test case to validate the tool's functionality.

```bash
❯ colima version
colima version 0.4.4
git commit: 8bb1101a861a8b6d2ef6e16aca97a835f65c4f8f
```
```


### Platforms where you need the tool

- [ ] Azure DevOps
- [ ] GitHub Actions

### Virtual environments where you need the tool

- [ ] Ubuntu 18.04
- [ ] Ubuntu 20.04
- [ ] Ubuntu 22.04
- [X] macOS 10.15
- [X] macOS 11
- [X] macOS 12
- [ ] Windows Server 2019
- [ ] Windows Server 2022

### Can this tool be installed during the build?

```bash
brew install docker colima
```

### Tool installation time in runtime

1 minute

### Are you willing to submit a PR?

Sure!

mikhailkoliada:
@mvgijssel we will decide on adding it and get back to you.

ddobranic:
hi @mvgijssel! thank you for your proposal. we will add colima to our macOS images.

mvgijssel:
That’s awesome! Let me know if you need any help!

ddobranic:
hello @mvgijssel! colima has been added to macOS images.

Jasperav:
@mvgijssel @ddobranic does any of you know why this is removed from macos 13?

mvgijssel:
No idea!

ejsmith:
Yeah, wondering the same thing. Why was it removed?

asottile-sentry:
it's also missing on the new macos-14 image as well

AkihiroSuda:
- https://github.com/cncf/toc/issues/1348#issuecomment-2578607870

We, the maintainers of [Lima](https://lima-vm.io) (the tool under the hood of Colima), are trying to promote Lima from Cloud Native Computing Foundation (CNCF) Sandbox to Incubating.

We need several project adopters to be interviewed by the Technical Oversight Committee of CNCF.
[...]

PR #6285, opened 2022-09-22
[macOS] Add Colima tool

bogdan-damian-bgd:
# Description

Colima tool added to macOS

#### Related issue:

#6216 

## Virtual environments affected
- [ ] macOS 10.15
- [x] macOS 11
- [x] macOS 12

## Check list

### DIFF
--- a/images/macos/provision/core/colima.sh
+++ b/images/macos/provision/core/colima.sh
@@ -1,0 +1,6 @@
+#!/bin/bash -e -o pipefail
+
+echo Installing Colima...
+brew install colima
+
+invoke_tests "Common" "Colima"
--- a/images/macos/provision/core/colima.sh
+++ b/images/macos/provision/core/colima.sh
@@ -1,6 +1,0 @@
-#!/bin/bash -e -o pipefail
-
-echo Installing Colima...
-brew install colima
-
-invoke_tests "Common" "Colima"
--- a/images/macos/software-report/SoftwareReport.Generator.ps1
+++ b/images/macos/software-report/SoftwareReport.Generator.ps1
@@ -182,7 +182,8 @@ $toolsList += @(
     (Get-GHCVersion),
     (Get-CabalVersion),
     (Get-StackVersion),
-    (Get-SwiftFormatVersion)
+    (Get-SwiftFormatVersion),
+    (Get-ColimaVersion)
 )
 
 $markdown += New-MDList -Style Unordered -Lines ($toolsList | Sort-Object)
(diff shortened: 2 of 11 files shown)


images/macos/provision/core/colima.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash -e -o pipefail
```
al-cheb, line 1: Remove this file and use brew section in toolset file - https://github.com/actions/runner-images/blob/main/images/macos/toolsets/toolset-12.json

images/macos/templates/macOS-10.15.json
```diff
@@ -203,7 +203,8 @@
                 "./provision/core/pypy.sh",
                 "./provision/core/pipx-packages.sh",
                 "./provision/core/bicep.sh",
-                "./provision/core/graalvm.sh"
+                "./provision/core/graalvm.sh",
+                "./provision/core/colima.sh"
```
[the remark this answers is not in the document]
bogdan-damian-bgd: Yes, I removed it for the 10.15 version

images/macos/provision/core/colima.sh
bogdan-damian-bgd: I switched to toolset json

al-cheb (approved):

images/macos/software-report/SoftwareReport.Generator.ps1
```diff
@@ -182,7 +182,8 @@ $toolsList += @(
     (Get-GHCVersion),
     (Get-CabalVersion),
     (Get-StackVersion),
-    (Get-SwiftFormatVersion)
+    (Get-SwiftFormatVersion),
+    (Get-ColimaVersion)
```
miketimofeev, line 186: @bogdan-damian-bgd should it be in the runtime section instead as Colima - container **runtimes** on macOS (and Linux) with minimal setup?

bogdan-damian-bgd: Yes, directly from a naming point of view it should be in 'Language and Runtimes'.

On the other hand, if we consider it as some kind of replacement for Docker (which is also labeled as a container **runtime**) and try to be consistent with the location of Docker within the ReadMe sections for Windows and Linux, then it also makes some sense to stay in this section as well?

miketimofeev: Yup, if we'd like to be consistent then it's fine

igorboskovic3 (approved):

PR #6295, opened 2022-09-23
Ensure VS MEF is also initialized in the VS experimental

kzu:
MEF initialization is very costly and can take up to 15' on first run of VS.

MEF is only initialized when some action in the IDE requires resolving a MEF  export in VS. The heaviest consumer of MEF in VS is the core editor itself, and  the existing initialization code was not causing the editor to open at all, leaving  MEF uninitialized. 

By opening even a (mostly) empty C# file, we cause most key components in 
MEF to be properly initialized once: editor and project system.

# Description

Performance issue bug fixing for VS experimental (for VS extension authoring 
integration testing).


Moorning (approved):

IcebergVerse (approved):
kzu:patch-1

mikhailkoliada:
Hello! Current VS configuration scripts are too fragile, we would not like to extend initialization commands as it may break lots of customers and affect them badly.

PR #6305, opened 2022-09-26
[Ubuntu] parse java semver from API

mikhailkoliada:
# Description

As we are parsing semver from the API server on macOS and Windows it is the easiest way to do so on Linux, too

## Check list

### DIFF
--- a/images/linux/scripts/installers/java-tools.sh
+++ b/images/linux/scripts/installers/java-tools.sh
@@ -67,24 +67,20 @@ installOpenJDK() {
     if [[ ${VENDOR_NAME} == "Temurin-Hotspot" ]]; then
         apt-get -y install temurin-${JAVA_VERSION}-jdk=\*
         javaVersionPath="/usr/lib/jvm/temurin-${JAVA_VERSION}-jdk-amd64"
+        assetUrl=$(curl -s "https://api.adoptium.net/v3/assets/latest/${JAVA_VERSION}/hotspot")
     elif [[ ${VENDOR_NAME} == "Adopt" ]]; then
         apt-get -y install adoptopenjdk-${JAVA_VERSION}-hotspot=\*
         javaVersionPath="/usr/lib/jvm/adoptopenjdk-${JAVA_VERSION}-hotspot-amd64"
+        assetUrl=$(curl -s "https://api.adoptopenjdk.net/v3/assets/latest/${JAVA_VERSION}/hotspot")
     else
         echo "${VENDOR_NAME} is invalid, valid names are: Temurin-Hotspot and Adopt"
         exit 1
     fi
 
+    asset=$(echo ${assetUrl} | jq -r '.[] | select(.binary.os=="linux" and .binary.image_type=="jdk" and .binary.architecture=="x64")')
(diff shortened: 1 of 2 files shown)


images/linux/scripts/installers/java-tools.sh
```diff
@@ -67,24 +67,20 @@ installOpenJDK() {
...
+        assetUrl=$(curl -s "https://api.adoptopenjdk.net/v3/assets/latest/${JAVA_VERSION}/hotspot")
     else
         echo "${VENDOR_NAME} is invalid, valid names are: Temurin-Hotspot and Adopt"
         exit 1
     fi
 
-    JAVA_TOOLCACHE_PATH="${AGENT_TOOLSDIRECTORY}/Java_${VENDOR_NAME}_jdk"
-
-    fullJavaVersion=$(cat "${javaVersionPath}/release" | grep "^SEMANTIC" | cut -d "=" -f 2 | tr -d "\"" | tr "+" "-")
+    asset=$(echo ${assetUrl} | jq -r '.[] | select(.binary.os=="linux" and .binary.image_type=="jdk" and .binary.architecture=="x64")')
```
al-cheb, line 80: How can we be sure that version obtained from the link is the same we install via apt ?

mikhailkoliada: @al-cheb I think we can introduce the comparison between `java -version` and the api

➜ echo $asset | jq -r '.version.openjdk_version'
17.0.4.1+1

images/linux/scripts/installers/java-tools.sh
```diff
@@ -67,24 +67,20 @@ installOpenJDK() {
     if [[ ${VENDOR_NAME} == "Temurin-Hotspot" ]]; then
         apt-get -y install temurin-${JAVA_VERSION}-jdk=\*
         javaVersionPath="/usr/lib/jvm/temurin-${JAVA_VERSION}-jdk-amd64"
+        assetUrl=$(curl -s "https://api.adoptium.net/v3/assets/latest/${JAVA_VERSION}/hotspot")
```
Smithbull001, line 70: Well done

images/linux/scripts/installers/java-tools.sh
```diff
@@ -67,24 +67,20 @@ installOpenJDK() {
...
     elif [[ ${VENDOR_NAME} == "Adopt" ]]; then
         apt-get -y install adoptopenjdk-${JAVA_VERSION}-hotspot=\*
         javaVersionPath="/usr/lib/jvm/adoptopenjdk-${JAVA_VERSION}-hotspot-amd64"
+        assetUrl=$(curl -s "https://api.adoptopenjdk.net/v3/assets/latest/${JAVA_VERSION}/hotspot")
     else
         echo "${VENDOR_NAME} is invalid, valid names are: Temurin-Hotspot and Adopt"
         exit 1
     fi
```
Smithbull001, line 79: Yes

images/linux/scripts/installers/java-tools.sh
Smithbull001: Awesome

PR #6332, opened 2022-09-30
[Windows] Test Chrome and Chrome Driver major versions are the same

miketimofeev:
# Description
Sometimes we have a situation when Chrome is already updated to a new major version but chromedriver isn't.
We need to fail a build in such cases rather than find it out only during the deployment.
This PR adds a Pester test that will fail the build in case of a version mismatch.

#### Related issue:
https://github.com/actions/runner-images-internal/issues/4392

## Check list

### DIFF
--- a/images/win/scripts/Tests/Browsers.Tests.ps1
+++ b/images/win/scripts/Tests/Browsers.Tests.ps1
@@ -49,7 +49,7 @@ Describe "Chrome" {
         }
 
         It "Chrome and Chrome Driver major versions are the same" {
-            $chromeMajor = (Get-Item $chromePath).VersionInfo.ProductVersion.Split(".")[0]
+            $chromeMajor = (Get-Item $chromePath).VersionInfo.ProductMajorPart
             $chromeDriverMajor = (Get-Content $env:ChromeWebDriver\versioninfo.txt).Split(".")[0]
             $chromeMajor | Should -BeExactly $chromeDriverMajor
         }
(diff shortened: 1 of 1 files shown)


ddobranic (approved):

images/win/scripts/Tests/Browsers.Tests.ps1
```diff
@@ -47,6 +47,12 @@ Describe "Chrome" {
             $chromePath | Should -Exist
             $chromeName | Should -BeExactly "chrome.exe"
         }
+
+        It "Chrome and Chrome Driver major versions are the same" {
+            $chromeMajor = (Get-Item $chromePath).VersionInfo.ProductVersion.Split(".")[0]
+            $chromeDriverMajor = (& $env:ChromeWebDriver\chromedriver.exe --version).Trim("ChromeDriver ").Split(".")[0]
```
al-cheb, line 53: nitpick: $env:ChromeWebDriver\versioninfo.txt

igorboskovic3 (approved):

images/win/scripts/Tests/Browsers.Tests.ps1
```diff
@@ -47,6 +47,12 @@ Describe "Chrome" {
             $chromePath | Should -Exist
             $chromeName | Should -BeExactly "chrome.exe"
         }
+
+        It "Chrome and Chrome Driver major versions are the same" {
+            $chromeMajor = (Get-Item $chromePath).VersionInfo.ProductVersion.Split(".")[0]
```
al-cheb, line 52: nitpick: .VersionInfo.ProductMajorPart

al-cheb (approved):

erik-bershel (dismissed):

mikhailkoliada (approved):

ddobranic (approved):

PR #6592, opened 2022-11-18
Update PowerShell Az Modules to 9.0.1

olandese:
# Description
Update installed Powershell Az Module, from 7.5.0 to 9.0.1. 
The 7.5.0 is seven months old, and version 9.x is out (1 month ago).
At the moment 9.1.1 is available, but because it's less than 1 day old (at the moment of writing) I prefer to pin to 9.0.1 which has been released one month ago.

### DIFF
--- a/images/linux/toolsets/toolset-2004.json
+++ b/images/linux/toolsets/toolset-2004.json
@@ -121,13 +121,15 @@
             "name": "az",
             "url" : "https://raw.githubusercontent.com/Azure/az-ps-module-versions/main/versions-manifest.json",
             "versions": [
-                "7.5.0"
+                "9.0.1"
             ],
             "zip_versions": [
                 "3.1.0",
                 "4.4.0",
                 "5.9.0",
-                "6.6.0"
+                "6.6.0",
+                "7.5.0"
+
             ]
         }
     ],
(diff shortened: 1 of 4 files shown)


LockTar (approved):
Looks good!

erik-bershel:
@olandese main idea is correct, but devil in the details. Please wait a bit I'll come back with instructions if you want to continue work on that.

olandese:
@erik-bershel glad to help! :)

erik-bershel:
@olandese so.
I suggest doing it all at once:
1) It is necessary to split changes for different OS into different pull requests.
2) You should not move the current version to the packaged version until there is a separate request from the community for this - we try to save space on virtual machines as much as possible. (main issue in the end)
3) It is highly undesirable to remove required fields from pull requests that were automatically suggested.
4) It is advisable to keep pull request names consistent with current practice.

I believe the best way would be to close this one and make two new pull requests with edits in mind. Like these: https://github.com/actions/runner-images/pull/6551, https://github.com/actions/runner-images/pull/6536, https://github.com/actions/runner-images/pull/6481
I understand that this looks like nitpicking. If you don't want to mess with it, then I can issue new pull requests myself.

olandese:
No problem, I will do that, 
You are absolutely not nitpicking, every repo has its own "housekeeping" rules :)

PR #6630, opened 2022-11-25
Update README for ubuntu-latest description

Connie18:
# Description

It is described that `Ubuntu 20.04` is used for `ubuntu-latest` YAML Label, but `Ubuntu 22.04` is now generally available according to the following blog.
https://github.blog/changelog/2022-08-09-github-actions-ubuntu-22-04-is-now-generally-available-on-github-hosted-runners/

I misunderstood by old `README`, so I created this PR.

## Check list


panekj:
Blog post says:
> Otherwise, our recommendation is to use `ubuntu-latest`, which currently utilizes Ubuntu 20.04

Issue #6399, opened 2022-10-13
Ubuntu-latest workflows will use Ubuntu-22.04


ddobranic:
### Breaking changes

Ubuntu 22.04 is ready to be the default version for the "ubuntu-latest" label in GitHub Actions and Azure DevOps.

### Target date

This change will be rolled out over a period of several weeks beginning in October, 3. We plan to complete the migration by December, 15th, 2022.

### The motivation for the changes

GitHub Actions and Azure DevOps have supported Ubuntu 22.04 in preview mode since May 2022, and starting from August 2022 Ubuntu 22.04 is generally available for all customers. We have monitored customer feedback to improve the Ubuntu 22.04 image stability and now we are ready to set it as the latest.

### Possible impact

The Ubuntu 20.04 image has a different set of software than Ubuntu 22.04. The most significant changes are listed in the table below:

|Tool name|Ubuntu 20.04|Ubuntu 22.04|Notes|
|-|:-:|-|-|
|Ruby toolcache | 2.5.9 <br>2.6.10 <br>2.7.6 <br>3.0.4 <br>3.1.2 | 3.* (latest) | Only latest available version will be cached on start.|
|Ruby | 2.7.0 | 3.* (latest) | The latest available version for each operating system provided in the APT repository is installed.|
|Clang <br>Clang-format <br>Clang-tidy | 10.0.0 <br>11.0.0 <br>12.0.0 | 12.* <br>13.* <br>14.* | Three latest available versions for each operating system provided in the APT repository are installed.|
|GCC <br>GFortran | 9.4.0 <br>10.3.0<br>- | 9.* <br>10.* <br>11.* | Three latest available versions for each operating system provided in the APT repository are installed.|
|Python | 3.8.10 | 3.10.6 | Fresh updates for Python and tools available for Ubuntu 22 in APT repository.|
|PHP | 7.4.32 <br>8.0.24 <br>8.1.11 | 8.* (latest) | Only one latest officially supported PHP version will be supported. See more [here]( https://github.com/actions/runner-images/issues/6331).|
|.NET Core SDK | 3.1.120 <br>3.1.202 <br>3.1.302 <br>3.1.423 <br>5.0.104 <br>5.0.214 <br>5.0.303 <br>5.0.408 <br>6.0.401 | 6.0.401 | Only the latest available version will be installed on start.|
[...]

PR #6641, opened 2022-11-27
[ubuntu] update latest to ubuntu-22.04

arol:
# Description
Documentation edit to reflect the changes on the ubuntu runner, now `latest` points to `22.04` instead of `20.04`.

#### Related issue:
#6399 

## Check list


mikhailkoliada:
It is not yet done, once it will, we update the documentation.

PR #6655, opened 2022-11-29
Add workflows to create SBOM and upload it to release asset

erik-bershel:
# Description
Workflow for generating SNOM on published images of runners. The result will be uploaded as an attachment to the corresponding release.

#### Related issue:
[#4405](https://github.com/actions/runner-images-internal/issues/4405)

## Check list

### DIFF
--- a/.github/workflows/create_sbom_report.yml
+++ b/.github/workflows/create_sbom_report.yml
@@ -1,0 +1,119 @@
+name: Create and upload a SBOM to release assets
+# Inherited variables:
+# github.event.client_payload.imageLabel - AzDO image/OS label
+# github.event.client_payload.GHreleaseID - Current release ID
+# github.event.client_payload.imageVersion - AzDO image/OS version "major.minor"
+# 
+# Current SYFT tool issues:
+# macOS (minor): very long cataloging process (https://github.com/anchore/syft/issues/1328), 
+# macOS (major): promt privilegies that blocking process indefinetely (https://github.com/anchore/syft/issues/1367)
+# Windows (major): process fails on some symlinks (https://github.com/anchore/syft/issues/1094)
+on:
+  repository_dispatch:
+    types: [generate-sbom]
+jobs:
+  #Converting image OS variable for the next steps
+  initialize: 
+    strategy:
+      matrix: #Couples AzDO label <-> GHA image spec
+        include:
+          - label: ubuntu22
+            os: ubuntu-22.04
+          - label: ubuntu20
+            os: ubuntu-20.04
+          - label: ubuntu18
+            os: ubuntu-18.04
+          - label: macOS-12
+            os: macos-12
+          - label: macOS-11
+            os: macos-11
+          - label: macOS-10.15
+            os: macos-10.15
+          - label: win22
+            os: windows-2022
+          - label: win19
+            os: windows-2019
+    runs-on: ubuntu-latest
+    steps:
+      - name: Convert image label variable
+        run: if ("${{ github.event.client_payload.imageLabel }}"" -eq "{{ $matrix.label }}"") {echo "current-os=${{ $matrix.os }}" >> $GITHUB_OUTPUT}
+  #Checking image version on available runner
+  version-check:
+    strategy:
+      matrix:
+        include:
+          - os: Windows
+            path: C:\
+          - os: macOS
+            path: /Users/runner/
+          - os: Linux
+            path: /imagegeneration/
+    defaults:
+      run:
+        shell: pwsh
+    needs: initialize
+    runs-on: ${{ needs.initialize.output.current-os }}
+    steps:
+      - name: Available image version check
+        run: |
+          if ("${{ runner.os }}" -eq "${{ matrix.os }}") {
+            if ((Get-Content ${{ matrix.path }}imagedata.json | jq '.[1] | .detail') -match '\d{8}\.+\d') { 
+              if ($Matches[0] -eq ${{ github.event.client_payload.imageLabel }}) {
+                echo "Current runner use ${{ github.event.client_payload.imageLabel }} image version."
+              } else {
+                echo "Error. Current runner image version don't match ${{ github.event.client_payload.imageLabel }} image version."
+                exit 1
+              }
+            } else {
+              echo "Cannot parse image version."
+              exit 1
+            }
+          }
+  #Install and run SYFT, compress SBOM, upload it to release assets
+  create-sbom:
+    strategy:
+      matrix:
+        include:
+          - os: Windows
+            path: D:/syft
+          - os: Linux
+            path: /usr/local/bin
+          - os: macOS
+            path: /usr/local/bin
+    defaults:
+      run:
+        shell: pwsh
+    needs: [initialize, version-check]
+    runs-on: ${{ initialize.output.current-os }}
+    steps:
+      - name: Install SYFT tool
+        run: |
+          echo "Trying to install SYFT tool from oficial sources"
+          if ("${{ runner.os }}" -eq "${{ matrix.os }}") {
+            echo "Installing SYFT for ${{ matrix.os }}"
+            curl -sSfL https://raw.githubusercontent.com/anchore/syft/main/install.sh | sh -s -- -b ${{ matrix.path }}
+          }
+      - name: Prepare SBOM report
+        run: |
+          if ("${{ runner.os }}" -eq "Windows") {
+            echo "Runnig SYFT tool on Windows on disk C:/"
+            D:/syft/syft dir:C:/ -vv -o spdx-json=SBOM.${{ github.event.client_payload.imageLabel }}.json
+          } elseif ("${{ runner.os }}" -eq "macOS") {
+              echo "Runnig SYFT tool on macOS with exludes"
+              syft dir:/ -vv -o spdx-json=SBOM.${{ github.event.client_payload.imageLabel }}.json --exclude ./Users --exclude ./System/Volumes --exclude ./private
+          } else {
+                echo "Runnig SYFT tool on Ubuntu with default settings"
+                syft dir:/ -vv -o spdx-json=SBOM.${{ github.event.client_payload.imageLabel }}.json
+          }
+          Compress-Archive sbom.${{ github.event.client_payload.imageLabel }}.json sbom.${{ github.event.client_payload.imageLabel }}.json.zip
+          echo "SBOM created successfully!"
+      #Might be changed to softprops/action-gh-release after additional check
+      - name: Upload release asset
+        uses: actions/upload-release-asset@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          upload_url: "https://uploads.github.com/repos/actions/runner-images/releases/${{ github.event.client_payload.GHreleaseID }}/assets{?name,label}"
+          asset_path: ./SBOM.${{ github.event.client_payload.imageLabel }}.json.zip
+          asset_name: SBOM.${{ github.event.client_payload.imageLabel }}.json.zip
+          asset_content_type: application/zip
--- a/.github/workflows/create_sbom_report.yml
+++ b/.github/workflows/create_sbom_report.yml
@@ -1,112 +1,81 @@
 name: Create and upload a SBOM to release assets
 # Inherited variables:
-# github.event.client_payload.imageLabel - AzDO image/OS label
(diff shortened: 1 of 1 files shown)


.github/workflows/create_sbom_report.yml
```diff
@@ -0,0 +1,119 @@
...
+          - label: macOS-10.15
+            os: macos-10.15
+          - label: win22
+            os: windows-2022
+          - label: win19
+            os: windows-2019
+    runs-on: ubuntu-latest
+    steps:
+      - name: Convert image label variable
+        run: if ("${{ github.event.client_payload.imageLabel }}"" -eq "{{ $matrix.label }}"") {echo "current-os=${{ $matrix.os }}" >> $GITHUB_OUTPUT}
```
maxim-lobanov, line 39: I am not sure that this line is correct.
I see 3 problems here:
1) `}}""` - typo with double quotes twice
2) `{{ $matrix.label }}` - we shouldn't use $ sign here: `{{ matrix.label }}`
3) Looks like we mixed two different languages here: bash and powershell:
Based on using `-eq` I think you want to use PowerShell here but:
- `shell` is not defined and bash is used by default
- we should use just `"current-os=${{ $matrix.os }}" >> $GITHUB_OUTPUT` instead of `echo "current-os=${{ $matrix.os }}" >> $GITHUB_OUTPUT` in powershell. Echo is not needed in powershell
- `>> $GITHUB_OUTPUT` won't work in PowerShell because we should use `>> $env:GITHUB_OUTPUT` (`$env:` to access environment variables) or even better `Out-File -Append -FilePath $env:GITHUB_OUTPUT`

.github/workflows/create_sbom_report.yml
```diff
@@ -0,0 +1,119 @@
...
+        include:
+          - os: Windows
+            path: C:\
+          - os: macOS
+            path: /Users/runner/
+          - os: Linux
+            path: /imagegeneration/
+    defaults:
+      run:
+        shell: pwsh
```
maxim-lobanov, line 53: nit: define it on workflow level to avoid code duplication in every job.

.github/workflows/create_sbom_report.yml
```diff
@@ -0,0 +1,119 @@
...
+    needs: initialize
+    runs-on: ${{ needs.initialize.output.current-os }}
```
maxim-lobanov, line 55: In current implementation you can't use this variable. Syntax `echo "var=value" >>$GITHUB_OUTPUT` sets variable on step level by default and it is not exposed to different jobs.
If you want to consume variable from different job you should define it as job output. See this example: https://docs.github.com/en/actions/using-jobs/defining-outputs-for-jobs

.github/workflows/create_sbom_report.yml
```diff
@@ -0,0 +1,119 @@
...
+                echo "Current runner use ${{ github.event.client_payload.imageLabel }} image version."
+              } else {
+                echo "Error. Current runner image version don't match ${{ github.event.client_payload.imageLabel }} image version."
+                exit 1
+              }
+            } else {
+              echo "Cannot parse image version."
+              exit 1
+            }
+          }
```
maxim-lobanov, line 71: nit: swap if branches to avoid nested code and improve readability:

```suggestion
          if ("${{ runner.os }}" -ne "${{ matrix.os }}") {
            exit 0;
          }
          
          if (-not ((Get-Content ${{ matrix.path }}imagedata.json | jq '.[1] | .detail') -match '\d{8}\.+\d')) { 
              echo "Cannot parse image version."
              exit 1
          }
          
          if ($Matches[0] -eq ${{ github.event.client_payload.imageLabel }}) {
                echo "Current runner use ${{ github.event.client_payload.imageLabel }} image version."
          } else {
                echo "Error. Current runner image version don't match ${{ github.event.client_payload.imageLabel }} image version."
                exit 1
          }
```

.github/workflows/create_sbom_report.yml
```diff
@@ -0,0 +1,119 @@
...
+    defaults:
+      run:
+        shell: pwsh
+    needs: initialize
+    runs-on: ${{ needs.initialize.output.current-os }}
+    steps:
+      - name: Available image version check
+        run: |
+          if ("${{ runner.os }}" -eq "${{ matrix.os }}") {
+            if ((Get-Content ${{ matrix.path }}imagedata.json | jq '.[1] | .detail') -match '\d{8}\.+\d') {
```
maxim-lobanov, line 60: We don't really need to parse the file to get image version. You can take it from environment variable:
```pwsh
$env:ImageVersion
```

This variable is defined on every image

.github/workflows/create_sbom_report.yml
```diff
@@ -0,0 +1,119 @@
...
+          } elseif ("${{ runner.os }}" -eq "macOS") {
+              echo "Runnig SYFT tool on macOS with exludes"
+              syft dir:/ -vv -o spdx-json=SBOM.${{ github.event.client_payload.imageLabel }}.json --exclude ./Users --exclude ./System/Volumes --exclude ./private
+          } else {
+                echo "Runnig SYFT tool on Ubuntu with default settings"
+                syft dir:/ -vv -o spdx-json=SBOM.${{ github.event.client_payload.imageLabel }}.json
+          }
+          Compress-Archive sbom.${{ github.event.client_payload.imageLabel }}.json sbom.${{ github.event.client_payload.imageLabel }}.json.zip
+          echo "SBOM created successfully!"
+      #Might be changed to softprops/action-gh-release after additional check
```
maxim-lobanov, line 110: Looks like we don't have common logic between images so may be split it to 3 different steps?
```yaml
steps:
 - name: Prepare SBOM Windows
    if: ${{ runner.os == 'Windows' }}"
    run: |
        echo "Runnig SYFT tool on Windows on disk C:/"
        D:/syft/syft dir:C:/ -vv -o spdx-json=SBOM.${{ github.event.client_payload.imageLabel }}.json
 - name: Prepare SBOM macOS
    if: ${{ runner.os == 'macOS' }}"
    run: |
        echo "Runnig SYFT tool on macOS with exludes"
        syft dir:/ -vv -o spdx-json=SBOM.${{ github.event.client_payload.imageLabel }}.json --exclude ./Users --exclude ./System/Volumes --exclude ./private

```

.github/workflows/create_sbom_report.yml
```diff
@@ -0,0 +1,119 @@
...
+      - name: Upload release asset
+        uses: actions/upload-release-asset@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          upload_url: "https://uploads.github.com/repos/actions/runner-images/releases/${{ github.event.client_payload.GHreleaseID }}/assets{?name,label}"
+          asset_path: ./SBOM.${{ github.event.client_payload.imageLabel }}.json.zip
+          asset_name: SBOM.${{ github.event.client_payload.imageLabel }}.json.zip
```
maxim-lobanov, line 118: Just for curiosity - why do we need to include imageLabel to file name?

.github/workflows/create_sbom_report.yml
```diff
@@ -0,0 +1,119 @@
...
+# github.event.client_payload.imageLabel - AzDO image/OS label
+# github.event.client_payload.GHreleaseID - Current release ID
+# github.event.client_payload.imageVersion - AzDO image/OS version "major.minor"
+# 
+# Current SYFT tool issues:
+# macOS (minor): very long cataloging process (https://github.com/anchore/syft/issues/1328), 
+# macOS (major): promt privilegies that blocking process indefinetely (https://github.com/anchore/syft/issues/1367)
+# Windows (major): process fails on some symlinks (https://github.com/anchore/syft/issues/1094)
+on:
+  repository_dispatch:
```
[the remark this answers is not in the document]
erik-bershel: After further discussion, we came to the following conclusion:
Postpone changing the workflow invocation type due to required changes on the caller side.
@mikhailkoliada topic to discuss about 2nd half of Dec.

.github/workflows/create_sbom_report.yml
maxim-lobanov: Discussed offline.

erik-bershel: After further discussion, we came to the following conclusion:
It is desirable to have unique report names for current versions of operating systems. Assumed applicability scenario: the user wants to compare the reports of two different operating systems, the downloaded files will have a visual difference, helping to determine "where to look."
On reflection, there is one interesting situation where it would be necessary to add another token with the current image version to the name.

.github/workflows/create_sbom_report.yml
```diff
@@ -0,0 +1,119 @@
...
+          - label: macOS-12
+            os: macos-12
+          - label: macOS-11
+            os: macos-11
+          - label: macOS-10.15
+            os: macos-10.15
+          - label: win22
+            os: windows-2022
+          - label: win19
+            os: windows-2019
```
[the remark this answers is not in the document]
erik-bershel: Accepted and corrected.

.github/workflows/create_sbom_report.yml
erik-bershel: Accepted and corrected.

.github/workflows/create_sbom_report.yml
erik-bershel: Accepted and corrected.

.github/workflows/create_sbom_report.yml
erik-bershel: Deleted part

.github/workflows/create_sbom_report.yml
erik-bershel: Accepted and corrected.

.github/workflows/create_sbom_report.yml
erik-bershel: Corrected with additionals

.github/workflows/create_sbom_report.yml
```diff
@@ -0,0 +1,88 @@
...
+          }
+          "current-os=$currentOS" | Out-File -FilePath $env:GITHUB_OUTPUT -Append
+  #Checking image version on available runner
+  version-check:
+    needs: initialize
+    runs-on: ${{ needs.initialize.output.agent-spec }}
+    steps:
+      - name: Available image version check
+        run: |
+          if ($env:ImageVersion -ne ${{ github.event.client_payload.imageVersion }}) {
```
maxim-lobanov, line 49: Use single quotes around payload variables: `if ($env:ImageVersion -ne '${{ github.event.client_payload.imageVersion }}') {`.
Otherwise, GH will compare string with double.

.github/workflows/create_sbom_report.yml
```diff
@@ -0,0 +1,88 @@
+name: Create and upload a SBOM to release assets
+# Inherited variables:
+# github.event.client_payload.imageLabel - AzDO image label
+# github.event.client_payload.GHreleaseID - Current release ID
+# github.event.client_payload.imageVersion - AzDO image version "major.minor"
+# github.event.client_payload.ReleaseBranchName - Bacause it wouldn't work withoit it, sorry
```
maxim-lobanov, line 6:
```suggestion
# github.event.client_payload.ReleaseBranchName - Necessary to identify workflow run
```

.github/workflows/create_sbom_report.yml
```diff
@@ -0,0 +1,88 @@
...
+# 
+# Current SYFT tool issues:
+# macOS (minor): very long cataloging process (more than 6 hours) (https://github.com/anchore/syft/issues/1328), 
+# macOS (major): promt privilegies that blocking process indefinetely (https://github.com/anchore/syft/issues/1367)
```
maxim-lobanov, line 10:
```suggestion
# macOS (major): prompt privilegies that blocking process indefinetely (https://github.com/anchore/syft/issues/1367)
```

maxim-lobanov (approved):

.github/workflows/create_sbom_report.yml
erik-bershel: Accepted and resolved

.github/workflows/create_sbom_report.yml
erik-bershel: Accepted and resolved

.github/workflows/create_sbom_report.yml
erik-bershel: Accepted and resolved

maxim-lobanov (approved):

mikhailkoliada (approved):

PR #6661, opened 2022-11-30
Rust: add support for mingw-w64 toolchains

jeroen:
# Description

The R project (and probably other OSS projects) are using mingw-w64 based compilers, that require the `x86_64-pc-windows-gnu` target for rust code. It would be great if we can preinstall this on the image.

## Check list


erik-bershel (approved):

mikhailkoliada (approved):

mikhailkoliada:
@jeroen looks like the target addition time is short, we'll add it.Thanks for the submission.

PR #6682, opened 2022-12-02
`ubuntu-latest` is now `ubuntu-22.04`

dougcalobrisi:
# Description
The `ubuntu-latest` label is now an alias for `ubuntu-22.04`.

## Check list
- [x ] Documentation is updated (if applicable)


mikhailkoliada:
Hello! The migration process is not yet done fully, we update docs automatically.

dougcalobrisi:
Makes sense to close, but perhaps there should be a note in the README to indicate migration is underway and `ubuntu-latest` is not necessarily `ubuntu-20.04`?

PR #6718, opened 2022-12-07
[macOS] Add xcode-14.2

mikhailkoliada:
# Description

Add xcode-14.2

## Check list

### DIFF
--- a/images/macos/toolsets/toolset-12.json
+++ b/images/macos/toolsets/toolset-12.json
@@ -2,6 +2,7 @@
     "xcode": {
         "default": "14.0.1",
         "versions": [
+            {"link": "14.2", "version": "14.2.0"},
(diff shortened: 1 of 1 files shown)


maxim-lobanov (approved):

bdrelling:
@maxim-lobanov your step does not work for `14.2` on `macos-12`, but you approved these changes so I'm assuming you helped test it. Do you happen to have further information on why it wouldn't be working? 

We're not getting a reply on #6762.

maxim-lobanov:
@bdrelling , The PR was merged and new image with this update is not deployed yet.
It can take some time to deploy the new image everywhere.
The deployment from https://github.com/actions/runner-images/pull/6762 was reverted due to some issues. So we have to wait for next deployment.
cc: @mikhailkoliada to share timeline for new image to be available.

PR #6754, opened 2022-12-12
[Windows] Install ghc and cabal using ghcup

ddobranic:
# Description
Use ghcup as primary installation method to install cabal and ghc

#### Related issue:
https://github.com/actions/runner-images-internal/issues/4682
## Check list


mudanzasIPROM:
sssss

images/win/scripts/Installers/Install-Haskell.ps1
```diff
@@ -18,7 +31,7 @@ ForEach ($version in $VersionsList)
     if ($version -eq "9.4.3"){ [version]$version = "9.4.2" }
```
mikhailkoliada, line 31: we should try to drop these versioning workaround

ddobranic: Removed.

igorboskovic3 (dismissed):

erik-bershel (dismissed):

mudanzasIPROM (approved):

mudanzasIPROM (dismissed):

mikhailkoliada:
CC: @hasufell for review

images/win/scripts/Installers/Install-Haskell.ps1
```diff
@@ -13,20 +26,14 @@ $VersionsList = $LatestMajorMinor | ForEach-Object { $_.Group | Select-Object -F
 # The latest version will be installed as a default
 ForEach ($version in $VersionsList)
 {
-     # 9.4.3 has failed choco builds, replace with 9.4.2 for the time being
```
hasufell, line 16: We're still using choco to get the available versions (a few lines up). We should use GHCup now.

images/win/scripts/Installers/Install-Haskell.ps1
```diff
@@ -13,20 +26,14 @@ $VersionsList = $LatestMajorMinor | ForEach-Object { $_.Group | Select-Object -F
...
-     # 9.2.5 has failed choco builds, replace with 9.2.4 for the time being
-    if ($version -eq "9.4.3"){ [version]$version = "9.4.2" }
-    if ($version -eq "9.2.5"){ [version]$version = "9.2.4" }
     Write-Host "Installing ghc $version..."
-    Choco-Install -PackageName ghc -ArgumentList '--version', $version, '-m'
+    ghcup install ghc $version
+    ghcup set ghc $version
```
hasufell, line 31: Can also use `ghcup install ghc --set $version` now.

hasufell (commented):
This looks correct so far, except that we need to use ghcup to get the available versions, not chocolatey.

The other concern I have is about testing this change. What do we test currently? We should test

1. The 3 GHC versions are all installed
2. A minimal test program can be built with all 3, see https://github.com/haskell/ghcup-metadata/blob/3b98df479a37430d2b9c61f3dd1e88d70bde5cb9/.github/workflows/install-bindist.sh#L23 for a very simple example
3. Cabal is installed
4. GHCUP_INSTALL_BASE_PREFIX is indeed set in the executing environment
4. Msys2 is accessible. This can be tested with something like: `ghcup run -- pacman --help`
5. The `cabal.config` was actually changed and contains entries for extra-lib-dirs, extra-include-dirs and extra-prog-paths. `cabal user-config diff` is helping here.

I think some of those tests already exist. I'm just putting this out here again for completeness sake.

images/win/scripts/Installers/Install-Haskell.ps1
```diff
@@ -16,32 +16,25 @@ Set-SystemVariable "CABAL_DIR" $cabalDir
...
 $DefaultGhcVersion = $VersionsList | Select-Object -Last 1
 $DefaultGhcShortVersion = ([version]$DefaultGhcVersion).ToString(3)
-$DefaultGhcPath = Join-Path $env:ChocolateyInstall "lib\ghc.$DefaultGhcVersion\tools\ghc-$DefaultGhcShortVersion\bin"
-# Starting from version 9 haskell installation directory is $env:ChocolateyToolsLocation instead of $env:ChocolateyInstall\lib
-if ($DefaultGhcShortVersion -notmatch '^[0-8]\.\d+.*')
-{
-    $DefaultGhcPath = Join-Path $env:ChocolateyToolsLocation "ghc-$DefaultGhcShortVersion\bin"
-}
-
+$DefaultGhcPath = Join-Path $env:GHCUP_INSTALL_BASE_PREFIX "ghcup\ghc\$DefaultGhcShortVersion\bin"
```
hasufell, line 35: this is not really needed... all binaries are in `Join-Path $env:GHCUP_INSTALL_BASE_PREFIX "ghcup\bin"` already, which we expose to path anyway. So I think we can remove this and the line below.

ddobranic: updated. please check once more. thank you!

ddobranic: one more thing, i think that '$cabalDir = "C:\cabal"' is now also false

hasufell: No... `$cabalDir` is where the cabal binary that resides in `/c/ghcup/bin/cabal` will write its runtime files into.

hasufell:
So the tests definitely need an update: 

https://github.com/actions/runner-images/blob/ad0880f33b9a3bef1c4ce7b9cdf6d93e072b704e/images/win/scripts/Tests/Haskell.Tests.ps1#L2-L9

This still uses chocolatey to discover the GHC versions.

Then we probably also want:

```ps1
    It "ghcup can access msys2" {
        "ghcup run -- pacman --version" | Should -ReturnZeroExitCode
    }
```

and:

1. cabal config change needs a new test (as indicated above in my comment)
2. building a minimal test program with all 3 GHC versions

hasufell:
> The cabal.config was actually changed and contains entries for extra-lib-dirs, extra-include-dirs and extra-prog-paths. cabal user-config diff is helping here.

Correction: we do not want it changed due to https://github.com/msys2/MINGW-packages/issues/10837#issuecomment-1047009539

The existing test seems to already imply that the config is not changed:

https://github.com/actions/runner-images/blob/b38865e056e5d6579fc8e973055387ae2b25f3a3/images/win/scripts/Tests/Haskell.Tests.ps1#L52-L54

So that seems fine.

----

Additionally I suggest to build a hello world program with each installed ghc version:

In bash, this would be something like:

```sh
cat <<EOF > main.hs
{- cabal:
build-depends: base
-}

main = print $ 1 + 1
EOF
```

And then execute it via cabal:

```sh
cabal --verbose=0 run --enable-profiling ./main.hs -- +RTS -s
```

And check that **stdout** is `2` and exit code is success.

----

Another test I suggest is a trick to verify that `cabal` actually sees the same dir we set via `CABAL_DIR`:

```ps1
        cabal help user-config | Select -Last 1 | ForEach-Object {$_.Trim()} | Should -MatchCommandOutput $env:CABAL_DIR
```

Something like that... not sure that's valid powershell test. I only tested `cabal help user-config | Select -Last 1 | ForEach-Object {$_.Trim()}`, which returns the cabal dir.

axljoker03 (approved):

axljoker03 (approved):

hasufell:
Does this need a rebase? Why are there so many commits?

Omishi09 (approved):

ddobranic:
@hasufell 
```
    It "ghcup can access msys2" {
        "ghcup run -- pacman --version" | Should -ReturnZeroExitCode
    }
```
is now failing with error:
```
   Command 'ghcup run -- pacman --version' has finished with exit code
    vhd:        [ Warn  ] New cabal version available. If you want to install this latest version, run 'ghcup install cabal 3.8.1.0'
    vhd:        [ Info  ] verifying digest of: gs.exe
    vhd:        ghcup: pacman: createProcess: does not exist (No such file or directory)
    vhd:    at "ghcup run -- pacman --version" | Should -ReturnZeroExitCode
```

hasufell:
@ddobranic that suggests `GHCUP_MSYS2` is not set. Try:

```sh
ghcup run -- env
```

hasufell:
Oops... I forgot, we need to do `ghcup run --mingw-path -- pacman --version`

mikhailkoliada:
@hasufell Hey Julian! I see that in ghcup cabal's version is 3.6 while in choco is 3.8, what's the actual difference?

hasufell:
> @hasufell Hey Julian! I see that in ghcup cabal's version is 3.6 while in choco is 3.8, what's the actual difference?

You can also install 3.8 via ghcup. We decided to keep 3.6 the recommended one since there are a number of regressions in 3.8: https://github.com/haskell/ghcup-metadata/issues/40

3.10 will be out soon, so my idea is to make that the recommended one eventually.

images/win/scripts/Installers/Install-Haskell.ps1
```diff
@@ -45,4 +17,27 @@ Set-SystemVariable "CABAL_DIR" $cabalDir
...
+    ghcup install ghc $version
+    ghcup set ghc $version
+}
+
+# Add default version of GHC to path
+$DefaultGhcVersion = $VersionsList | Select-Object -Last 1
+ghcup set ghc $DefaultGhcVersion
+
+Write-Host 'Installing cabal...'
+ghcup install cabal 3.8.1.0
```
hasufell, line 41: You can also do `ghcup install cabal latest`. Unless you want to track cabal releases.

ddobranic: OK, thank you @hasufell

mudanzasIPROM (dismissed):

images/win/scripts/Installers/Install-Haskell.ps1
```diff
@@ -45,4 +17,27 @@ Set-SystemVariable "CABAL_DIR" $cabalDir
 Add-MachinePathItem "$ghcupPrefix\ghcup\bin"
 Add-MachinePathItem "$cabalDir\bin"
 
+Update-SessionEnvironment
+
+# Get 3 latest versions of GHC
+$Versions = ghcup list -t ghc -r | Where-Object {$_ -notlike "prerelease"}
+$VersionsOutput = [Version[]]($Versions | %{ $_.Split(' ')[1]; })
```
mikhailkoliada, line 24: Do not use the alias

mikhailkoliada (changes requested):

images/win/scripts/Installers/Install-Haskell.ps1
ddobranic: Updated.

mudanzasIPROM:
z

mikhailkoliada (approved):

PR #6756, opened 2022-12-12
[Ubuntu] Allow image_sku to be passed in as variable

ifutant:
# Description
This Pull Request seeks to modify the latest Ubuntu 22.04 Packer template to allow an override of the hard-coded image_sku. This change introduces a new variable, and contains the hard-coded value as its default. 
The reason I desire to perform this override is to allow the deployment of a Generation 2 VM while consuming the packer template. I only targeted the images/linux/ubuntu2204.pkr.hcl file, because I didn't want to introduce a new variable dependency. I anticipate that this change will not impact current users of this packer template. 

#### Additional Information:
[Gen 2 VM Details](https://learn.microsoft.com/en-us/azure/virtual-machines/generation-2)

## Check list

### DIFF
--- a/images/linux/ubuntu2204.pkr.hcl
+++ b/images/linux/ubuntu2204.pkr.hcl
@@ -70,6 +70,11 @@ variable "image_version" {
   default = "dev"
 }
 
+variable "image_sku" {
+  type    = string
+  default = "22_04-lts"
+}
+
 variable "imagedata_file" {
   type    = string
   default = "/imagegeneration/imagedata.json"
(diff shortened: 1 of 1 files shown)


ifutant:
This is my first contribution to this repo. Is there anything else that I need to complete to obtain a review of this PR?

mikhailkoliada:
Hello! We do not plan using any alternative sku, so it makes little sense in general for us to use the variable here. So, we would not like to change it anyhow.

ifutant:
Why not introduce additional flexibility especially since Gen2 virtual machines are more modern? Just because your team doesn't plan to utilize an alternate SKU at the moment, that may not always be the case. What is the harm in making a small modification that facilities SKU parametrization while also maintaining existing functionality without impacting current users?

PR #6770, opened 2022-12-14
[Haskell] Make sure GHCUP_INSTALL_BASE_PREFIX is set

hasufell:
We use it during installation and add it to PATH, but then don't instruct ghcup to actually use this directory at runtime.

This leads to:

1. ghcup is in /usr/local/.ghcup/bin/
2. `ghcup install cabal latest` will actually install into ~/.ghcup/bin/
3. since ~/.ghcup/bin/ is not in PATH, but /usr/local/.ghcup/bin/ is, the new binary is not visible to the user

This issues does not exist on darwin, because there we use ~/.ghcup/bin/.

----

An alternative approach is matching the macOS install way and use the default `~/.ghcup`. Why was `/usr/local/.ghcup` chosen for linux?

### DIFF
--- a/images/linux/scripts/installers/haskell.sh
+++ b/images/linux/scripts/installers/haskell.sh
@@ -14,6 +14,7 @@ export GHCUP_INSTALL_BASE_PREFIX=/usr/local
 export BOOTSTRAP_HASKELL_GHC_VERSION=0
 ghcup_bin=$GHCUP_INSTALL_BASE_PREFIX/.ghcup/bin
 setEtcEnvironmentVariable "BOOTSTRAP_HASKELL_NONINTERACTIVE" $BOOTSTRAP_HASKELL_NONINTERACTIVE
+setEtcEnvironmentVariable "GHCUP_INSTALL_BASE_PREFIX" $GHCUP_INSTALL_BASE_PREFIX
 
 # Install GHCup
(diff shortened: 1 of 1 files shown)


hasufell:
@al-cheb @mikhailkoliada

images/linux/scripts/installers/haskell.sh
```diff
@@ -14,6 +14,7 @@ export GHCUP_INSTALL_BASE_PREFIX=/usr/local
 export BOOTSTRAP_HASKELL_GHC_VERSION=0
 ghcup_bin=$GHCUP_INSTALL_BASE_PREFIX/.ghcup/bin
 setEtcEnvironmentVariable "BOOTSTRAP_HASKELL_NONINTERACTIVE" $BOOTSTRAP_HASKELL_NONINTERACTIVE
+setEtcEnvironmentVariable "GHCUP_INSTALL_BASE_PREFIX" $GHCUP_INSTALL_BASE_PREFIX
```
hasufell, line 17: This will make sure it's exposed to all shells, right?

mikhailkoliada: @hasufell it affects the /etc/environment file, so yes

ddobranic (approved):

hasufell:
Good to go?

mikhailkoliada (approved):

mikhailkoliada:
@hasufell yes, sorry for the delay!

PR #6860, opened 2023-01-02
Use apt.llvm.org repo for clang on ubuntu & add clang 15

TerrorJack:
# Description

This PR uses the [official](https://apt.llvm.org/) LLVM apt repository to provide the clang packages. It's much more up-to-date than ubuntu's own repos, and above all, it provides clang 15 (the latest official release) for all ubuntu LTS versions that we support. Therefore we include it in the clang installation script, and add `15` to the installed versions. To avoid breakage, `15` is not set as the default version yet.

## Check list


mikhailkoliada:
Hello! We  do not use 3rd party resources to install packages on linux until absolutely needed, that's why llvm 15 is not yet in the image, we have no plans to add it uless it is backported ti ubuntu's lts repos.

PR #6884, opened 2023-01-06
[Linux, macOS, Windows] Download two prior CodeQL bundles

angelapwen:
# Description
Previously, our runner image only included the most recent release of CodeQL. In order to improve stability of rollout of new CodeQL versions, we are now downloading the most recent two releases.

The change updates the Linux and Windows script variable names to match that of the Mac script, because they are mostly functionally identical. It also adds the same log messages to all scripts and ensures Mac and Windows test parity. It tests that there are two bundles in each of the OS's as well. 

This will approximately double the total size and installation time of the existing CodeQL tool. 

## Check list
- [x] Tests are written
- [x] Documentation is updated


axljoker03 (approved):

images/linux/scripts/installers/codeql-bundle.sh
```diff
@@ -4,22 +4,37 @@
...
 
 # Retrieve the name of the CodeQL bundle preferred by the Action (in the format codeql-bundle-YYYYMMDD).
-codeql_bundle_name="$(curl -sSL https://raw.githubusercontent.com/github/codeql-action/v2/src/defaults.json | jq -r .bundleVersion)"
-# Convert the bundle name to a version number (0.0.0-YYYYMMDD).
-codeql_bundle_version="0.0.0-${codeql_bundle_name##*-}"
+defaults="$(curl -sSL https://raw.githubusercontent.com/github/codeql-action/v2/src/defaults.json)"
+codeql_bundle_name="$(${defaults} | jq -r .bundleVersion)"
+codeql_cli_version="$(${defaults} | jq -r .cliVersion)"
+prior_codeql_bundle_name="$(${defaults} | jq -r .priorBundleVersion)"
+prior_codeql_cli_version="$(${defaults} | jq -r .priorCliVersion)"
```
henrymercer, line 14: Hmm, this looks like it'll try to execute `defaults.json`.  I think we might want to do `jq -r '.bundleVersion' <<< "${defaults}"` instead.

images/linux/scripts/installers/codeql-bundle.sh
```diff
@@ -4,22 +4,37 @@
...
+codeql_bundle_version="${codeql_cli_version}-${codeql_bundle_name##*-}"
+prior_codeql_bundle_version="${prior_codeql_cli_version}-${prior_codeql_bundle_name##*-}"
 
-echo "Downloading CodeQL bundle $codeql_bundle_version..."
-download_with_retries "https://github.com/github/codeql-action/releases/download/$codeql_bundle_name/codeql-bundle.tar.gz" "/tmp" "codeql-bundle.tar.gz"
-tar -xzf "/tmp/codeql-bundle.tar.gz" -C "$extraction_directory"
+# For both of the bundle versions, download and name appropriately.
+for bundle in "${codeql_bundle_version} ${codeql_bundle_name}" "${prior_codeql_bundle_version} ${prior_codeql_bundle_name}"
+do
+    set -- "$bundle"
```
henrymercer, line 23: This assumes that there are no spaces in `codeql_bundle_version` or `codeql_bundle_name`.  This should be a safe assumption.

images/linux/scripts/installers/codeql-bundle.sh
```diff
@@ -4,22 +4,37 @@
...
 
-source $HELPER_SCRIPTS/install.sh
+source "$HELPER_SCRIPTS/install.sh"
 
 # Retrieve the name of the CodeQL bundle preferred by the Action (in the format codeql-bundle-YYYYMMDD).
-codeql_bundle_name="$(curl -sSL https://raw.githubusercontent.com/github/codeql-action/v2/src/defaults.json | jq -r .bundleVersion)"
-# Convert the bundle name to a version number (0.0.0-YYYYMMDD).
-codeql_bundle_version="0.0.0-${codeql_bundle_name##*-}"
+defaults="$(curl -sSL https://raw.githubusercontent.com/github/codeql-action/v2/src/defaults.json)"
+codeql_bundle_name="$(${defaults} | jq -r .bundleVersion)"
```
henrymercer, line 11: We might consider calling this the "tag name", since it's the tag of the GitHub Release.

images/linux/scripts/installers/codeql-bundle.sh
```diff
@@ -4,22 +4,37 @@
...
-# Touch a file to indicate to the toolcache that setting up CodeQL is complete.
-touch "$extraction_directory.complete"
+    codeqlToolcachePath="$AGENT_TOOLSDIRECTORY/codeql/$1/x64"
+    mkdir -p "$codeqlToolcachePath"
+
+    echo "Unpacking the downloaded CodeQL bundle archive..."
+    tar -xzf "$codeqlArchive" -C "$codeqlToolcachePath"
+
+    # Touch a special file that indicates to the CodeQL Action that this bundle was baked-in to the hosted runner images.
+    touch "$codeqlToolcachePath/pinned-version"
```
henrymercer, line 36: I believe this only affects a codepath that relies on the toolcache only containing a single version of `CodeQL`.  Therefore we should be able to remove this. https://github.com/github/codeql-action/blob/cf1437a51414bcd7b1ce9aa9704fa8b55b2c2926/src/codeql.ts#L477

images/linux/toolsets/toolset-1804.json
```diff
@@ -67,6 +67,7 @@
             "platform" : "linux",
             "arch": "x64",
             "versions": [
+                "*",
                 "*"
```
henrymercer, line 71: What's the implication of having two `*`s here?  Also, do we want to update the macOS toolsets too?

images/macos/tests/Common.Tests.ps1
```diff
@@ -149,6 +149,16 @@ Describe "CodeQL" -Skip:($os.IsCatalina) {
         $CodeQLPacksPath = Join-Path $CodeQLVersionPath -ChildPath "x64" | Join-Path -ChildPath "codeql" | Join-Path -ChildPath "qlpacks"
         $CodeQLPacksPath | Should -Exist
     }
+
+    It "prior codeql" {
```
henrymercer, line 153: Are there any corresponding tests we could update for Linux?

images/macos/tests/Common.Tests.ps1
```diff
@@ -149,6 +149,16 @@ Describe "CodeQL" -Skip:($os.IsCatalina) {
         $CodeQLPacksPath = Join-Path $CodeQLVersionPath -ChildPath "x64" | Join-Path -ChildPath "codeql" | Join-Path -ChildPath "qlpacks"
         $CodeQLPacksPath | Should -Exist
     }
+
+    It "prior codeql" {
+        $PriorCodeQLVersionsWildcard = Join-Path $Env:AGENT_TOOLSDIRECTORY -ChildPath "CodeQL" | Join-Path -ChildPath "*"
+        $PriorCodeQLVersionPath = Get-ChildItem $PriorCodeQLVersionsWildcard | Select-Object -Last 1 -Expand FullName
+        $PriorCodeQLPath = Join-Path $PriorCodeQLVersionPath -ChildPath "x64" | Join-Path -ChildPath "codeql" | Join-Path -ChildPath "codeql"
+        "$PriorCodeQLPath version --quiet" | Should -ReturnZeroExitCode
```
henrymercer, line 157: Optional: We could potentially have a test that the current and prior versions of CodeQL are different.

henrymercer (commented):
Nice!  A few comments around the installation script and testing.

images/linux/scripts/installers/codeql-bundle.sh
angelapwen: Ah, that makes sense

images/linux/scripts/installers/codeql-bundle.sh
angelapwen: Oh, interesting, nice catch!

images/linux/toolsets/toolset-1804.json
angelapwen: Hm, I was hoping that the two `*`s would test that there are two versions (each of which can be any version). I'm not sure if just one `*` would have the same meaning.

It looks like CodeQL doesn't exist in the macOS toolsets right now, which is why I haven't added them. I think for macOS we are relying on the `Common.Tests` file (and invoking the test at the end of the script for the macOS file directly).

images/linux/scripts/installers/codeql-bundle.sh
angelapwen: Done!

images/linux/scripts/installers/codeql-bundle.sh
angelapwen: Done 👍

images/linux/scripts/installers/codeql-bundle.sh
angelapwen: Done!

images/macos/tests/Common.Tests.ps1
angelapwen: I've added this test separately. It does add additional time in now calling `codeql version` two more times (so four total calls) for the extra test in order to compare the versions to one another. 

Alternatively I could collapse all three tests into one that would together test that the versions are different, and that each pack path exists, which would decrease the amount of repeated work/calls to `codeql version`. What do you think?

images/linux/scripts/installers/codeql-bundle.sh
```diff
@@ -4,22 +4,34 @@
...
 
-source $HELPER_SCRIPTS/install.sh
+source "$HELPER_SCRIPTS/install.sh"
 
 # Retrieve the name of the CodeQL bundle preferred by the Action (in the format codeql-bundle-YYYYMMDD).
-codeql_bundle_name="$(curl -sSL https://raw.githubusercontent.com/github/codeql-action/v2/src/defaults.json | jq -r .bundleVersion)"
-# Convert the bundle name to a version number (0.0.0-YYYYMMDD).
-codeql_bundle_version="0.0.0-${codeql_bundle_name##*-}"
+defaults="$(curl -sSL https://raw.githubusercontent.com/github/codeql-action/v2/src/defaults.json)"
+codeql_tag_name=${jq -r '.bundleVersion' <<< "${defaults}"}
```
adityasharad, line 11: Should the outermost brackets be `$(...)` rather than `${...}`, since we're running a subshell command rather than interpolating a variable? Here and on the 3 lines below.

images/macos/provision/core/codeql-bundle.sh
```diff
@@ -2,24 +2,33 @@
...
 
 # Retrieve the name of the CodeQL bundle preferred by the Action (in the format codeql-bundle-YYYYMMDD).
-codeql_bundle_name="$(curl -sSL https://raw.githubusercontent.com/github/codeql-action/v2/src/defaults.json | jq -r .bundleVersion)"
-# Convert the bundle name to a version number (0.0.0-YYYYMMDD).
-codeql_bundle_version="0.0.0-${codeql_bundle_name##*-}"
+defaults="$(curl -sSL https://raw.githubusercontent.com/github/codeql-action/v2/src/defaults.json)"
+codeql_tag_name=${jq -r '.bundleVersion' <<< "${defaults}"}
+codeql_cli_version=${jq -r '.cliVersion' <<< "${defaults}"}
+prior_codeql_tag_name=${jq -r '.priorBundleVersion' <<< "${defaults}"}
+prior_codeql_cli_version=${jq -r '.priorCliVersion' <<< "${defaults}"}
```
adityasharad, line 9: Same question here.

images/linux/scripts/SoftwareReport/SoftwareReport.Generator.ps1
```diff
@@ -103,6 +103,7 @@ $tools.AddToolVersion("Bazelisk", $(Get-BazeliskVersion))
 $tools.AddToolVersion("Bicep", $(Get-BicepVersion))
 $tools.AddToolVersion("Buildah", $(Get-BuildahVersion))
 $tools.AddToolVersion("CMake", $(Get-CMakeVersion))
+$tools.AddToolVersion("CodeQL Action Bundle", $(Get-PriorCodeQLBundleVersion))
 $tools.AddToolVersion("CodeQL Action Bundle", $(Get-CodeQLBundleVersion))
```
adityasharad, line 107: I don't see other examples where we call `AddToolVersion` with the same tool name but different versions. Just want to be sure this will append the second version rather than overwriting the first version.

images/linux/scripts/installers/codeql-bundle.sh
angelapwen: Oh yes, thank you!

axljoker03 (dismissed):

images/linux/scripts/SoftwareReport/SoftwareReport.Generator.ps1
angelapwen: Oh, nice catch, I think you're right. Looks like `AddToolVersion` is defined [here](https://github.com/actions/runner-images/blob/9ee93111224473826a8168cc6b2951f11fdba197/helpers/software-report-base/SoftwareReport.Nodes.psm1#L79) which calls [`AddNode`](https://github.com/actions/runner-images/blob/9ee93111224473826a8168cc6b2951f11fdba197/helpers/software-report-base/SoftwareReport.Nodes.psm1#L41), which in turns throws if the node name has been found. 

I'll update the name.

images/linux/scripts/SoftwareReport/SoftwareReport.Generator.ps1
```diff
@@ -103,7 +103,8 @@ $tools.AddToolVersion("Bazelisk", $(Get-BazeliskVersion))
 $tools.AddToolVersion("Bicep", $(Get-BicepVersion))
 $tools.AddToolVersion("Buildah", $(Get-BuildahVersion))
 $tools.AddToolVersion("CMake", $(Get-CMakeVersion))
-$tools.AddToolVersion("CodeQL Action Bundle", $(Get-CodeQLBundleVersion))
+$tools.AddToolVersion("CodeQL Action Bundle (Prior Release)", $(Get-PriorCodeQLBundleVersion))
+$tools.AddToolVersion("CodeQL Action Bundle (Latest Release)", $(Get-CodeQLBundleVersion))
```
angelapwen, line 107: Open to suggestions on how to name these more effectively!

images/linux/scripts/installers/codeql-bundle.sh
adityasharad: Also don't forget the quotes around `"$(...)"`. Not essential here since the output doesn't have newlines, but good to be consistent.

images/linux/scripts/SoftwareReport/SoftwareReport.Generator.ps1
Nomiio: sorry no idea

henrymercer: It's a little redundant but we specify the actual version number, e.g. "CodeQL Action Bundle v2.12.1" and "CodeQL Action Bundle v2.12.0".  This would be consistent with "Docker Compose v1" and "Docker Compose v2" below.

images/linux/scripts/installers/codeql-bundle.sh
henrymercer: On reflection, I think we want to keep this, but only for the latest CodeQL version.  This means we can continue supporting the behaviour on Enterprise where the toolcache version of CodeQL can override the defaults.json version on Enterprise.  We'll just need to modify the CodeQL Action to allow the override when the toolcache contains a single pinned version of CodeQL, vs just a single version of CodeQL.

images/linux/scripts/SoftwareReport/SoftwareReport.Generator.ps1
angelapwen: That would look like:

`CodeQL Action Bundle v2.12.1, 2.12.1` right? I'm not sure I like the redundancy 😆 but happy to do it if you think that's clearer 👍

images/linux/scripts/installers/codeql-bundle.sh
angelapwen: Oh, I see, that makes sense. Thanks for explaining the logic! I'll make the change 👍

images/linux/scripts/installers/codeql-bundle.sh
```diff
@@ -4,22 +4,39 @@
 ##  Desc:  Install the CodeQL CLI Bundle to the toolcache.
 ################################################################################
 
-source $HELPER_SCRIPTS/install.sh
+source "$HELPER_SCRIPTS/install.sh"
```
mikhailkoliada, line 7: I do not think we need quotes here :)

images/linux/scripts/installers/codeql-bundle.sh
```diff
@@ -4,22 +4,39 @@
...
 
 # Retrieve the name of the CodeQL bundle preferred by the Action (in the format codeql-bundle-YYYYMMDD).
-codeql_bundle_name="$(curl -sSL https://raw.githubusercontent.com/github/codeql-action/v2/src/defaults.json | jq -r .bundleVersion)"
-# Convert the bundle name to a version number (0.0.0-YYYYMMDD).
-codeql_bundle_version="0.0.0-${codeql_bundle_name##*-}"
+defaults="$(curl -sSL https://raw.githubusercontent.com/github/codeql-action/v2/src/defaults.json)"
```
mikhailkoliada, line 10: the link seems to be wrong, it does not have all the properties you list later

mikhailkoliada, line 10: rename to `base_url`

images/linux/scripts/installers/codeql-bundle.sh
```diff
@@ -4,22 +4,39 @@
...
+codeql_tag_name="$(jq -r '.bundleVersion' <<< "${defaults}")"
```
mikhailkoliada, line 11: it is quite hard to read, I'd go with something like:

`codeql_tag_name=$(echo $base_url | jq -r '.bundleVersion')`

images/linux/scripts/installers/codeql-bundle.sh
```diff
@@ -4,22 +4,39 @@
...
-mkdir -p "$extraction_directory"
+# Convert the tag names to bundles with a version number (x.y.z-YYYYMMDD).
+codeql_bundle_version="${codeql_cli_version}-${codeql_tag_name##*-}"
+prior_codeql_bundle_version="${prior_codeql_cli_version}-${prior_codeql_tag_name##*-}"
 
-echo "Downloading CodeQL bundle $codeql_bundle_version..."
-download_with_retries "https://github.com/github/codeql-action/releases/download/$codeql_bundle_name/codeql-bundle.tar.gz" "/tmp" "codeql-bundle.tar.gz"
-tar -xzf "/tmp/codeql-bundle.tar.gz" -C "$extraction_directory"
+# For both of the bundle versions, download and name appropriately.
+for bundle in "${codeql_bundle_version} ${codeql_tag_name}" "${prior_codeql_bundle_version} ${prior_codeql_tag_name}"
```
mikhailkoliada, line 21: make an array of these versions, then iterate :)

images/linux/scripts/installers/codeql-bundle.sh
```diff
@@ -4,22 +4,39 @@
...
+do
+    set -- "$bundle"
 
-# Touch a special file that indicates to the CodeQL Action that this bundle was baked-in to the hosted runner images.
-touch "$extraction_directory/pinned-version"
+    echo "Downloading CodeQL bundle $1..."
```
mikhailkoliada, line 25: Why do we need $1 here?

images/linux/scripts/installers/codeql-bundle.sh
```diff
@@ -4,22 +4,39 @@
...
+    download_with_retries "https://github.com/github/codeql-action/releases/download/$2/codeql-bundle.tar.gz" "/tmp" "codeql-bundle.tar.gz"
```
mikhailkoliada, line 26: ... and  $2 here?

images/linux/toolsets/toolset-1804.json
```diff
@@ -67,6 +67,7 @@
             "platform" : "linux",
             "arch": "x64",
             "versions": [
+                "*",
```
mikhailkoliada, line 70: you do not need a second asterisk here, it literally means "any version it finds" and it is used for the toolcache pester test mostly

images/linux/scripts/installers/codeql-bundle.sh
angelapwen: Ah yes, we need to create a new release of the CodeQL Action to update the file. We'll make sure to hold on merging this PR until then.

images/linux/scripts/installers/codeql-bundle.sh
angelapwen: Hm I'm not totally clear on what you mean here — I thought that `"${codeql_bundle_version} ${codeql_tag_name}" "${prior_codeql_bundle_version} ${prior_codeql_tag_name}"` was essentially making an array of pairs (where the first item is the version and second is tag name). If I only make an array of versions I won't have the associated tag name. 

In the PowerShell version I was able to create an array of `PSCustomObject`s but I'm not sure how I would do that in `bash` 👀

images/linux/scripts/installers/codeql-bundle.sh
angelapwen: It refers to the first item in the `bundle` pair (ie. just the bundle version rather than the entire pair)

images/linux/toolsets/toolset-1804.json
angelapwen: Ah, ok, thank you!

images/linux/scripts/installers/codeql-bundle.sh
angelapwen: Done!

images/linux/scripts/installers/codeql-bundle.sh
angelapwen: Done :)

images/linux/scripts/installers/codeql-bundle.sh
angelapwen: Thanks for the suggestion, done!

images/linux/toolsets/toolset-1804.json
angelapwen: Removed 👍

angelapwen: Do you have any suggestions for how we could test that the Linux toolsets contain exactly 2 versions of CodeQL? We do it in a separate `Common.Tests` file for macOS — do you recommend reusing the same logic in the equivalent file for Linux?

images/linux/scripts/installers/codeql-bundle.sh
angelapwen: Ah yes, I see.. so somewhat similar but a lot cleaner. Thank you! Done.

images/linux/scripts/installers/codeql-bundle.sh
angelapwen: Resolved using the bash arrays.

images/linux/scripts/installers/codeql-bundle.sh
angelapwen: Resolved using the bash arrays.

images/linux/scripts/SoftwareReport/SoftwareReport.Generator.ps1
Nomiio: All I know is that I dont't need linux on mu machine. Win 11. thats it.

angelapwen: We decided to go with `CodeQL Action Bundle (Prior)` and `CodeQL Action Bundle` 😄

angelapwen:
@mikhailkoliada do you mind taking another look when you have time? 

I plan on cleaning up the commits before merge, but didn't want to rebase yet for ease of review. Please let me know if you prefer I clean them up sooner!

images/linux/toolsets/toolset-1804.json
angelapwen: Resolved below 👍

images/macos/software-report/SoftwareReport.Common.psm1
```diff
@@ -632,6 +632,14 @@ function Build-GraalVMTable {
     }
 }
 
+function Get-PriorCodeQLBundleVersion {
```
mikhailkoliada, line 635: Do we really need these 2 functions? Given that we are using the same toolcache directory, I think we can simply iterate over versions with `For-Each` loop and then make something like `CodeQL versions: x.x.x, y.y.y`

images/macos/provision/core/codeql-bundle.sh
```diff
@@ -2,24 +2,41 @@
...
 
-codeqlToolcachePath="$AGENT_TOOLSDIRECTORY/codeql/$codeql_bundle_version/x64"
-mkdir -p $codeqlToolcachePath
+# Download and name both CodeQL bundles.
+codeql_bundle_versions=("${codeql_bundle_version}" "${prior_codeql_bundle_version}")
+codeql_tag_names=("${codeql_tag_name}" "${prior_codeql_tag_name}")
 
-echo "Unpacking the downloaded CodeQL bundle archive..."
-tar -xzf $codeqlArchive -C $codeqlToolcachePath
+for ((i=0;i<=1;i++)); do
```
mikhailkoliada, line 19: Could be something like `for $version in $codeql_bundle_versions[@]; do ... done`

images/macos/software-report/SoftwareReport.Common.psm1
angelapwen: Thanks, I've made this change. Please take a look!

henrymercer:
For more context, we'd like to get this change in before the next runner image update cycle starts if possible as part of our work on controlled rollout of the CodeQL Action (see backlinked issue for more information).

angelapwen:
Thank you @mikhailkoliada !! I see I have some build failures; do you think you could help me find the error messages? I see [internal link](https://github.visualstudio.com/runner-images/_build/results?buildId=130467&view=logs&j=747417c8-a650-5aed-afed-d9b52dbd7d24&t=f245de8a-85db-5d11-6f6b-a3b45c2b2a6a) but cannot find the relevant failure.

images/macos/tests/Common.Tests.ps1
henrymercer: I think repeated calls to `codeql version` are probably fine, and we should do whatever's most readable.  But it probably makes sense to do this as a followup rather than in this PR.

angelapwen: Sounds good — I will leave this thread unresolved so I remember to do the followup.

angelapwen:
Upstream Actions failure. Will re-run later

angelapwen:
Sorry about the spam... tests were failing due to other changes, had to rebase on `main`

mikhailkoliada (approved):

PR #6888, opened 2023-01-07
docs: update chart to reflect `macos-latest`

ckipp01:
The chart makes it seem like `macos-latest` is 11 when it's actually 12.


mikhailkoliada:
Hello! We update docs manually as soon as it is fully migrated.

PR #6906, opened 2023-01-12
README: no active migrations

mikhailkoliada:
We no longer have active migrations by now

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,7 @@
 
 - [About](#about)
 - [Available Images](#available-images)
+- [Ongoing migrations](#ongoing-migrations)
 - [Image Definitions](#image-definitions)
 - [Image Releases](#image-releases)
 - [Software and Image Support](#software-and-image-support)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -5,7 +5,6 @@
 
 - [About](#about)
 - [Available Images](#available-images)
-- [Ongoing migrations](#ongoing-migrations)
```
samvantran, line 8: Let's not delete this since it's a regular event

README.md
```diff
@@ -45,12 +44,6 @@ To build a VM machine from this repo's source, see the [instructions](docs/creat
 [macOS-10.15]: https://github.com/actions/runner-images/blob/main/images/macos/macos-10.15-Readme.md
 [self-hosted runners]: https://help.github.com/en/actions/hosting-your-own-runners
 
-## Ongoing migrations
-
-There are currently two migrations in progress:
```
samvantran, line 50: Let's change this line to 
```
There are currently no migrations.
```
and delete the lines below.

We should also add a doc (can be outside of this PR) that explains migrations and what users should expect during these time periods.

samvantran (changes requested):

README.md
mikhailkoliada: done

samvantran (approved):

PR #6950, opened 2023-01-20
macos: installing awscli, aws-sam-cli via brew

chenrui333:
# Description
Simplify the awscli, aws-sam-cli for macos runner

## Check list

### DIFF
--- a/images/macos/provision/core/aws.sh
+++ b/images/macos/provision/core/aws.sh
@@ -2,12 +2,9 @@
 source ~/utils/utils.sh
 
 echo Installing aws...
-AWS_CLI_URL="https://awscli.amazonaws.com/AWSCLIV2.pkg"
-download_with_retries $AWS_CLI_URL "/tmp"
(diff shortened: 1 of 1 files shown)


mikhailkoliada:
Hello! What would be the reason of the change? We generally prefer official sources if it is available and easy enough to install.

chenrui333:
> Hello! What would be the reason of the change? We generally prefer official sources if it is available and easy enough to install.

Mostly for python dependency, for brew core formulae installation ecosystem, the official one is pretty outdated using python@3.8, since we included aws-sam-cli recently to streamline all the aws tools. I find it would be better to change both. (lots of dependencies between the two are overlapped), let me know if that makes better sense.

mikhailkoliada:
@chenrui333 hello again and sorry for the delays!

We prefer not to install awscli tooling from brew as it tends to bring unwanted dependencies to the images we try to avoid, we used brew in the past and then switched to the official pkg. Given quite small benefit of changing the installation method again and taking all the risks into account we would not like getting back to brew.

PR #7008, opened 2023-01-30
Allow to use existing resource group for pipelines generating Linux and Windows image variants

SheGe:
# Description

Enhancements in the pipeline (`image-generation.yml`) and in the `build-image.ps1` script to allow using a static resource group as an option to the temporary resource group.

This change allows generating an image in situations when given service principal does not have enough permissions to create resource groups in Azure.

According to the [Packer AzureRM documentation](https://developer.hashicorp.com/packer/plugins/builders/azure/arm):

> By default a temporary resource group will be created and destroyed as part of the build. If you do not have permissions to do so, use `build_resource_group_name` to specify an existing resource group to run the build in.

> Providing `temp_resource_group_name` or `location` in combination with `build_resource_group_name` is not allowed.

> The Azure builder works under the assumption that it creates everything it needs to execute a build. When the build has completed it simply deletes the resource group to cleanup any runtime resources. 

This change was tested and is backward compatible with existing variable groups. The `BUILD_RESOURCE_GROUP_NAME` is a new variable which existence in variable group is optional.

***REMARK** When using an existing resource group the cleanup process needs to be covered by custom cleaning logic which is not part of this repository.*

Example pipeline design showing where the custom cleanup logic could be created:

```yaml
resources:
  repositories:
    - repository: RunnerImagesGitHub
      type: github
      name: actions/runner-images
      endpoint: ***
      ref: releases/ubuntu22/20221212

...
  - stage: 'Build'
    dependsOn: []
    displayName: 'Build'
    jobs:
      - template: images.CI/linux-and-win/azure-pipelines/image-generation.yml@RunnerImagesGitHub
        parameters:
          job_id: 'GenerateVHD'
          image_type: 'ubuntu2204'
          image_readme_name: 'Ubuntu2204-Readme.md'
          agent_pool:
            name: 'ci-agent-pool'
          variable_group_name: 'Image Generation Variables'
          create_release: false
          repository_ref: 'RunnerImagesGitHub'

      - job: 'Cleanup'
        dependsOn:
          - 'GenerateVHD'
        condition: always()
        displayName: 'Cleanup Packer Resources'
        pool:
          vmImage: 'ubuntu-latest'
        steps:
          - checkout: none

          - task: AzureCLI@2
            displayName: 'Cleanup all resources'
            inputs:
              azureSubscription: '***'
              scriptType: 'pscore'
              scriptLocation: inlineScript
              inlineScript: |
                # Custom cleanup logic here. 
                Write-Host "Cleanup completed."
              powerShellErrorActionPreference: stop
```

*Hint: All resources created by packer in case of `runner-images` repository will have names starting from `$(Build.BuildId)` pipeline variable.*

## Check list

### DIFF
--- a/images.CI/linux-and-win/azure-pipelines/image-generation.yml
+++ b/images.CI/linux-and-win/azure-pipelines/image-generation.yml
@@ -40,6 +40,8 @@ jobs:
   pool: ${{ parameters.agent_pool }}
   variables:
   - group: ${{ parameters.variable_group_name }}
+  - name: 'BUILD_RESOURCE_GROUP_NAME_RUNTIME'
+    value: $[variables.BUILD_RESOURCE_GROUP_NAME]
 
   steps:
   - checkout: ${{ parameters.repository_ref }}
@@ -84,8 +86,9 @@ jobs:
                         -Location $(AZURE_LOCATION) `
                         -VirtualNetworkName $(BUILD_AGENT_VNET_NAME) `
                         -VirtualNetworkRG $(BUILD_AGENT_VNET_RESOURCE_GROUP) `
-                        -VirtualNetworkSubnet $(BUILD_AGENT_SUBNET_NAME)
-
+                        -VirtualNetworkSubnet $(BUILD_AGENT_SUBNET_NAME) `
+                        -BuildResourceGroupName "$(BUILD_RESOURCE_GROUP_NAME_RUNTIME)"
+                        
     env:
       PACKER_LOG: 1
       PACKER_LOG_PATH: $(Build.ArtifactStagingDirectory)/packer-log.txt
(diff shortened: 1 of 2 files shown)


Alexey-Ayupov:
Hello @SheGe, looks good so far however, the "build resource group" should already exist if we would like to have a successful run. Some users could assume that filling in the "BUILD_RESOURCE_GROUP_NAME" variable provides resource group creation, but it doesn't. At least we need to check the mentioned resource group's existence and what significantly increases script difficulty. 
Given all that we would not like to introduce the change.

SheGe:
@Alexey-Ayupov regarding the naming convention of variables I was also confused, because of how packer azurerm was naming arguments and documenting them. At the end I just decided to follow same name for new argument as in packer provider because most arguments of the wrapper script are already referring to packer/packer azurm provider arguments.

From resource group existence check perspective - packer will do the check and it will fail with appropriate error when resource group does not exist at beginning of execution. If you would like to have a check in code - it will be not complex as in existing cleanup there is already a code possible to re-use.

This change is minimal (it is extending existing logic only). If still, because of the confusion, you would like to block this PR I'll be happy to refactor existing wrapper script to have arguments named differently.

SheGe:
@mikhailkoliada May I ask you also for second opinion?

mikhailkoliada:
@SheGe Hello!

I tend to agree with @Alexey-Ayupov, from usage point of view we must introduce additional checks in order that everything goes smooth, we do not observe high demand in implementing this, so would not like to add the code. If the situation changes it might be reconsidered.

PR #7043, opened 2023-02-04
fix issue 6251

erik-de-bont:
# Description
Bug fixing:
When Microsoft Defender for Endpoint is active in an Azure Subscription, it can cause interference when sysprep is run at the end of the VM Build causing the **IIMAGE_STATE_UNDEPLOYABLE** error. To avoid interference the parameter "azure_tags": { "ExcludeMdeAutoProvisioning": "True"  has been added to the builder type parameter., 

Related issue: [https://github.com/actions/runner-images/discussions/6251](https://github.com/actions/runner-images/discussions/6251)

## Check list

### DIFF
--- a/images/win/windows2019.json
+++ b/images/win/windows2019.json
@@ -34,7 +34,7 @@
     "builders": [
         {
             "name": "vhd",
-            "type": "azure-arm",
+            "type": "azure-arm","azure_tags": { "ExcludeMdeAutoProvisioning": "True" },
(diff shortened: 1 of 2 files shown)


mikhailkoliada:
Given this tag is provided by a very specific azure policy and the problem is not common we would not like to add it on a permanent basis, it is pretty much case-dependent and would not solve the problem globally.

Issue #7046, opened 2023-02-06
Deploying Ubuntu 22.04 with Tags fails


JoshuaVanDaalen:
### Description

Within the PowerShell script _(GenerateResourcesAndImage.ps1)_, the if statement for tags uses some ConverTo-Json and ConverFrom-Json, this will fail when choosing the packer file with extension pkr.hcl

### Platforms affected

- [ ] Azure DevOps
- [ ] GitHub Actions - Standard Runners
- [ ] GitHub Actions - Larger Runners

### Runner images affected

- [ ] Ubuntu 18.04
- [ ] Ubuntu 20.04
- [X] Ubuntu 22.04
- [ ] macOS 11
- [ ] macOS 12
- [ ] Windows Server 2019
- [ ] Windows Server 2022

### Image version and build link

​

### Is it regression?

​

### Expected behavior

Expected not to fail

### Actual behavior

It fails

### Repro steps

```powershell

$params = @{
    SubscriptionId                = $SubscriptionId
    ResourceGroupName             = $StorageAccountResourceGroupName
    ImageGenerationRepositoryRoot = "$pwd"
    ImageType                     = 'Ubuntu2204'
    AzureLocation                 = 'Australia Southeast'
    AzureClientId                 = $Credential.UserName
    AzureClientSecret             = $Credential.GetNetworkCredential().Password
    AzureTenantId                 = 'a47bd588-f2c9-4146-b0b2-d39f56758e0e'
    RestrictToAgentIpAddress      = $true
    AllowBlobPublicAccess         = $False
    EnableHttpsTrafficOnly        = $true
    SecondsToWaitForServicePrincipalSetup = 30
    OnError = "ask"
    Tags                          = @{
        Department  = 'P&E Infrastructure & Cloud'
        Environment = 'Non-Production-Shared'
        Platform    = 'Networking'
        Project     = 'BAU'
    }
}


.\GenerateResourcesAndImage.ps1 @params
```

JoshuaVanDaalen:
See PR: [7047](https://github.com/actions/runner-images/pull/7047)

ddobranic:
Hello @JoshuaVanDaalen, thank you for your contribution. We will take a look.

erik-bershel:
Hi @JoshuaVanDaalen!
Thank you for your contribution again. I'll work on it near time. But I should close that issue due to it's a duplicate for https://github.com/actions/runner-images/issues/6946. We'll continue discussion in the #6946 issue and PR as well.

PR #7047, opened 2023-02-06
Update GenerateResourcesAndImage.ps1 #7046

JoshuaVanDaalen:
# Description
Update script to handle Tags with the pkr.hcl file

#### Related issue:
[7046](https://github.com/actions/runner-images/issues/7046)

## Check list


erik-bershel:
@JoshuaVanDaalen that part of code should be rewritten in a bit other way. I'll publish changes a bit later. But main your idea is good and helpful. Thank you for contribution!

JoshuaVanDaalen:
There is a lot about this script that should be rewritten, we just want things to work today.
The script is literally doing the same thing for the IP address for the agent.

PR #7051, opened 2023-02-06
install only the latest patch version of .net core

gonchalo620:
To comply with the guidelines https://github.com/gonchalo620/virtual-environments/blob/main/README.md#software-guidelines, the Install-DotnetSDK.ps1 script was modified so that it only installs the latest patch version of each major.minor version of .net core

# Description
Improvement

To comply with the guidelines https://github.com/gonchalo620/virtual-environments/blob/main/README.md#software-guidelines in which it is mentioned that only the latest version of each major.secondary version will be installed. Therefore, the Install-DotnetSDK.ps1 script was modified so that it only installs the latest patch version of each major.minor version of .net core.

image

Currently the script is scheduled to install all patch versions other than preview/rc releases. Therefore it installs older versions, which contain uncorrected detected vulnerabilities.

### DIFF
--- a/images/win/scripts/Installers/Install-DotnetSDK.ps1
+++ b/images/win/scripts/Installers/Install-DotnetSDK.ps1
@@ -18,8 +18,9 @@ function Get-SDKVersionsToInstall (
     $releaseJson = "https://dotnetcli.blob.core.windows.net/dotnet/release-metadata/${DotnetVersion}/releases.json"
     $releasesJsonPath = Start-DownloadWithRetry -Url $releaseJson -Name "releases-${DotnetVersion}.json"
     $currentReleases = Get-Content -Path $releasesJsonPath | ConvertFrom-Json
-    # filtering out the preview/rc releases
-    $currentReleases = $currentReleases.'releases' | Where-Object { !$_.'release-version'.Contains('-') }
+    # filtering the latest patch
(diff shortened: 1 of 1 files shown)


maxim-lobanov:
Hello @gonchalo620 , thank you for contribution but I think the current installation process is fully compliant with our policy.

Our policy - `For each feature version only latest patch is installed`. It is not equal to `only the latest version of each major.secondary version will be installed` for .NET Core.

According to [.NET Core documentation](https://learn.microsoft.com/en-us/dotnet/core/releases-and-support#feature-bands-sdk-only) `feature version` is `x.y.z*` (major + minor + first digit of patch version):
> the .NET SDK uses the concept of feature bands. For example, the first .NET 5 SDK was 5.0.100. This release corresponds to the 3.1.1xx feature band. Feature bands are defined in the hundreds groups in the third section of the version number. For example, 5.0.101 and 5.0.201 are versions in two different feature bands while 5.0.101 and 5.0.199 are in the same feature band. When .NET SDK 5.0.101 is installed, .NET SDK 5.1.100 is removed from the machine if it exists. When .NET SDK 5.0.200 is installed on the same machine, .NET SDK 5.0.101 isn't removed.

So I think we follow .NET Core versioning process correctly.

Issue #7061, opened 2023-02-08
ghcup broken


hasufell:
### Description

Failure reports:

* https://github.com/input-output-hk/ouroboros-network/actions/runs/4117108345/jobs/7108005081
* https://github.com/haskell/cabal/actions/runs/4118707642/jobs/7112381525#step:3:135

Error is:

```
ghcup: /usr/local/.ghcup/cache/ghcup-0.0.7.yaml: setModificationTime:setFileTimes: permission denied (Operation not permitted)
```

----

GHCup needs the cache and other directories writable. This is ensured in

https://github.com/actions/runner-images/blob/8361bb299ba9290b70b6ed8bf337456714c6516d/images/linux/scripts/installers/haskell.sh#L45

However, for some reason this seems to not be enough. 

This PR might have triggered the bug, because before, ghcup would (incorrectly) use `~/.ghcup`:

https://github.com/actions/runner-images/pull/6770

@mikhailkoliada @angerman @Mikolaj 

### Platforms affected

- [ ] Azure DevOps
- [X] GitHub Actions - Standard Runners
- [ ] GitHub Actions - Larger Runners

### Runner images affected

- [X] Ubuntu 18.04
- [X] Ubuntu 20.04
- [X] Ubuntu 22.04
- [ ] macOS 11
- [ ] macOS 12
- [ ] Windows Server 2019
- [ ] Windows Server 2022

### Image version and build link

Runner version 2.301.1

Runner image version 20230206.1

### Is it regression?

Yes

### Expected behavior

ghcup can write to `/usr/local/.ghcup`

### Actual behavior

ghcup can't write to `/usr/local/.ghcup`

### Repro steps

Run `ghcup install ghc 9.4.4` in a standard ubuntu image.

ddobranic:
@hasufell, thank you for reporting this issue. we already have ghc 9.4.4 installed on the image and image build logs doesn't contain any write permission errors. do you think we should revert [#6770](https://github.com/actions/runner-images/pull/6770)?

hasufell:
I think we need to investigate why there are permission errors and fix it.

I also noticed newer pipeline runs don't use 20230206.1 image anymore but 20230129.2. Has someone reverted already?

ddobranic:
20230206.1 is being released currently

hasufell:
Some things to investigate:

1. do the tests run as superuser and change the permissions implicitly? Why did the tests not catch this?
2. there should be a permission test
[...]

PR #7068, opened 2023-02-08
Fix GHCup wrt #7061

hasufell:
# Description

Unbreak the permission on `/usr/local/.ghcup` that cause most new Haskell CIs (non-container envs) to fail at the moment.

## Check list

### DIFF
--- a/images/linux/scripts/installers/haskell.sh
+++ b/images/linux/scripts/installers/haskell.sh
@@ -43,6 +43,7 @@ echo "install cabal..."
 ghcup install cabal
 
 chmod -R 777 $GHCUP_INSTALL_BASE_PREFIX/.ghcup
+chown -R runner $GHCUP_INSTALL_BASE_PREFIX/.ghcup
 ln -s $GHCUP_INSTALL_BASE_PREFIX/.ghcup /etc/skel/.ghcup
 
 # Install the latest stable release of haskell stack
--- a/images/linux/scripts/installers/haskell.sh
+++ b/images/linux/scripts/installers/haskell.sh
@@ -43,7 +43,7 @@ echo "install cabal..."
 ghcup install cabal
 
 chmod -R 777 $GHCUP_INSTALL_BASE_PREFIX/.ghcup
-chown -R runner $GHCUP_INSTALL_BASE_PREFIX/.ghcup
+chown -R 1001 $GHCUP_INSTALL_BASE_PREFIX/.ghcup
 ln -s $GHCUP_INSTALL_BASE_PREFIX/.ghcup /etc/skel/.ghcup
 
 # Install the latest stable release of haskell stack
--- a/images/linux/scripts/tests/Haskell.Tests.ps1
+++ b/images/linux/scripts/tests/Haskell.Tests.ps1
@@ -1,4 +1,5 @@
 Describe "Haskell" {
+    $GHCupCommonPath = "/usr/local/.ghcup"
     $GHCCommonPath = "/usr/local/.ghcup/ghc"
     $GHCVersions = Get-ChildItem -Path $GHCCommonPath | Where-Object { $_.Name -match "\d+\.\d+" }
 
@@ -30,6 +31,10 @@ Describe "Haskell" {
 
     It "Stack" {
         "stack --version" | Should -ReturnZeroExitCode
+    }
+
+    It "GHCup dir has correct user permissions" {
+        "stat --format '%U' $GHCupCommonPath" | Should -Be "runner"
     }
 
     It "Stack hook is not installed" {


images/linux/scripts/tests/Haskell.Tests.ps1
```diff
@@ -32,6 +33,10 @@ Describe "Haskell" {
         "stack --version" | Should -ReturnZeroExitCode
     }
 
+    It "GHCup dir has correct user permissions" {
+        "stat --format '%U' $GHCupCommonPath" | Should -Be "${env:USER}"
```
hasufell, line 37: I'm not sure if this is correct powershell and will work... but this is the idea at least

images/linux/scripts/installers/haskell.sh
```diff
@@ -43,6 +43,7 @@ echo "install cabal..."
 ghcup install cabal
 
 chmod -R 777 $GHCUP_INSTALL_BASE_PREFIX/.ghcup
+chown -R "$USER" $GHCUP_INSTALL_BASE_PREFIX/.ghcup
```
hasufell, line 46: `$USER` is the default user in the bare ubuntu VM in the github action, right?

ddobranic: it is correct.

images/linux/scripts/tests/Haskell.Tests.ps1
ddobranic: variable name is `GHCCommonPath`

hasufell: No, this is a new variable

ddobranic: @hasufell test is failing because of $USER in the moment of changing permissions and performing the test is not the same as the <actionuser>. action user is 'runner'

hasufell: So what is the right way to invoke test as if you're executing a github action?

ddobranic: When changing owner permissions we should do:
`chown -R "runner" $GHCUP_INSTALL_BASE_PREFIX/.ghcup`
And for the test:
`"stat --format '%U' $GHCupCommonPath" | Should -Be "runner"`

hasufell: Can you add a commit fixing that? I'm unsure when I'll be at my PC.

hasufell: Fixed

hasufell:
Please re-run tests @ddobranic

ddobranic:
@hasufell unfortunately, runner user doesn't exist at the point where we are trying to change folder owner.
```
chown: invalid user: ‘runner’
```

ddobranic:
> Are the UIDs predictable/stable?

Should be 1001

hasufell:
@ddobranic updated

hasufell:
```
    azure-arm.build_vhd:   [-] GHCup dir has correct user permissions 302ms (300ms|2ms)
    azure-arm.build_vhd:    Expected strings to be the same, but they were different.
    azure-arm.build_vhd:    Expected length: 6
    azure-arm.build_vhd:    Actual length:   19
    azure-arm.build_vhd:    Strings differ at index 0.
    azure-arm.build_vhd:    Expected: 'runner'
    azure-arm.build_vhd:    But was:  'stat --format '%U' '
    azure-arm.build_vhd:               ^
    azure-arm.build_vhd:    at "stat --format '%U' $GHCupCommonPath" | Should -Be "runner", /imagegeneration/tests/Haskell.Tests.ps1:37
    azure-arm.build_vhd:    at <ScriptBlock>, /imagegeneration/tests/Haskell.Tests.ps1:37
```

hasufell:
So what now? I don't have enough knowledge of the test suite to figure out how to execute `stat` successfully there.

ddobranic:
I have tested on current VM and got this result:
```
$ GHCupCommonPath=/usr/local/.ghcup;stat --format '%U' $GHCupCommonPath
root
```
So, the test looks fine. Will check with the team for a fix.

hasufell:
Any news?

mikhailkoliada:
@hasufell the test you've proposed will not work, the pester test is being executed during the image generation where the `runner` user does not exist, there is no easy way of checking the permission on the image-spec level.

hasufell:
> @hasufell the test you've proposed will not work, the pester test is being executed during the image generation where the `runner` user does not exist, there is no easy way of checking the permission on the image-spec level.

The user doesn't have to exist, we can just check the UID?

But the whole 'stat' thing doesn't even seem to execute properly.

hasufell:
I just released a GHCup version that does not error out if it can't change modification or access time:

* https://github.com/haskell/ghcup-hs/releases/tag/v0.1.19.1
* https://github.com/haskell/ghcup-hs/commit/e03c5ee4a1b84d397e202fc924e9308ce6a50f1a

This might already fix this problem with a minor side-effect that it may re-download the metadata file more often. 

However, I still believe we want the owners of the directory set to the runner.

mikhailkoliada:
Left a comment in the original ticket.

PR #7175, opened 2023-02-22
[macOS] update macOS 13 template

Alexey-Ayupov:
# Description
Update macOS 13 template

## Check list

### DIFF
--- a/images/macos/toolsets/toolset-13.json
+++ b/images/macos/toolsets/toolset-13.json
@@ -2,10 +2,9 @@
     "xcode": {
         "default": "14.2",
         "versions": [
-            {"link": "14.3", "version": "14.3.0"},
-            {"link": "14.2", "version": "14.2.0"},
-            { "link": "14.1", "version": "14.1.0"},
-            { "link": "14.0.1", "version": "14.0.1", "symlinks": ["14.0"] }
+            { "link": "14.3", "version": "14.3.0" },
+            { "link": "14.2", "version": "14.2.0" },
+            { "link": "14.1", "version": "14.1.0" }
         ]
     },
     "xamarin": {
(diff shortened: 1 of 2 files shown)


images/macos/toolsets/toolset-13.json
```diff
@@ -4,8 +4,7 @@
         "versions": [
             {"link": "14.3", "version": "14.3.0"},
             {"link": "14.2", "version": "14.2.0"},
```
maxim-lobanov, line 6: nit:
```suggestion
            { "link": "14.3", "version": "14.3.0" },
            { "link": "14.2", "version": "14.2.0" },
```

images/macos/toolsets/toolset-13.json
```diff
@@ -112,15 +91,14 @@
...
     },
     "go": {
-        "default": "1.17"
+        "default": "1.18"
     },
     "node": {
-        "default": "16",
+        "default": "18",
         "nvm_versions": [
             "14",
```
maxim-lobanov, line 103: Should we remove `nvm_versions` since we don't install nvm to images?

maxim-lobanov: The same for `mongodb` and `postgresql`

images/macos/toolsets/toolset-13.json
Alexey-Ayupov: Fixed

images/macos/toolsets/toolset-13.json
Alexey-Ayupov: I believe it is ok to remove packages from the list that will not be installed.

maxim-lobanov (approved):

PR #7182, opened 2023-02-23
Add GitHub actions for image generation

Alexey-Ayupov:
# Description
In order to move "unstable" image generation from Azure pipelines to GitHub actions. 

## Check list

### DIFF
--- a/.github/workflows/macos-generation.yml
+++ b/.github/workflows/macos-generation.yml
@@ -1,0 +1,212 @@
+name: macOS image generation
+on:
+  workflow_call:
+    inputs:
+      image_label:
+        type: string
+        description: macOS codename
+        required: true
+      base_image_name:
+        type: string
+        description: Base clean image
+        required: true
+      template_path:
+        type: string
+        description: Packer template path
+        required: true
+      target_datastore:
+        type: string
+        description: Image datastore
+        required: true
+      custom_repo:
+        type: string
+        description: Custom repo to checkout
+        required: false
+      custom_repo_commit_hash:
+        type: string
+        description: Custom repo commit hash
+        required: false
+
+env:
+  KEYVAULT: imagegeneration
+  ESXI_CLUSTER: mcv2-build-unstable
+  VCENTER_DATACENTER: imagegen
+  OUTPUT_FOLDER: mms-output
+  BUILD_DATASTORE: ds-image
+
+defaults:
+  run:
+    shell: pwsh
+
+jobs:
+  build:
+    runs-on: macos-vmware
+    timeout-minutes: 1200
+    steps:
+    - uses: azure/login@v1
+      with:
+        creds: ${{ secrets.AZURE_CREDENTIALS }}
+
+    - name: Set image variables
+      run: |
+        $currentDate = Get-Date -Format "yyyyMMdd"
+        $templatePath = "${{ inputs.template_path }}"
+        $osName = $(($templatePath.Split("/")[-1]).Split(".")[0])
+        $virtualMachineName = "${osName}_${currentDate}_unstable.${{ github.run_id }}.${{ github.run_attempt }}"
+        $GitHubFeed = az keyvault secret show -n "github-feed-token" --vault-name "${{ env.KEYVAULT }}" --query value -o tsv
+        $VIUserName = az keyvault secret show -n "vcenter-username-v2" --vault-name "${{ env.KEYVAULT }}" --query value -o tsv
+        $VIPassword = az keyvault secret show -n "vcenter-password-v2" --vault-name "${{ env.KEYVAULT }}" --query value -o tsv
+        $XcodeUser = az keyvault secret show -n "xcode-installation-user" --vault-name "${{ env.KEYVAULT }}" --query value -o tsv
+        $XcodePassword =  az keyvault secret show -n "xcode-installation-password" --vault-name "${{ env.KEYVAULT }}" --query value -o tsv
+        echo "::add-mask::$GitHubFeed"
+        echo "::add-mask::$VIUserName"
+        echo "::add-mask::$VIPassword"
+        echo "::add-mask::$XcodeUser"
+        echo "::add-mask::$XcodePassword"
+        "GH_FEED=$GitHubFeed" | Out-File -Append -FilePath $env:GITHUB_ENV
+        "VI_USER_NAME=$VIUserName" | Out-File -Append -FilePath $env:GITHUB_ENV
+        "VI_PASSWORD=$VIPassword" | Out-File -Append -FilePath $env:GITHUB_ENV
+        "XCODE_USER=$XcodeUser" | Out-File -Append -FilePath $env:GITHUB_ENV
+        "XCODE_PASSWORD=$XcodePassword" | Out-File -Append -FilePath $env:GITHUB_ENV
+        "VM_NAME=$virtualMachineName" | Out-File -Append -FilePath $env:GITHUB_ENV
+
+    - name: Determine checkout type
+      run: |
+        if ("${{ inputs.custom_repo }}" -and "${{ inputs.custom_repo_commit_hash }}") {
+            $checkoutType = "custom_repo"
+        } elseif (("${{ github.event_name }}" -eq "pull_request_target") -and ("${{ github.event.action }}" -eq "labeled" )) {
+            $checkoutType = "pull_request"
+        } else {
+            $checkoutType = "main"
+        }
+        "CHECKOUT_TYPE=$checkoutType" | Out-File -Append $env:GITHUB_ENV
+
+    - name: Checkout repository
+      if: ${{ env.CHECKOUT_TYPE == 'main' }}
+      uses: actions/checkout@v3
+      with:
+        repository: actions/runner-images
+
+    - name: Checkout PR
+      if: ${{ env.CHECKOUT_TYPE == 'pull_request' }}
+      uses: actions/checkout@v3
+      with:
+        ref: ${{ github.event.pull_request.head.sha }}
+
+    - name: Checkout custom repository
+      if: ${{ env.CHECKOUT_TYPE == 'custom_repo' }}
+      uses: actions/checkout@v3
+      with:
+        repository: '${{ inputs.custom_repo }}'
+        ref: '${{ inputs.custom_repo_commit_hash }}'
+
+    - name: Validate contributor permissions
+      if: ${{ github.event_name == 'pull_request_target' }}
+      run: |
+        
+        [string]$contributorAllowList = "${{ vars.CONTRIBUTOR_ALLOWLIST }}"
+        ./images.CI/macos/validate-contributor.ps1 `
+          -RepositoryName ${{ github.repository }} `
+          -AccessToken ${{ env.GH_FEED }} `
+          -SourceBranch "refs/pull/${{ github.event.pull_request.number }}/merge" `
+          -ContributorAllowList $contributorAllowList
+
+    - name: Select datastore
+      run: |
+        ./images.CI/macos/select-datastore.ps1 `
+          -VMName "${{ env.VM_NAME }}" `
+          -VIServer ${{ secrets.VISERVER_V2 }} `
+          -VIUserName ${{ env.VI_USER_NAME }} `
+          -VIPassword ${{ env.VI_PASSWORD }} `
+          -Cluster ${{ env.ESXI_CLUSTER }}
+
+    - name: Build VM
+      run: |
+        $SensitiveData = @(
+          'IP address:',
+          'Using ssh communicator to connect:'
+        )
+         packer build -on-error=abort `
+        -var="vcenter_server=${{ secrets.VISERVER_V2 }}" `
+        -var="vcenter_username=${{ env.VI_USER_NAME }}" `
+        -var="vcenter_password=${{ env.VI_PASSWORD }}" `
+        -var="vcenter_datacenter=${{ env.VCENTER_DATACENTER }}" `
+        -var="cluster_or_esxi_host=${{ env.ESXI_CLUSTER }}" `
+        -var="esxi_datastore=${{ env.BUILD_DATASTORE }}" `
+        -var="output_folder=${{ env.OUTPUT_FOLDER }}" `
+        -var="vm_username=${{ secrets.VM_USERNAME }}" `
+        -var="vm_password=${{ secrets.VM_PASSWORD }}" `
+        -var="github_api_pat=${{ secrets.GH_FEED_TOKEN }}" `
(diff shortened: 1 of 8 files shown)


Doni108108 (approved):

.github/workflows/macos-generation.yml
```diff
@@ -0,0 +1,198 @@
...
+    shell: pwsh
+
+jobs:
+  build:
+    runs-on: maccloudv2
+    timeout-minutes: 1200
+    steps:
+    #- uses: actions/checkout@v3
+    - name: Checkout repository
+      if: ${{ (inputs.custom_repo == '' && inputs.custom_repo_hash == '') || github.event_name != 'pull_request_target'  }}
```
maxim-lobanov, line 52: Are you sure that this condition is correct?
I see a couple of cases where it won't work properly:
- `workflow_dispatch` event: in this case both `checkout repository` and `Checkout custom repository` will be invoked.
- `pull_request_target` event - the same.

May be move the calculation to the separate step?
```yml
- name: Determine checkout type
  run: |
    if ("${{ inputs.custom_repo }}" -and "${{ inputs.custom_repo_hash }}") {
        $checkoutType = "custom_repo"
    } elseif (("${{ github.event_name }}" -eq "pull_request_target") -and (${{ github.event.action }} -eq "labeled" )) {
        $checkoutType = "pull_request"
    } else {
        $checkoutType = "main"
    }
    "CHECKOUT_TYPE=$checkoutType" | Out-File -Append $env:GITHUB_ENV
```
And then use this variable in `if` conditions for checkout steps?
Right now, you have 3 separate conditions and it looks very unclear. Also, if you change one of them, you have to change two other conditions too.

.github/workflows/macos-generation.yml
```diff
@@ -0,0 +1,198 @@
...
+          -VIPassword "$VIPassword" `
+          -CpuCount "$cpuCount" `
+          -CoresPerSocketCount "$coresPerSocketCount" `
+          -Memory "$memory"
+
+    - name: Destroy VM (if build canceled only)
+      if: ${{ cancelled() }}
+      run: |
+        $VIUserName = az keyvault secret show -n "vcenter-username-v2" --vault-name "${{ env.KEYVAULT }}" --query value -o tsv
+        $VIPassword = az keyvault secret show -n "vcenter-password-v2" --vault-name "${{ env.KEYVAULT }}" --query value -o tsv
```
maxim-lobanov, line 192: Can we save these variables as environment variables to not read in every step?

.github/workflows/macos11.yml
```diff
@@ -0,0 +1,46 @@
+run-name: macOS-11_${{ vars.current_date }}_unstable.${{ github.run_id }}.${{ github.run_attempt }}
+on:
+  workflow_dispatch:
+    inputs:
+      CUSTOM_REPOSITORY:
+        description: 'Custom repository (owner/repo)'
+        required: false
+      CUSTOM_REPOSITORY_HASH:
```
maxim-lobanov, line 8: nit: may be `CUSTOM_REPOSITORY_REF` or `CUSTOM_REPOSITORY_COMMIT`?

.github/workflows/macos11.yml
```diff
@@ -0,0 +1,46 @@
...
+        shell: pwsh
+        run: |
+          $currentDate = Get-Date -Format "yyyyMMdd"
+          $uri = "https://api.github.com/repos/AleAiu/simpletest/actions/variables/CURRENT_DATE"
+          $header = @{"Authorization" = "Bearer ${{ secrets.REPO_TOKEN }}"}
+          $body = "{`"name`":`"CURRENT_DATE`",`"value`":`"$currentDate`"}"
+          Invoke-WebRequest -Method PATCH -Header $header -Uri $uri -Body $body -ContentType "application/json"
+
+  macOS_11:
+    if: ${{ (github.event.label.name == 'macos-all') || (github.event.label.name == 'macos-11') || (inputs.CUSTOM_REPOSITORY != '' && inputs.CUSTOM_REPOSITORY_HASH != '') || (github.event_name == 'schedule') }}
```
maxim-lobanov, line 34: Do we really need to repeat the full condition here?
This job depends on the first time so it should be auto-cancelled if first one is skipped.

.github/workflows/macos11.yml
```diff
@@ -0,0 +1,46 @@
...
+    runs-on: windows-latest
+    steps:
+      - name: run api
+        shell: pwsh
+        run: |
+          $currentDate = Get-Date -Format "yyyyMMdd"
+          $uri = "https://api.github.com/repos/AleAiu/simpletest/actions/variables/CURRENT_DATE"
+          $header = @{"Authorization" = "Bearer ${{ secrets.REPO_TOKEN }}"}
+          $body = "{`"name`":`"CURRENT_DATE`",`"value`":`"$currentDate`"}"
+          Invoke-WebRequest -Method PATCH -Header $header -Uri $uri -Body $body -ContentType "application/json"
```
maxim-lobanov, line 31: Why do we need this step? What is its purpose?

.github/workflows/macos11.yml
```diff
@@ -0,0 +1,46 @@
...
+    paths:
+    - 'images/macos/**'
+  schedule:
+    - cron: '0 0 * * *'
+
+jobs:
+  update_vars:
+    if: ${{ (github.event.label.name == 'macos-all') || (github.event.label.name == 'macos-11') || (inputs.CUSTOM_REPOSITORY != '' && inputs.CUSTOM_REPOSITORY_HASH != '') || (github.event_name == 'schedule') }}
+    name: Update Variables
+    runs-on: windows-latest
```
maxim-lobanov, line 22: Any reason to run this job on `windows-latest`? can we run it on `ubuntu-latest` instead? Assignment time for ubuntu is shorter and the number of free ubuntu vms is bigger too.

.github/workflows/macos-generation.yml
```diff
@@ -0,0 +1,198 @@
+name: macOS image generation
```
maxim-lobanov, line 1: Should we move reusable steps to the separate folder? `.github/workflows/steps`?

.github/workflows/ubuntu-win-generation.yml
```diff
@@ -0,0 +1,123 @@
...
+
+          if ( -not (Test-Path $TemplatePath) ) {
+            $TemplatePath = Join-Path $TemplateDirectoryPath "$ImageType.json"
+          }
+
+          echo "::set-env name=TemplatePath::$TemplatePath"
+          echo "::set-env name=TemplateDirectoryPath::$TemplateDirectoryPath"
+          echo "::set-env name=ImageType::$ImageType"
+        env:
+           ACTIONS_ALLOW_UNSECURE_COMMANDS: true
```
maxim-lobanov, line 66: Any reason why we can't use secure commands to set environment variables?

.github/workflows/ubuntu-win-generation.yml
```diff
@@ -0,0 +1,123 @@
...
+            -BuildId ${{ github.run_number }} `
+            -Organization ${{ env.RELEASE_TARGET_ORGANIZATION }} `
+            -DefinitionId ${{ env.RELEASE_TARGET_DEFINITION_ID }} `
+            -Project ${{ env.RELEASE_TARGET_PROJECT }} `
+            -ImageName ${{ env.ImageType }} `
+            -AccessToken ${{ secrets.RELEASE_TARGET_TOKEN }}
+        env:
+          RELEASE_TARGET_ORGANIZATION: mseng
+          RELEASE_TARGET_DEFINITION_ID: '3851'
+          RELEASE_TARGET_PROJECT: AzureDevOps
```
maxim-lobanov, line 111: Can we move these values to variables in repo settings? It is internal org and don't want to show any of these values in our YAMLs

.github/workflows/macos11.yml
maxim-lobanov: Discussed offline. It is a workaround to use date in run-name. 
I recommend to avoid this approach and just use `run_id` in run_name.

.github/workflows/macos-generation.yml
mikhailkoliada: @maxim-lobanov no, they will be revealed during step initialization as plain text

.github/workflows/macos-generation.yml
```diff
@@ -0,0 +1,198 @@
...
+        required: true
+      target_datastore:
+        type: string
+        description: Image datastore
+        required: true
+      custom_repo:
+        type: string
+        description: Custom repo to checkout
+        required: false
+      custom_repo_hash:
```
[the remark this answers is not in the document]
Alexey-Ayupov: What about `custom_repo_commit_hash`? Does it sound more clear?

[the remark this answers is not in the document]
maxim-lobanov: Yep, sounds good to me.

.github/workflows/macos-generation.yml
Alexey-Ayupov: It looks like reusable workflows must still reside in the .github/workflows directory
https://github.com/actions/runner/issues/1493

.github/workflows/ubuntu-win-generation.yml
Alexey-Ayupov: Do you mean using something like `"VAR=value" >> $env:GITHUB_ENV` ?

.github/workflows/macos-generation.yml
maxim-lobanov: Oh, you are right. Only reusable actions can be moved to different folder 😢

.github/workflows/ubuntu-win-generation.yml
maxim-lobanov: Yes, exactly.
```pwsh
"VAR=value" | Out-File -Append -FilePath $env:GITHUB_ENV
```

I think this old unsecure way will be deprecated soon or later so it makes sense to switch to the new approach.

.github/workflows/macos-generation.yml
```diff
@@ -0,0 +1,198 @@
...
+defaults:
+  run:
+    shell: pwsh
+
+jobs:
+  build:
+    runs-on: maccloudv2
+    timeout-minutes: 1200
+    steps:
+    #- uses: actions/checkout@v3
```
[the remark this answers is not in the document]
Alexey-Ayupov: Removed

.github/workflows/macos-generation.yml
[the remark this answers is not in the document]
Alexey-Ayupov: Changed to custom_repo_commit_hash

.github/workflows/macos-generation.yml
Alexey-Ayupov: Updated according to suggestion

.github/workflows/macos11.yml
Alexey-Ayupov: Changed to CUSTOM_REPOSITORY_COMMIT_HASH

.github/workflows/macos11.yml
Alexey-Ayupov: Removed

.github/workflows/macos11.yml
Alexey-Ayupov: Removed (belong to the removed job)

.github/workflows/ubuntu-win-generation.yml
Alexey-Ayupov: Updated

.github/workflows/ubuntu-win-generation.yml
Alexey-Ayupov: Changed to secrets

.github/workflows/macos-generation.yml
```diff
@@ -0,0 +1,212 @@
...
+  BUILD_DATASTORE: ds-image
+
+defaults:
+  run:
+    shell: pwsh
+
+jobs:
+  build:
+    runs-on: maccloudv2
+    timeout-minutes: 1200
```
mikhailkoliada, line 44: we should change the agent tags, lets pick `azure-ubuntu` for ubuntu builds `azure-windows` for windows ones and `macos-vmware` for current macs.

@Alexey-Ayupov @maxim-lobanov ideas?

Alexey-Ayupov: I believe the tag should represent the environment where it locates. I suggest `agent-vmware` and `agent-azure` or something similar.

.github/workflows/macos-generation.yml
```diff
@@ -0,0 +1,212 @@
...
+    steps:
+    - name: Set image varibles
```
maxim-lobanov, line 46:
```suggestion
    - name: Set image variables
```

.github/workflows/macos-generation.yml
```diff
@@ -0,0 +1,212 @@
...
+        echo "Copy image output files"
+        cp -R "images/image-output/software-report/." "${{ runner.temp }}/artifacts"
+
+        echo "Put VM name to 'VM_Done_Name' file"
+        echo "${{ env.VM_NAME }}" > "${{ runner.temp }}/artifacts/VM_Done_Name"
+
+    - name: Print markdown software report
+      shell: bash
+      run: |
+        cat "${{ runner.temp }}/artifacts/systeminfo.md"
```
maxim-lobanov, line 165:
```suggestion
      run: |
        Get-Content "${{ runner.temp }}/artifacts/systeminfo.md"
```

.github/workflows/macos-generation.yml
```diff
@@ -0,0 +1,212 @@
...
+
+    - name: Print json software report
+      shell: bash
+      run: |
+        cat "${{ runner.temp }}/artifacts/systeminfo.json"
```
maxim-lobanov, line 170:
```suggestion
      run: |
        Get-Content "${{ runner.temp }}/artifacts/systeminfo.json"
```

.github/workflows/macos-generation.yml
maxim-lobanov: Why can't we mark them as secret?
https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#example-masking-an-environment-variable

maxim-lobanov (approved):

.github/workflows/macos-generation.yml
Alexey-Ayupov: fixed

.github/workflows/macos-generation.yml
Alexey-Ayupov: Updated

.github/workflows/macos-generation.yml
Alexey-Ayupov: Updated

maxim-lobanov (approved):

PR #7214, opened 2023-03-01
[macOS] Use variable instead of hard coded value

ddobranic:
# Description
Use the interval value in retry function
#### Related issue:
https://github.com/actions/runner-images/issues/7208
## Check list

### DIFF
--- a/images/macos/helpers/Common.Helpers.psm1
+++ b/images/macos/helpers/Common.Helpers.psm1
@@ -131,7 +131,7 @@ function Start-DownloadWithRetry {
                 exit 1
             }
 
-            Write-Host "Waiting '$Interval' seconds before retrying. Retries left: $Retries"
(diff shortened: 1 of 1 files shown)


images/macos/helpers/Common.Helpers.psm1
```diff
@@ -130,8 +131,8 @@ function Start-DownloadWithRetry {
                 exit 1
             }
 
-            Write-Host "Waiting 30 seconds before retrying. Retries left: $Retries"
-            Start-Sleep -Seconds 30
+            Write-Host "Waiting '$Interval' seconds before retrying. Retries left: $Retries"
```
mikhailkoliada, line 134: we can drop quotes on $Interval

ddobranic: Updated.

sergei-pyshnoi (approved):

mikhailkoliada (approved):

PR #7337, opened 2023-03-24
[Windows] Add script for OpenSSL installation

vpolikarpov-akvelon:
# Description

Add script to install OpenSSL using executable installer rather then choco package. This is mainly to be able to select which version to install and to switch between light and full variants. The problem with choco is that light package doesn't provide up-to-date patch for OpenSSL 1.1.1 and full package doesn't have versions for OpenSSL 3.x.

Script uses repo https://github.com/slproweb/opensslhashes to get information about currently available versions of Win-OpenSSL installers, filters out installers that don't satisfy the requirements provided and selects the latest version among remaining ones.

Currently it is configured to install full variant of the version 1.1.1.

## Check list

### DIFF
--- a/images/win/scripts/Installers/Install-OpenSSL.ps1
+++ b/images/win/scripts/Installers/Install-OpenSSL.ps1
@@ -1,0 +1,37 @@
+################################################################################
+##  File:  Install-OpenSSL.ps1
+##  Desc:  Install win64-openssl.
+################################################################################
+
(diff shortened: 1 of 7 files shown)


mikhailkoliada (approved):

MehdiChinoune:
This breaks building packages on MSYS2, please remove it from PATH.
Add an environment variable for its location instead.

PR #7407, opened 2023-04-07
[Ubuntu22] Allow Managed Image builds

fabiocastagnino:
# Description
Improvement
Added options to allow Managed Image builds

## Check list

### DIFF
--- a/images/linux/ubuntu2204.pkr.hcl
+++ b/images/linux/ubuntu2204.pkr.hcl
@@ -14,6 +14,11 @@ variable "build_resource_group_name" {
   default = "${env("BUILD_RESOURCE_GROUP_NAME")}"
 }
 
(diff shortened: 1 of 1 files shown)


mikhailkoliada:
Hello! Managed images is not really a frequent case or demand, that is why we would not like adding it to the existing specs.

fabiocastagnino:
Hello! 
The actual build method are deprecated: 
- https://learn.microsoft.com/en-us/azure/virtual-machines/unmanaged-disks-deprecation.
In our Azure subscription, we start getting error building with storage account method, with Managed Image, we can build the runners.
My PR are small improvement, and keep the compatibility with both build method. This upgrade should be done.

I hope you can reconsider this PR.

Thank you

PR #7409, opened 2023-04-07
Fix ip address whiltelist value for hcl2 templates

vpolikarpov-akvelon:
# Description
The value that is used for allowed_inbound_ip_addresses argument with HCL2 template is wrong so GenerateResourcesAndImage function fails when RestrictToAgentIpAddress switch is on and ImageType is Ubuntu2204. This request is to fix the value.

## Check list


erik-bershel (approved):

MrLucifer92:
Received

mikhailkoliada (approved):

PR #7453, opened 2023-04-18
updates for MacOS M1/arm64 runners

finaroux:
# Description
As discussed internally with @mikhailkoliada, submitting changes required to build MacOS images for ARM64 runners.
It was tested with an image generated locally on a test rack and unblocked us for validating our Github workflows.

## Check list

### DIFF
--- a/images/macos/provision/bootstrap-provisioner/installNewProvisioner.sh
+++ b/images/macos/provision/bootstrap-provisioner/installNewProvisioner.sh
@@ -5,8 +5,13 @@ ProvisionerScriptUri="$3"
 ScriptName="$4"
 ScriptParam="$5"
 Username="$6"
+arch=$(arch)
 
-export PATH=/usr/bin:/usr/sbin:/usr/local/bin:/bin:/sbin:/opt/homebrew/bin
+if [[ $arch == "arm64" ]]; then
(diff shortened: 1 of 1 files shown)


mikhailkoliada (dismissed):

images/macos/provision/bootstrap-provisioner/installNewProvisioner.sh
```diff
@@ -6,11 +6,11 @@ ScriptName="$4"
 ScriptParam="$5"
 Username="$6"
 
-export PATH=/usr/bin:/usr/sbin:/usr/local/bin:/bin:/sbin
+export PATH=/usr/bin:/usr/sbin:/usr/local/bin:/bin:/sbin:/opt/homebrew/bin
```
mikhailkoliada, line 9: @finaroux I am wondering if we should make PATH arch-conditional and add /opt/homebrew/bin only if `$(arch) == arm64`?

finaroux: If it is what you are doing in your other PR that makes sense. I will update it.

mikhailkoliada (approved):

makslevental:
I'm sorry if this is the wrong place but I found this PR by googling around about M1 runners entering public beta (according to https://github.com/github/roadmap/issues/528). I can glean from this PR that something is working somewhere but I'm not sure where or how. So is there any info using M1 runners around?

PR #7478, opened 2023-04-24
[ubuntu] script after reboot need more time before executing

nadavshaham:
# Description

When trying to build the ubuntu 22.04 image (also relevant for ubuntu20.04), there is a provisioner in  `images/linux/ubuntu2204.pkr.hcl` which does a reboot:

    provisioner "shell" {
    execute_command   = "/bin/sh -c '{{ .Vars }} {{ .Path }}'"
    expect_disconnect = true
    scripts           = ["${local.scripts_folder}/base/reboot.sh"]
    }

the next provisioner has pause_before of "1m0s":

    provisioner "shell" {
    execute_command     = "sudo sh -c '{{ .Vars }} {{ .Path }}'"
    pause_before        = "1m0s"
    scripts             = ["${local.scripts_folder}/installers/cleanup.sh"]
    start_retry_timeout = "10m"
  }

When running the pipeline I get:
`Error uploading script: Error copying input data into local temporary file. Check that TEMPDIR has enough space`.
After running this multiple times the pipeline succeeded once, i figured that the "pause_before" was too short and the VM hasn't finished rebooting and ssh service was still down, after changing the value in cleanup.sh to `pause_before = "2m0s" `  the code consistently works.

## Check list

### DIFF
--- a/images/linux/ubuntu2004.json
+++ b/images/linux/ubuntu2004.json
@@ -312,7 +312,7 @@
         },
         {
             "type": "shell",
-            "pause_before": "60s",
+            "pause_before": "2m0s",
             "start_retry_timeout": "10m",
             "scripts": [
                 "{{template_dir}}/scripts/installers/cleanup.sh"
--- a/images/linux/ubuntu2204.pkr.hcl
+++ b/images/linux/ubuntu2204.pkr.hcl
@@ -369,7 +369,7 @@ build {
 
   provisioner "shell" {
     execute_command     = "sudo sh -c '{{ .Vars }} {{ .Path }}'"
-    pause_before        = "1m0s"
+    pause_before        = "2m0s"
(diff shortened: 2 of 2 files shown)


Swampen:
The issue was for Ubuntu 20.04, but I guess the same applies to 22.04. 
Could you include the same change for 20.04 as well?

nadavshaham:
@Swampen done

nadavshaham:
this also looks like a lousy solution, i am currently trying to check for alternative ways that wait until the instance is actually up and ready rather then hoping it takes some arbitrary time to reboot

mikhailkoliada:
Hello! We do not observe anything connected with the error in nightly builds. Please apply this change locally if you still need it. If you have questions please feel free to reach us out again.

PR #7568, opened 2023-05-10
README: distinguish announcements section

mikhailkoliada:
# Description

For visibility

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,7 @@
 - [About](#about)
 - [Available Images](#available-images)
 - [Ongoing migrations](#ongoing-migrations)
+- [Announcements](#announcements)
 - [Image Definitions](#image-definitions)
 - [Image Releases](#image-releases)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -47,6 +48,10 @@ To build a VM machine from this repo's source, see the [instructions](docs/creat
 
 There are no migrations in progress.
 
+## Announcements
+
+You can also view notable upcoming changes by viewing the [Announcement](https://github.com/actions/runner-images/labels/Announcement) label.
+
```
samvantran, line 54: Wondering if we should rename `Ongoing migrations` to `Announcements` and when we have a migration we can simply make an announcement issue for it. wdyt?

```suggestion
## Announcements

See notable upcoming changes by viewing the issues with the [Announcement](https://github.com/actions/runner-images/labels/Announcement) label.


```

mikhailkoliada: @samvantran works for me

PR #7602, opened 2023-05-16
CodeQL: Simplify the toolcache version number for bundles tagged using semver

henrymercer:
# Description

We are migrating to tagging the CodeQL bundle using semantic versions (like `codeql-bundle-v2.13.3`) rather than a date-based versions (like `codeql-bundle-20230516`). This lets us simplify how we version these bundles within the toolcache.

Previously, we used a version number like `2.13.3-20230516` to allow us to look up the version in the toolcache by either the CLI version number (2.13.3) or the bundle tag name (`codeql-bundle-20230516`).  This PR changes how we version bundles in the toolcache to take advantage of the new bundle tag name.  For bundles versioned as `codeql-bundle-v2.13.3`, we can just version the bundle in the toolcache as `2.13.3` since the bundle tag name is derivable from the CLI version number.

## Check list
- [x] Related issue / work item is attached
- [x] Tests are written (if applicable)
- [x] Documentation is updated (if applicable)
- [x] Changes are tested and related VM images are successfully generated

PR #7717, opened 2023-06-13
[macOS] CodeQL: Simplify the toolcache version number for bundles tagged using semver

henrymercer:
# Description

(Originally opened as a cross-platform PR here: #7602)

We are migrating to tagging the CodeQL bundle using semantic versions (like `codeql-bundle-v2.13.3`) rather than a date-based versions (like `codeql-bundle-20230516`). This lets us simplify how we version these bundles within the toolcache.

Previously, we used a version number like `2.13.3-20230516` to allow us to look up the version in the toolcache by either the CLI version number (2.13.3) or the bundle tag name (`codeql-bundle-20230516`).  This PR changes how we version bundles in the toolcache to take advantage of the new bundle tag name.  For bundles versioned as `codeql-bundle-v2.13.3`, we can just version the bundle in the toolcache as `2.13.3` since the bundle tag name is derivable from the CLI version number.

## Check list

### DIFF
--- a/images/macos/provision/core/codeql-bundle.sh
+++ b/images/macos/provision/core/codeql-bundle.sh
@@ -1,16 +1,36 @@
 #!/bin/bash -e -o pipefail
 source ~/utils/utils.sh
 
-# Retrieve the name of the CodeQL bundle preferred by the Action (in the format codeql-bundle-YYYYMMDD).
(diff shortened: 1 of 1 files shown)


mikhailkoliada:
(gh workflows slightly not ready here yet)

mikhailkoliada:
(now GitHub ready too)

Issue #7855, opened 2023-07-05
"brew update" on MacOS 13 fails with an error: error: https://github.com/Homebrew/homebrew-cask did not send all necessary objects


lsh123:
### Description

MacOSX 13 builds fail (e.g. https://github.com/lsh123/xmlsec/actions/runs/5458349250):

```
Run brew update
error: Could not read ffac597fdb6ce8236b5e2da77ab6fbae3d9d9cb7
fatal: revision walk setup failed
error: https://github.com/Homebrew/homebrew-cask did not send all necessary objects

error: Could not read 623226b77a148708197a10c0d1eab3f8d6ca005b
fatal: revision walk setup failed
error: https://github.com/Homebrew/homebrew-core did not send all necessary objects

Error: Fetching /usr/local/Homebrew/Library/Taps/homebrew/homebrew-cask failed!
```

### Platforms affected

- [ ] Azure DevOps
- [x] GitHub Actions - Standard Runners
- [ ] GitHub Actions - Larger Runners

### Runner images affected

- [ ] Ubuntu 20.04
- [ ] Ubuntu 22.04
- [ ] macOS 11
- [ ] macOS 12
- [X] macOS 13
- [ ] Windows Server 2019
- [ ] Windows Server 2022

### Image version and build link

https://github.com/lsh123/xmlsec/actions/runs/5458349250

```
Current runner version: '2.305.0'
Operating System
Runner Image
Runner Image Provisioner
GITHUB_TOKEN Permissions
Secret source: Actions
Prepare workflow directory
Prepare all required actions
Getting action download info
Download action repository 'actions/checkout@v2' (SHA:ee0669bd1cc54295c223e0bb666b733df41de1c5)
Complete job name: check-osx
```

### Is it regression?

It worked on same version last week (https://github.com/lsh123/xmlsec/actions/runs/5384482840/jobs/9772463248)

### Expected behavior

Builds work

### Actual behavior

Builds fail (see details above)

### Repro steps

See https://github.com/lsh123/xmlsec/actions/runs/5458349250

shamil-mubarakshin:
Hello @lsh123,
The seems to be an issue when cloning `homebrew-cask`. You can track in https://github.com/Homebrew/homebrew-cask/issues/150323, which also provides workaround options. E.g. adding following action allowed `brew update` to complete:
```
    - name: Set up Homebrew
      id: set-up-homebrew
      uses: Homebrew/actions/setup-homebrew@master
```

[...]

Issue #7134, opened 2023-02-17
Add homebrew-core and homebrew-cask taps


shivammathur:
### Description

After Homebrew 4.0.0 release, the brew install script does not install the `homebrew-core` and the `homebrew-cask` taps. This is fine for installing software since homebrew now uses an API to get the formula info.

But since these images are also used for building/releasing software using brew, these taps are required on disk and should be added to the images.

Unless added back, installing them during runtime takes 4 to 5 minutes in total.

### Platforms affected

- [ ] Azure DevOps
- [X] GitHub Actions - Standard Runners
- [X] GitHub Actions - Larger Runners

### Runner images affected

- [X] Ubuntu 18.04
- [X] Ubuntu 20.04
- [X] Ubuntu 22.04
- [X] macOS 11
- [X] macOS 12
- [ ] Windows Server 2019
- [ ] Windows Server 2022

### Image version and build link

Currently, the taps are missing in ubuntu-22.04 (20230217.1), but they will be missing in all new ubuntu and macOS images, unless added after homebrew is installed or `HOMEBREW_NO_INSTALL_FROM_API` is set before installation.

### Is it regression?

Yes

### Expected behavior

The `homebrew-core` and `homebrew-cask` taps are installed by default on Ubuntu and macOS images.

### Actual behavior

Since 4.0.0 Homebrew release the install script does not install the `homebrew-core` and `homebrew-cask` taps, so they are missing on the new images.

### Repro steps

These directories should have the taps

```bash
# homebrew-core
ls $(brew --prefix)/Homebrew/Library/Taps/homebrew/homebrew-core

# homebrew-cask
ls $(brew --prefix)/Homebrew/Library/Taps/homebrew/homebrew-cask
```

mikhailkoliada:
@shivammathur fixed on macOS, on Ubuntu we would not like to add it back as we generally try to minimize brew's presence on the image in favour of using the [actions/setup-homebrew](https://github.com/Homebrew/actions/tree/master/setup-homebrew) action. Homebrew on linux is a subject of further deprecation.

mikhailkoliada:
Deployed :)

MikeMcQuaid:
> But since these images are also used for building/releasing software using brew, the core tap is required on disk and should be added to the images.

This isn't true. Can you please consider looping in Homebrew folks in future to discuss these sorts of changes?

PR #7861, opened 2023-07-05
[macOS] Don't set HOMEBREW_NO_INSTALL_FROM_API by default

Bo98:
# Description

While I agree this probably made sense a few months ago, there is no little reason to use `HOMEBREW_NO_INSTALL_FROM_API` given brew now automatically sets it where necessary (https://github.com/Homebrew/brew/commit/68289f1165f9cb285eb3a164fe285529cbc8b9fb and https://github.com/Homebrew/brew/commit/bb7aef85db959a7b1e9007f601515702b7633782), plus we currently have fixed outstanding issues that may have required switching off API, in particular from a refactor to the dependency system (https://github.com/Homebrew/brew/commit/a5a728a4df6f5d20befe21859df37d8bf8a887f6).

If anything still requires `HOMEBREW_NO_INSTALL_FROM_API` it will be a bug, and we're not aware of any such cases. You will only need to set it if you are modifying core formulae and casks locally, [as per the documentation](https://docs.brew.sh/FAQ#can-i-edit-formulae-myself). Even third-party taps do not need to set it for their own formulae and casks.

The change here gives parity to how Homebrew works for users on their machines. The API is the default mode. During the ongoing issues with the git repos that are currently being worked on by the git systems team (see #7855), we have seen comparatively little reports of breakages for local Homebrew usage. GitHub Actions has been where most issue reports have come from. (If you have issues with `brew tap` in image generation, feel free to add retries or ask for an update from the people at GitHub working on that issue.)

Note the original request of #7134 has been retained. A local version of homebrew-core and homebrew-cask is still available, should anyone need it.

## Check list


MikeMcQuaid (approved):
Homebrew Project Leader here 👋🏻: this would be actively desirable for our users (who report issues to us based on this configuration).

Issue #7933, opened 2023-07-19
Ubuntu 22.04 VHD fails to build due to issue during install of ChromeDriver


PottsSG:
### Description

On or around July 18, 2023 the latest stable version of Chrome and ChromeDriver changed from major version 114 to major version 115.  As per an announcement on the [ChromeDriver page](https://chromedriver.chromium.org/), the Downloads page only has versions up to 114.

When building a VHD for Ubuntu 22.04 (and possibly other linux vhds), the image creation fails with an error 

````
azure-arm.build_vhd: Downloading chromedriver v<?xml version='1.0' encoding='UTF-8'?><Error><Code>NoSuchKey</Code><Message>The specified key does not exist.</Message><Details>No such object: chromedriver/LATEST_RELEASE_115.0.5790</Details></Error>...
    azure-arm.build_vhd: Downloading 'https://chromedriver.storage.googleapis.com/<?xml' to 'version='1.0'/encoding='UTF-8'?><Error><Code>NoSuchKey</Code><Message>The'...
    azure-arm.build_vhd: Error — Either HTTP response code for 'https://chromedriver.storage.googleapis.com/<?xml' is wrong - '' or exit code is not 0 - '1'. Waiting 30 seconds before the next attempt, 19 attempts left
````

### Platforms affected

- [ ] Azure DevOps
- [X] GitHub Actions - Standard Runners
- [ ] GitHub Actions - Larger Runners

### Runner images affected

- [x] Ubuntu 20.04
- [X] Ubuntu 22.04
- [ ] macOS 11
- [ ] macOS 12
- [ ] macOS 13
- [ ] Windows Server 2019
- [x] Windows Server 2022

### Image version and build link

Latest from the main branch.
````
Latest commit:
Date: Tue Jul 18 17:47:39 2023 +0200; Commit: 0f7d7a82caa1a89cac0b656cf840ada34caaa37e - [macos] download XCode installer with retries (#7898); Author: ilia-shipitsin <125650415+ilia-shipitsin@users.noreply.github.com>
````

### Is it regression?

No

### Expected behavior

When running `/helpers/GenerateResourcesAndImage.ps1` with `-ImageType ubuntu2204` the image generation process should complete successfully.  Part way through the log the ChromeDriver download succeeds:

````
    azure-arm.build_vhd: Downloading chromedriver v114.0.5735.90...
[...]

PR #7944, opened 2023-07-20
Updated for changes to Chrome/ChromeDriver 115

ManagedUx:
# Description
Fixes issue #7933 due to chromedrive changes


ManagedUx:
Also handles the downloads in the same way as the previous version such that the issue in the initial fix in #7942 does not occur.  The issue with the original fix is that it's pulling the full build number, however ChromeEnterprise is currently at version 115.0.5790.99, but ChromeDriver is currently 115.0.5790.98.  This fix utilizes the 'latest-patch-versions-per-build-with-downloads.json' file that will provide the latest ChromeDriver version based on Major.Minor.Build.

PR #7975, opened 2023-07-25
macOS bashrc: don't cleanup Homebrew on installation.

MikeMcQuaid:
This is unnecessary in ephemeral environments like GitHub Actions. `HOMEBREW_CLEANUP_PERIODIC_FULL_DAYS` avoided the longer periodic cleanup but this is a better fix for speeding up Homebrew a bit in GitHub Actions.


mikhailkoliada (approved):

MikeMcQuaid:
Thanks for merge @mikhailkoliada !

PR #8106, opened 2023-08-17
Check only kubectl tool version - not the server version

luka5:
# Description
When reading the error message in https://github.com/actions/runner-images/issues/8094, it is clearly fixed in https://github.com/actions/runner-images/pull/8098. However, we still do try to connect to a server endpoint. It's not a bug but not nice either.

By using `--client` we no longer get `The connection to the server localhost:8080 was refused - did you specify the right host or port?`. See https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#version for more details

## Check list

### DIFF
--- a/images/linux/scripts/tests/Tools.Tests.ps1
+++ b/images/linux/scripts/tests/Tools.Tests.ps1
@@ -284,7 +284,7 @@ Describe "Kubernetes tools" {
     }
 
     It "kubectl" {
-        "kubectl version" | Should -MatchCommandOutput "Client Version: v"
+        "kubectl version --client" | Should -MatchCommandOutput "Client Version: v"
     }
 
     It "helm" {


images/linux/scripts/tests/Tools.Tests.ps1
```diff
@@ -284,7 +284,7 @@ Describe "Kubernetes tools" {
     }
 
     It "kubectl" {
-        "kubectl version" | Should -MatchCommandOutput "Client Version: v"
+        "kubectl version --client" | Should -MatchCommandOutput "Client Version: v"
```
luka5, line 287: I just saw it in the window script: https://github.com/actions/runner-images/blob/main/images/win/scripts/Tests/Tools.Tests.ps1#L98

We could also change it to `"kubectl version --client=true"`, if you want it to be the same.

vpolikarpov-akvelon:
Hey @luka5. Thank you for your contribution. Unfortunately we are unable to accept your PR directly due to some temporary technical issues. Sorry for that. I've created #8112 with your suggestion and some other related fixes.

PR #8132, opened 2023-08-22
Switch to build managed image instead of VHD

vpolikarpov-akvelon:
# Description

As Microsoft is going to begin deprecating unmanaged disks on September 13, 2022 we have to update our packer templates and pipelines to build managed images instead of VHDs

## Check list


images.CI/linux-and-win/InstallAzCopy.ps1
```diff
@@ -0,0 +1,28 @@
+Write-Host 'Acquiring AzCopy executable...'
```
mikhailkoliada, line 1: lets not install azcopy in runtime and pre-install it onto the agents, I think @Alexey-Ayupov can help with this

images.CI/linux-and-win/convert-to-vhd.ps1
```diff
@@ -0,0 +1,211 @@
...
+
+    [string] [Parameter(Mandatory = $true)] $ResourceGroupName,
+    [string] [Parameter(Mandatory = $true)] $StorageAccountName,
+    [string] [Parameter(Mandatory = $true)] $ContainerName,
+    [string] [Parameter(Mandatory = $true)] $ImageBlobName,
+    
+    [string] [Parameter(Mandatory = $true)] $TenantId
+  )
+
+  Write-Host "List the access keys or Kerberos keys for a $StorageAccountName storage account."
```
mikhailkoliada, line 33: I we probably can use SP rather than sas token to avoid security concerns as we had in prod, at least worth checking

images.CI/linux-and-win/convert-to-vhd.ps1
```diff
@@ -0,0 +1,211 @@
...
+# Create Azure Managed Disk from Shared Image Gallery
+Write-Host "Creating Azure Managed Disk '$ManagedImageName' from Shared Image Gallery..."
+az disk create `
+  --resource-group $ResourceGroupName `
+  --name $ManagedImageName `
+  --location $Location `
+  --gallery-image-reference "/subscriptions/$SubscriptionId/resourceGroups/$ResourceGroupName/providers/Microsoft.Compute/galleries/$GalleryName/images/$imageDefinitionName/versions/$GalleryImageVersion"
+
+# Generate SAS URL for the Managed Disk
+Write-Host "Generating SAS URL for the Managed Disk '$ManagedImageName'..."
```
mikhailkoliada, line 116: same as above, hmm, but not sure we can get rid of sas here exactly

images.CI/linux-and-win/convert-to-vhd.ps1
```diff
@@ -0,0 +1,211 @@
...
+az disk delete `
+  --resource-group $ResourceGroupName `
+  --name $ManagedImageName `
+  --only-show-errors `
+  --yes
+if ($LastExitCode) {
+  Write-Host "Warning: Failed to delete the Managed Disk '$ManagedImageName'."
+}
+
+# Delete Image Version from Shared Image Gallery
```
mikhailkoliada, line 197: I think we can leave the version, do we have gallery limits? cc @Alexey-Ayupov

images.CI/linux-and-win/InstallAzCopy.ps1
Alexey-Ayupov: We can get rid of azcopy and use `az storage blob copy start` command if we would like. (https://learn.microsoft.com/en-us/azure/virtual-machines/scripts/copy-managed-disks-vhd-to-storage-account)

images.CI/linux-and-win/convert-to-vhd.ps1
Alexey-Ayupov: It looks like SAS is the only option here.

images.CI/linux-and-win/convert-to-vhd.ps1
Alexey-Ayupov: It will overcomplicate our CI. Our SPN has no certificate authorization, so firstly we need to create a cert for it, store it in some KV, etc. 
To simplify the solution we can use `az storage blob copy start` command, by login to Azure using OIDC.

images.CI/linux-and-win/create-release.ps1
Alexey-Ayupov, line 1: I would think about getting rid of the ADO part here. Workflow based on `Compute Gallery image versions` can replace the old approach with `pipeline releases` in my opinion.

images.CI/linux-and-win/convert-to-vhd.ps1
vpolikarpov-akvelon: What do you mean? If we just print value here, then nobody will track the value and hitting the limit won't be expected by anyone.

images.CI/linux-and-win/InstallAzCopy.ps1
vpolikarpov-akvelon: Replaced `azcopy` with `az storage blob copy start`

images.CI/linux-and-win/azure-pipelines/image-generation.yml
```diff
@@ -71,28 +71,71 @@ jobs:
...
+                 -ClientId $(CLIENT_ID) `
+                 -ClientSecret $(CLIENT_SECRET) `
+                 -TenantId $(AZURE_TENANT)
+
+  - task: PowerShell@2
+    displayName: 'Remove managed image'
+    inputs:
+      targetType: 'inline'
+      script: |
+        az login --service-principal `
```
Alexey-Ayupov, line 129: Can we move this task above "Convert managed image to VHD" and connect to Azure only once?

images.CI/linux-and-win/azure-pipelines/image-generation.yml
```diff
@@ -71,28 +71,71 @@ jobs:
...
+          --username "$(CLIENT_ID)" `
+          --password "$(CLIENT_SECRET)" `
+          --tenant "$(AZURE_TENANT)" | Out-Null
+
+        az image delete `
```
Alexey-Ayupov, line 134: Should we move it to the "cleanup.ps1" script?

vpolikarpov-akvelon: I think it'd be better to move it to `convert-to-vhd.ps1` as it relates to image conversion while `cleanup.ps1` is about removing resources that were possibly left by packer.

images.CI/linux-and-win/convert-to-vhd.ps1
Alexey-Ayupov: Ok. Maybe we should set the retention period to 7 days and delete image versions after the date expiration? For example `Get-AzGalleryImageVersion` returns "PublishedDate" that can be used.

images.CI/linux-and-win/azure-pipelines/image-generation.yml
vpolikarpov-akvelon: It's not essential since I moved image removal to conversion script.

images.CI/linux-and-win/convert-to-vhd.ps1
```diff
@@ -0,0 +1,173 @@
...
+  --gallery-name $GalleryName `
+  --gallery-image-definition $imageDefinitionName `
+  --gallery-image-version $GalleryImageVersion `
+  --only-show-errors
+if ($LastExitCode) {
+  Write-Host "Warning: Failed to delete the Image Version '$GalleryImageVersion'."
+}
+
+# Logout from Azure
+az logout
```
Alexey-Ayupov, line 186: Do we need `az logout`?

vpolikarpov-akvelon: Well, it looks useless here indeed

images.CI/linux-and-win/convert-to-vhd.ps1
vpolikarpov-akvelon: I think it's better to implement this outside of build pipeline. Don't we already have something to remove outdated resources from our subscription?

vpolikarpov-akvelon: Anyway, why do we even want to keep image version in gallery?

Alexey-Ayupov (approved):

PR #8170, opened 2023-08-26
[macos] add visionOS SDK on macos-13

ilia-shipitsin:
# Description

add visionOS SDK on macos-13

#### Related issue:

https://github.com/actions/runner-images/issues/8144

## Check list

### DIFF
--- a/images/macos/toolsets/toolset-13.json
+++ b/images/macos/toolsets/toolset-13.json
@@ -3,7 +3,7 @@
         "default": "14.3.1",
         "x64": {
             "versions": [
-                { "link": "15.0", "version": "15.0.0-Beta.7+15A5229h" },
+                { "link": "15.0", "version": "15.0.0-Beta.7+15A5229h", "runtimes": [ "visionOS 1.0-beta2" ] },
                 { "link": "14.3.1", "version": "14.3.1+14E300c","symlinks": ["14.3"] },
                 { "link": "14.2", "version": "14.2.0+14C18" },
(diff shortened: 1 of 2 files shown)


images/macos/toolsets/toolset-13.json
```diff
@@ -3,7 +3,7 @@
         "default": "14.2",
         "x64": {
             "versions": [
```
mikhailkoliada, line 5: if we pin VersionOS version it should be done in a different part of toolset so we could call it like `Xcode.Versionos.Version` in jq, lets pull it out

images/macos/toolsets/toolset-13.json
```diff
@@ -75,7 +75,8 @@
             "gmp",
             "r",
             "yq",
-            "unxip"
+            "unxip",
+            "xcodesorg/made/xcodes"
```
mikhailkoliada, line 79: I kinda do not like Xcode's to be involved, the tool is way too unpredictable, can we get rid of it and simply download the platform separately with xcodebuild?

images/macos/toolsets/toolset-13.json
ilia-shipitsin: it makes no sense in different part of a toolset, because it is tightly paired with XCode

images/macos/toolsets/toolset-13.json
ilia-shipitsin: I kinda not like too, but xcodebuild just timed out downloading that particular SDK

PR #8189, opened 2023-08-28
Give codespell blessing: config, action + typos fixed

yarikoptic:
# Description

https://github.com/codespell-project/codespell is great at finding typos and has low false positive rate.  With this PR changes you can detect typos before they expose themselves to users etc.


yarikoptic:
I wonder why it was silently closed: @sergei-pyshnoi cherishes the typos, or is there a policy to not have external contributions, or ...?

PR #8261, opened 2023-09-11
AWS arm64 support for macOS 13

tuxity:
# Description

Here our contribution to support building AMIs on AWS. Only macOS 13 and arm64 (mac2.metal) for now.

I know you are not officially supporting PRs for macOS yet, but I will still submit as a draft so you guys are aware of some changes required and maybe can take decisions based on those.

I will gladly take feedbacks to improve it in the meanwhile.

Thanks!

## Check list

### DIFF
--- a/images/macos/helpers/Xcode.Installer.psm1
+++ b/images/macos/helpers/Xcode.Installer.psm1
@@ -29,7 +29,7 @@ function Invoke-DownloadXcodeArchive {
 
     $tempXipDirectory = New-Item -Path $DownloadDirectory -Name "Xcode$Version" -ItemType "Directory"
 
-    $xcodeFileName = 'Xcode-{0}.xip' -f $Version
+    $xcodeFileName = [System.Web.HttpUtility]::UrlEncode('Xcode-{0}.xip' -f $Version)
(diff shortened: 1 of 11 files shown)


mikhailkoliada:
Hello! Given we can not test image build on AWS we do not accept changes made other than to Azure, apart from that we have no plans to support anything than Azure either officially or non-officially.

ilia-shipitsin:
@tuxity , it is interesting how did manage TCC permissions, the following line is related to virtualization used in Azure https://github.com/bitso-oss/runner-images/blob/aws-support/images/macos/provision/configuration/configure-tccdb-macos.sh#L22

that is used for image generation itself, without modifying TCC we were not able to run image generation.

and, apparently, you are not adding anything instead. I'm surpised.

tuxity:
@mikhailkoliada alright, we will continue to provide support for AWS (packer template and public AMI) on our side then.

@ilia-shipitsin good catch, unfortunately for now it's a manual step when running the packer build locally (As we mentioned in a blog post [here](https://medium.com/bitso-engineering/our-journey-moving-the-ios-ci-to-aws-mac2-metal-arm-instances-351294bc3c00)). I hope to fix this later once changes happen from AWS (possibility to disable SIP) or Apple

PR #8326, opened 2023-09-19
helpers/GenerateResourcesAndImage.ps1: check for interactive mode

ilia-shipitsin:
# Description

helpers/GenerateResourcesAndImage.ps1: check for interactive mode

#### Related issue:

https://github.com/actions/runner-images/issues/8298

## Check list

### DIFF
--- a/helpers/GenerateResourcesAndImage.ps1
+++ b/helpers/GenerateResourcesAndImage.ps1
@@ -269,7 +269,7 @@ Function GenerateResourcesAndImage {
                 # are we running in a non-interactive session?
                 # https://stackoverflow.com/questions/9738535/powershell-test-for-noninteractive-mode
                 if ([System.Console]::IsOutputRedirected -or ![Environment]::UserInteractive -or !!([Environment]::GetCommandLineArgs() | Where-Object { $_ -ilike '-noni*' })) {
-                    throw "Non-interactive mode, resource group '$ResourceGroupName' already exists, specify -Force to delete it."
+                    throw "Non-interactive mode, resource group '$ResourceGroupName' already exists, either specify -Force to delete it, or -ReuseResourceGroup to reuse."
                 }
                 else {
                     # Resource group already exists, ask the user what to do
--- a/helpers/GenerateResourcesAndImage.ps1
+++ b/helpers/GenerateResourcesAndImage.ps1
@@ -266,16 +266,23 @@ Function GenerateResourcesAndImage {
                 $ResourceGroupExists = $false
             }
             else {
-                # Resource group already exists, ask the user what to do
-                $title = "Resource group '$ResourceGroupName' already exists"
-                $message = "Do you want to delete the resource group and all resources in it?"
+                # are we running in a non-interactive session?
+                # https://stackoverflow.com/questions/9738535/powershell-test-for-noninteractive-mode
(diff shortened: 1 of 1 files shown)


helpers/GenerateResourcesAndImage.ps1
```diff
@@ -266,16 +266,23 @@ Function GenerateResourcesAndImage {
                 $ResourceGroupExists = $false
             }
             else {
-                # Resource group already exists, ask the user what to do
-                $title = "Resource group '$ResourceGroupName' already exists"
-                $message = "Do you want to delete the resource group and all resources in it?"
+                # are we running in a non-interactive session?
+                # https://stackoverflow.com/questions/9738535/powershell-test-for-noninteractive-mode
+                if ([System.Console]::IsOutputRedirected -or ![Environment]::UserInteractive -or !!([Environment]::GetCommandLineArgs() | Where-Object { $_ -ilike '-noni*' })) {
```
Alexey-Ayupov, line 271: Do we need double `-not` here -  (**!!**([Environment]::GetCommandLineArgs() | Where-Object { $_ -ilike '-noni*' })))?

ilia-shipitsin: can you elaborate on some example when it comes to

Alexey-Ayupov: I got it. It turns the value from expression to the boolean (True or False)

Alexey-Ayupov (approved):

helpers/GenerateResourcesAndImage.ps1
```diff
@@ -266,16 +266,23 @@ Function GenerateResourcesAndImage {
...
+                    throw "Non-interactive mode, resource group '$ResourceGroupName' already exists, specify -Force to delete it."
```
vpolikarpov-akvelon, line 272: There is also 'ReuseResourceGroup' flag that is worth mentioning.

helpers/GenerateResourcesAndImage.ps1
```diff
@@ -266,16 +266,23 @@ Function GenerateResourcesAndImage {
...
+                }
+                else {
```
vpolikarpov-akvelon, line 274: You don't need `else` here because code inside `if` block interrupts execution

ilia-shipitsin: technically, I agree with you, but may "else" improve readability , it wraps code block which requires interactive mode.

helpers/GenerateResourcesAndImage.ps1
```diff
@@ -376,4 +383,4 @@ Function GenerateResourcesAndImage {
         }
         Write-Verbose "Cleanup completed."
     }
-}
+}
```
vpolikarpov-akvelon, line 386: Please keep newline at the end of the file

helpers/GenerateResourcesAndImage.ps1
ilia-shipitsin: makes sense

helpers/GenerateResourcesAndImage.ps1
ilia-shipitsin: do we have ".vscode" setting for that ?

mikhailkoliada (approved):

PR #8338, opened 2023-09-20
[windows] implement PyPy checksum validation

ilia-shipitsin:
# Description

implementing checksums for PyPy

#### Related issue:

https://github.com/github/c2c-actions-akvelon/issues/38

## Check list

### DIFF
--- a/images/win/scripts/Installers/Install-PyPy.ps1
+++ b/images/win/scripts/Installers/Install-PyPy.ps1
@@ -2,6 +2,7 @@
 ##  File:  Install-PyPy.ps1
 ##  Team:  CI-Build
 ##  Desc:  Install PyPy
+##  Supply chain security: checksum validation
 ################################################################################
 function Install-PyPy
 {
@@ -82,6 +83,11 @@ $toolsetVersions = Get-ToolsetContent | Select-Object -ExpandProperty toolcache 
 # Get PyPy releases
 $pypyVersions = Invoke-RestMethod https://downloads.python.org/pypy/versions.json
 
+# required for html parsing
+Install-Module PowerHTML -Scope CurrentUser 
+Import-Module PowerHTML
+$checksums = (Invoke-RestMethod -Uri 'https://www.pypy.org/checksums.html' | ConvertFrom-HTML).SelectNodes('//*[@id="content"]/article/div/pre')
+
 Write-Host "Starting installation PyPy..."
 foreach($toolsetVersion in $toolsetVersions.versions)
 {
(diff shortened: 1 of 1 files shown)


images/win/scripts/Installers/Install-PyPy.ps1
```diff
@@ -93,8 +100,23 @@ foreach($toolsetVersion in $toolsetVersions.versions)
...
     if ($latestMajorPyPyVersion)
     {
-        Write-Host "Found PyPy '$($latestMajorPyPyVersion.filename)' package"
-        $tempPyPyPackagePath = Start-DownloadWithRetry -Url $latestMajorPyPyVersion.download_url -Name $latestMajorPyPyVersion.filename
+        $filename = $latestMajorPyPyVersion.filename
+        Write-Host "Found PyPy '$filename' package"
+        $tempPyPyPackagePath = Start-DownloadWithRetry -Url $latestMajorPyPyVersion.download_url -Name $filename
+
+        #region Supply chain security
+        $local_file_hash = (Get-FileHash -Path $tempPyPyPackagePath -Algorithm SHA256).Hash
```
erik-bershel, line 108: I’m not sure myself, but it seems like a camelСase was expected to name the variables.

images/win/scripts/Installers/Install-PyPy.ps1
```diff
@@ -82,6 +83,12 @@ $toolsetVersions = Get-ToolsetContent | Select-Object -ExpandProperty toolcache
 # Get PyPy releases
 $pypyVersions = Invoke-RestMethod https://downloads.python.org/pypy/versions.json
 
+# required for html parsing
+Install-Module PowerHTML -Scope CurrentUser 
+Import-Module PowerHTML
+$checksums = (Invoke-RestMethod -Uri 'https://www.pypy.org/checksums.html' | ConvertFrom-HTML).SelectNodes('//*[@id="content"]/article/div/pre')
```
vpolikarpov-akvelon, line 89: This looks extremely fragile. Do we really need to parse HTML? Why not to treat page as plain text and search for proper "checksum-filename" pair instead?

ilia-shipitsin: I do not agree with you. Page is html, treating it as plain text is more fragile from my point of view

images/win/scripts/Installers/Install-PyPy.ps1
ilia-shipitsin: thank you for suggestion, much better indeed

images/win/scripts/Installers/Install-PyPy.ps1
ilia-shipitsin: that's how parsing works. if that has changed, image generation will fail. that is expected.
I'm not writing uber parser that will survive such changes, I think it is fine

Alexey-Ayupov (approved):

PR #8389, opened 2023-09-26
[Windows] Replace misleading mingw version in software report

vpolikarpov-akvelon:
# Description

There is a line in software report currently that states for mingw version, but in fact it represents version of gcc instead of mingw. This sometimes causes confusion and misunderstanding. This PR replaces line mentioned with some more specific.

## Check list


images/win/scripts/SoftwareReport/SoftwareReport.Tools.psm1
```diff
@@ -107,8 +107,20 @@ function Get-KindVersion {
     return $kindVersion
 }
 
-function Get-MinGWVersion {
-    (gcc --version | Select-String -Pattern "MinGW-W64") -match "(?<version>\d+\.\d+\.\d+)" | Out-Null
+function Get-GCCVersion {
+    (gcc --version | Select-String -Pattern "gcc.exe") -match "(?<version>\d+\.\d+\.\d+)" | Out-Null
```
codybak4, line 111:
```suggestion
    (gcc --version | Select-String -Pattern "gcc.exe") -match "(?<version>\d+\.\d+\.\d+)" | Out-Null
FUNCTION GET-VERSION "kind json-node"
```
- - - VSTESTLOCATION🤔  SETTINGSTEST@IPHONE SE VERSION 17.0.5.5

codybak4 (approved):
I am not sure what you are referring to but I think it was my version of device?? I was thinking about using my phone device files that allow you to send the file for approval

mikhailkoliada (approved):

PR #8401, opened 2023-09-28
feat: Allow overriding Linux and Windows vm image SKUs via Packer variables

tmeckel:
# Description

This pull request contains adaptations to the image definitions for Windows and Linux to be able to adjust the image SKU by a packer variable at runtime.

This can be used, for example, to create Gen2 VM images.

## Check list


mikhailkoliada:
Hello, we do not have plans to support more SKUs at the moment so we do not modify and nat parametrize it, please apply your changes locally.

PR #8413, opened 2023-09-29
[macos] add retries to brew install

ilia-shipitsin:
# Description

add retries to Homebrew install process

## Check list

### DIFF
--- a/images/macos/tests/Python.Tests.ps1
+++ b/images/macos/tests/Python.Tests.ps1
@@ -30,19 +30,11 @@ Describe "Python3" -Skip:($os.IsVenturaArm64) {
 
 Describe "Python2" -Skip:($os.IsVenturaArm64 -or $os.IsVentura) {
     It "Python 2 is available" {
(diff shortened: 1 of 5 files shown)


images/macos/tests/Python.Tests.ps1
```diff
@@ -8,8 +8,8 @@ Describe "Python3" -Skip:($os.IsVenturaArm64) {
         "python3 --version" | Should -ReturnZeroExitCode
     }
 
-    It "Python 3 is installed under /usr/local/bin" {
-        Get-WhichTool "python3" | Should -BeLike "/usr/local/bin*"
+    It "Python 3 is installed under /usr/bin" {
```
mikhailkoliada, line 11: we test exactly `/usr/local` here

ilia-shipitsin: thus, this should be addressed in other way.
"python3" lives in /usr/bin 

if that's unexpected, ok, let's fix that

Alexey-Ayupov (approved):

PR #8415, opened 2023-09-29
Download actions/action-versions latest release on Windows and set ACTIONS_RUNNER_ACTION_ARCHIVE_CACHE

TingluoHuang:
# Description
New tool

Fetching latest release from https://github.com/actions/action-versions

The latest release contain all tar.gz or zip of all versions of the top 5 used first-party actions (ex: `actions/checkout`).

With these cached on the hosted runner image, the hosted runner won't need to reach out to codeload to download those same assets over and over again.

The asset is about 120MB for both Windows and non-Windows, the download and unzip normally is less than 10s.
Since the assets is an archive of other archive, we only need to unzip the outer layer, after unzip the total size on disk are pretty much the same.

image

#### Related issue:

https://github.com/github/actions-broker/issues/17

## Check list

### DIFF
--- a/images/win/scripts/Installers/Install-ActionArchiveCache.ps1
+++ b/images/win/scripts/Installers/Install-ActionArchiveCache.ps1
@@ -1,0 +1,18 @@
+################################################################################
+##  File:  Install-ActionArchiveCache.ps1
+##  Desc:  Install Action Archive Cache
+##         from https://github.com/actions/action-versions
+################################################################################
+
+if (-not (Test-Path $env:ACTIONS_RUNNER_ACTION_ARCHIVE_CACHE))
+{
+    Write-Host "Creating action archive cache folder"
+    New-Item -ItemType Directory -Path $env:ACTIONS_RUNNER_ACTION_ARCHIVE_CACHE | Out-Null
+}
+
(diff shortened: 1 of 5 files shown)


images/win/scripts/Installers/Install-ActionArchiveCache.ps1
```diff
@@ -0,0 +1,19 @@
...
+    Write-Host "Creating action archive cache folder"
+    New-Item -ItemType Directory -Path $env:ACTIONS_RUNNER_ACTION_ARCHIVE_CACHE | Out-Null
+}
+
+$downloadUrl = Get-GitHubPackageDownloadUrl -RepoOwner "actions" -RepoName "action-versions" -Version "latest" -UrlFilter "*/v{Version}/action-versions.zip"
+Write-Host "Download Latest action-versions archive from $downloadUrl"
+$actionVersionsArchivePath = Start-DownloadWithRetry -Url $downloadUrl -Name "action-versions.zip"
+
+Write-Host "Expand action-versions archive"
+Extract-7Zip -Path $actionVersionsArchivePath -DestinationPath $env:ACTIONS_RUNNER_ACTION_ARCHIVE_CACHE
```
mikhailkoliada, line 19: add the pester test invocation here :)

TingluoHuang: fixed.

mikhailkoliada (approved):

PR #8427, opened 2023-10-02
[windows] warmup Azure CLI

ilia-shipitsin:
# Description

warmup Azure CLI during image generation to improve first run timing

#### Related issue:

https://github.com/actions/runner-images/issues/8296

## Check list


jessehouwing (approved):

jessehouwing:
Replaces the part of this pull request that performs the redirection of the `.azure` folder and the warmup of `az`:

 * https://github.com/actions/runner-images/pull/8294

images/win/scripts/Installers/Install-AzureCli.ps1
```diff
@@ -3,13 +3,32 @@
...
+# Warm-up Azure CLI
+
+Write-Host "Warmup 'az'"
 
-[Environment]::SetEnvironmentVariable("AZURE_EXTENSION_DIR", $azureCliExtensionPath, [System.EnvironmentVariableTarget]::Machine)
+az --help
+if ($LASTEXITCODE -ne 0)
+{
+   throw "Command 'az --help' failed"
+}
```
vpolikarpov-akvelon, line 32: Don't we have tests that do the same?

ilia-shipitsin: even if we do, it would be a leaked abstraction.
tests do testing, warmup does warmup

jessehouwing: Current test does `az --version`. Which is nice too, as it causes the caching of the version information on disk (as `az --version` does an update check and caches the result).

ilia-shipitsin: someone once may want to change tests behavior. I would not like to depend on those changes, thus dedicated warm up makes sense (even if it's identical)

vpolikarpov-akvelon: Ok. I would only hide az's output and put braces according to1TBS rule.

ilia-shipitsin: I used VS Code with whatever we have in .vscode, I wonder why do you think code is not compliant

mikhailkoliada (approved):

PR #8474, opened 2023-10-05
add runner-package to curated images

pshepherd10:
# Description
The plan is to cache the runner package onto our curated images just in case the CDN that we use to download the package is experiencing downtime. This will help speed up VM setup times in some cases when the latest version is already cached. 

The size of the package on linux would be 180Mb and windows would be 84Mb.
Related: 
[ADR](https://github.com/github/actions-larger-runners/blob/main/docs/adrs/bake-runner-package-into-image.md)
https://github.com/github/actions-larger-runners/issues/2318

## Check list


images/linux/scripts/installers/runner-package.sh
```diff
@@ -0,0 +1,20 @@
...
+    curl -L \ -H "Accept: application/vnd.github+json" \ -H "X-GitHub-Api-Version: 2022-11-28" \ https://api.github.com/repos/actions/runner/releases/latest 
+    | jq '.name' 
+    | tr -d "v" 
+)
+URL = "https://github.com/actions/runner/releases/download/v${LATEST_VERSION}/actions-runner-linux-x64-${LATEST_VERSION}.tar.gz"
+FILE_NAME = "actions-runner-linux-x64-${LATEST_VERSION}.tar.gz"
+
+
+download_with_retries "${URL}" "/tmp" "${FILE_NAME}"
+tar xzf "/temp/${FILE_NAME}"
```
ritchxu, line 20: As discussed in Slack, we don't need to unpack the tar ball, just download and cache the package as it is.

maxim-lobanov (approved):

images/linux/scripts/installers/runner-package.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash -e
+################################################################################
+##  File:  runner-package.sh
+##  Desc:  Downloads and Installs runner package
+################################################################################
+
+# Source the helpers for use with the script
+source $HELPER_SCRIPTS/install.sh
+
+LATEST_VERSION = $(
```
mikhailkoliada, line 10: we should use the `get_github_package_download_url` helper here

images/linux/ubuntuminimal.pkr.hcl
```diff
@@ -258,7 +258,8 @@ build {
     scripts          = [
                         "${path.root}/scripts/installers/git.sh",
                         "${path.root}/scripts/installers/github-cli.sh",
-                        "${path.root}/scripts/installers/zstd.sh"
+                        "${path.root}/scripts/installers/zstd.sh",
```
mikhailkoliada, line 261: I do not think ubuntu minimal image need runner agent installed, it is for those who build their own images

images/linux/scripts/installers/runner-package.sh
```diff
@@ -0,0 +1,19 @@
...
+    curl -L \ -H "Accept: application/vnd.github+json" \ -H "X-GitHub-Api-Version: 2022-11-28" \ https://api.github.com/repos/actions/runner/releases/latest 
+    | jq '.name' 
+    | tr -d "v" 
+)
+URL = "https://github.com/actions/runner/releases/download/v${LATEST_VERSION}/actions-runner-linux-x64-${LATEST_VERSION}.tar.gz"
+FILE_NAME = "actions-runner-linux-x64-${LATEST_VERSION}.tar.gz"
+
+
+download_with_retries "${URL}" "/opt/runner-cache" "${FILE_NAME}"
```
mikhailkoliada, line 19: this directory does not exist in the runner so we should at least ` mkdir -p /opt/runner-cache` it

images/linux/scripts/installers/runner-package.sh
pshepherd10: Would it be something like: `downloadUrl=$(get_github_package_download_url "actions/runner" "contains(\"actions-runner-linux-x64-\")" "latest")`

images/linux/scripts/installers/runner-package.sh
```diff
@@ -0,0 +1,17 @@
...
+##  Desc:  Downloads and Installs runner package
+################################################################################
+
+# Source the helpers for use with the script
+source $HELPER_SCRIPTS/install.sh
+
+DOWNLOAD_URL = $(get_github_package_download_url "actions/runner" "contains(\"actions-runner-linux-x64-\") and endswith(\".tar.gz\")" "latest")
+
+LATEST_VERSION = echo ${DOWNLOAD_URL##*/}
+FILE_NAME = "actions-runner-linux-x64-${LATEST_VERSION}.tar.gz"
```
ritchxu, line 13: This only covers Linux platform, should we add a similar step for Windows too?

mikhailkoliada: @ritchxu I guess we need but better w/ another windows-specific PR

mikhailkoliada:
@pshepherd10 please make a separate pr for windows

mikhailkoliada:
@pshepherd10 actually I was thinking, lets introduce the test on both windows and ubuntu runners to make sure that the runner directory exist and not empty

pshepherd10:
@mikhailkoliada ok just added the test, let me know if that is right.

pshepherd10:
@mikhailkoliada removed some spaces that was causing it to break. Can you try rerunning it?

pshepherd10:
@mikhailkoliada updated it again, fixed a typo

mikhailkoliada:
@pshepherd10 could you rebase?

pshepherd10:
@mikhailkoliada branch is updated now

pshepherd10:
@mikhailkoliada not sure what the issue is since it didn't give a reason for the failure

```
==> azure-arm: Provisioning with shell script: /home/vsts/Agents/image-generation-1/_work/1/s/images/linux/scripts/installers/runner-package.sh
==> azure-arm: Provisioning step had errors: Running the cleanup provisioner, if present...

```

mikhailkoliada:
@pshepherd10 hmm you might need use `sudo` before the mkdir call

pshepherd10:
@mikhailkoliada ok added sudo lets see if that works

images/linux/scripts/installers/runner-package.sh
```diff
@@ -0,0 +1,16 @@
...
+# Source the helpers for use with the script
+source $HELPER_SCRIPTS/install.sh
+
+DOWNLOAD_URL=$(get_github_package_download_url "actions/runner" "contains(\"actions-runner-linux-x64-\") and endswith(\".tar.gz\")" "latest")
+
+FILE_NAME=echo ${DOWNLOAD_URL##*/}
+
+sudo mkdir -p /opt/runner-cache
+
+download_with_retries "${URL}" "/opt/runner-cache" "${FILE_NAME}"
```
mikhailkoliada, line 16: @pshepherd10 also `${URL}` should be `${DOWNLOAD_URL}`

pshepherd10: woops! ok fixed it!

pshepherd10:
https://jqplay.org/s/qYBTIZGg_lV looks like it should work

pshepherd10:
fixed more syntax issues

images/linux/scripts/installers/runner-package.sh
```diff
@@ -0,0 +1,16 @@
...
+# Source the helpers for use with the script
+source $HELPER_SCRIPTS/install.sh
+
+DOWNLOAD_URL=$(get_github_package_download_url "actions/runner" "test(\"actions-runner-linux-x64-[0-9].[0-9]{3}.[0-9].tar.gz\")" "latest")
+
+FILE_NAME="${DOWNLOAD_URL##*/}"
+
+sudo mkdir -p /opt/runner-cache
+
+download_with_retries "${DOWNLOAD_URL}" "/opt/runner-cache" "${FILE_NAME}"
```
mikhailkoliada, line 16: you can not save directly to `/opt/runner-cache` as it is write-protected, so you need to save it to `/tmp` and only then `sudo mv` it from `/tmp` to target destination

maxim-lobanov: I think the current implementation should work because the full script `runner-package.sh` is run from `root` user: https://github.com/actions/runner-images/blob/433178adf5d64cca48a8bf96525b38385e3b7682/images/linux/ubuntu2204.pkr.hcl#L281

images/linux/scripts/installers/runner-package.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash -e
+################################################################################
+##  File:  runner-package.sh
+##  Desc:  Downloads and Installs runner package
+################################################################################
+
+# Source the helpers for use with the script
+source $HELPER_SCRIPTS/install.sh
+
+DOWNLOAD_URL=$(get_github_package_download_url "actions/runner" "test(\"actions-runner-linux-x64-[0-9].[0-9]{3}.[0-9].tar.gz\")" "latest")
```
erik-bershel, line 10:
```suggestion
DOWNLOAD_URL=$(get_github_package_download_url "actions/runner" 'test("actions-runner-linux-x64-[0-9]+\\.[0-9]{3}\\.[0-9]+\\.tar\\.gz")')
```

erik-bershel: Hey @pshepherd10! Should help with the error. The syntax of JQ, bash and pipe with special characters make things painful to troubleshoot here. Also possible to replace it with the release tag:
```bash
TAG=$(curl -fsSL "https://api.github.com/repos/actions/runner/releases/latest" | jq -r ".tag_name" | tr -cd '0-9.')
DOWNLOAD_URL="https://github.com/actions/runner/releases/latest/download/actions-runner-linux-x64-${TAG}.tar.gz"
```

images/linux/scripts/installers/runner-package.sh
ritchxu: @mikhailkoliada could you help me understand why `/opt/runner-cache` is write-protected? 😃 and if it is still a blocker if, as Maxim mentioned above, we are running image generation as root

mikhailkoliada (approved):

PR #8484, opened 2023-10-06
[windows] Install Microsoft OLE DB Driver for SQL Server 19.3.1

chrisdecker1201:
# Description

I mentioned that there is a `Microsoft OLE DB Driver for SQL Server` 18.6.4 installed inside the GitHub Runners. At least we should install the 18.6.6, but I think the best is the 19.3.1


Alexey-Ayupov:
Hello, @chrisdecker1201. We are afraid that the latest version of the Microsoft OLE DB driver could potentially break someone, however, we will update a minor version of the driver v18 in another PR.

Alexey-Ayupov:
PR with updated driver version - https://github.com/actions/runner-images/pull/8553

PR #8533, opened 2023-10-12
Expose client_jwt to enable oidc authentication

Triodes:
# Description
In the most recent version of the azure-arm packer plugin OIDC auth is supported by setting the client_jwt parameter to the OIDC token. This PR exposes this parameter as a template variable in the templates using the azure-arm builder.

This change has been inspired by the [following article](https://www.hashicorp.com/blog/version-2-packer-azure-plugin-now-available) and the more recent support in Azure Devops for [workload identity federation](https://devblogs.microsoft.com/devops/public-preview-of-workload-identity-federation-for-azure-pipelines/)

## Check list
- [ ] Related issue / work item is attached (n.a.)
- [ ] Tests are written (if applicable) (n.a.)


Alexey-Ayupov:
Hello @Triodes and thank you for your suggestion. However, due to the small demand for OIDC authentication, we would like to not include this method in our packer templates.
If you want to share this improvement with the community feel free to create a new discussion.

Triodes:
Hey Alexey,

Thanks for your prompt response.

I don't see a good reason to not include this. Workload identity federation is now supported in both Github Actions and Azure Devops. In the latter it is now the default way for creating service connections with active push to move away from service connections using client credentials
image

The biggest advantage here is not having to rotate client secrets periodically. Using identity federation will most likely become the de-facto standard for authenticating to AzureRM from these type of pipelines.

The changes in this PR are minor and backwards compatible. They would allow anyone to choose their preferred way of authenticating instead of forcing a specific method. Packer supports it, so why not allow others to use that functionality?

For me personally this would solve the issue of manually having to rotate the client secret every 6 months. It is a manual task which, if not done, will cause the packer build to fail.

Would I be able to get you to reconsider this? If not, would you be willing to explain in a bit more detail the reason for not including this?

P.S. I have already created [a discussion here](https://github.com/actions/runner-images/discussions/8527)

PR #8552, opened 2023-10-13
[Ubuntu] Add helpers to work with checksums

erik-bershel:
# Description

Adding following functions to work with packages hashes:
`get_github_package_hash()` will be used to get package hash from GitHub release body, possible to use with direct release link or with just `repo_owner` and `repo_name` pair.
`use_checksum_comparison()` will be used to compare known checksum with the package hash

#### Related issue:

https://github.com/actions/runner-images-internal/issues/5491

## Check list

### DIFF
--- a/images/linux/scripts/helpers/install.sh
+++ b/images/linux/scripts/helpers/install.sh
@@ -95,3 +95,68 @@ get_github_package_download_url() {
     fi
     echo $downloadUrl
 }
+
+get_github_package_hash() {
+    local repo_owner=$1
+    local repo_name=$2
+    local file_name=$3
+    local url=$4
+    local version=${5:-"latest"}
+    local prerelease=${6:-false}
+    local delimiter=${8:-'|'}
+    local word_number=${9:-2}
+
+    if [ -n "$url" ]; then
+        release_url="$url"
+    else
+        if [ "$version" == "latest" ]; then
+            release_url="https://api.github.com/repos/${repo_owner}/${repo_name}/releases/latest"
+        else
+            json=$(curl -s "https://api.github.com/repos/${repo_owner}/${repo_name}/releases?per_page=100")
+            tags=$(echo "$json" | jq -r --arg prerelease "$prerelease" '.[] | select(.prerelease == ($prerelease | test("true"; "i"))) | .tag_name')
+            tag=$(echo "$tags" | grep -o "$version")
+            if [ -z "$tag" ]; then
+                echo "Failed to get a tag name for version $version."
+                exit 1
+            fi
+            release_url="https://api.github.com/repos/${repo_owner}/${repo_name}/releases/tags/$tag"
+        fi
+    fi
+
+    body=$(curl -s "$release_url" | jq -r '.body' | tr -d '`')
+    matching_line=$(echo "$body" | grep "$file_name")
+    if [ -z "$matching_line" ]; then
+        echo "File name '$file_name' not found in release body."
+        exit 1
+    fi
+
+    result=$(echo "$matching_line" | cut -d "$delimiter" -f "$word_number" | tr -d -c '[:alnum:]')
+    if [ -z "$result" ]; then
+        echo "Empty result. Check parameters Delimiter and/or WordNumber for the matching line."
+        exit 1
+    fi
+
+    echo "$result"
+}
+
+use_checksum_comparison() {
+    local file_path=$1
+    local checksum=$2
+    local sha_type=${3-"256"}
+
+    echo "Performing checksum verification"
+
+    if [ ! -f "$file_path" ]; then
+        echo "File not found: $file_path"
+        exit 1
+    fi
+
+    local_file_hash=$(shasum --algorithm "$sha_type" "$file_path" | awk '{print $1}')
+
+    if [ "$local_file_hash" != "$checksum" ]; then
+        echo "Checksum verification failed. Expected hash: $checksum; Actual hash: $local_file_hash."
+        exit 1
+    else
+        echo "Checksum verification passed"
+    fi
+}
--- a/images/linux/scripts/helpers/install.sh
+++ b/images/linux/scripts/helpers/install.sh
@@ -103,8 +103,13 @@ get_github_package_hash() {
     local url=$4
     local version=${5:-"latest"}
     local prerelease=${6:-false}
-    local delimiter=${8:-'|'}
-    local word_number=${9:-2}
+    local delimiter=${7:-'|'}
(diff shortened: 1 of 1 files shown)


images/linux/scripts/helpers/install.sh
```diff
@@ -95,3 +95,68 @@ get_github_package_download_url() {
...
+    local delimiter=${8:-'|'}
+    local word_number=${9:-2}
+
+    if [ -n "$url" ]; then
+        release_url="$url"
+    else
+        if [ "$version" == "latest" ]; then
+            release_url="https://api.github.com/repos/${repo_owner}/${repo_name}/releases/latest"
+        else
+            json=$(curl -s "https://api.github.com/repos/${repo_owner}/${repo_name}/releases?per_page=100")
```
vpolikarpov-akvelon, line 115:
```suggestion
            json=$(curl -fsSL "https://api.github.com/repos/${repo_owner}/${repo_name}/releases?per_page=100")
```

images/linux/scripts/helpers/install.sh
```diff
@@ -95,3 +95,68 @@ get_github_package_download_url() {
...
 }
+
+get_github_package_hash() {
+    local repo_owner=$1
+    local repo_name=$2
+    local file_name=$3
+    local url=$4
+    local version=${5:-"latest"}
+    local prerelease=${6:-false}
+    local delimiter=${8:-'|'}
```
vpolikarpov-akvelon, line 106: Did you skip 7th argument intentionally?
```suggestion
    local prerelease=${6:-false}
    local search_in_count=${7:-100}
    local delimiter=${8:-'|'}
```

images/linux/scripts/helpers/install.sh
```diff
@@ -95,3 +95,68 @@ get_github_package_download_url() {
...
+    local word_number=${9:-2}
```
vpolikarpov-akvelon, line 107: Similar PowerShell function has default value `1` here
```suggestion
    local word_number=${9:-1}
```

images/linux/scripts/helpers/install.sh
```diff
@@ -95,3 +95,68 @@ get_github_package_download_url() {
...
+            tag=$(echo "$tags" | grep -o "$version")
+            if [ -z "$tag" ]; then
+                echo "Failed to get a tag name for version $version."
+                exit 1
+            fi
+            release_url="https://api.github.com/repos/${repo_owner}/${repo_name}/releases/tags/$tag"
+        fi
+    fi
+
+    body=$(curl -s "$release_url" | jq -r '.body' | tr -d '`')
```
vpolikarpov-akvelon, line 126:
```suggestion
    body=$(curl -fsSL "$release_url" | jq -r '.body' | tr -d '`')
```

images/linux/scripts/helpers/install.sh
```diff
@@ -95,3 +95,68 @@ get_github_package_download_url() {
...
+    matching_line=$(echo "$body" | grep "$file_name")
+    if [ -z "$matching_line" ]; then
```
vpolikarpov-akvelon, line 128: What if several lines have matched (e.g. when `$file_name` is empty of not quite well defined)?

images/linux/scripts/helpers/install.sh
```diff
@@ -95,3 +95,68 @@ get_github_package_download_url() {
...
+
+    if [ -n "$url" ]; then
+        release_url="$url"
+    else
+        if [ "$version" == "latest" ]; then
+            release_url="https://api.github.com/repos/${repo_owner}/${repo_name}/releases/latest"
+        else
+            json=$(curl -s "https://api.github.com/repos/${repo_owner}/${repo_name}/releases?per_page=100")
+            tags=$(echo "$json" | jq -r --arg prerelease "$prerelease" '.[] | select(.prerelease == ($prerelease | test("true"; "i"))) | .tag_name')
+            tag=$(echo "$tags" | grep -o "$version")
```
vpolikarpov-akvelon, line 117: What if several tags pass this filter? E.g. when version specifies only major and minor part but there are multiple patches available.

images/linux/scripts/helpers/install.sh
```diff
@@ -95,3 +95,68 @@ get_github_package_download_url() {
...
+        exit 1
+    fi
+
+    echo "$result"
+}
+
+use_checksum_comparison() {
+    local file_path=$1
+    local checksum=$2
+    local sha_type=${3-"256"}
```
vpolikarpov-akvelon, line 145: Function `Use-ChecksumComparison` has different arguments set and behavior. Are you going to update that function as well so they match?

Also colon is missing in `${3:-"256"}`

vpolikarpov-akvelon (changes requested):

images/linux/scripts/helpers/install.sh
erik-bershel: I spent about an hour to find that PS start counting from 0 but Bash from 1, so let it be 2 to be the same as in PS. 🤣

images/linux/scripts/helpers/install.sh
erik-bershel: I'll think so. It will save for us couple raws.

images/linux/scripts/helpers/install.sh
erik-bershel: Well, in general, a mechanism with a version is just for specifying the exact value. If someone indicates a range of versions, then of course they will do it in vain.
But sounds reasonable to save future us from such a mistake. I'll add something here.

images/linux/scripts/helpers/install.sh
```diff
@@ -95,3 +95,81 @@ get_github_package_download_url() {
...
+    if [ -n "$url" ]; then
+        release_url="$url"
+    else
+        if [ "$version" == "latest" ]; then
+            release_url="https://api.github.com/repos/${repo_owner}/${repo_name}/releases/latest"
+        else
+            json=$(curl -fsSL "https://api.github.com/repos/${repo_owner}/${repo_name}/releases?per_page=100")
+            tags=$(echo "$json" | jq -r --arg prerelease "$prerelease" '.[] | select(.prerelease == ($prerelease | test("true"; "i"))) | .tag_name')
+            tag=$(echo "$tags" | grep -o "$version")
+            if [[ "$(echo "$tag" | wc -l)" -gt 1 ]]; then
```
mikhailkoliada, line 123: nit: we should either use ` [ condition ]` or `[[ "$condition" ]]` here I see we are using both overall in the helper

mikhailkoliada (approved):

PR #8582, opened 2023-10-18
[macos] add reties to CodeQL download step

ilia-shipitsin:
# Description

add reties to CodeQL download step

## Check list

### DIFF
--- a/images/macos/provision/core/codeql-bundle.sh
+++ b/images/macos/provision/core/codeql-bundle.sh
@@ -2,8 +2,8 @@
 source ~/utils/utils.sh
 
 # Retrieve the CLI version of the latest CodeQL bundle.
-base_url="$(curl -fsSL https://raw.githubusercontent.com/github/codeql-action/v2/src/defaults.json)"
-bundle_version="$(echo "$base_url" | jq -r '.cliVersion')"
+download_with_retries https://raw.githubusercontent.com/github/codeql-action/v2/src/defaults.json "/tmp" "codeql-defaults.json"
+bundle_version="$(cat /tmp/codeql-defaults.json | jq -r '.cliVersion')"
 bundle_tag_name="codeql-bundle-v$bundle_version"
 
 echo "Downloading CodeQL bundle $bundle_version..."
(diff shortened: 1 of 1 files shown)


images/macos/provision/core/codeql-bundle.sh
```diff
@@ -2,8 +2,8 @@
 source ~/utils/utils.sh
 
 # Retrieve the CLI version of the latest CodeQL bundle.
-base_url="$(curl -fsSL https://raw.githubusercontent.com/github/codeql-action/v2/src/defaults.json)"
-bundle_version="$(echo "$base_url" | jq -r '.cliVersion')"
+download_with_retries https://raw.githubusercontent.com/github/codeql-action/v2/src/defaults.json "/tmp" "codeql-defaults.json"
+bundle_version="$(cat /tmp/codeql-defaults.json | jq -r '.cliVersion')"
```
henrymercer, line 6: We can simplify this:
```suggestion
bundle_version="$(jq -r '.cliVersion' /tmp/codeql-defaults.json)"
```

images/macos/provision/core/codeql-bundle.sh
```diff
@@ -2,8 +2,8 @@
 source ~/utils/utils.sh
 
 # Retrieve the CLI version of the latest CodeQL bundle.
-base_url="$(curl -fsSL https://raw.githubusercontent.com/github/codeql-action/v2/src/defaults.json)"
-bundle_version="$(echo "$base_url" | jq -r '.cliVersion')"
+download_with_retries https://raw.githubusercontent.com/github/codeql-action/v2/src/defaults.json "/tmp" "codeql-defaults.json"
```
henrymercer, line 5: Could you confirm that `download_with_retries` follows redirects?

henrymercer (approved):
Thanks!

images/macos/provision/core/codeql-bundle.sh
ilia-shipitsin: no idea on that. it uses curl under the hood. hopefully the same behaviour as regular curl

do we have tests on codeql (triggered on image generation) ?

ilia-shipitsin: I hope that if something will go wrong with redirect, we'll catch it with https://github.com/actions/runner-images/blob/main/images/macos/provision/core/codeql-bundle.sh#L28

henrymercer: Looks like it does — https://github.com/actions/runner-images/blob/e113e2f6b829f0e6fd528b93a1241c52410873a2/images/macos/provision/utils/utils.sh#L15C39-L15C40 passes `-L`.

The runner image tests will check the download works, I just wanted to be robust to the case where the download doesn't include a redirect today but starts including a redirect in the future.

henrymercer (approved):

Alexey-Ayupov (approved):

PR #8608, opened 2023-10-19
[macos] add retries to "brew install"

ilia-shipitsin:
# Description

sometimes "brew install" fails due to network timeout, let's use "brew --cache" with retries to make it more robust

## Check list

### DIFF
--- a/images/macos/provision/utils/utils.sh
+++ b/images/macos/provision/utils/utils.sh
@@ -192,6 +192,7 @@ brew_smart_install() {
 
         for dep in `cat /tmp/$tool_name`; do
 
+            failed=true
             for i in {1..10};
             do
                 brew --cache $dep && failed=false || sleep 60
--- a/images/macos/provision/utils/utils.sh
+++ b/images/macos/provision/utils/utils.sh
@@ -177,8 +177,7 @@ brew_smart_install() {
         # get deps & cache em
 
         failed=true
-        for i in {1..10};
-        do
+        for i in {1..10}; do
             brew deps $tool_name > /tmp/$tool_name && failed=false || sleep 60
             if [ "$failed" = false ]; then
                 break
@@ -193,8 +192,7 @@ brew_smart_install() {
         for dep in `cat /tmp/$tool_name`; do
 
             failed=true
-            for i in {1..10};
-            do
+            for i in {1..10}; do
                 brew --cache $dep && failed=false || sleep 60
                 if [ "$failed" = false ]; then
                     break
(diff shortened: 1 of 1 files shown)


images/macos/provision/utils/utils.sh
```diff
@@ -173,6 +173,38 @@ brew_smart_install() {
...
+            brew deps $tool_name > /tmp/$tool_name && failed=false || sleep 60
+            if [ "$failed" = false ]; then
+                break
+            fi
+        done
+
+        if [ "$failed" = true ]; then
+           echo "Failed: brew deps $tool_name"
+           exit 1;
+        fi
```
vpolikarpov-akvelon, line 190: It's possible to implement this logic simpler:

- place `break` in place of `failed=false`;
- change condition in the inner 'if' block to `$i -eq 10` and put `exit` inside;
- remove `failed` variable and outer 'if' block

Something like this:
```
for i in {1..10}; do
    brew deps $tool_name > /tmp/$tool_name && break || sleep 60
    if [ $i -eq 10 ]; then
        echo "Failed: brew deps $tool_name"
        exit 1
    fi
done
```

images/macos/provision/utils/utils.sh
```diff
@@ -173,6 +173,38 @@ brew_smart_install() {
...
+
+        for dep in `cat /tmp/$tool_name`; do
```
vpolikarpov-akvelon, line 192:
```suggestion
        for dep in $(cat /tmp/$tool_name); do
```

images/macos/provision/utils/utils.sh
```diff
@@ -173,6 +173,38 @@ brew_smart_install() {
         brew install --build-from-source $tool_name
     else
         echo "Downloading $tool_name..."
+
+        # get deps & cache em
```
vpolikarpov-akvelon, line 177: This comment is not clear

images/macos/provision/utils/utils.sh
ilia-shipitsin: I tried inline "break", it did not work on my syntetic tests.
if you end with something working, feel free to submit follow up PR

images/macos/provision/utils/utils.sh
ilia-shipitsin: "comment is not clear" does not help to improve it.
feel free to suggest better wording

images/macos/provision/utils/utils.sh
ilia-shipitsin: personally, I feel that `if [ $i -eq 10 ]; then` is not "readable", it may be harder to understand it's intention

Alexey-Ayupov (approved):

PR #8657, opened 2023-10-25
[Ubuntu] fix the condition in docker.sh

yassi-github:
# Description
Corrected omission in https://github.com/actions/runner-images/pull/7595.

Fix the condition in the file docker.sh
Instead of "-eq" we will use "=="
"==" is for string comparisons and "-eq" is for numeric ones.

#### Related issue:
https://github.com/actions/runner-images/pull/7595

## Check list

### DIFF
--- a/images/linux/scripts/installers/docker.sh
+++ b/images/linux/scripts/installers/docker.sh
@@ -73,6 +73,6 @@ rm $gpg_key
 rm $repo_path
 
 invoke_tests "Tools" "Docker"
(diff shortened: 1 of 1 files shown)


vpolikarpov-akvelon (approved):

ilia-shipitsin (approved):

mikhailkoliada:
I see it is going to be fixed as a part of the https://github.com/actions/runner-images/pull/8660/files, but thanks a lot for the submission!

PR #8683, opened 2023-10-30
[macOS] Remove AWS cli from macOS 13

Alexey-Ayupov:
# Description
Remove AWS cli from macOS 13

## Check list


jozefizso:
@Alexey-Ayupov why was `aws-cli` removed only from the premium priced M1 macOS runners and not from all the other runners?

PR #8704, opened 2023-11-01
fix: checksum search for Kotlin

frittsy:
The [latest release notes](https://github.com/JetBrains/kotlin/releases/tag/v1.9.20) for Kotlin v1.9.20 includes multiple lines that contain the search string "kotlin-compiler", one of which is not the line that contains the checksum. This tweaks the search string to only look for the filename ending in `.zip`, where the checksum should be present.


frittsy:
There appears to be a similar issue with Windows too, so looking at that as well and may update this PR with the fix for both.

mikhailkoliada (approved):

PR #8757, opened 2023-11-07
[Mac OS] Update Anka Helpers to build Mac OS 14 ARM image

vpolikarpov-akvelon:
# Description
This pull request modifies `Get-AvailableIPSWVersions` function to return Mac OS Build ID instead of version. It also adds `--compatible` flag to the `mist list firmware` command.

## Check list

### DIFF
--- a/images.CI/macos/anka/Service.Helpers.psm1
+++ b/images.CI/macos/anka/Service.Helpers.psm1
@@ -63,24 +63,24 @@ function Get-AvailableIPSWVersions {
     )
 
     if ($IsBeta) {
-        $command = { mist list firmware "$MacOSCodeNameOrVersion" --include-betas --latest --export "/Applications/export.json"}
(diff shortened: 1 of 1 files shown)


images.CI/macos/anka/Service.Helpers.psm1
```diff
@@ -302,7 +305,13 @@ function Invoke-SSHPassCommand {
         "${env:SSHUSER}@${HostName}"
     )
     $sshPassOptions = $sshArg -join " "
-    $result = bash -c "$sshPassOptions \""$Command\"" 2>&1"
+    if ($PSVersionTable.PSVersion.Major -eq 5) {
```
mikhailkoliada, line 308: this script is always being run on Mac hosts i.e. they do not have powershell other than push-core which is always 7 and never 5, do we need the native powershell condition?

vpolikarpov-akvelon: Good point, thank you

mikhailkoliada (approved):

PR #8767, opened 2023-11-09
[macOS] fix 2 bad rules in configure-tccdb-macos.sh

liam-x:
# Description

Bug fixing

The TCC rule is broken for kTCCServiceScreenCapture /usr/local/opt/runner/provisioner/provisioner in macos image. 

It should be 1 instead of 0.  
0 is for bundle id. 1 is for executable path.

The consequence of these bad rules are child process of provisioner may lose the ability to take screenshot on macos runner.

## Check list


shamil-mubarakshin:
Hello @liam-x, thanks for contributing.
External PRs for macOS [are not](https://github.com/actions/runner-images#faqs) accepted for now. We will check suggested changes internally.

Issue #8004, opened 2023-07-31
Unable to install MySQLCli


mnieto:
### Description

Executing Install-MysqlCli.ps1, line 
```
    $MysqlVersion = (Invoke-RestMethod -Uri "https://dev.mysql.com/downloads/mysql/${MysqlVersionMajorMinor}.html" |
        Select-String -Pattern "${MysqlVersionMajorMinor}\.\d+").Matches.Value
```

is returning most of the times (a few times is getting the correct response) something like this:

```
             This site https://dev.mysql.com/downloads/mysql/8.0.html  is experiencing technical difficulty. We are
aware of the issue and are working as quick as possible to correct the issue.
```

See full output in the attached file for reference: [mysqlcli_error.txt](https://github.com/actions/runner-images/files/12215042/mysqlcli_error.txt)

That makes the Select-String cmdlet to fail. I managed to solve the problem adding a custom header `Accept=text/html`

```
    $MysqlVersion = (Invoke-RestMethod -Uri "https://dev.mysql.com/downloads/mysql/${MysqlVersionMajorMinor}.html" -Headers @{"Accept"="text/html"} |
        Select-String -Pattern "${MysqlVersionMajorMinor}\.\d+").Matches.Value
```

### Platforms affected

- [X] Azure DevOps
- [ ] GitHub Actions - Standard Runners
- [ ] GitHub Actions - Larger Runners

### Runner images affected

- [ ] Ubuntu 20.04
- [ ] Ubuntu 22.04
- [ ] macOS 11
- [ ] macOS 12
- [ ] macOS 13
- [ ] Windows Server 2019
- [X] Windows Server 2022

### Image version and build link

win22/20230724.1

### Is it regression?

No

### Expected behavior

MySQLCli is installed

### Actual behavior

Getting powershell error during this Install-MySQLCli.ps1 script

```
==> azure-arm: At C:\Windows\Temp\script-64c789b3-6529-7f0a-8e5d-0ab63b6da3b7.ps1:18 char:22
==> azure-arm: + ... lVersion = (Invoke-RestMethod -Uri "https://dev.mysql.com/downloads/m ...
==> azure-arm: +                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
==> azure-arm:     + CategoryInfo          : InvalidOperation: (System.Net.HttpWebRequest:HttpWebRequest) [Invoke-RestMethod], WebExc
[...]

PR #8773, opened 2023-11-09
[Windows] Set user agent header to get mysql patch version

frittsy:
# Description
For those of us building an image from this repo in AWS, the `Invoke-WebRequest` call to https://dev.mysql.com/downloads/mysql/8.0.html has been failing for awhile with a 403 Forbidden, even with the `Accept` header that was added back in #8009. The protection/filtering they have on that page from their CDN provider seems to have greater restrictions on requests coming from EC2 IPs. This issue is easily replicated in the free tier of AWS.

We tested setting various combinations of headers and cookies, including this version which works with pwsh but not PowerShell 5.1 (can't customize the `Connection` header in 5.1)
```
Invoke-RestMethod -Uri "https://dev.mysql.com/downloads/mysql/8.0.html" -Headers @{
    'Accept'='text/html';
    'Accept-Encoding'='gzip';
    'Connection'='keep-alive';
    'Accept-Language'='en-US'
}
```

We'd be happy with the above solution too, I could either edit this PR or submit a different one if this is preferred. But the proposed solution here to set the user agent to act as `curl` was the simplest one we could come up with, while also ensuring it can continue to build in Azure/GitHub too.

I know supporting AWS isn't a primary concern of this repository, but my hope here is that it would make builds in Azure/GitHub more resilient in the future too. If this change isn't desired, we are able to override the script coming from this repo, so it isn't a huge deal for us.

#### Related issue:
#8004

## Check list

### DIFF
--- a/images/win/scripts/Installers/Install-MysqlCli.ps1
+++ b/images/win/scripts/Installers/Install-MysqlCli.ps1
@@ -16,7 +16,7 @@ Install-Binary -Url $InstallerURI -Name $InstallerName -ArgumentList $ArgumentLi
 $MysqlVersionMajorMinor = $MysqlVersion.ToString(2)
 
 if ($MysqlVersion.Build -lt 0) {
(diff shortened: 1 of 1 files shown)


mikhailkoliada (approved):

mikhailkoliada:
Thanks for submission!

PR #8778, opened 2023-11-10
[macos11] Unlink brew ghc and cabal-install

shamil-mubarakshin:
# Description
With brew deprecation on macos-11, `ghc` and `cabal-install` are installed by brew as dependency of `yq` when building from source. Links cause conflicts with `ghc` from `ghcup`. This PR unlinks brew packages

## Check list

### DIFF
--- a/images/macos/provision/core/haskell.sh
+++ b/images/macos/provision/core/haskell.sh
@@ -1,5 +1,13 @@
 #!/bin/bash -e -o pipefail
 
+source ~/utils/utils.sh
+
+# https://github.com/actions/runner-images/issues/8738
+if is_BigSur; then
+    (brew list ghc > /dev/null 2>&1) && brew unlink ghc
+    (brew list cabal-install > /dev/null 2>&1) && brew unlink cabal-install
+fi
+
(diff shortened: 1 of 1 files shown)


images/macos/provision/core/haskell.sh
```diff
@@ -1,5 +1,13 @@
 #!/bin/bash -e -o pipefail
 
+source ~/utils/utils.sh
+
+# https://github.com/actions/runner-images/issues/8738
+if is_BigSur; then
+    (brew list ghc > /dev/null 2>&1) && brew unlink ghc
+    (brew list cabal-install > /dev/null 2>&1) && brew unlink cabal-install
```
vpolikarpov-akvelon, line 8: I don't think you need `brew list` here. It doesn't change behavior at all. For example, about `ghc`:

- If you have `ghc` installed then both commands succeed and execution proceeds either with or without `brew list`
- If you don't have `ghc` installed then both commands fail and execution stops either with or without `brew list`

vpolikarpov-akvelon (approved):

mikhailkoliada (approved):

PR #8798, opened 2023-11-14
[Windows] Move .NET 7.0 to the toolset due to the new Visual Studio version

Alexey-Ayupov:
# Description
Move .NET 7.0 to the toolset due to the new Visual Studio version

#### Related issue:
https://github.com/actions/runner-images/issues/8797

## Check list


Sergio0694:
If the change is specific to VS 17.8, shouldn't it only be for the windows20222 image?
My understanding (as per https://github.com/actions/runner-images/issues/7612#issuecomment-1552723883) is that windows2019 doesn't have VS2022? 🤔

PR #8871, opened 2023-11-22
[Windows] Unify services handling

vpolikarpov-akvelon:
# Description

This pull request includes refactoring the code to directly use commands instead of custom functions for managing services, removing unnecessary commands, and replacing parameters with direct variables.

## Check list


images/windows/scripts/build/Configure-System.ps1
```diff
@@ -105,10 +105,10 @@ $servicesToDisable = @(
...
     'gupdatem'
     'StorSvc'
-)
-
-$servicesToDisable | Stop-SvcWithErrHandling
-$servicesToDisable | Set-SvcWithErrHandling -Arguments @{StartupType = "Disabled"}
+) | Get-Service -ErrorAction SilentlyContinue
+Stop-Service $servicesToDisable
+$servicesToDisable.WaitForStatus('Stopped', "00:01:00")
+$servicesToDisable | Set-Service -StartupType Disabled
```
shamil-mubarakshin, line 111: We had issue in the past, https://github.com/actions/runner-images/issues/8770, when similar approach wasn't handling non-existent services for custom builds. Is it ok to revert changes?

vpolikarpov-akvelon: There is a difference: I added `| Get-Service -ErrorAction SilentlyContinue` to handle that.

mikhailkoliada (approved):

PR #8883, opened 2023-11-23
[Windows] Add .NET 8 support

Gh0stWalk3r:
# Description
This pull request adds .NET 8 support in both windows json definitions. Although mentioned in https://github.com/actions/runner-images/issues/8864#issuecomment-1822301037 that there are new images available, it does not seem to be added to the windows json definitions on main.

## Check list


mikhailkoliada:
Hello! This should not be in the json file, latest .NET is always a part of VS installation :)

PR #8912, opened 2023-11-28
[Ubuntu] Rewrite function download_with_retry

vpolikarpov-akvelon:
# Description

This pull request implements simplified interface and more straightforward behavior for the `download_with_retry` function. The new implementation matches the the function with the same name in Windows image helpers (https://github.com/actions/runner-images/pull/8906), so these changes would improve overall code consistency and maintainability.

## Check list

### DIFF
--- a/images/ubuntu/scripts/helpers/install.sh
+++ b/images/ubuntu/scripts/helpers/install.sh
@@ -32,7 +32,7 @@ download_with_retry() {
             echo "Package download failed in $attempt_seconds seconds" >&2
         fi
 
-        if [ $retries -eq 0 ]; then
+        if [ "$retries" -le 1 ]; then
             total_seconds=$(($(date +%s) - download_start_time))
             echo "Package download failed after $total_seconds seconds" >&2
             exit 1
--- a/images/ubuntu/scripts/helpers/install.sh
+++ b/images/ubuntu/scripts/helpers/install.sh
@@ -4,50 +4,45 @@
 ##  Desc:  Helper functions for installing tools
 ################################################################################
 
-download_with_retries() {
-# Due to restrictions of bash functions, positional arguments are used here.
-# In case if you using latest argument NAME, you should also set value to all previous parameters.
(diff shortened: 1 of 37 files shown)


images/ubuntu/scripts/helpers/install.sh
```diff
@@ -4,50 +4,45 @@
...
+                echo "Received HTTP status code $http_code after $attempt_seconds seconds" >&2
+            fi
         else
-            echo "Error — Either HTTP response code for '$URL' is wrong - '$http_code' or exit code is not 0 - '$exit_code'. Waiting $interval seconds before the next attempt, $retries attempts left"
-            sleep $interval
+            attempt_seconds=$(($(date +%s) - attempt_start_time))
+            echo "Package download failed in $attempt_seconds seconds" >&2
+        fi
+
+        if [ $retries -eq 0 ]; then
```
shamil-mubarakshin, line 35: Will this condition be ever met? Can it be `retries -eq 1`?

images/ubuntu/toolsets/toolset-2204.json
```diff
@@ -289,8 +289,7 @@
         {"name": "fastlane"}
     ],
     "selenium": {
-        "version": "4",
-        "binary_name": "selenium-server"
```
shamil-mubarakshin, line 293: Should we really remove this key-value pair? It is used in more than 1 place and, should we need to change something, seems easier to manage

vpolikarpov-akvelon: I doubt it will ever change. So I consider this as redundant complexity that we might want to rid off. I can rollback it if you think it's better to keep this value.

images/ubuntu/scripts/helpers/install.sh
vpolikarpov-akvelon:
```suggestion
        if [ "$retries" -le 1 ]; then
```

shamil-mubarakshin (approved):

mikhailkoliada (approved):

PR #8963, opened 2023-12-06
docker: install buildx from GitHub Releases

crazy-max:
# Description

related to https://dockercommunity.slack.com/archives/C7S7A40MP/p1701735186713239

image

Installation of Docker Buildx currently relies on docker-ce packages: https://github.com/actions/runner-images/blob/3a0da34c4894c3a3013966edfbf528f98ba0d5db/images/ubuntu/scripts/build/install-docker.sh#L19

As stated in https://github.com/docker/docker-ce-packaging/issues/859#issuecomment-1558958080

> Releases of plugins (more factually: publishing of `.deb` and `.rpm` packages for those on download.docker.com) are currently coupled with releases of the Docker Engine and CLI. This is a limitation in the current release-pipeline, which was created when "docker" was a single package, and the only "release vehicle" was the Docker Engine release. Docker being a single package is no longer the case (e.g., we now have separate packages for the `docker` CLI, the Docker Engine (`dockerd`), as  `containerd`, `compose`, `buildx`, etc).

Therefore plugins such as Buildx will be released only when a new version of Docker is published. For latest Buildx 0.12.0 it means it will be for next Docker 25 which will not happen any time soon (afaik).

To mitigate this issue and make sure customers are on latest stable when using GitHub Runners we can install Docker Buildx from GitHub Releases like it's currently done for Docker Compose.

cc @shamil-mubarakshin @tonistiigi @thaJeztah @colinhemmings

## Check list


thaJeztah:
I'm a bit on the fence; download.docker.com are the official packages we provide; if we can't ship things in time, we should fix that.

shamil-mubarakshin:
Hey @crazy-max, Thank you for the contribution! We would not like to change current installation method due to maintenance concerns.

crazy-max:
> Hey @crazy-max, Thank you for the contribution! We would not like to change current installation method due to maintenance concerns.

Hey! So what about compose then? I guess you should use `docker-compose-plugin` package instead of grabbing from GH release?

mikhailkoliada:
@crazy-max we slowly moving towards providing non-latest but pinned package versions for docker toolset (including docker-ce, compose etc), due to latest versions may break customers some unexpected ways, technically we can install compose from the apt repo instead of the GitHub, but I guess the cost of that change would be nothing in terms of gains? (I mean there is no difference between docker compose packages as a deb package in the apt repo vs docker compose distributed as a GitHub release).

PR #8967, opened 2023-12-06
Prevent Install-WindowsUpdates.ps1 from failing

stefanes:
# Description
Prevent the script `Install-WindowsUpdates.ps1` from failing if there is already a `WindowsUpdateDone.txt` file create from a previous execution.

## Check list


mikhailkoliada:
Hello, please modify it on your side, we should track the failure during the image generation (we have no previous runs and we always generate everything from scratch).

PR #8968, opened 2023-12-06
Prevent Configure-DeveloperMode.ps1 from failing

stefanes:
# Description
Prevent the script `Configure-DeveloperMode.ps1` from failing if the `AllowDevelopmentWithoutDevLicense` reg key already has been set.

## Check list


mikhailkoliada:
Hello, please modify it on your side, we should track the failure during the image generation (we have no previous runs and we always generate everything from scratch).

PR #8995, opened 2023-12-11
Add required plugin section for ubuntu22 and ubuntu-minimal

Triodes:
# Description
Adds the required plugin block for ubuntu22 and ubuntu-minimal so the azure plugin is installed on `packer init`. I'm not 100% sure the version constraint is the right one for you folks, but that can easily be changed.

## Check list


mikhailkoliada:
Duplicate of https://github.com/actions/runner-images/pull/8994

Triodes:
I think you probably meant #8991 ?
But agreed! I'll wait for that PR 😄

Issue #5493, opened 2022-05-05
Kubectl in ubuntu20-20220503.1 breaks EKS integration


shallax:
### Description

When trying to use kubectl to communicate with AWS EKS, it reports the following:

```error: exec plugin: invalid apiVersion "client.authentication.k8s.io/v1alpha1"```

I can't follow the Kubectl version numbers. The previous image used Kubectl v1.23.6, this new one uses Kubectl 4.5.4?

Issue is similar to this: https://www.reddit.com/r/k8s/comments/ui46zf/deploying_on_eks_cia_travis_ci_fails/

### Virtual environments affected

- [ ] Ubuntu 18.04
- [X] Ubuntu 20.04
- [ ] Ubuntu 22.04
- [ ] macOS 10.15
- [ ] macOS 11
- [ ] macOS 12
- [ ] Windows Server 2019
- [ ] Windows Server 2022

### Image version and build link

No public example

### Expected behavior

Kubectl authentication should work.

### Actual behavior

Kubectl authentication does not work.

### Repro steps

Try using kubectl to communicate with an EKS cluster in any way.

al-cheb:
We have updated kubectl version to 1.24.0. Looks like it is a known issue - https://github.com/kubernetes/kubectl/issues/1210

shallax:
Thanks for the quick response. I went with the former set of sudo commands to fudge the version in the image and it worked.

Suspect this is going to cause a storm, so is there any chance of downgrading in the image?

bendotexe:
Great to see this being sorted so quickly! Is there an Azure DevOps Pipelines equivalent to that task?

shallax:
`steps:
- script: |
    which kubectl
    kubectl version || true
    sudo curl -o /usr/local/bin/kubectl -LO "https://dl.k8s.io/release/v1.23.6/bin/linux/amd64/kubectl"
    sudo chmod +x /usr/local/bin/kubectl
    kubectl version || true
    exit 0
  displayName: Downgrade kubectl`

then in each of your jobs (joy!):

`jobs:
- job: build_and_publish
  displayName: Build and publish
  dependsOn: ${{parameters.dependsOn}}
  steps:
  - template: /pipelines/jobs/downgrade_kubectl.yaml
....`

bendotexe:
An ADO supported root level spec would be best yeah but in the meantime that's going to unblock us by the looks of things so thanks!

ahilmathew:
[...]

PR #9021, opened 2023-12-13
[macOS] Force ruby gem to check gems

erik-bershel:
# Description

Small security improvement for macOS images

#### Related issue:

[#5493](https://github.com/actions/runner-images-internal/issues/5493)

## Check list


erik-bershel:
Increasing the security level causes the images build process to be blocked. For each gem that is installed, there is a dependent one with an expired or insecure signature.

PR #9114, opened 2024-01-05
Avoid static path definition

luka5:
# Description

Since the latest refactoring the ubuntu installer scripts are no longer in `installers` but in `build`. In the VM / during the build we still use `/imagegeneration/installers` as defined in the Packerfiles.

This PR changes the only place left where the path is defined and not used via ENV.

Afterwards I would recommend to change the envs in the Packerfiles to `/imagegeneration/build`

### DIFF
--- a/images/ubuntu/scripts/helpers/install.sh
+++ b/images/ubuntu/scripts/helpers/install.sh
@@ -46,7 +46,7 @@ download_with_retry() {
 }
 
 get_toolset_value() {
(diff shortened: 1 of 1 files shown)


shamil-mubarakshin:
Hello @luka5, thank you for the contribution! I have opened https://github.com/actions/runner-images/pull/9138 with a few modifications.
We will also take changing envs in packer templates suggestion for internal assessment.

PR #9265, opened 2024-01-31
[macOS] Install only latest runtime for Xcode - compatibility  bug

erik-bershel:
# Description
On the latest builds of `macOS-13` images, strange behaviour is observed in the installed simulators. The behaviour exists only for `Xcode 15`. If you set more than one runtime, the simulators are duplicated, which leads to errors in the future. As a possible workaround, you can only install the latest available runtime for `Xcode 15`. Users will need to download/install previous versions at runtime.

UPD: Also possible to install any other of runtimes, for `Xcode 15.0` for example.

## Check list

### DIFF
--- a/images/macos/toolsets/toolset-13.json
+++ b/images/macos/toolsets/toolset-13.json
@@ -4,9 +4,9 @@
         "x64": {
             "versions": [
                 { "link": "15.3", "version": "15.3.0-Beta+15E5178i", "install_runtimes": "true", "sha256": "CE08F7B13B072AE901C3D455C02BF374FA47F678A85B180532C5D80EC4B619C0"},
-                { "link": "15.2", "version": "15.2.0+15C500b", "install_runtimes": "true", "sha256": "04E93680C6DDBEC84666531BE412DE778AFC8EAC6AB2037F4C2BE7290818B59B"},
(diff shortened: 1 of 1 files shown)


simlay:
This probably isn't the perfect place to report this but on latest builds of `macOS-14` images, I am also seeing what I'd call strange behavior with simulators. It seems that `xcrun simctl launch <device> <app bundle>` hangs with my setup. At this point, I've filed https://github.com/sonos/dinghy/issues/207 (`cargo-dinghy` can take/build a rust binary for iOS and bundle it up as an iOS app and run it on the simulator) as it's possibly a bug in `cargo-dinghy` but when using my local m1, things work as expected.

I'm not sure how to debug this behavior. Would it be helpful for me to setup a more refined example (using a simple xcode project, building the app and using `xcrun simctl launch`?

erik-bershel:
@simlay sounds as a possible `good first issue` to be honest. If you are currently not sure that the problem is in the simulators presented in the image, then yes, the best way to check is to build a simple application and run it through the simulator. Even if it freezes, this is an alarm bell. After it you may report it as an issue - we'll investigate it. It would be great if you could provide a sample project and links to successful launches in other environments.

PR #9432, opened 2024-03-01
[Ubuntu] 22.04: install libssl1.1 via apt

vicamo:
# Description

Install libssl1.1 for jammy using apt instead, so that no upstream versioning bump would fail the build again.

#### Related issue:
It has been resolved in #9428, which updated the new version string and hash.

## Check list


mikhailkoliada (approved):

mikhailkoliada:
Thanks for the submission!

Issue #9227, opened 2024-01-24
Add Android SDK Platform-Tools on Ubuntu2204


s1204IT:
### Tool name

Android SDK Platform-Tools

### Tool license

Apache License, Version 2.0

### Add or update?

- [X] Add
- [ ] Update

### Desired version

34.0.5

### Approximate size

16M

### Brief description of tool

![](https://github.com/actions/runner-images/assets/52069677/4fb8ad3f-ba5f-4c9c-808d-9993163a51cf)

This behavior occurs in repositories that are building with Gradle using the latest AGP.

I hope you will respond in the same way as #9219.

### URL for tool's homepage

https://developer.android.com/tools/releases/platform-tools#downloads

### Provide a basic test case to validate the tool's functionality.

Building an Android app with Gradle

Gradle: 8.5
AGP: 8.2.1

### Platforms where you need the tool

- [ ] Azure DevOps
- [X] GitHub Actions

### Runner images where you need the tool

- [ ] Ubuntu 20.04
- [X] Ubuntu 22.04
- [ ] macOS 11
- [ ] macOS 12
- [ ] macOS 13
- [ ] macOS 13 Arm64
- [ ] Windows Server 2019
- [ ] Windows Server 2022

### Can this tool be installed during the build?

Yes. But, we would like it to in included with runner-image.

### Tool installation time in runtime

2s

### Are you willing to submit a PR?

Sorry, but I don't have the skills to fix it.

shamil-mubarakshin:
@s1204IT,
It has been decided not to bake emulator android component into the image at this time due to maintenance concerns and relatively short runtime installation time.
Component can be installed on runtime:
```
SDKMANAGER=${ANDROID_SDK_ROOT}/cmdline-tools/latest/bin/sdkmanager
echo "y" | ${SDKMANAGER} "platform-tools"
```

Issue #9219, opened 2024-01-23
Add Android emulator on Ubuntu2204


yschimke:
### Tool name

Android emulator on Ubuntu2204

### Tool license

Android SDK license

### Add or update?

- [ ] Add
- [X] Update

### Desired version

Same as macos-latest

### Brief description of tool

https://github.com/actions/runner-images/blob/main/images/macos/macos-13-Readme.md includes the android emulator (33.1.24)

But https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md

### Platforms where you need the tool

- [ ] Azure DevOps
- [ ] GitHub Actions

### Runner images where you need the tool

- [ ] Ubuntu 20.04
- [X] Ubuntu 22.04
- [ ] macOS 11
- [ ] macOS 12
- [ ] macOS 13
- [ ] macOS 13 Arm64
- [ ] Windows Server 2019
- [ ] Windows Server 2022

yschimke:
Hit this switching from macos-latest to ubuntu-latest in order to use the new KVM support.

https://issuetracker.google.com/issues/321876773

Example PR - https://github.com/android/wear-os-samples/pull/989/files

Without this, each build must install the emulator in order to use the new Android emulator launcher in gradle managed devices.

shamil-mubarakshin:
Hey @yschimke, we will take a look

shamil-mubarakshin:
@yschimke,
It has been decided not to bake emulator android component into the image at this time due to maintenance concerns and relatively short runtime installation time.
Component can be installed on runtime:
```
SDKMANAGER=${ANDROID_SDK_ROOT}/cmdline-tools/latest/bin/sdkmanager
echo "y" | ${SDKMANAGER} "tools"
```

shamil-mubarakshin:
No plans as of yet

NavDevganMistplay:
I am running macos-latest VM for running android emulators. I only have to install JDK 17 and appium version to run my script on emulators. For switching to ubuntu, do we have an ticket open or steps to follow to setup emulator environment so my selenium java can be run on ubuntu as well ? Thanks in advance.

NavDevganMistplay:
I am running macos-latest with no issues to start my selenium java appium scripts on the emulators. I am trying to switch to ubuntu VM but being faced with issues like appium driver not being triggered. What changes do i have to do to my .yml script to get my script get running on ubuntu emulator as well.Thanks

karlhorky:
[...]

PR #9493, opened 2024-03-12
Add emulator and platform-tools of Android SDK to Ubuntu

s1204IT:
# Description

We has discussed this topic before, but we were not satisfied with the outcome.

The main reasons for this are:
- When we build an Android app or library using Gradle, the packages I added are reinstalled every time.
- This isn't just happening in a few repositories, but in a very large number of repositories.

Also, as a counterargument to the previous answer:

> relatively short runtime installation time.

It seems odd to reject this based on the time it takes to install, and if this reason was acceptable they should remove a ton of other non-Android packages.

> not to bake emulator android component into the image at this time due to maintenance concerns

`emulator` and `platform-tools` do not need to look back at previous versions, and unless otherwise specified, the latest version of the packages will always be required regardless of the AGP version.

> Component can be installed on runtime:

I'm sure most people know how to install it, and I wasn't asking how to do it.  
We just wished the package would be included with the image.

---

#### Related issue:
- #9227
- #9219

## Check list

### DIFF
--- a/images/ubuntu/scripts/build/install-android-sdk.sh
+++ b/images/ubuntu/scripts/build/install-android-sdk.sh
@@ -103,6 +103,9 @@ for ndk_major_version in "${android_ndk_major_versions[@]}"; do
     components+=("ndk;$ndk_full_version")
 done
 
+components+=("platform-tools;")
+components+=("emulator;")
+
 available_platforms=($($SDKMANAGER --list | sed -n '/Available Packages:/,/^$/p' | grep "platforms;android-[0-9]" | cut -d"|" -f 1))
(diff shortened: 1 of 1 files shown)


MaksimZhukov:
Hello @s1204IT!
In addition to all the above reasons that prevent us from pre-installing android `emulator` and `platform-tools`, there is another significant reason - free space. The fact is that after installing these tools, the `ANDROID_HOME` directory increases by 800 MB. We already have a [report](https://github.com/actions/runner-images/issues/9494) related to lack of free space, so we don't want to install these tools to avoid cutting down on free space even more. These tools can be easily installed in runtime as needed.
Thank you for your understanding and sorry for the inconvenience!

PR #9509, opened 2024-03-13
Fixed typos

lzandman:
# Description
Fixed some typos in the source code. Note that I also touched some files in `macos` folders. I understand that that may be off limits. I could revert those changes.


mikhailkoliada (approved):

mikhailkoliada:
@lzandman in this specific case macOS submission is ok, thanks for getting all these fixed!

PR #9562, opened 2024-03-25
Fix regex "version" capture group name case.

lzandman:
# Description
These regular expressions define a fully lowercase capture group name `version`. However, that capture group is later referred to using a `Version` property, with a capital `V`.

The underlying Powershell `$Matches` object contains the lowercase `version` property, as you can see in this picture:

version-property

PowerShell, unfortunately, ignores the case and works either way. However, I'd recommend just doing it right and using the correct capitalization.

Note: I also edited macOS source files. If that's a problem I could revert those.


sergei-pyshnoi:
Hello @lzandman . Thanks for your suggestion but we do not have plans for those changes for now.

Issue #9551, opened 2024-03-23
WiX toolkit is not in the PATH


kevgo:
### Description

WiX provides several binaries in `C:\Program Files (x86)\WiX Toolset v3.14\bin`. This directory is not in the PATH environment variable.

### Platforms affected

- [ ] Azure DevOps
- [X] GitHub Actions - Standard Runners
- [ ] GitHub Actions - Larger Runners

### Runner images affected

- [ ] Ubuntu 20.04
- [ ] Ubuntu 22.04
- [ ] macOS 11
- [ ] macOS 12
- [ ] macOS 13
- [ ] macOS 13 Arm64
- [ ] macOS 14
- [ ] macOS 14 Arm64
- [ ] Windows Server 2019
- [X] Windows Server 2022

### Image version and build link

```
Image: windows-2022
  Version: 20240317.1.0
  Included Software: https://github.com/actions/runner-images/blob/win22/20240317.1/images/windows/Windows2022-Readme.md
  Image Release: https://github.com/actions/runner-images/releases/tag/win22%2F20240317.1
```

### Is it regression?

no, it was always broken afaik

### Expected behavior

`C:\Program Files (x86)\WiX Toolset v3.14\bin` is in the PATH. I can call executables from the WiX toolkit, like `C:\Program Files (x86)\WiX Toolset v3.14\bin\candle.exe` from any directory.

### Actual behavior

`C:\Program Files (x86)\WiX Toolset v3.14\bin` is not in the PATH. I can call executables from the WiX toolkit, like `C:\Program Files (x86)\WiX Toolset v3.14\bin\candle.exe` from any directory.

### Repro steps

To reproduce, I try to call the `candle.exe` executable provided by WiX:

```pwsh
> candle.exe -?

'candle' is not recognized as an internal or external command,
operable program or batch file.
```

Checking the PATH environment variable reveals that WiX is indeed not in it:

```pwsh
> $env:PATH

[...]

PR #9567, opened 2024-03-26
Add WiX `bin` folder to `PATH`.

lzandman:
# Description
As mentioned in issue "WiX toolkit is not in the PATH" #9551 the WiX installer does not add its `bin` folder to the `PATH` environment variable. However, it does set the `WIX` environment variable, pointing to the WiX install folder. This code adds WiX `bin` folder using the path that's defined in the `WIX` environment variable, so WiX commands can be called directly from the command-line.

#### Related issue:
#9551

### DIFF
--- a/images/windows/scripts/build/Install-Wix.ps1
+++ b/images/windows/scripts/build/Install-Wix.ps1
@@ -5,4 +5,9 @@
 
 Install-ChocoPackage wixtoolset -ArgumentList "--force"
 
+# Add WiX bin folder to the PATH
+$currentPath = [Environment]::GetEnvironmentVariable("PATH", "Machine")
+$NewPath = "$currentPath;$($Env:WIX)\bin"
+[Environment]::SetEnvironmentVariable("PATH", $NewPath, "Machine")
+
 Invoke-PesterTests -TestFile "Wix"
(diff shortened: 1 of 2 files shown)


lzandman:
Should a test for this also be added?

lzandman:
> I sketched out a quick example of what we would like to see.

Why is the WiX version retrieved using a somewhat complicated registry query? Many other apps are tested by calling the tool itself. The way I see it WiX's install could be tested by this single test:

```powershell
Describe "Wix" {
    It "Wix Toolset has been installed and is available in PATH" {
        "candle.exe -?" | Should -ReturnZeroExitCode
    }
}
```

erik-bershel:
Hey @lzandman 
> Why is the WiX version retrieved using a somewhat complicated registry query?

Because on the Chocolatey side there were occasional weird packaging errors that would result in us getting unexpected versions of the WIX toolset.

erik-bershel:
I'm closing this PR because I do not have enough rights to make changes to save time, and I would like these changes to be included in the image next week.

lzandman:
> I'm closing this PR because I do not have enough rights to make changes to save time, and I would like these changes to be included in the image next week.

I don't understand. This PR fixes a real issue a user is experiencing. Is the plan to not fix this issue? Or is it just that my code isn't OK?

PR #9783, opened 2024-05-03
[Ubuntu] Use the latest version of Docker on Ubuntu

taspeotis:
# Description

This unpegs Docker from the old v24 and allows it to be latest. This is for https://github.com/actions/runner-images/issues/9773 and I've made it in case you are open to installing the latest version. It can be closed if there are reasons not to update Docker. (I'm aware of some GitHub issues in this repo about later versions of Docker on Windows, but not aware of any problems with the latest version of Docker on Ubuntu. But I understand if there are problems and you don't want to upgrade it.)

## Check list
- [ ] ~~Tests are written (if applicable)~~
- [ ] ~~Documentation is updated (if applicable)~~

### DIFF
--- a/images/ubuntu/toolsets/toolset-2004.json
+++ b/images/ubuntu/toolsets/toolset-2004.json
@@ -236,11 +236,11 @@
             },
             {
                 "package": "docker-ce-cli",
-                "version": "24.0.9"
+                "version": "latest"
             },
             {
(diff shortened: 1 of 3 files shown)


vishalkhode1:
Hi @taspeotis
Any tentative timelines when Ubuntu 24.04 image will be available ?
Thanks.

Issue #9780, opened 2024-05-02
Ubuntu-22.04 image build failing at Report Generation


bvaughan527:
### Description

The current Ubuntu-22.04 runner image is failing when executing the Generate-SoftwareReport.ps1 script with the below error:
```
==> azure-arm.build_image: /imagegeneration/SoftwareReport/Generate-SoftwareReport.ps1 : Cannot bind argument to parameter 'Path' because it is null.
==> azure-arm.build_image: + CategoryInfo          : InvalidData: (:) [Generate-SoftwareReport.ps1], ParameterBindingValidationException
==> azure-arm.build_image: + FullyQualifiedErrorId : ParameterArgumentValidationErrorNullNotAllowed,Generate-SoftwareReport.ps1
```

### Platforms affected

- [X] Azure DevOps
- [ ] GitHub Actions - Standard Runners
- [ ] GitHub Actions - Larger Runners

### Runner images affected

- [ ] Ubuntu 20.04
- [X] Ubuntu 22.04
- [ ] macOS 11
- [ ] macOS 12
- [ ] macOS 13
- [ ] macOS 13 Arm64
- [ ] macOS 14
- [ ] macOS 14 Arm64
- [ ] Windows Server 2019
- [ ] Windows Server 2022

### Image version and build link

Image Version: 20240422.1

### Is it regression?

Yes: 20240324.2

### Expected behavior

The Generate-SoftwareReport.ps1 PowerShell script should complete successfully.

### Actual behavior

The Generate-SoftwareReport.ps1 is throwing the below error:

```
==> azure-arm.build_image: /imagegeneration/SoftwareReport/Generate-SoftwareReport.ps1 : Cannot bind argument to parameter 'Path' because it is null.
==> azure-arm.build_image: + CategoryInfo          : InvalidData: (:) [Generate-SoftwareReport.ps1], ParameterBindingValidationException
==> azure-arm.build_image: + FullyQualifiedErrorId : ParameterArgumentValidationErrorNullNotAllowed,Generate-SoftwareReport.ps1
```

### Repro steps

Repro steps:
1. Follow the instructions at the below link to build a runner agent image
    Link: https://github.com/actions/runner-images/blob/ubuntu22/20240324.2/docs/create-image-and-azure-resources.md

egg-r:
[...]

PR #9792, opened 2024-05-06
[ubuntu] fix image generation

vidyasagarnimmagaddi:
modified configure-apt.sh file : moving apt-fast back to sbin right after installation

## Check list
- [ X] Related issue / work item is attached

### DIFF
--- a/images/ubuntu/scripts/build/configure-apt.sh
+++ b/images/ubuntu/scripts/build/configure-apt.sh
@@ -47,4 +47,3 @@ if ! is_ubuntu24; then
     # https://github.com/ilikenwf/apt-fast
     bash -c "$(curl -fsSL https://raw.githubusercontent.com/ilikenwf/apt-fast/master/quick-install.sh)"
 fi
-mv /usr/local/bin/apt-fast /usr/local/sbin/apt-fast
--- a/images/ubuntu/scripts/build/configure-apt.sh
+++ b/images/ubuntu/scripts/build/configure-apt.sh
@@ -47,3 +47,4 @@ if ! is_ubuntu24; then
     # https://github.com/ilikenwf/apt-fast
     bash -c "$(curl -fsSL https://raw.githubusercontent.com/ilikenwf/apt-fast/master/quick-install.sh)"
 fi
+mv /usr/local/bin/apt-fast /usr/local/sbin/apt-fast


images/ubuntu/scripts/build/configure-apt.sh
```diff
@@ -47,3 +47,4 @@ if ! is_ubuntu24; then
     # https://github.com/ilikenwf/apt-fast
     bash -c "$(curl -fsSL https://raw.githubusercontent.com/ilikenwf/apt-fast/master/quick-install.sh)"
 fi
```
mikhailkoliada, line 49: @vidyasagarnimmagaddi why do you need to move this file?

bryan-bar: Please check issue `#9780`, which states why it broke and the debugging steps we took.

mikhailkoliada: the investigation there does not seem to be correct `which` mentioned in the issue works fine for both `/usr/local/bin` and /usr/local/sbin` (I have just checked and it works fine)

images/ubuntu/scripts/build/configure-apt.sh
```diff
@@ -47,3 +47,4 @@ if ! is_ubuntu24; then
...
+mv /usr/local/bin/apt-fast /usr/local/sbin/apt-fast
```
bryan-bar, line 50: This should be within the `if` statement.

images/ubuntu/scripts/build/configure-apt.sh
mikhailkoliada: @bryan-bar both `/usr/local/bin` and `/usr/local/sbin` are in $PATH, for software report we are using the `which apt-fast` command, `which` in turn scan everything it finds in `$PATH` you can see my runtime execution in the screenshot (silly step names if you do not mind, just a debug.

image


1. line 7 - I check `which apt-fast`
2. I remove current apt-fast installation from the current image
3. I reinstall it in runtime
4. I check `which apt-fast` again, it is now in the new path
5. I run the

```
function Get-AptFastVersion {
    $versionFileContent = Get-Content (which apt-fast) -Raw
    $match = [Regex]::Match($versionFileContent, '# apt-fast v(.+)\n')
    return $match.Groups[1].Value
}
```
function, which is used in the software generator and get the version for the newly installed path, my take is that `/usr/local/bin` and `/usr/local/sbin dirs are using different permission set during generation, i.e. moving will work but it is not a root cause

bryan-bar: I notice that it says you are using version `1.10.0` under the `fix` job. That might be why it is succeeding for you but failing for us. The latest version is now `1.11.0`.

mikhailkoliada: @bryan-bar that is the mess by the tool author, we do really parse the content of the apt-fast script downloaded from the main branch here: https://github.com/ilikenwf/apt-fast/blob/master/apt-fast it still [pints at 1.10.0, but I guess I found the root of the issue and will come up with the solution soon-ish

PR #9906, opened 2024-05-21
Add nginx to MacOS 13 and 14

nyurik:
Any type of testing that requires an http server would reliably use nginx as already available on macos 11 and 12 runners, plus it will be consistent with linux and win platforms.

# Description

## Check list


Alexey-Ayupov:
Hello @nyurik.  We do not accept customers pull requests on macOS.

radeva:
Hi @nyurik ,
My name is Veselina, and I’m part of [FlyCI](https://www.flyci.net?utm_source=github&utm_medium=community&utm_campaign=recommend) - macOS runners for GitHub Actions. We currently do not have nginx embedded in the runner images, but we can add it if we find a common-case scenario. 

We would be happy to discuss your specific needs and see how we can help. Feel free to contact us at [contact@flyci.net](mailto:contact@flyci.net).

PR #9930, opened 2024-05-24
modified ports for Apache and Nginx

hemanthmanga:
The Apache and Nginx services are conflicting with the default port 80, which is typically used in Windows server,8080 for Apache and 8081 for [Nginx]This allows the user to run both Apache and Nginx services on the same Windows servers (19 and 22) without any issues.[9915](https://github.com/actions/runner-images/issues/9915)

## Check list


hemanthmanga:
Port Updation is not required.

PR #9964, opened 2024-05-30
tree package included

vidyasagarnimmagaddi:
Add tree command in ubuntu-24.04 runner [9935](https://github.com/actions/runner-images/issues/9935)

## Check list


images/ubuntu/toolsets/toolset-2404.json
```diff
@@ -134,6 +134,7 @@
             "rpm",
             "texinfo",
             "tk",
+            "tree",
```
kachick, line 137: Same diff with https://github.com/actions/runner-images/pull/9936?

PR #9979, opened 2024-06-03
[Ubuntu] do not restart provisioner on 24.04 automatically

mikhailkoliada:
# Description

Apply to 24.04 for now as it is the only impacted version, but later can be backported or exposed further as needed.

## Check list

### DIFF
--- a/images/ubuntu/scripts/build/configure-system.sh
+++ b/images/ubuntu/scripts/build/configure-system.sh
@@ -34,7 +34,7 @@ fi
 
 if is_ubuntu24; then
 # Prevent needrestart from restarting the provisioner service.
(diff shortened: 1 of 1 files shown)


images/ubuntu/scripts/build/configure-system.sh
```diff
@@ -31,3 +31,10 @@ fi
 if npm --version; then
     npm cache clean --force
 fi
+
+if is_ubuntu24; then
+# Prevent needrestart from restarting the provisioner service.
+# Currently only happens on Ubuntu 20.04, so make it conditional for the time being
```
kasper93, line 37:
```suggestion
# Currently only happens on Ubuntu 24.04, so make it conditional for the time being
```

Alexey-Ayupov (approved):

Issue #10001, opened 2024-06-05
LLVM tools are not up to date in `windows-latest` image


mgovers:
### Description

The latest Windows Server 2022 image with Visual Studio build tools

```log
** Visual Studio 2022 Developer PowerShell v17.10.1
** Copyright (c) 2022 Microsoft Corporation
```

contain Clang CL version `16.0.6`, but Clang CL version `>=17.0.0` is expected for that build tool set.

Likely related to #9990

### Platforms affected

- [ ] Azure DevOps
- [x] GitHub Actions - Standard Runners
- [ ] GitHub Actions - Larger Runners

### Runner images affected

- [ ] Ubuntu 20.04
- [ ] Ubuntu 22.04
- [ ] Ubuntu 24.04
- [ ] macOS 11
- [ ] macOS 12
- [ ] macOS 13
- [ ] macOS 13 Arm64
- [ ] macOS 14
- [ ] macOS 14 Arm64
- [ ] Windows Server 2019
- [X] Windows Server 2022

### Image version and build link

`20240603.1.0`

example of failing build: https://github.com/PowerGridModel/power-grid-model/actions/runs/9381448994/job/25830720214#step:6:51

### Is it regression?

yes,

latest passing image version: `20240514.3.0`

example of passing build: https://github.com/PowerGridModel/power-grid-model/actions/runs/9380566997/job/25827885168#step:6:51

### Expected behavior

* Clang 17.0.3 is found (according to https://learn.microsoft.com/en-us/visualstudio/install/workload-component-id-vs-build-tools?view=vs-2022 )
* Build does not fail on STL `static_assert` checks.

### Actual behavior

* Clang 16.0.6 found
```log
-- The C compiler identification is Clang 16.0.6 with MSVC-like command-line
-- The CXX compiler identification is Clang 16.0.6 with MSVC-like command-line
```
* Build fails on STL `static_assert` that checks that the Clang version `>= 17.0.0`
```log
In file included from C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Tools\MSVC\14.40.33807\include\concepts:8:
C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Tools\MSVC\14.40.33807\include\yvals_core.h(898,1): error: static assertion failed: error STL1000: Unexpected compiler version, expected Clang 17.0.0 or newer.
_EMIT_STL_ERROR(STL1000, "Unexpected compiler version, expected Clang 17.0.0 or newer.");
[...]

PR #10005, opened 2024-06-06
[Windows] Remove LLVM standalone installation

kasper93:
LLVM is already included in the Visual Studio installation. This duplication creates conflicts between versions. The current MSVC C++ standard library requires at least Clang 17, while the standalone LLVM is 16.

There is no reason to include two different LLVM versions. Use the one from the Visual Studio installation. Should anyone need standalone LLVM, they can install it with a single command: `choco install llvm`.

Fixes: #10001

## Check list


FuzzysTodd (approved):
yu8t

FKNCOIN (commented):
Okay

Kichura (approved):

kasper93:
Closing, as it seems to be not wanted change. I still think duplicating software with conflicting versions is not making things easier.

Issue #10015, opened 2024-06-07
apparmor should be disabled by default on Ubuntu


daandemeyer:
### Description

apparmor.service being enabled by default in the images causes various issues (e.g. https://gitlab.com/apparmor/apparmor/-/issues/402). Given these are ephemeral build VMs where users have full root access already and can trivially disable apparmor anyway, there's no real point in having apparmor enabled. It only serves to cause hard to debug issues. To avoid others running into these issues, I propose to disable apparmor in the runner images.

### Platforms affected

- [ ] Azure DevOps
- [X] GitHub Actions - Standard Runners
- [ ] GitHub Actions - Larger Runners

### Runner images affected

- [X] Ubuntu 20.04
- [X] Ubuntu 22.04
- [X] Ubuntu 24.04
- [ ] macOS 11
- [ ] macOS 12
- [ ] macOS 13
- [ ] macOS 13 Arm64
- [ ] macOS 14
- [ ] macOS 14 Arm64
- [ ] Windows Server 2019
- [ ] Windows Server 2022

### Image version and build link

latest

### Is it regression?

no

### Expected behavior

apparmor is disabled by default

### Actual behavior

apparmor is enabled by default

### Repro steps

- Run job
- systemctl status apparmor will show that it is running

ijunaidm:
@DaanDeMeyer - Thank you for bringing this issue to us, we are investigating on this issue and we will update you on this issue after our findings.

kit-ty-kate:
any update?

hartwork:
Here's the two projects (or pull requests) where I needed to tame AppArmor to unbreak CI for Ubuntu >=24.04 so far:
- https://github.com/hartwork/sandwine/pull/93
- https://github.com/kevin1024/vcrpy/pull/904

jsf9k:
FWIW, AppArmor [also bit me](https://github.com/fedora-cloud/docker-brew-fedora/issues/117).

saz:
Stopping apparmor and disabling the service wasn't fixing it for me.
Adding `sudo aa-teardown` helped.

```
sudo aa-teardown || true
sudo systemctl disable --now apparmor.service
```
Note: there's [a bug](https://bugs.launchpad.net/ubuntu/+source/apparmor/+bug/2093797) in aa-teardown currently, that's why I've added the `|| true`

hartwork:
[...]

PR #10024, opened 2024-06-10
[Ubuntu] disable apparmor

mikhailkoliada:
# Description

Strictly speaking we do not need any means of 3rd party hardening on runners (selinux/apparmor) so lets just disable it so it does not change system's behaviour.

## Check list

### DIFF
--- a/images/ubuntu/scripts/build/configure-environment.sh
+++ b/images/ubuntu/scripts/build/configure-environment.sh
@@ -68,3 +68,7 @@ fi
 if is_ubuntu22; then
     sed -i 's/openssl_conf = openssl_init/#openssl_conf = openssl_init/g' /etc/ssl/openssl.cnf
 fi
+
+# Disable apparmor
+systemctl is-active --quiet apparmor.service && systemctl stop apparmor.service
+systemctl disable apparmor.service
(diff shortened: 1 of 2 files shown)


DaanDeMeyer:
@mikhailkoliada Note that `systemctl stop apparmor.service` was not sufficient in my case and only running `aa-teardown` fully disabled `apparmor`. Note that the exit status of `aa-teardown` has to be ignored because of https://gitlab.com/apparmor/apparmor/-/issues/403.

DaanDeMeyer:
@mikhailkoliada I am not sure, I'm not very familiar with apparmor. As a workaround I currently run aa-teardown and then uninstall the apparmor package. That would prevent anything from ever getting enabled again. Note that `snapd` has a hard dependency on apparmor so it would be uninstalled as well, not sure if that's acceptable for the base images.

mikhailkoliada:
@DaanDeMeyer hmm, from what I've read in the docs aa-teardown is indeed needed to unload apparmor profiles but it seems if only you do not need / want to reboot your system while we can just reboot. I'm gonna run more tests to see if disabling pn systemd's side + rebot does the trick.

maxim-lobanov (approved):

Chalakadir:
> # Description
> Strictly speaking we do not need any means of 3rd party hardening on runners (selinux/apparmor) so lets just disable it so it does not change system's behaviour.
> 
> #### Related issue: #10015
> ## Check list
> * [x]  Related issue / work item is attached
> * [ ]  Tests are written (if applicable)
> * [ ]  Documentation is updated (if applicable)
> * [ ]  Changes are tested and related VM images are successfully generated

PR #10053, opened 2024-06-12
[ubuntu 24] added rlang.sh  dependency for gfortran

vidyasagarnimmagaddi:
during the issue it was observed gfortran alias not available only on ubuntu 24.04 , hence  added rlang.sh to resolve the issue. [10025](https://github.com/actions/runner-images/issues/10025)

## Check list


vidyasagarnimmagaddi:
rlang should'nt be a part of ubuntu-24 . Have to try another alternative for Gfortan issue.

PR #10202, opened 2024-07-10
Allow array of signatures for Visual Studio

ManagedUx:
# Description
It appears that there are two common Authenticode certificates that are being used to sign the Visual Studio installer.  This can be seen in the following Thumbprints and commits:

- C2048FB509F1C37A8C3E9EC6648118458AA01780 from commits 2a7b21b, 82a37d1
-  F9A7CF9FBE13BAC767F4781061332DA6E8B4E0EE from commit 3afd153

This PR updates the toolset json and helper VS installer script to allow for an array of signatures to prevent the need to continually update back and forth between the two Thumbprints.

## Check list
- [x] Tests are written (if applicable)  [NA]
- [x] Documentation is updated (if applicable) [NA]
- [x] Changes are tested and related VM images are successfully generated [Tested in org's local build environment]

### DIFF
--- a/images/windows/scripts/helpers/VisualStudioHelpers.ps1
+++ b/images/windows/scripts/helpers/VisualStudioHelpers.ps1
@@ -29,7 +29,7 @@ Function Install-VisualStudio {
         [Parameter(Mandatory)] [String] $Channel,
         [Parameter(Mandatory)] [String[]] $RequiredComponents,
         [String] $ExtraArgs = "",
-        [Parameter(Mandatory)] [String] $SignatureThumbprint
+        [Parameter(Mandatory)] [String[]] $SignatureThumbprint
     )
 
     $bootstrapperUrl = "https://aka.ms/vs/${Version}/${Channel}/vs_${Edition}.exe"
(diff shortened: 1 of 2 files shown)


kishorekumar-anchala:
taken care in [10205](https://github.com/actions/runner-images/pull/10205)

ManagedUx:
@kishorekumar-anchala Would it not be better to implement a solution that prevents having to continually flip the value back and forth between the various thumbprint values the VS team is using to sign the exe?  There are a series of commits where the setting is flipped between C2048FB509F1C37A8C3E9EC6648118458AA01780 and F9A7CF9FBE13BAC767F4781061332DA6E8B4E0EE.  This solution prevents having to update the build files each time the bootstrapper is updated and potentially signed with the alternate cert.

MichalDron:
I agree with @ManagedUx , after few hours of building it fails because new version of VS was released signed with different thumbprint.
The same for AzureCLI tool (which uses the same thumbprint). 
Just in last 1 day there are 2 commits (1 for azure cli and 1 for vs) to flip it.

PR #10220, opened 2024-07-12
Switch null builder to the SSH password auth

paveliak:
# Description

We encountered some issues with the SSH RSA key auth.  After VM is rebooted Packer fails to reconnect to it. Password auth works just fine.

We decided to not spend much time troubleshooting it because Packer scripts still require VM admin password in some places so we have to provide it anyway. Once we get rid of those places we can switch back to the RSA key auth.

## Check list

### DIFF
--- a/images/macos/templates/macOS-12.anka.pkr.hcl
+++ b/images/macos/templates/macOS-12.anka.pkr.hcl
@@ -53,6 +53,11 @@ variable "vm_password" {
   sensitive = true
 }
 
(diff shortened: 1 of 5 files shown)


vinniespina (approved):

vinniespina (approved):
agree

vinniespina (approved):
commit

vinniespina (approved):
commit

ijunaidm (approved):

PR #10355, opened 2024-07-26
Vs enterprise signature update

ManagedUx:
It appears that there are three common Authenticode certificates that are being used to sign the Visual Studio installer. This can be seen in the following Thumbprints and commits:

- C2048FB509F1C37A8C3E9EC6648118458AA01780 from commits https://github.com/actions/runner-images/commit/2a7b21b9ba118d8e468e55e427823b8176dabe4c, https://github.com/actions/runner-images/commit/82a37d19facbc55fc7adcb53fe608db0200c61b6
- F9A7CF9FBE13BAC767F4781061332DA6E8B4E0EE from commit https://github.com/actions/runner-images/commit/3afd153301d1a498c8d2be9c7abfed7cfa8e5498
- 72105B6D5F370B62FD5C82F1512F7AD7DEE5F2C0 not in any prior build here but found in the bootstrapper on the [releases page](https://learn.microsoft.com/en-us/visualstudio/releases/2022/release-history)

This PR updates the toolset json and helper VS installer script to allow for an array of signatures to prevent the need to continually update back and forth between the two Thumbprints.  

This was previously proposed in https://github.com/actions/runner-images/pull/10202, but I'm proposing it again because we ran into the same issue again yesterday where the hash was switched back to C2048FB509F1C37A8C3E9EC6648118458AA01780. 

Related issue: 
https://github.com/actions/runner-images/issues/10201
https://github.com/actions/runner-images/pull/10351

Check list
 [x] Related issue / work item is attached
 [x] Tests are written (if applicable) [NA]
 [x] Documentation is updated (if applicable) [NA]
 [x] Changes are tested and related VM images are successfully generated [Tested in org's local build environment]


kishorekumar-anchala:
HI @ManagedUx ,

  Thanks for the workaround , we made changes and tested. it will taken care with this [PR](https://github.com/actions/runner-images/pull/10593)

PR #10525, opened 2024-09-02
[Macos] Add ruby 3.2 & 3.3.

susmitamane:
# Description

- Add latest ruby versions 3.2,3.3.

## Check list


erik-bershel:
The same situation as with NDK. Tool removal and deprecation should be properly announced before applying.

sarathrajsrinivasan (approved):

PR #10556, opened 2024-09-05
Added try catch block added for azlogin

RaviAkshintala:
This PR adds a try-catch block for the `azlogin` function. * If an account is already logged in, it prints a  message "Already logged in...". * If the login fails, it prints an error message "Login failed...". This change will help to handle the login process more gracefully and provide better user feedback.

#### Related issue:
https://github.com/actions/runner-images/issues/10236

## Check list

### DIFF
--- a/helpers/GenerateResourcesAndImage.ps1
+++ b/helpers/GenerateResourcesAndImage.ps1
@@ -241,18 +241,18 @@ Function GenerateResourcesAndImage {
     try {
         # Login to Azure subscription
         try {
-            $azAccount = az account show -o none
-            Write-Warning "Already logged in..."
+             az account show -o none
+             Write-Verbose "Already logged in..."
         }
         catch {
-        if ([string]::IsNullOrEmpty($AzureClientId)) {
-            Write-Verbose "No AzureClientId was provided, will use interactive login."
-            az login --output none
-        }
-        else {
-            Write-Verbose "AzureClientId was provided, will use service principal login."
-            az login --service-principal --username $AzureClientId --password=$AzureClientSecret --tenant $AzureTenantId --output none
-          }
+            if ([string]::IsNullOrEmpty($AzureClientId)) {
+                Write-Verbose "No AzureClientId was provided, will use interactive login."
+                az login --output none
+            }
+            else {
+                Write-Verbose "AzureClientId was provided, will use service principal login."
+                az login --service-principal --username $AzureClientId --password=$AzureClientSecret --tenant $AzureTenantId --output none
+            }
         }
         az account set --subscription $SubscriptionId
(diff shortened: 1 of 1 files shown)


vidyasagarnimmagaddi (approved):

helpers/GenerateResourcesAndImage.ps1
```diff
@@ -240,13 +240,19 @@ Function GenerateResourcesAndImage {
 
     try {
         # Login to Azure subscription
+        try {
+            $azAccount = az account show -o none
```
maxim-lobanov, line 244: Where `azAccount` variable is used? Why do we need it?
`az account show -o none` always returns nothing because `-o none` means `--output none`. What do we expect to set to `azAccount` variable?

helpers/GenerateResourcesAndImage.ps1
```diff
@@ -240,13 +240,19 @@ Function GenerateResourcesAndImage {
...
+            Write-Warning "Already logged in..."
```
maxim-lobanov, line 245: Why is it warning? It looks like expected behaviour for me. Should it be `Write-Verbose`?

helpers/GenerateResourcesAndImage.ps1
```diff
@@ -240,13 +240,19 @@ Function GenerateResourcesAndImage {
 
     try {
         # Login to Azure subscription
+        try {
```
maxim-lobanov, line 243: This script will never enter `catch` block because `az account show -o none` command will never throw exception.
Please check other places in this script to see how to handle command output correctly.

helpers/GenerateResourcesAndImage.ps1
```diff
@@ -240,13 +240,19 @@ Function GenerateResourcesAndImage {
...
+            $azAccount = az account show -o none
+            Write-Warning "Already logged in..."
+        }
```
maxim-lobanov, line 246: Could you please follow PowerShell styleguide as it is recommended in this repo readme? See https://github.com/actions/runner-images/blob/main/CONTRIBUTING.md for more details on correct style.

1) `}` and `catch` should be on the same line
2) content of `catch` block should have different indentation from `catch` block itself.

helpers/GenerateResourcesAndImage.ps1
feliasson: @maxim-lobanov this is not correct, the `az account show -o none` gives error if not logged in.
Easy to reproduce locally:
```
az logout
az account show -o none
```
`No subscription found. Run 'az account set' to select a subscription.`

See my comment [here](https://github.com/actions/runner-images/pull/10556#issuecomment-2346729920)  why it would never get into the `catch` block

PR #10594, opened 2024-09-11
[macOS] Increase base disk size by 25 GB

erik-bershel:
# Description

- Add 25 GB to the base images
- Remove disk_size from the template name

#### Related issue:
https://github.com/actions/runner-images/issues/10511

## Check list


sarathrajsrinivasan (approved):

erik-bershel:
Hey @oreillyjf :wave:
There is no difference in the disk area between larger runners and standard runners. For example: https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories

This PR did not change any service contracts - I just extended base disk size in the images.

PR #10634, opened 2024-09-17
[ubuntu] Add new Swift GPG key to unlock Ubuntu builds

kishorekumar-anchala:
# Description
Ubuntu-20.04 image generation is failing due to the swift failure. This PR fixes the issue by updating the swift version signature for the latest version.

## Check list

### DIFF
--- a/images/ubuntu/scripts/build/install-swift.sh
+++ b/images/ubuntu/scripts/build/install-swift.sh
@@ -27,7 +27,8 @@ gpg --keyserver hkp://keyserver.ubuntu.com \
       '8513 444E 2DA3 6B7C 1659  AF4D 7638 F1FB 2B2B 08C4' \
       'A62A E125 BBBF BB96 A6E0  42EC 925C C1CC ED3D 1561' \
       '8A74 9566 2C3C D4AE 18D9  5637 FAF6 989E 1BC1 6FEA' \
-      'E813 C892 820A 6FA1 3755  B268 F167 DF1A CF9C E069'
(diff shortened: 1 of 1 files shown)


images/ubuntu/scripts/build/install-swift.sh
```diff
@@ -27,7 +27,8 @@ gpg --keyserver hkp://keyserver.ubuntu.com \
       '8513 444E 2DA3 6B7C 1659  AF4D 7638 F1FB 2B2B 08C4' \
       'A62A E125 BBBF BB96 A6E0  42EC 925C C1CC ED3D 1561' \
       '8A74 9566 2C3C D4AE 18D9  5637 FAF6 989E 1BC1 6FEA' \
-      'E813 C892 820A 6FA1 3755  B268 F167 DF1A CF9C E069'
+      'E813 C892 820A 6FA1 3755  B268 F167 DF1A CF9C E069' \
```
nassah, line 30: @kishorekumar-anchala thanks, we are waiting for this! 
Is it not wiser to do the following instead (line 21):
```
wget -q -O - https://swift.org/keys/all-keys.asc | gpg --import
```

This way you do not need to update keys per version (major or when devs sign with new keys).

sanderv32:
+1

images/ubuntu/scripts/build/install-swift.sh
erik-bershel: This method is more stable, but slightly less secure. We would not like to go back to public key uploading.

erik-bershel (approved):

ijunaidm (approved):

PR #10684, opened 2024-09-25
[macOS] Fix automationmodetool for macOS14.7

sarathrajsrinivasan:
# Description
Details of the two changes done below:

1. Replace 'osascript' with 'expect' for enabling automationmodetool. osascript was not working as expected with macOS 14.7. Hence replacing for all images.

2. Include `configure-autologin.sh` script for all images and set `autoLoginUserScreenLocked` value to 'false'.

## Check list

### DIFF
--- a/images/macos/scripts/build/configure-autologin.sh
+++ b/images/macos/scripts/build/configure-autologin.sh
@@ -9,6 +9,7 @@
 echo "Enabling automatic GUI login for the '$USERNAME' user.."
 python3 $HOME/bootstrap/kcpassword.py "$PASSWORD"
 /usr/bin/defaults write /Library/Preferences/com.apple.loginwindow autoLoginUser "$USERNAME"
+/usr/bin/defaults write /Library/Preferences/com.apple.loginwindow autoLoginUserScreenLocked -bool false
 
 : '
 The MIT License (MIT)
--- a/images/macos/scripts/build/configure-machine.sh
+++ b/images/macos/scripts/build/configure-machine.sh
@@ -50,21 +50,18 @@ done
 rm -f ./add-certificate
 
 # enable-automationmode-without-authentication
-
+brew install expect
 retry=10
 while [[ $retry -gt 0 ]]; do
 {
-osascript <<EOF
-    tell application "Terminal"
-        activate
-        do script "automationmodetool enable-automationmode-without-authentication"
-        delay 2
-        tell application "System Events"
-            keystroke "${PASSWORD}"
-            keystroke return
-        end tell
-    end tell
-    delay 5
+    /usr/bin/expect <<EOF
+        spawn automationmodetool enable-automationmode-without-authentication
+        expect "password"
+        send "${PASSWORD}\r"
+        expect {
+            "succeeded." { puts "Automation mode enabled successfully"; exit 0 }
+            eof
+        }
 EOF
 } && break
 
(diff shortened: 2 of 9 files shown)


images/macos/scripts/build/configure-machine.sh
```diff
@@ -50,21 +50,18 @@ done
 rm -f ./add-certificate
 
 # enable-automationmode-without-authentication
-
+brew install expect
```
paveliak, line 53: Would this tool appear in SBOM? If not then we need to make sure it's removed from the image

images/macos/scripts/build/configure-autologin.sh
```diff
@@ -6,9 +6,20 @@
...
-/usr/bin/defaults write /Library/Preferences/com.apple.loginwindow autoLoginUser "$USERNAME"
+source ~/utils/utils.sh
+
+if is_Sonoma; then
+    echo "Enabling autologin for Sonoma"
+    sudo defaults delete /Library/Preferences/com.apple.loginwindow autoLoginUser
+    sudo rm /etc/kcpassword
+    sudo sysadminctl -autologin set -userName "$USERNAME" -password "$PASSWORD"
+    /usr/bin/defaults read /Library/Preferences/com.apple.loginwindow autoLoginUser
+    echo "Validated autologin for Sonoma"
```
paveliak, line 17: There is `sysadminctl -autologin status` command. Should we use it instead to validate autologin?

images/macos/scripts/build/configure-autologin.sh
```diff
@@ -6,9 +6,20 @@
...
-echo "Enabling automatic GUI login for the '$USERNAME' user.."
-python3 $HOME/bootstrap/kcpassword.py "$PASSWORD"
-/usr/bin/defaults write /Library/Preferences/com.apple.loginwindow autoLoginUser "$USERNAME"
+source ~/utils/utils.sh
+
+if is_Sonoma; then
+    echo "Enabling autologin for Sonoma"
+    sudo defaults delete /Library/Preferences/com.apple.loginwindow autoLoginUser
+    sudo rm /etc/kcpassword
+    sudo sysadminctl -autologin set -userName "$USERNAME" -password "$PASSWORD"
```
paveliak, line 15: `sysadminctl` is available since OS13. For consistency reasons should we switch all images except OS12 to using it?

paveliak: Basically the proposal is
```
ii IsMonterey
  <old behavior>
else
  <use sysadminctl>
end
```

images/macos/scripts/build/configure-machine.sh
```diff
@@ -76,11 +73,6 @@ EOF
     sleep 10
 done
 
-echo "Getting terminal windows"
```
paveliak, line 79: Is this code deleted deliberately?

images/macos/scripts/build/configure-machine.sh
sarathrajsrinivasan: Got it. Removing from the image.

images/macos/scripts/build/configure-machine.sh
sarathrajsrinivasan: Yes..since we use "expect" command now it is no longer opening a terminal to run the code for enabling automationmodetool. Tested if there are any other terminal sessions open and dont see any.

images/macos/scripts/build/configure-autologin.sh
sarathrajsrinivasan: `sudo sysadminctl -autologin set -userName "$USERNAME" -password "$PASSWORD`  This is not working as expected. Still seeing the login window. Hence validating few items.

sarathrajsrinivasan: The issue is resolved if we set this flag to false : `autoLoginUserScreenLocked`

images/macos/scripts/build/configure-autologin.sh
sarathrajsrinivasan: Currently we are not using autologin.sh in any images other than macOS12. Will create an internal item to check how it is handled and update that accordingly.

erik-bershel (approved):

PR #10696, opened 2024-09-27
[macOS] Install packer using tap

sarathrajsrinivasan:
# Description
Packer is deprecated from Homebrew `Warning: packer has been deprecated! It will be disabled on 2024-09-27.`

Hence we are installing packer using `hashicorp/tap/packer`.

## Check list

### DIFF
--- a/images/macos/scripts/build/install-common-utils.sh
+++ b/images/macos/scripts/build/install-common-utils.sh
@@ -17,7 +17,12 @@ for package in $common_packages; do
         xcbeautify_path=$(download_with_retry "https://raw.githubusercontent.com/Homebrew/homebrew-core/d3653e83f9c029a3fddb828ac804b07ac32f7b3b/Formula/x/xcbeautify.rb")
         brew install "$xcbeautify_path"
     else
-        brew_smart_install "$package"
+        if [[ $package == "packer" ]]; then
+            # Packer has been deprecated in Homebrew. Use tap to install Packer.
+            brew install hashicorp/tap/packer
+        else
+            brew_smart_install "$package"
+        fi
     fi
 done
 


images/macos/scripts/build/install-common-utils.sh
```diff
@@ -17,7 +17,12 @@ for package in $common_packages; do
         xcbeautify_path=$(download_with_retry "https://raw.githubusercontent.com/Homebrew/homebrew-core/d3653e83f9c029a3fddb828ac804b07ac32f7b3b/Formula/x/xcbeautify.rb")
         brew install "$xcbeautify_path"
     else
-        brew_smart_install "$package"
+        if [[ $package == "packer" ]]; then
```
paveliak, line 20: Can this logic be folded into `brew_smart_install`? I assume that internally it invokes `brew install` for the provided package so if we specify `hashicorp/tap/packer` in the template then it would install the thing. Perhaps some changes might be needed in the `brew_smart_install` to handle compound name with `/`.

erik-bershel: > Can this logic be folded into brew_smart_install?

In general, yes, but it doesn't make much sense - the complication will be quite large given that this is the corner case and will exhaust itself soon. Now for `packer` there is both a formula and a tap, which makes the situation quite funny. It's easier to hardcode until the formula is deprecated completely - after that it will be possible to simplify the installation and remove the exception.
Later it will be necessary to make a separate work item to improve the function for working with taps. But now we need to quickly unlock the generation. 😞

erik-bershel (approved):

images/macos/scripts/build/install-common-utils.sh
paveliak: > Later it will be necessary to make a separate work item to improve the function for working with tap

@erik-bershel How do we track this work? Is there an issue for it?

erik-bershel: > How do we track this work? Is there an issue for it?

I placed an issue on the internal board to check on it once formulae deprecated.

PR #10765, opened 2024-10-11
Use environment variable instead of a fixed path

sujithq:
# Description
Improvement: the environment variable is defined has a default value. It worked when you use the default value but it fails when you change it. Using the environment variable the script is in sync with the template variable

## Check list


Kichura (approved):

invisiblepancake (approved):

sujithq:
Could you explain why it is closed?

invisiblepancake (commented):
[](url)

PR #10794, opened 2024-10-15
[Ubuntu 2404] Allow pip to install packages in default environment

subir0071:
# Description
This is a temporary workaround put in place to allow users of ubuntu-24.04 runner images use pip install in the default environment.

#### Related issue:

[runner-images/issues#10781](https://github.com/actions/runner-images/issues/10781)
[c2c-actions-support/issues#4413](https://github.com/github/c2c-actions-support/issues/4413)

VM image is successfully [generated](https://github.com/actions/runner-images-ci/actions/runs/11354988709).
VM image is successfully [deployed](https://github.com/actions/runner-images-ci/actions/runs/11355755511).

## Check list

### DIFF
--- a/images/ubuntu/scripts/build/install-python.sh
+++ b/images/ubuntu/scripts/build/install-python.sh
@@ -19,6 +19,13 @@ export PIPX_HOME=/opt/pipx
 if is_ubuntu24; then
     apt-get install --no-install-recommends pipx
     pipx ensurepath
+
+# Create temporary workaround to allow user to continue using pip
+    sudo cat <<EOF > /etc/pip.conf
+[global]
+break-system-packages = true
+EOF
+
 else
     python3 -m pip install pipx
     python3 -m pipx ensurepath


images/ubuntu/scripts/build/install-python.sh
```diff
@@ -19,6 +19,13 @@ export PIPX_HOME=/opt/pipx
 if is_ubuntu24; then
     apt-get install --no-install-recommends pipx
     pipx ensurepath
+
+# Create temporary workaround to allow user to continue using pip
+    sudo cat <<EOF > /etc/pip.conf
```
erik-bershel, line 24:
```suggestion
sudo cat <<EOF > /etc/pip.conf
```

ijunaidm (approved):

notatallshaw:
The better approach to this would have to just delete the `EXTERNALLY-MANAGED` file, this approach has two major problems:

1. It affects pip only, and not any other Python package tool that is PEP 668 compliant
2. If a user wants to put an `EXTERNALLY-MANAGED` file in their own environment this overrides pip checking for that

The second one is the main reason that pip's CI has been stuck on Ubuntu 22.04 worker for the last year because I was unable to determine why PEP 668 tests were failing (https://github.com/pypa/pip/pull/13152 and https://github.com/pypa/pip/pull/13853).

PR #10930, opened 2024-11-08
Revert "Ubuntu 22.04 (20241103) Image Update"

Liabaemt:
Reverts actions/runner-images#10899


Kichura:
What's the reasoning for the revert? also this is not a python project as this project primarily focuses on powershell instead.

PR #10970, opened 2024-11-14
Disable `rustup` self-updating

madsmtm:
# Description
`rustup` [has a mechanism for self-updating](https://rust-lang.github.io/rustup/basics.html#keeping-rustup-up-to-date), which triggers whenever you request it to install a new toolchain. This is annoying to have happen in CI though, since it increases build time for no benefit (both the check itself, and the eventual update if rustup deems it necessary).

So let's disable this!

Since this is a behavioural change, it [could be considered a breaking change](https://xkcd.com/1172/), though I will argue that it isn't, people shouldn't need to update `rustup` itself during a build (`rustup update` to fetch a newer `rustc`/`cargo` still works fine).

#### Related issue:
https://github.com/actions/runner-images/issues/246, the proposed workaround in there (`--no-self-update`) wouldn't be necessary any more.

## Check list

### DIFF
--- a/images/macos/scripts/build/install-rust.sh
+++ b/images/macos/scripts/build/install-rust.sh
@@ -15,6 +15,15 @@ rustup-init -y --no-modify-path --default-toolchain=stable --profile=minimal
 echo "Initialize environment variables..."
 CARGO_HOME=$HOME/.cargo
 
+echo "Disable rustup self-updating"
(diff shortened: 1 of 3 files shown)


vidyasagarnimmagaddi:
Closing this as not planned.

madsmtm:
Could you elaborate on why?

PR #10971, opened 2024-11-14
[macOS] Pin version for tcl/tk and pkg-config

sarathrajsrinivasan:
# Description

1. Pin version for tcl/tk to 8.
2. Pin version for pkg-config to 0.29.2. Check and migrate to "pkgconf" if basic functionality is identical.

## Check list


images/macos/toolsets/toolset-13.json
```diff
@@ -72,7 +72,6 @@
             "p7zip",
             "packer",
             "perl",
-            "pkg-config",
             "swiftformat",
             "tcl-tk",
```
erik-bershel, line 76: What will be changed if we just use a package with a version here instead of another condition? 🤔

sarathrajsrinivasan: We could update as "tcl-tk@8" directly here as well..just thought this would be neat and similar format to earlier pinned versions.

erik-bershel (approved):

sureshe456 (approved):

images/macos/toolsets/toolset-13.json
erik-bershel: > earlier pinned versions

Ah, we simply didn't have the opportunity to pin it normally through the toolset - there are edge cases or methods that you can't call popular. It's better to control versioning through the toolset - it's the most visual file.

PR #11015, opened 2024-11-21
[macOS] Replace pkg-config with pkgconf in all macOS images

sarathrajsrinivasan:
# Description

Replace pkg-config with pkgconf in all macOS images.

## Check list


erik-bershel (approved):

erik-bershel (approved):

moratom:
@erik-bershel when roughly will this get deployed?

We've just pushed a new release on https://github.com/luxonis/depthai-core and the CI fails due to the issue, so I'm trying to decide if we just wait for the new version to be deployed or if we force push the release with the fixed CI.

PR #11173, opened 2024-12-12
Implement az-cli warmup on ubuntu

### LINKED ISSUE #10110: First run of az-cli is slow
### Description

For quite some time we have been impacted by poor performance due to the nature of az-cli's command and module caching on it's first run. It would seem that az-cli was not written with ephemeral runners in mind, and it doesn't look like they could change this easily. Hence why we are opening this issue here.

We can see that the issue is fixed on windows images as a simple warmup has been implemented
https://github.com/actions/runner-images/blob/c5c9a7cd7b2cfd20d740f7667fdaca8a412f2b52/images/windows/scripts/build/Install-AzureCli.ps1#L22

Can you please implement a simple warm up like on the windows runners?

From a sample of our logs we have measured the following impact
Minimum 2 s
Maximum 31 s
Average 5 s

We estimate wasting roughly 6500 minutes per month which is more than 10% of our of total usage

### Platforms affected

- [x] Azure DevOps
- [X] GitHub Actions - Standard Runners
- [X] GitHub Actions - Larger Runners

### Runner images affected

- [X] Ubuntu 20.04
- [X] Ubuntu 22.04
- [X] Ubuntu 24.04
- [ ] macOS 11
- [ ] macOS 12
- [ ] macOS 13
- [ ] macOS 13 Arm64
- [ ] macOS 14
- [ ] macOS 14 Arm64
- [ ] Windows Server 2019
- [ ] Windows Server 2022

https://github.com/actions/runner-images/releases/tag/ubuntu22%2F20240609.1

### Is it regression?

no

### Expected behavior

A build-in warmup in the ubuntu image creation similar to the windows image

### Actual behavior

No warmup

### Repro steps

None needed

lanni-energinet:
# Description
This PR fixes #10110.

- Config + cli-extensions directory is placed in /opt but not in installation dir at /opt/az/ as this is bad practice.
- Removed cli-extensions directory setup from devops extension as this setting belongs in az-cli setup
- Added cleanup for telemetry / logs in config directory

#### Related issue:

#10110

## Check list

### DIFF
--- a/images/ubuntu/scripts/build/install-azure-cli.sh
+++ b/images/ubuntu/scripts/build/install-azure-cli.sh
@@ -4,12 +4,40 @@
 ##  Desc:  Install Azure CLI (az)
 ################################################################################
 
+# Source the helpers for use with the script
+source $HELPER_SCRIPTS/etc-environment.sh
+
+# AZURE_CONFIG_DIR shell variable defines where the CLI configuration file for managing behavior are stored
+# https://learn.microsoft.com/en-us/cli/azure/azure-cli-configuration#cli-configuration-file
+# This path SHOULD be different from the installation directory /opt/az/
+export AZURE_CONFIG_DIR="/opt/az-config"
+mkdir -p $AZURE_CONFIG_DIR
+set_etc_environment_variable "AZURE_CONFIG_DIR" "${AZURE_CONFIG_DIR}"
+
+# AZURE_EXTENSION_DIR shell variable defines where modules are installed
+# https://docs.microsoft.com/en-us/cli/azure/azure-cli-extensions-overview
+# This path SHOULD be different from the installation directory /opt/az/
+export AZURE_EXTENSION_DIR="/opt/az-extension"
+mkdir -p $AZURE_EXTENSION_DIR
+set_etc_environment_variable "AZURE_EXTENSION_DIR" "${AZURE_EXTENSION_DIR}"
+
 # Install Azure CLI (instructions taken from https://docs.microsoft.com/en-us/cli/azure/install-azure-cli)
 curl -fsSL https://aka.ms/InstallAzureCLIDeb | sudo bash
 
 echo "azure-cli https://docs.microsoft.com/en-us/cli/azure/install-azure-cli-linux?pivots=apt" >> $HELPER_SCRIPTS/apt-sources.txt
 
-rm -f /etc/apt/sources.list.d/azure-cli.list
-rm -f /etc/apt/sources.list.d/azure-cli.list.save
+# Remove Azure CLI repository (instructions taken from https://learn.microsoft.com/en-us/cli/azure/install-azure-cli-linux?pivots=apt#uninstall-azure-cli)
+rm -f /etc/apt/sources.list.d/azure-cli.sources
+
+echo "Warmup 'az'"
+az --help > /dev/null
+if [ $? -ne 0 ]; then
+    echo "Command 'az --help' failed"
+    exit 1
+fi
+
+# Hand over the ownership of the directories and files to the non-root user
+chown -R "$SUDO_USER:$SUDO_USER" $AZURE_CONFIG_DIR
+chown -R "$SUDO_USER:$SUDO_USER" $AZURE_EXTENSION_DIR
 
 invoke_tests "CLI.Tools" "Azure CLI"
--- a/images/ubuntu/scripts/build/cleanup.sh
+++ b/images/ubuntu/scripts/build/cleanup.sh
@@ -19,6 +19,13 @@ if command -v journalctl; then
     journalctl --vacuum-time=1s
 fi
 
+# remove redundant folders from azcli
+if [[ -z "${AZURE_CONFIG_DIR}" ]]; then
+    rm -rf $AZURE_CONFIG_DIR/logs
+    rm -rf $AZURE_CONFIG_DIR/commands
+    rm -rf $AZURE_CONFIG_DIR/telemetry
+fi
+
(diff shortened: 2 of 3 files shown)


subir0071 (approved):
Thanks a lot for your contribution @lanni-energinet .
We are merging this PR based on our internal test results.

Tolsto:
This change breaks `azure/cli` with `azure/login` for me. See https://github.com/Azure/cli/issues/175

images/ubuntu/scripts/build/install-azure-cli.sh
```diff
@@ -4,12 +4,40 @@
 ##  Desc:  Install Azure CLI (az)
 ################################################################################
 
+# Source the helpers for use with the script
+source $HELPER_SCRIPTS/etc-environment.sh
+
+# AZURE_CONFIG_DIR shell variable defines where the CLI configuration file for managing behavior are stored
+# https://learn.microsoft.com/en-us/cli/azure/azure-cli-configuration#cli-configuration-file
+# This path SHOULD be different from the installation directory /opt/az/
```
jiasli, line 12: The configuration directory of Azure CLI on an agent is `/home/runner/.azure`.

You may see this using this action:

```yaml
name: Run Azure Login

on:
  workflow_dispatch:

jobs:
  test:
    runs-on: ubuntu-24.04
    steps:   
    - uses: azure/login@v2
      with:
         creds: ${{ secrets.AZURE_CREDENTIALS }}

    - run: az account get-access-token --debug
```

The debug log shows:

```
DEBUG: cli.azure.cli.core.auth.persistence: build_persistence: location='/home/runner/.azure/service_principal_entries.json', encrypt=False
```

jiasli:
### ⚠️ **The Azure CLI product team demands an immediate revert of this PR!**

This PR was merged without fully testing and communication with Azure CLI product team. It breaks [`azure/cli`](https://github.com/marketplace/actions/azure-cli-action) action (https://github.com/Azure/cli/issues/175) because:

- `azure/cli` action assumes `AZURE_CONFIG_DIR` points to `$HOME/.azure`
- `azure/cli` action directly mounts `$HOME/.azure` into the docker container, instead of via `AZURE_CONFIG_DIR`

According to my experience, `azure/cli` action is not the only user of `$HOME/.azure`. Moving `AZURE_CONFIG_DIR` away from `$HOME/.azure` to another location may unexpectedly break many other users who rely on `$HOME/.azure`.

Therefore, please immediately revert this PR before this change is rolled out to more users and causes more damage.

lanni-energinet:
Thanks for quickly reverting the PR - the goal was to make the ship sail faster, not sink it.

As @jiasli points out there's probably even more actions that assumes `AZURE_CONFIG_DIR=$Home/.azure` or even `AZURE_CONFIG_DIR=/runner/.azure` thus the paths need to remain as-is. I'm puzzled by the lack of impact on the windows runners given the similarity of how the [windows warmup](https://github.com/actions/runner-images/blob/main/images/windows/scripts/build/Install-AzureCli.ps1) is implemented - But I'm guessing that none of the affected downstream actions support windows runners.

@subir0071 @kishorekumar-anchala  
Can either of you reopen https://github.com/actions/runner-images/issues/10110

@jiasli If you could lend you knowledge in any way to fix the original issue it would be much appreciated

jiasli:
> Config + cli-extensions directory is placed in /opt but not in installation dir at /opt/az/ as this is bad practice.

This statement is not 100% true. As I pointed out in https://github.com/actions/runner-images/pull/11173#discussion_r1916144221, the configuration directory of Azure CLI on an agent is `/home/runner/.azure`, not `/opt/az/`.

As for why `AZURE_EXTENSION_DIR` is modified to `/opt/az/azcliextensions`, I have no context. By default, it should be `~/.azure/cliextensions`: https://github.com/Azure/azure-cli/blob/d1a5844544ba30da428505867ad2f1ea39f2b950/src/azure-cli-core/azure/cli/core/extension/__init__.py#L22-L23

lanni-energinet:
The reason for moving the config directory (as it has been done on windows), is because of the poor performance of azure-cli and github runners

Calling `az version` can sometimes take as long as 2m40s.

Here is the context 
https://github.com/actions/runner-images/pull/8410
https://github.com/actions/runner-images/pull/8441
https://github.com/actions/runner-images/issues/10110

After investigating and looking through https://github.com/Azure/azure-cli and https://github.com/microsoft/knack it looks like the issue is related to the generation of commandIndex.json or close by. As I've only been able to reproduce this on the first command invocation. But only on github runners, never my local machine

Try running this workflow - it will show how inconsistent it is
```
name: 'az version is slow on first run'

on: [push]

jobs:
  test:
    name: test
    runs-on: ubuntu-latest   
    strategy:
      matrix:
        run: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20]
        
    steps:
    - run: echo ${{ matrix.run }}
    - run: az version
```

bebound:
Another issue in this PR: the `az` warmup should run after the `az extension add -n azure-devops`, because installing extension clears the command index.

MoChilia:
Is it necessary to change the `AZURE_CONFIG_DIR`? Azure/cli action has adapted to this change, but other actions or tasks—such as the `AzureContainerAppsV1` task referenced in [https://github.com/microsoft/azure-pipelines-tasks/issues/20781](https://github.com/microsoft/azure-pipelines-tasks/issues/20781) might not be compatible and could potentially break again. Can you only "Warmup 'az'" without changing environment variables?

PR #11318, opened 2025-01-07
Revert "[Windows] Update Service Fabric Versions"

subir0071:
Reverts actions/runner-images#11048

As we need to check with 1ES folks.


lawrencegripper (commented):
LGTM

lawrencegripper (approved):

Issue #11391, opened 2025-01-15
Add Ninja


purpleKarrot:
### Tool name

Ninja

### Tool license

Apache-2.0

### Add or update?

- [x] Add
- [ ] Update

### Desired version

1.11 or later

### Approximate size

> 400 kB

### Brief description of tool

Ninja is a small build system with a focus on speed. It differs from other build systems in two major respects: it is designed to have its input files generated by a higher-level build system, and it is designed to run builds as fast as possible.

### URL for tool's homepage

https://ninja-build.org/

### Provide a basic test case to validate the tool's functionality.

```bash
echo 'cmake_minimum_required(VERSION 3.10)' >> CMakeLists.txt
echo 'project(NinjaTest NONE)' >> CMakeLists.txt
cmake -GNinja .
```

### Platforms where you need the tool

- [ ] Azure DevOps
- [x] GitHub Actions

### Runner images where you need the tool

- [ ] Ubuntu 20.04
- [ ] Ubuntu 22.04
- [x] Ubuntu 24.04
- [ ] macOS 12
- [ ] macOS 13
- [ ] macOS 13 Arm64
- [ ] macOS 14
- [ ] macOS 14 Arm64
- [x] macOS 15
- [x] macOS 15 Arm64
- [ ] Windows Server 2019
- [ ] Windows Server 2022
- [ ] Windows Server 2025

### Can this tool be installed during the build?

```bash
apt-get install ninja-build
```

### Are you willing to submit a PR?

Sure, if it has a chance of getting merged.

purpleKarrot:
The page about [C++ Modules](https://cmake.org/cmake/help/latest/manual/cmake-cxxmodules.7.html) in the CMake documentation lists the following requirements:

#### Compiler

* MSVC toolset 14.34 and newer (provided with Visual Studio 17.4 and newer)
* LLVM/Clang 16.0 and newer
* GCC 14 (for the in-development branch, after 2023-09-20) and newer

#### Build System Generator

* Ninja
* Visual Studio 17 2022

---

The latest runner images all contain both CMake and a compiler that is capable to use C++20 Modules.
The images for Windows contain a capable build system.

The images for Mac and Linux are missing a capable build system. Adding Ninja solves that issue.

hemanthmanga:
Hi @purpleKarrot, Thank you for bringing this issue to our attention. We are looking into this issue and will update you on this issue after investigating.

Margen67:
#514

[...]

PR #11393, opened 2025-01-15
Add Ninja

purpleKarrot:
fixes: #11391

# Description

See linked issue for motivation. Adding it to additional toolsets should be considered.

## Check list

### DIFF
--- a/images/ubuntu/toolsets/toolset-2404.json
+++ b/images/ubuntu/toolsets/toolset-2404.json
@@ -107,6 +107,7 @@
             "g++",
             "gcc",
             "make",
+            "ninja-build",
             "jq",
             "tar",
             "unzip",


erik-bershel:
Hey @purpleKarrot!

Thank you for rising PR to cover the feature request you provided. 👍 Unfortunately we do not accept contributions to macOS branch currently. 😞 In this regard I need to ask you to correct this PR in order to keep only Ubuntu. After further review, we may request or suggest additional changes.

purpleKarrot:
> ninja can be installed very easily as a step in the workflow. Takes around 10 sec. Here is an [example](https://github.com/subir0071/hello-world/actions/runs/12840499443/job/35809204480).

Did you even have a look at this motivation? https://github.com/actions/runner-images/issues/11391#issuecomment-2592321390
I know that it can be installed as a step in the workflow, but the same is true for any other tool that is installed by default.

PR #11518, opened 2025-01-31
Add workflows to require `pinnedDetails` when minor version of a tool is pinned in a toolset file

lawrencegripper:
# Description

To improve handling of temporary minor version pinning this PR adds a process where we can track and resolve temporary minor version pins due to downstream issues.

To do this we add `pinnedDetails` to the toolset file and tooling to validate and nudge to fix these up when the date is reached.

```json
  "mingw": {
        "version": "12.2.0",
        "runtime": "ucrt",
        "pinnedDetails": {
            "link": "https://github.com/some/issue",
            "reason": "There is a downstream installer problem.",
            "review-at": "2025-06-01"
        }
    },
```

To do this the PR  adds:
- Json Schema to require `pinnedDetails` object on all Toolset properties which pin a minor `version` field
- Validate Json Schema Action which will ensure that this requirement is met before merge
- Check Pinned Versions Action which will auto create issues when the `date` of specified in `pinnedDetails` reaches near it's expiry. Runs weekly

Examples:
- A run where schema is not met https://github.com/lawrencegripper/runner-images/actions/runs/13071293659/job/36473329139
- The issue created when dates reach expiry
 - https://github.com/lawrencegripper/runner-images/issues/3

#### Related issue:

- https://github.com/actions/runner-images-internal/pull/6702

## Check list


.gitignore
```diff
@@ -397,3 +397,4 @@ launch.json
 
 # Ignore dynamic template
 images/*/*-temp.json
+.github/workflows/validate-json-schema.yml
```
lawrencegripper, line 400:
```suggestion
```

.github/workflows/validate-json-schema.yml
```diff
@@ -0,0 +1,20 @@
...
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  validate-json:
```
maxim-lobanov, line 12: We already have JSON linter: https://github.com/actions/runner-images/blob/b0de0a7049c04caed4a5e181fc29a1883ce37797/.github/workflows/linter.yml#L29C11-L29C24

I am surprised that it doesn't throw on invalid JSON files

maxim-lobanov (dismissed):

.github/workflows/check-pinned-versions.yml
```diff
@@ -0,0 +1,22 @@
...
+on:
+  schedule:
+    - cron: '0 12 * * 1'  # Run at 12:00 UTC every Monday
+
+permissions:
+  issues: write
+  contents: read
+
+jobs:
+  validate-json:
```
lawrencegripper, line 12:
```suggestion
  check-pinning-dates:
```

.github/workflows/validate-json-schema.yml
lawrencegripper, line 12:
```suggestion
  validate-json-schema:
```

lawrencegripper: That's bad naming from me, this is validating the json schema matches in `./schema/toolset-schema.json` rather than validating the json, basically another layer on top, yeah the file is valid json but does it meet the patterns and rules in the json schema.

lawrencegripper: I've renamed the job to make this clearer

maxim-lobanov (approved):

PR #11587, opened 2025-02-12
Add the ability to specify Packer Azure plugin version

Alexey-Ayupov:
# Description
This pull request removes the `packer` block from the related templates. 
Also, a new parameter `PluginVersion` is introduced for the scripts `GenerateResourcesAndImage.ps1` and `build-image.ps1`. It will allow the version of the Packer Azure plugin to be specified."

To install the Packer Azure plugin, the command `packer init` was changed to `packer plugins install`. 

#### Related issue:
https://github.com/actions/runner-images/issues/10560

## Check list

### DIFF
--- a/helpers/GenerateResourcesAndImage.ps1
+++ b/helpers/GenerateResourcesAndImage.ps1
@@ -119,6 +119,8 @@ Function GenerateResourcesAndImage {
             The default is 'ask'.
         .PARAMETER Tags
             Tags to be applied to the Azure resources created.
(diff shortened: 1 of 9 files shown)


SaloneePandeyKPMG:
Hi @Alexey-Ayupov -- after this push, do we need to include packer plugins in our script as well? Because my build is failing because of missing plugin now. I have been referring the readme documentation and I see my pipeline failed. Would be grateful if you could help

PR #11699, opened 2025-03-01
Revert "macOS 15 arm64 (20250226) Image Update"

erik-bershel:
Reverts actions/runner-images#11681


daltonclaybrook:
Will this be merged today? My builds are taking twice or three times as long as normal since I have to install 18.2 on each run.

ijunaidm1 (approved):

PR #11763, opened 2025-03-10
[macOS] Update Xcode runtimes installer helper and toolset format

erik-bershel:
# Description

Modification of the Xcode runtime installer function. Added the ability to install multiple versions for one platform, improved the mechanism for working with the toolset, added validation of options. Additionally, the formatting of the toolset section has been updated to be more readable, and the documentation has been updated accordingly. The set of runtimes remains unchanged.

## Check list

### DIFF
--- a/images/macos/toolsets/Readme.md
+++ b/images/macos/toolsets/Readme.md
@@ -6,9 +6,19 @@
   - `link` property points to the place where Xcode will be located on image. `/Applications/Xcode_<link>.app`  
   - `version` points to Xcode version that will be downloaded and installed
   - `symlinks` describes the list of aliases where symlinks will be created to
-  - `install_runtimes` is an array or boolean function to control over the related simulator runtimes, set of possible values: [ `iOS`, `watchOS`, `visionOS`, `tvOS` ], use `true` if you want to install all runtimes, use `false` if you want to skip runtimes installation
   - `sha256` used to check integrity of the Xcode installer file
-- `default` - version of Xcode to set as default (should be metched with any `link` in `versions` property)
+  - `install_runtimes` – controls the installation of simulator runtimes:
+    - `default` – installs all default runtimes.
+    - `none` – skips runtime installation.
+    - **Hashtable** – allows manual selection:
+      - Mandatory keys: `[ "iOS", "watchOS", "tvOS" ]`, plus `visionOS` for arm64 images.
+      - Values [array of string]:  
+        - `"default"` – installs the default runtime.  
+        - `"skip"` – skips installation.  
+        - Specific version numbers, e.g., `"18.2"`, `"2.2"`, `"18.3.1"`.
+        - Apple release version, e.g., `"22E5216h"`, `"17A577"`.
+
+- `default` - version of Xcode to set as default (should be matched with any `link` in `versions` property)
     **Example:** `"11.2"`  
 
 **Note:**
@@ -22,10 +32,36 @@
 
 **Example:**
 
+String format:
+
 ```json
 "versions": [
-    { "link": "16_beta_4", "version": "16.0.0-Beta.4+16A5211f", "symlinks": ["16.0"], "install_runtimes": "false", "sha256": "4270cd8021b2f7f512ce91bfc4423b25bccab36cdab21834709d798c8daade72"},
-    { "link": "15.4", "version": "15.4.0+15F31d", "install_runtimes": "true", "sha256": "82d3d61804ff3f4c7c82085e91dc701037ddaa770e542848b2477e22f4e8aa7a"}
+    { "link": "16_beta_4", "version": "16.0.0-Beta.4+16A5211f", "symlinks": ["16.0"], "install_runtimes": "none", "sha256": "4270cd8021b2f7f512ce91bfc4423b25bccab36cdab21834709d798c8daade72"},
+    { "link": "15.4", "version": "15.4.0+15F31d", "install_runtimes": "default", "sha256": "82d3d61804ff3f4c7c82085e91dc701037ddaa770e542848b2477e22f4e8aa7a"}
+]
+```
+
+Block format:
+
+```json
+"versions": [
+    {
+        "link": "16.2",
+        "version": "16.2+16C5032a",
+        "sha256": "0e367d06eb7c334ea143bada5e4422f56688aabff571bedf0d2ad9434b7290de",
+        "install_runtimes": [
+          { "iOS": ["18.0", "18.1", "18.2"] },
+          { "watchOS": "default" },
+          { "tvOS": "default" },
+          { "visionOS": "2.2" }
+        ]
+    },
+    {
+        "link": "16.1",
+        "version": "16.1+16B40",
+        "sha256": "8ca961d55981f983d21b99a95a6b0ac04905b837f6e11346ee86d28f12afe720",
+        "install_runtimes": "default"
+    }
 ]
 ```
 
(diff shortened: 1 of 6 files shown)


Alexey-Ayupov (dismissed):

ijunaidm1 (dismissed):

ijunaidm1 (approved):

images/macos/toolsets/toolset-15.json
```diff
@@ -3,18 +3,69 @@
...
-                { "link": "16.2", "version": "16.2+16C5032a", "install_runtimes": ["iOS -buildVersion 18.2", "watchOS", "tvOS" ], "sha256": "0e367d06eb7c334ea143bada5e4422f56688aabff571bedf0d2ad9434b7290de"},
-                { "link": "16.1", "version": "16.1+16B40", "install_runtimes": "true", "sha256": "8ca961d55981f983d21b99a95a6b0ac04905b837f6e11346ee86d28f12afe720"},
-                { "link": "16", "version": "16.0.0+16A242d", "symlinks": ["16.0"], "install_runtimes": "true", "sha256": "4a26c3d102a55c7222fb145e0ee1503249c9c26c6e02dc64d783c8810b37b1e3"},
-                { "link": "15.4", "version": "15.4.0+15F31d", "install_runtimes": "true", "sha256": "82d3d61804ff3f4c7c82085e91dc701037ddaa770e542848b2477e22f4e8aa7a"}
+                {
+                    "link": "16.2",
+                    "version": "16.2+16C5032a",
+                    "sha256": "0e367d06eb7c334ea143bada5e4422f56688aabff571bedf0d2ad9434b7290de",
+                    "install_runtimes": [
+                        { "iOS": "18.2" },
```
paveliak, line 11: Can I specify array here like `{"iOS": ["18.2", "18.3.1"]}`?

images/macos/toolsets/Readme.md
```diff
@@ -6,8 +6,18 @@
...
-  - `install_runtimes` is an array or boolean function to control over the related simulator runtimes, set of possible values: [ `iOS`, `watchOS`, `visionOS`, `tvOS` ], use `true` if you want to install all runtimes, use `false` if you want to skip runtimes installation
   - `sha256` used to check integrity of the Xcode installer file
+  - `install_runtimes` – controls the installation of simulator runtimes:
+    - `"all"` – installs all default runtimes.
+    - `false` – skips runtime installation.
+    - **Hashtable** – allows manual selection:
+      - Mandatory keys: `[ "iOS", "watchOS", "tvOS" ]`, plus `visionOS` for arm64 images.
+      - Values [string]:  
+        - `"default"` – installs the default runtime.  
+        - `"skip"` – skips installation.
```
paveliak, line 17: nit: Would simply omitting the runtime have the same meaning as "skip"? Just trying to think if we can avoid introducing entities that may add to cognitive load with a small ROI

images/macos/toolsets/Readme.md
```diff
@@ -6,8 +6,18 @@
   - `link` property points to the place where Xcode will be located on image. `/Applications/Xcode_<link>.app`  
   - `version` points to Xcode version that will be downloaded and installed
   - `symlinks` describes the list of aliases where symlinks will be created to
-  - `install_runtimes` is an array or boolean function to control over the related simulator runtimes, set of possible values: [ `iOS`, `watchOS`, `visionOS`, `tvOS` ], use `true` if you want to install all runtimes, use `false` if you want to skip runtimes installation
   - `sha256` used to check integrity of the Xcode installer file
+  - `install_runtimes` – controls the installation of simulator runtimes:
+    - `"all"` – installs all default runtimes.
+    - `false` – skips runtime installation.
```
paveliak, line 12: nit: Consider using same kinds of words like `all`/`none` or `true`/`false`. It is also confusing that `"all"` is quoted but `false` is not (I understand why, but still...)

images/macos/toolsets/toolset-15.json
erik-bershel: Yep

images/macos/toolsets/Readme.md
erik-bershel: In general, of course, it is possible. Why not. I just didn't want us to forget something one day. At least the script will remind us.

images/macos/toolsets/toolset-15.json
```diff
@@ -4,40 +4,82 @@
...
+                    "link": "16.2",
+                    "version": "16.2+16C5032a",
+                    "sha256": "0e367d06eb7c334ea143bada5e4422f56688aabff571bedf0d2ad9434b7290de",
+                    "install_runtimes": ["none"]
+                },
+                {
+                    "link": "16.1",
+                    "version": "16.1+16B40",
+                    "sha256": "8ca961d55981f983d21b99a95a6b0ac04905b837f6e11346ee86d28f12afe720",
+                    "install_runtimes": ["none"]
```
paveliak, line 27: There seems to be multiple way on how to specify the things:
- `"install_runtimes": ["none"]`
- `"install_runtimes": "all"`
- `{ "iOS": ["18.0", "18.1", "18.2", "18.3.1", "22E5216h"] }`
- `{ "iOS": "18.2" }`

Should we standardize on when string constants should be used in place of objects and arrays?

I think it's ok to support both `{ "iOS": "18.2" }` and `{ "iOS": ["18.2"] }`

But the notation `"install_runtimes": ["none"]` feels a bit strange. I mean the `array of "none"`

images/macos/toolsets/toolset-14.json
```diff
@@ -3,24 +3,105 @@
...
+                        { "watchOS": "default" },
+                        { "tvOS": "default" },
+                        { "visionOS": "2.2" }
+                    ]
+                },
+                {
+                    "link": "16.1",
+                    "version": "16.1+16B40",
+                    "sha256": "8ca961d55981f983d21b99a95a6b0ac04905b837f6e11346ee86d28f12afe720",
+                    "install_runtimes": ["all"]
```
paveliak, line 72: Thinking more about it maybe we could use work `default` to install everything and install nothing if `install_runtimes` property is omitted? I do not have a strong opinion here but I am trying to reduce amount of "key words" that someone would have to keep in mind...
- "default" feels like a good word and it used to be a default behavior
- and omitting  `install_runtimes` could be a good signal to not install anything

paveliak (approved):

images/macos/toolsets/toolset-15.json
erik-bershel: > I think it's ok to support both { "iOS": "18.2" } and { "iOS": ["18.2"] }
> But the notation "install_runtimes": ["none"] feels a bit strange. I mean the array of "none"

It will also work in both ways. Array of one string and one string are just the same for PS script, yeah. 
The only reason why I left arrays in json, even of one element, is so that everyone understands that there can be an arbitrary number of elements.

images/macos/toolsets/toolset-14.json
erik-bershel: > "default" feels like a good word and it used to be a default behavior

We have documentation that describes the behaviour. We know that "by default" does not always behave predictably, for example. Whatever word we choose, there will be misinterpretations until it is explained in an accessible way.

> and omitting install_runtimes could be a good signal to not install anything

If we leave the ability to remove the field, then sooner or later someone will forget about adding it and we will only discover it on release. It is better to let the script slap our hands for every possible error at the build stage.

ijunaidm1 (approved):

paveliak (approved):

PR #11946, opened 2025-04-04
[macOS] Always use the latest major version of CodeQL Action

angelapwen:
Instead of hardcoding the CodeQL Action major version, we now always use the latest major version of CodeQL Action. 

## Check list
- [N/A] Tests are written (if applicable)
- [N/A] Documentation is updated (if applicable)

### DIFF
--- a/images/macos/scripts/build/install-codeql-bundle.sh
+++ b/images/macos/scripts/build/install-codeql-bundle.sh
@@ -7,7 +7,8 @@
 source ~/utils/utils.sh
 
 # Retrieve the latest major version of the CodeQL Action to use in the base URL for downloading the bundle.
-releases=$(curl -s "https://api.github.com/repos/github/codeql-action/releases")
+[ -n "$API_PAT" ] && authString=(-H "Authorization: token ${API_PAT}")
+releases=$(curl "${authString[@]}" -s "https://api.github.com/repos/github/codeql-action/releases")
 
 # Get the release tags starting with v[0-9] and sort them in descending order, then parse the first one to get the major version.
 codeql_action_latest_major_version=$(echo "$releases" |
--- a/images/macos/scripts/build/install-codeql-bundle.sh
+++ b/images/macos/scripts/build/install-codeql-bundle.sh
@@ -23,6 +23,7 @@ fi
 
 # Retrieve the CLI version of the latest CodeQL bundle.
 defaults_json_path=$(download_with_retry "https://raw.githubusercontent.com/github/codeql-action/$codeql_action_latest_major_version/src/defaults.json")
+bundle_version=$(jq -r '.cliVersion' "$defaults_json_path")
 bundle_tag_name="codeql-bundle-v$bundle_version"
 
 echo "Downloading CodeQL bundle $bundle_version..."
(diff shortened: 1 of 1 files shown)


images/macos/scripts/build/install-codeql-bundle.sh
```diff
@@ -6,9 +6,24 @@
 
 source ~/utils/utils.sh
 
+# Retrieve the latest major version of the CodeQL Action to use in the base URL for downloading the bundle.
+releases=$(curl -s "https://api.github.com/repos/github/codeql-action/releases")
```
paveliak, line 10: This needs to use a PAT [like we do here](https://github.com/actions/runner-images/blob/79ad381addf50d2f8cc28e998e27d25894cac64f/images/macos/scripts/build/install-ruby.sh#L31) or image generation could be throttled and fail

images/macos/scripts/build/install-codeql-bundle.sh
```diff
@@ -6,9 +6,24 @@
...
+
+# Get the release tags starting with v[0-9] and sort them in descending order, then parse the first one to get the major version.
+codeql_action_latest_major_version=$(echo "$releases" |
```
paveliak, line 13: Unconditionally updating to the next major version poses compatibility risks because major versions often bring breaking changes that need to be weighed in before putting them into the image. Can we leverage [toolset-XY.json](https://github.com/actions/runner-images/blob/79ad381addf50d2f8cc28e998e27d25894cac64f/images/macos/toolsets/toolset-15.json#L154) for specifying CodeQL version so we can have a control over it?

images/macos/scripts/build/install-codeql-bundle.sh
angelapwen: Thanks! Previously we already used `curl` on Ubuntu or `download_with_retry` on macOS for another URL without the PAT (see line 25 in this file). Does `download_with_retry` help prevent the throttling?

Also, is this specific to macOS or do you recommend I use the PAT in the corresponding [Ubuntu](https://github.com/actions/runner-images/pull/11931) & [Windows](https://github.com/actions/runner-images/pull/11947) PRs?

paveliak: MacOS image generation has a higher risk of throttling so we have to use PATs. It is less of a problem for Windows/Linux images. You may build Url with this code https://github.com/actions/runner-images/blob/79ad381addf50d2f8cc28e998e27d25894cac64f/images/macos/scripts/helpers/utils.sh#L151

images/macos/scripts/build/install-codeql-bundle.sh
angelapwen: Quick clarification: we only parse the latest major version of the Action in this line in order to populate the URL in the following lines, so that we can download the correct patch version of the CodeQL bundle. 

To address the larger concern, we currently run a series of CodeQL-specific canary test workflows during image rollouts (internal repo but I'm sure you know which one I'm talking about 😸). These workflows test the latest patch versions of the CodeQL Action & CodeQL CLI and serve to make sure that changes in CodeQL are not breaking the image; and also that other changes in the image are not breaking CodeQL.

Jago8989 (commented):
IDK

prasanjitsahoo:
Related Issue : https://github.com/actions/runner-images/issues/11925

Issue #9193, opened 2024-01-18
Ubuntu 20.04 - feature request - parameterize os_disk_size_gb


bmv0161:
### Description

Recently, Our Ubuntu 20.04 image generation is failing the at the test "Image has more than 31GB free space". We have had success resolving the issue by increasing the value for [os_disk_size_gb](https://github.com/actions/runner-images/blob/f046bcaec35305f006ce3cb6fdebb161d0ea5577/images/ubuntu/templates/ubuntu-20.04.pkr.hcl#L158C25-L158C25). 

We would like a feature request to be able to pass a different value for os_disk_size_gb to the template through a Packer variable.

### Platforms affected

- [X] Azure DevOps
- [ ] GitHub Actions - Standard Runners
- [ ] GitHub Actions - Larger Runners

### Runner images affected

- [X] Ubuntu 20.04
- [ ] Ubuntu 22.04
- [ ] macOS 11
- [ ] macOS 12
- [ ] macOS 13
- [ ] macOS 13 Arm64
- [ ] Windows Server 2019
- [ ] Windows Server 2022

### Image version and build link

https://github.com/actions/runner-images/tree/f046bcaec35305f006ce3cb6fdebb161d0ea5577/images/ubuntu

### Is it regression?

Yes, likely introduced here: https://github.com/actions/runner-images/pull/9109

### Expected behavior

Successful image generation.

### Actual behavior

Pipeline errors out due to [System.Test.ps1](https://github.com/actions/runner-images/blob/f046bcaec35305f006ce3cb6fdebb161d0ea5577/images/ubuntu/scripts/tests/System.Tests.ps1). Based on [this PR](https://github.com/actions/runner-images/pull/9153), we would expect this test to be skipped. We are running the image generation on a self-hosted Azure Devops agent, using an older version of the Ubuntu 20.04 image. 

We have validated that $env:AGENT_NAME is set on the DevOps agent prior to the packer build step on the Agent, however this is not passed to the Packer build VM.

    azure-arm.build_image: Running tests from '/imagegeneration/tests/System.Tests.ps1'
    azure-arm.build_image: Describing Disk free space
    azure-arm.build_image:   [-] Image has more than 31GB free space 34ms (33ms|2ms)
    azure-arm.build_image:    Expected the actual value to be greater than or equal to 33285996544, but got 31900524544.
[...]

PR #12087, opened 2025-04-22
Make os_disk_size_gb a variable in Packer templates.

philkild:
# Description
This is a minimum change to ensure that os_disk_size_gb is declared as a variable in each of the Packer templates rather than being hard coded.  This will allow users to override the variable in the packer template using the environment variable PKR_VAR_os_disk_size_gb [see here](https://developer.hashicorp.com/packer/guides/hcl/variables#from-environment-variables).  

Issues around this have been raised before and the advice was to use sed to modify the templates as part of the build process but this feels like a much cleaner and more user friendly approach without requiring extra parameters to be passed to the templates or to the helper scripts.  

The purpose of this change is to allow users to easily create images with larger disks to suit their needs with no breaking changes or impact to existing users.

#### Related issue:
#8503
#9193 

## Check list

### DIFF
--- a/images/ubuntu/templates/ubuntu-20.04.pkr.hcl
+++ b/images/ubuntu/templates/ubuntu-20.04.pkr.hcl
@@ -94,6 +94,11 @@ variable "managed_image_resource_group_name" {
   default = "${env("ARM_RESOURCE_GROUP")}"
 }
 
+variable "os_disk_size_gb" {
+  type    = string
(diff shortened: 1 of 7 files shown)


vidyasagarnimmagaddi:
Hi @philkild , Thanks for contribution in raising the PR .At present we are not proceeding with your PR .
We will likely to implement this in our stipulated timelines . .

PR #12101, opened 2025-04-28
[Ubuntu] Use mozilla official apt repo for installing firefox

TerrorJack:
The mozillateam ppa contains outdated firefox version. Mozilla has an official apt repo that provides up-to-date firefox package in sync with other OTA release channels, so use that instead.

# Description

## Check list


subir0071:
Hi @TerrorJack - Thanks for the PR. However, at this moment we are would not like to change over to the Mozilla apt repo as there has no significant push for the user community for teh same.
We agree that there is slight delay in catching up with the latest version but that is not seen to have any noticeable impact on the user workloads.

PR #12214, opened 2025-05-17
set rustup profile to default

usamoi:
`rustup` in GitHub Actions has been updated to `1.28.2`. Since `1.28.2` (https://github.com/rust-lang/rustup/commit/f439f4837650703f17e63d3b58011218b18eb07c), when a repository uses `rust-toolchain.toml` to automatically install the nightly toolchain, rustup consults the profile specified in `~/.rustup/settings.toml` instead of defaulting to the default profile. This behavior is reasonable, as it aligns with how `rustup toolchain install` works. However, it breaks usage for nightly toolchain users on GitHub Actions because the profile in GitHub Actions is set to `minimal`, which means `rustfmt` and `clippy` are no longer installed automatically. This PR changes the profile to `default` to fix it.

### DIFF
--- a/images/macos/scripts/build/install-rust.sh
+++ b/images/macos/scripts/build/install-rust.sh
@@ -10,13 +10,10 @@ echo "Installing Rustup..."
 brew_smart_install "rustup-init"
 
 echo "Installing Rust language..."
-rustup-init -y --no-modify-path --default-toolchain=stable --profile=minimal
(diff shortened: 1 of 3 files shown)


usamoi:
Would this be accepted, or rejected?

usamoi:
It seems that nobody cares, so I just close it.

PR #12317, opened 2025-06-04
fix: wait for network config to settle

gjasny:
Sometimes the install-actions-cache.sh fails to resolve api.github.com. This is probably happening due to the reboot in the provisioning step before. Add a 30 seconds pause like in the other block.

I read the section about macOS PRs. Feel free to pick that commit and and incorporate it in any form into your code base.


erik-bershel:
Hey @gjasny!

>Sometimes the install-actions-cache.sh fails to resolve api.github.com. This is probably happening due to the reboot in the provisioning step before. Add a 30 seconds pause like in the other block.

We don't encounter this kind of error. A few questions. Do you use the API PAT in your builds? If not, we highly recommend you do - you are most likely exhausting the limit of requests per hour for the IP address in that case. If yes, what is the build based on? Is it your local hardware or some service provider?

ivanarr429:
> roku

ivanarr429:
[Escáner QR - Wifi.pdf](https://github.com/user-attachments/files/21092188/Escaner.QR.-.Wifi.pdf)
chrome_qrcode_1749411023745

ivanarr429:
### 

> `_****_`

PR #12344, opened 2025-06-09
Update labeled PR CI checks

shamil-mubarakshin:
# Description
This PR:
1. Adds 2 jobs to [trigger-ubuntu-win-build.yml](https://github.com/actions/runner-images/compare/main...shamil-mubarakshin:update-build-triggers-01?expand=1#diff-85839106dda5deb61f20672976d566041d07cf61c53f893da5032b89f274fd24):
    - `wait-completion` will periodically check status of the remote workflow run
    - `cancel-workflow` will cancel remote workflow run if check is cancelled
 2. Adds [GitHubApi.psm1](https://github.com/actions/runner-images/compare/main...shamil-mubarakshin:update-build-triggers-01?expand=1#diff-f2101d180c33401a1229089a2772e3e36ca003fd471428d7ab8e6fe1363226c5) and [WaitWorkflowCompletion.ps1](https://github.com/actions/runner-images/compare/main...shamil-mubarakshin:update-build-triggers-01?expand=1#diff-c3339c1e68bb788f916b75373684f9002bb0a815927d77b8950076ca286e6382) to support new functionality
 3. Refactors `trigger-workflow` job

## Check list

### DIFF
--- a/.github/workflows/trigger-ubuntu-win-build.yml
+++ b/.github/workflows/trigger-ubuntu-win-build.yml
@@ -31,13 +31,15 @@ jobs:
           $gitHubApi = Get-GithubApi -Repository "${env:CI_REPO}" -AccessToken "${env:CI_PR_TOKEN}"
 
           $eventType = "trigger-${{ inputs.image_type }}-build"
+          [string] $prGuid = New-Guid
           $clientPayload = @{
-              pr_title                = "${env:PR_TITLE}"
+              pr_title                = "${env:PR_TITLE} - " + $prGuid
               custom_repo             = "${{ github.event.pull_request.head.repo.full_name }}"
               custom_repo_commit_hash = "${{ github.event.pull_request.head.sha }}"
           }
 
           $gitHubApi.DispatchWorkflow($eventType, $clientPayload)
+          "PR_GUID=$prGuid" | Out-File -Append -FilePath $env:GITHUB_ENV
 
       - name: Resolve Workflow Run ID
         id: resolve
@@ -46,20 +48,30 @@ jobs:
           $gitHubApi = Get-GithubApi -Repository "${env:CI_REPO}" -AccessToken "${env:CI_PR_TOKEN}"
 
           $workflowFileName = $("{0}.yml" -f "${{ inputs.image_type }}").ToLower()
-          $WorkflowSearchPattern = "${env:PR_TITLE}"
+          $WorkflowSearchPattern = "${env:PR_GUID}"
 
-          Start-Sleep -Seconds 30
-          $workflowRuns = $gitHubApi.GetWorkflowRuns($WorkflowFileName).workflow_runs
-          $workflowRunId = ($workflowRuns | Where-Object {$_.display_title -match $WorkflowSearchPattern}).id | Select-Object -First 1
+          # It might take a few minutes for the action to start
+          $attempt = 1
+          do {
+            $workflowRuns = $gitHubApi.GetWorkflowRuns($WorkflowFileName).workflow_runs
+            $workflowRunId = ($workflowRuns | Where-Object {$_.display_title -match $WorkflowSearchPattern}).id | Select-Object -First 1
+
+            if (-not ([string]::IsNullOrEmpty($workflowRunId))) {
+              $workflowRun = $gitHubApi.GetWorkflowRun($workflowRunId)
+              Write-Host "Found the workflow run with ID $workflowRunId on attempt $attempt. Workflow run link: $($workflowRun.html_url)"
+              "ci_workflow_run_id=$workflowRunId" | Out-File -Append -FilePath $env:GITHUB_OUTPUT
+              "ci_workflow_run_url=$($workflowRun.html_url)" | Out-File -Append -FilePath $env:GITHUB_OUTPUT
+              break
+            }
+
+            Write-Host "Workflow run for $WorkflowSearchPattern pattern not found on attempt $attempt."
+            $attempt += 1
+            Start-Sleep 30
+          } until ($attempt -eq 10)
 
           if ([string]::IsNullOrEmpty($workflowRunId)) {
               throw "Failed to find a workflow run for '$WorkflowSearchPattern'."
           }
-
-          $workflowRun = $gitHubApi.GetWorkflowRun($workflowRunId)
-          Write-Host "Found the workflow run with ID $workflowRunId. Workflow run link: $($workflowRun.html_url)"
-          "ci_workflow_run_id=$workflowRunId" | Out-File -Append -FilePath $env:GITHUB_OUTPUT
-          "ci_workflow_run_url=$($workflowRun.html_url)" | Out-File -Append -FilePath $env:GITHUB_OUTPUT
 
   wait-completion:
     runs-on: ubuntu-latest


.github/workflows/trigger-ubuntu-win-build.yml
```diff
@@ -15,33 +15,74 @@ jobs:
...
     steps:
+      - name: Checkout Code
+        uses: actions/checkout@v4
+
       - name: Trigger Build workflow
         env:
           CI_PR_TOKEN: ${{ secrets.CI_PR_TOKEN }}
           PR_TITLE: ${{ github.event.pull_request.title }}
-          CI_PR: ${{ secrets.CI_REPO }}
+          CI_REPO: ${{ vars.CI_REPO }}
```
paveliak: This would expose CI repo name. Probably not a big deal but curious if this is a deliberate change?

helpers/GitHubApi.psm1
```diff
@@ -0,0 +1,97 @@
...
+        $this.AuthHeader = $this.BuildAuth($AccessToken)
+    }
+
+    [object] hidden BuildAuth([string]$AccessToken) {
+        if ([string]::IsNullOrEmpty($AccessToken)) {
+            return $null
+        }
+        $base64AuthInfo = [Convert]::ToBase64String([Text.Encoding]::ASCII.GetBytes("'':${AccessToken}"))
+        return @{
+            Authorization = "Basic ${base64AuthInfo}"
```
paveliak, line 20: Previously we used Bearer auth and passed token directly, which also helps with secret masking because token value is unchanged. Is there a particular reason in switching to Basic auth? It looks like you use the same API.

helpers/WaitWorkflowCompletion.ps1
```diff
@@ -0,0 +1,76 @@
...
+    }
+} while ($true)
+
+Write-Host "Last result: $($finishedWorkflowRun.conclusion)."
+"CI_WORKFLOW_RUN_RESULT=$($finishedWorkflowRun.conclusion)" | Out-File -Append -FilePath $env:GITHUB_ENV
+"CI_WORKFLOW_RUN_URL=$($workflowRun.html_url)" | Out-File -Append -FilePath $env:GITHUB_ENV
+
+if ($finishedWorkflowRun.conclusion -eq "success") {
+    exit 0
+} elseif ($finishedWorkflowRun.conclusion -in ("failure", "cancelled")) {
```
paveliak: Are there any other choices for the workflow conclusion that could fall through into `else`?

helpers/WaitWorkflowCompletion.ps1
```diff
@@ -0,0 +1,76 @@
...
+function Wait-ForWorkflowCompletion($WorkflowRunId, $RetryIntervalSeconds) {
+    $attemptEndTime = (Get-Date).AddMinutes($AttemptTimeoutInMinutes)
+    do {
+        Start-Sleep -Seconds $RetryIntervalSeconds
+        $workflowRun = $gitHubApi.GetWorkflowRun($WorkflowRunId)
+        if ((Get-Date) -gt $attemptEndTime) {
+            "CI_WORKFLOW_RUN_RESULT=$($workflowRun.status)/timeout" | Out-File -Append -FilePath $env:GITHUB_ENV
+            "CI_WORKFLOW_RUN_URL=$($workflowRun.html_url)" | Out-File -Append -FilePath $env:GITHUB_ENV
+            $gitHubApi.CancelWorkflowRun($workflowRunId)
+            throw "The workflow run is still in the '$($workflowRun.status)' state for $AttemptTimeoutInMinutes minutes. Cancelling the run."
```
paveliak: There is a way to define workflow timeout on the yaml. If we want to terminate CI earlier than the default 6 hours then we probably should define the timeout on the CI workflow. I do not think this code  should act as a watchdog for the CI, it just needs to wait...

.github/workflows/trigger-ubuntu-win-build.yml
```diff
@@ -15,33 +15,74 @@ jobs:
...
-          try {
-            Invoke-WebRequest -Uri "$apiRepoUrl/dispatches" -Method Post -Headers $headers -Body $bodyString | Out-Null
-          } catch {
-            throw "$($_.exception[0].message)"
-          }
+  wait-completion:
+    runs-on: ubuntu-latest
+    needs: trigger-workflow
+    outputs:
+      ci_workflow_run_id: ${{ steps.wait.outputs.ci_workflow_run_id }}
```
paveliak: nit: Consider resolving workflow run ID in the `trigger` job. The this job is solely responsible for awaiting the given workflow run and cancellation job would not need to depend on it (except for ordering reasons).

Ideally repo dispatch event would return workflow run id (which is a feature community has been longing for) and once it is supported in Actions we would simply switch to using it instead of resolving workflow id from the title.

helpers/WaitWorkflowCompletion.ps1
```diff
@@ -0,0 +1,76 @@
...
+$attempt = 1
+do {
+    $finishedWorkflowRun = Wait-ForWorkflowCompletion -WorkflowRunId $workflowRunId -RetryIntervalSeconds $RetryIntervalSeconds
+    Write-Host "Workflow run finished with result: $($finishedWorkflowRun.conclusion)"
+    if ($finishedWorkflowRun.conclusion -in ("success", "cancelled")) {
+        break
+    } elseif ($finishedWorkflowRun.conclusion -eq "failure") {
+        if ($attempt -le $MaxRetryCount) {
+            Write-Host "Workflow run will be restarted. Attempt $attempt of $MaxRetryCount"
+            $gitHubApi.ReRunFailedJobs($workflowRunId)
```
paveliak: Can we restart this job by restarting the failed check run on the PR? This would make it a natural fit into the PR experience. Normally when a check fails it requires a manual rerun. And if it already works today then we do not need these retries (besides I do not think you use them anyway)

helpers/WaitWorkflowCompletion.ps1
```diff
@@ -0,0 +1,76 @@
...
+    [int] $AttemptTimeoutInMinutes = 340,
+    [int] $RetryIntervalSeconds = 300,
+    [int] $MaxRetryCount = 0
+)
+
+Import-Module (Join-Path $PSScriptRoot "GitHubApi.psm1")
+
+function Find-TriggeredWorkflow($WorkflowFileName, $WorkflowSearchPattern) {
+    $workflowRuns = $gitHubApi.GetWorkflowRuns($WorkflowFileName).workflow_runs
+    $workflowRunId = ($workflowRuns | Where-Object {$_.display_title -match $WorkflowSearchPattern}).id | Select-Object -First 1
```
paveliak: If job is retried I assume there will be multiple runs that match the pattern? Will it pick latest? Or should we filter only the "running" workflows?

.github/workflows/trigger-ubuntu-win-build.yml
shamil-mubarakshin: Change is intentional, as we want to get run URL for easier troubleshooting

helpers/GitHubApi.psm1
shamil-mubarakshin: There is no particular reason other than reusing the code. As far as I know, Actions mask popular encodings of the secret in the logs with `base64` being one of them

helpers/WaitWorkflowCompletion.ps1
shamil-mubarakshin: Retry in this case refers to re-running the workflow run, which increases attempt count. Filtering will pick up the latest

helpers/WaitWorkflowCompletion.ps1
shamil-mubarakshin: [There are](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/about-status-checks#check-statuses-and-conclusions), should probably add `timed_out` as fail case

helpers/WaitWorkflowCompletion.ps1
shamil-mubarakshin: From what I have seen, while workflow run is in progress its jobs cannot be restarted. Such functionality will require something like a separate workflow, which is run on failure of the PR check and restarts it.
It will surely make things more robust if we want to consider retries in the future, in cases such as windows build failing due to sporadic issue 3 hours into the run, but will add complexity

helpers/WaitWorkflowCompletion.ps1
shamil-mubarakshin: Makes sense, don't really see $AttemptTimeoutInMinutes usefulness at the moment, will remove related code

Alexey-Ayupov (approved):

PR #12366, opened 2025-06-11
[Windows] Fix PowerShell parameter binding error by splitting New-Item commands

kishorekumar-anchala:
# Description

Split the New-Item commands to avoid specifying the -Path parameter multiple times. This resolves the PowerShell parameter binding error encountered during provisioning.

## Check list


kishorekumar-anchala:
it is fixed now.

PR #12383, opened 2025-06-12
[Ubuntu] Resolve TODOs, refactor /etc/environment parsing

peppapig450:
# Description
Building an Ubuntu image currently relies on **scripts/…/etc-environment.sh** that  

1. uses multiple `sed/grep` passes,  
2. edits `/etc/environment` in place (risk of truncation on failure), 
3. cannot correctly prepend/append when the variable is initially empty, and
4. cannot handle spaces in environment variables

This PR rewrites the helper in pure Bash with array-based parsing, solving  
those issues while keeping the public API  
`get_/set_/append_/prepend_/reload_` intact.

### Motivation / Context
* Resolves the TODOs in the script.
* Allows better handling of edge cases such as spaces in the environment variable.
* Improves build robustness by writing `/etc/environment` atomically.  
* Simplifies future maintenance by avoiding fragile `sed` expressions.  

### What does this PR do?

* Replaces the old sed-driven helper with a single, tested implementation.  
* Adds atomic write (`mktemp && mv`) to avoid corrupting `/etc/environment`  
  on build interruption.  
* Automatically quotes values containing spaces.  
* **No behavioral change** for any variables runner-images currently writes.

### Size / Timing Impact

| variant        | mean ± σ (10 runs, `hyperfine`) | Δ time vs old |
| -------------- | ------------------------------- | ------------- |
| **new helper** | **1.353 s ± 0.032 s**           | **−24.5 %**   |
| old helper     | 1.793 s ± 0.041 s               | baseline      |

So the refactor shaves off about 24.5% off the test's runtime.

### How was this tested?

* **Local Docker harness** — built `Dockerfile` that mimics the
  runner-images layout, then ran:

  ```bash
  # Build container
  docker build -t env-test -f Dockerfile .

  # Start it detached
  cid=$(docker run -d env-test sleep infinity)

  # Copy helper scripts into the container
  docker cp images/ubuntu/scripts/helpers/etc-environment.sh \
        "$cid":/opt/runner-images/images/ubuntu/scripts/helpers/etc-environment.sh
  docker cp etc-environment-original.sh \
        "$cid":/opt/runner-images/images/ubuntu/scripts/helpers/etc-environment-og.sh

  # Functional tests + benchmark
  docker exec "$cid" bash /opt/runner-images/test-env-only.sh
  docker exec "$cid" bash /opt/runner-images/test-env-original.sh
  docker exec "$cid" hyperfine --warmup 3 \
        'bash /opt/runner-images/test-env-only.sh' \
        'bash /opt/runner-images/test-env-original.sh'
 
 Both helpers pass; the refactor shows the speed-up reported above.

> **Reproduce locally**:
> Full Dockerfile, helper scripts are in this Gist: https://gist.github.com/peppapig450/ff3bd4f5a80a16f8366c503314bf473b

### Backwards-Compatibility / Risk Assessment
* Removed functions were dead code (`add_etc_environment_variable`,
  `replace_etc_environment_variable`) – no call sites in repo.
* Variables starting with non-alpha  chars would be skipped the first time
  the file is rewritten; runner-images never emits such keys.
* Atomic write eliminates the “empty file if build step exits mid-sed” edge case.

### Unhandled Edge Cases
* Keys containing digits or dots first:
  - This is not used in the calling scripts so this should be okay.
* Comments at line end are currently treated as part of the value:
  - Also not used in the calling scripts.
 
#### Related issue:
N/A

## Check list

### DIFF
--- a/images/ubuntu/scripts/helpers/etc-environment.sh
+++ b/images/ubuntu/scripts/helpers/etc-environment.sh
@@ -4,86 +4,212 @@
 ##  Desc:  Helper functions for source and modify /etc/environment
 ################################################################################
 
-# NB: sed expression use '%' as a delimiter in order to simplify handling
-#     values containing slashes (i.e. directory path)
-#     The values containing '%' will break the functions
+ETC_ENV="/etc/environment"
 
-get_etc_environment_variable() {
-    local variable_name=$1
+# ------------------------------------------------------------------------------
+# parse_etc_environment <file> <keys_array_ref> <map_assoc_ref>
+#   Reads key=value lines from a file into an indexed array (for order) and an
+#   associative array (for lookup). Uses a mapfile callback for processing.
+#
+#
+# Arguments:
+#   $1: file_path - The file to parse (e.g., /etc/environment).
+#   $2: keys_array_ref - The name of the indexed array to store keys in order.
+#   $3: map_assoc_ref - The name of the associative array for key-value pairs.
+# ------------------------------------------------------------------------------
+parse_etc_environment() {
+    local file_path="$1"
+    # Use namerefs (-n) to pass arrays/maps by reference for modification
+    local -n _keys="$2"
+    local -n _map="$3"
+    # Ensure the callback function is unset when this function returns
+    trap 'unset -f _parse_cb' RETURN
 
-    # remove `variable_name=` and possible quotes from the line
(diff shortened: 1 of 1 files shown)


peppapig450:
Hi @subir0071,
I just noticed this was closed without any feedback, and I wanted to check if there was a specific reason? 

If it's out of scope or not something the team’s looking to change, totally understand—just figured I’d ask before fully abandoning it.

Thanks!

PR #12439, opened 2025-06-22
fix of insalling haskell stack

kazan417:
# Description
Bug fixing.
Fix installing of haskell stack. (error is: get.haskellstack.org: Name or service not known)
First  reason is : get.haskellstack.org now is inaccesible.
Second reason it is better to use previously installed specialized package manager for haskell ghcup.


subir0071:
`get.haskellstack.org` is not inaccessible.

kazan417:
> `get.haskellstack.org` is not inaccessible.
How about second reason?
- it is better to use previously installed SPECIALIZED package manager for haskell:  ghcup.
we are using ghcup for installing cabal, but  script for installing stack. It is no really good.

PR #12469, opened 2025-06-24
[Windows] Stop pinning of Visual Studio version

subir0071:
# Description
This Change proposes to remove the pinned version of Visual Studio from the `windows-2022` and `windows-2025` runner images.
The version available in the `postGRO` should automatically get installed into the runner images, without any manual intervention. 

#### Related issue:

https://github.com/github/hosted-runners-images/issues/116

## Check list

### DIFF
--- a/images/ubuntu/templates/variable.ubuntu.pkr.hcl
+++ b/images/ubuntu/templates/variable.ubuntu.pkr.hcl
@@ -96,7 +96,7 @@ variable "private_virtual_network_with_public_ip" {
 }
 variable "os_disk_size_gb" {
   type    = number
-  default = 0
+  default = null
 }
 variable "source_image_version" {
   type    = string
--- a/images/windows/templates/variable.windows.pkr.hcl
+++ b/images/windows/templates/variable.windows.pkr.hcl
@@ -104,7 +104,7 @@ variable "private_virtual_network_with_public_ip" {
 }
 variable "os_disk_size_gb" {
   type    = number
-  default = 0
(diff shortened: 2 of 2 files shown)


images/windows/scripts/helpers/VisualStudioHelpers.ps1
```diff
@@ -45,7 +41,7 @@ Function Install-VisualStudio {
     $channelUri = "https://aka.ms/vs/${Version}/${Channel}/channel"
     $channelId = "VisualStudio.${Version}.Release"
     $productId = "Microsoft.VisualStudio.Product.${Edition}"
-    $installChannelUri = "https://aka.ms/vs/${Version}/${Channel}/${installchannel}/channel"
+    $installChannelUri = "https://aka.ms/vs/${Version}/${Channel}/channel"
```
paveliak: `installChannelUri` is the same as `channelUri`. Curious, do we still need providing it in the request? And if so, then you might consider changing this line to `$installChannelUri = $channelUri` for better visibility

paveliak (approved):

paveliak (changes requested):
Let's have a chat in context of https://developercommunity.visualstudio.com/t/Code-optimization-bug-SIMD-std::transf/10912292

vidyasagarnimmagaddi (approved):

subir0071:
Added back `installChannel` which can be empty string for the latest release or can be pinned to specific version.

PR #12474, opened 2025-06-25
Validate authenticode signature using the certificate Subject

paveliak:
# Description

This PR switches verification of authenticode signature to use the certificate subject instead of the thumbprint. This prevents  occasional verification failures caused by certficate renewal.

## Check list


thresheek:
So, ugh, an attacker can now sign the packages with any certificate that has a matching CN, and the build will happily succeed?  That does not sound like a great idea?

PR #12529, opened 2025-07-04
Validate MS Defender is not installed

Alexey-Ayupov:
# Description
This PR validate MS Defender is not installed

## Check list


brgsstm:
@Alexey-Ayupov - Are you able to add any context as to why Defender is not to be installed?  We are going to have to justify this to cyber, we have policy pushing Defender so we will need to exclude our build pipelines.

PR #12573, opened 2025-07-10
Revert "[Windows 2022] Update `strawberryperl` to `5.40.2.2`"

subir0071:
Reverts actions/runner-images#12508

Breaking cmake default version.
Hence reverting


alexbal29:
> Reverts #12508
> 
> Breaking cmake default version. Hence reverting

The 2022 image update on pull request update the perl 
IMG_0164

PR #12679, opened 2025-07-31
Add Node.js 24 and set as default across images

salmanmkc:
Updated macOS, Ubuntu, and Windows image readmes and toolset JSON files to include Node.js 24.0.0 as an available version and set it as the default Node.js version. This ensures the latest Node.js is available and used by default in all supported environments.

# Description

## Check list


erik-bershel:
We do not plan to implement these changes now - they are outside the support policy and carry some unacceptable risks. We will add new versions when the time comes in the working order.

PR #12791, opened 2025-08-19
[macOS] Pin CMake to 3.31.6 due to a backward compatibility issue for all macOS images.

sureshe456:
# Description
1. Pin CMake to version 3.31.6 via custom Homebrew tap due to compatibility issue for all macOS images.

## Check list

### DIFF
--- a/images/macos/scripts/build/install-common-utils.sh
+++ b/images/macos/scripts/build/install-common-utils.sh
@@ -15,39 +15,26 @@ for package in $common_packages; do
             brew install hashicorp/tap/packer
             ;;
 
-        #cmake)
-            # Pin cmake bottle to 3.31.6 due to a backward compatibility issue with the latest version
-            # https://github.com/actions/runner-images/issues/11926
-            #cmake_commit="b4e46db74e74a8c1650b38b1da222284ce1ec5ce"
-            #cmake_rb_link="https://raw.githubusercontent.com/Homebrew/homebrew-core/$cmake_commit/Formula/c/cmake.rb"
-            #cmake_rb_path=$(download_with_retry "$cmake_rb_link")
-            #brew install "$cmake_rb_path"
-            #;;
-
         cmake)
             # Pin cmake to 3.31.6 due to a backward compatibility issue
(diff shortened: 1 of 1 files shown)


erik-bershel (approved):

xavier2k6:
Could latest CMake 3.31.8 not be used?

erik-bershel:
Hey @xavier2k6!

Unfortunately 3.31.6 is the latest available version in brew: https://github.com/Homebrew/homebrew-core/commits/0da738a3fdbb58b610a3918af076fd4ea6433e3a/Formula/c/cmake.rb
😢

kintel:
I feel that this fix breaks more than it fixes: Some Homebrew packages depend on CMake for building, and will try to install it. Since CMake now belongs to the `local/pinned` tap, brew will fail to install such packages:
```
Error: cmake is already installed from local/pinned!
Please `brew uninstall cmake` first."
```

With this change, we now have to `brew uninstall cmake; brew untap local/pinned; brew install cmake` to work around this.

Since all of Homebrew follows their main branch, it feels weird to make CMake an exception.

Ref. https://github.com/openscad/openscad/issues/6143

erik-bershel:
Hey @kintel!

Good point. Please open an issue - we'll work on the options.

kintel:
> Hey @kintel!
> 
> Good point. Please open an issue - we'll work on the options.

#12912

ReenigneArcher:
> due to a backward compatibility issue for all macOS images

What is the backward compatibility issue that this solves? It seems to me this causes far more issues than it solves.

The biggest issue is that no homebrew package that depends on cmake can be installed.

ghost:
> > due to a backward compatibility issue for all macOS images
> 
> What is the backward compatibility issue that this solves? It seems to me this causes far more issues than it solves.
> 
> The biggest issue is that no homebrew package that depends on cmake can be installed.

It will be fixed next week when cmake will be unpin.

vkedwardli:
> What is the backward compatibility issue that this solves?

I'm curious on this one also

PR #12864, opened 2025-08-21
Improvement/conditional mde installation validation

basabaris:
# Description
Motivation:
As defined in the related issue, the change  #12529 causes that the build of images fail on the temporary virtual machines which are created on a subscription where Microsoft Defender For Cloud is enabled. Microsoft Defender For Cloud installs the MDE by default even if the  following tag is set ExcludeMdeAutoProvisioning=false.

This gives the possibility to the users to decide if they want to have MDE in their images.

Since it is a public repository which is used by people to build their own images for their self hosted runners, closing an issue by mentioning that you don't need MDE in youra agents is not a solution.

## Check list


subir0071:
Ubuntu images should not be having MDE.

PR #12911, opened 2025-08-30
Fix tools version extraction in Install Android SDK scripts

alejandrorosas:
Modify version extraction logic to capture the last version number after '-' or ';'.

# Description

Bug fixing

This change updates the method for parsing the version number for Android platform tools.

The previous method used shell parameter expansion (`${item##*[-;]}`) to get the string segment after the last `-` or `;`. This approach failed for tool names with extensions, such as `platforms;android-33-ext4`, where it would incorrectly extract `ext4` instead of the actual version, `33`.

For example current full list of tools_array is:
```
tools_array=('platforms;android-10' 'platforms;android-11' 'platforms;android-12' 'platforms;android-13' 'platforms;android-14' 'platforms;android-15' 'platforms;android-16' 'platforms;android-17' 'platforms;android-18' 'platforms;android-19' 'platforms;android-20' 'platforms;android-21' 'platforms;android-22' 'platforms;android-23' 'platforms;android-24' 'platforms;android-25' 'platforms;android-26' 'platforms;android-27' 'platforms;android-28' 'platforms;android-29' 'platforms;android-30' 'platforms;android-31' 'platforms;android-32' 'platforms;android-33' 'platforms;android-33-ext4' 'platforms;android-33-ext5' 'platforms;android-34' 'platforms;android-34-ext10' 'platforms;android-34-ext11' 'platforms;android-34-ext12' 'platforms;android-34-ext8' 'platforms;android-35' 'platforms;android-35-ext14' 'platforms;android-35-ext15' 'platforms;android-36' 'platforms;android-36-ext18' 'platforms;android-7' 'platforms;android-8' 'platforms;android-9')
```

## Check list

### DIFF
--- a/images/ubuntu/scripts/build/install-android-sdk.sh
+++ b/images/ubuntu/scripts/build/install-android-sdk.sh
@@ -15,8 +15,8 @@ add_filtered_installation_components() {
     local tools_array=("$@")
 
     for item in ${tools_array[@]}; do
-        # Take the last argument after splitting string by ';'' and '-''
-        item_version=$(echo "${item##*[-;]}")
+        # Take the last version number that appears after the last '-' or ';'
+        item_version=$(echo "$item" | grep -oP '(?<=[-;])[0-9.]+')
 
         # Semver 'comparison'. Add item to components array, if item's version is greater than or equal to minimum version
         if [[ "$(printf "${minimum_version}\n${item_version}\n" | sort -V | head -n1)" == "$minimum_version" ]]; then


erik-bershel:
We do not encounter the problem described. 🤷‍♂️ 

For example: https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md#android

alejandrorosas:
Hi @erik-bershel, thanks for taking a look.

The issue is visible in the latest `ubuntu-24.04` image. The configuration sets the minimum Android platform version to 34 ([here](https://github.com/actions/runner-images/blob/main/images/ubuntu/toolsets/toolset-2404.json#L76)).

However, the build report summary show that packages for version `33` are still being installed, specifically `platforms;android-33-ext4` and `platforms;android-33-ext5`.

image

This happens because the old parsing logic (`${item##*[-;]}`) incorrectly extracts `ext4` as the version number from `platforms;android-33-ext4`.

My PR fixes this by ensuring only the correct numeric version (`33`) is extracted, so the comparison works as intended.

images/ubuntu/scripts/build/install-android-sdk.sh
```diff
@@ -15,8 +15,8 @@ add_filtered_installation_components() {
     local tools_array=("$@")
 
     for item in ${tools_array[@]}; do
-        # Take the last argument after splitting string by ';'' and '-''
-        item_version=$(echo "${item##*[-;]}")
+        # Take the last version number that appears after the last '-' or ';'
+        item_version=$(echo "$item" | grep -oP '(?<=[-;])[0-9.]+')
```
shamil-mubarakshin: @alejandrorosas, your thoughts/concerns on updating the filter to something like below?
```bash
item_version=$(echo "$item" | grep -oE '[-;][^;-]*' | grep -oE '[0-9]+([.][0-9]+)*' | head -n1)
```
The reasoning is MacOS has the same issue, but it doesn't support `grep -P`. We would like to be consistent where possible

alejandrorosas: Hey @shamil-mubarakshin, yes! In that case we can just split the grep i proposed into two:
```
item_version=$(echo "$item" | grep -oE '[-;][0-9.]+' | grep -oE '[0-9.]+')
```
Let me update also the MacOS script

shamil-mubarakshin: Thanks! Test builds completed successfully

shamil-mubarakshin (approved):

PR #13051, opened 2025-09-21
Update ruby-builder URLs

eregon:
#### Related issue:

https://github.com/actions/runner-images/issues/13050

## Check list


erik-bershel:
It won't work due to the naming pattern in toolset. It's 3.4.* for example, but your script expects it to be 3.4.6 or whatever version is current latest. I also should "fork" macOS changes in a separate branch due to our processes.

erik-bershel:
We will most likely be forced to open a separate PR for Ubuntu within the next 24 hours to unlock generation and deployment as a result.
But in any case, we really appreciate your participation! You saved us a lot of time here.

eregon:
> It won't work due to the naming pattern in toolset. It's 3.4.* for example, but your script expects it to be 3.4.6 or whatever version is current latest.

Ah I see. If you need a list of versions you can find it https://github.com/ruby/setup-ruby/blob/master/ruby-builder-versions.json (valid for Linux & macOS)

PR #13196, opened 2025-10-21
Document Ubuntu ARM64 images

uNetworkingAB:
# Description
Document the fact you provide ARM64 variants of Ubuntu runner images


erik-bershel:
It doesn't belong to our repo.

Doc: https://docs.github.com/en/actions/concepts/runners/github-hosted-runners#runner-images
Repo: https://github.com/actions/partner-runner-images

PR #13203, opened 2025-10-22
openssl: do not install DLLs globally to system32 on Windows

jcar87:
# Description
Pass `/tasks="copytobin"` to the OpenSSL installer on Windows so that the DLLs required to run `openssl.exe` are only local to `openssl.exe` and not globally exposed to all processes in `C:/Windows/system32`.

There are other copies of the same DLLs `libssl-3-x64.dll` and `libcrypto-3-x64.dll` across the system, used by different tools, and being in `C:/Windows/system32` gives them a higher priority, which can result in runtime incompatibilities or unexpected results in some scenarios (e.g. if user expects them to be found in `PATH` environment variable, or if the expected version is newer than the one installed at the system level).

This change should **NOT** impact the ability to run and invoke the `openssl.exe` tool, which is what the existing test already covered.

#### Related issue:
https://github.com/actions/runner-images/issues/13198
## Check list

### DIFF
--- a/images/windows/scripts/build/Install-OpenSSL.ps1
+++ b/images/windows/scripts/build/Install-OpenSSL.ps1
@@ -34,7 +34,7 @@ if ($null -eq $installerUrl) {
 
 Install-Binary `
     -Url $installerUrl `
-    -InstallArgs @('/silent', '/sp-', '/suppressmsgboxes', "/DIR=`"$installDir`"") `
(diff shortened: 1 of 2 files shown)


erik-bershel:
Hey @jcar87!

Thank you so much for your contribution! 🙇 
CI run is in progress, I will merge it once got green builds!

erik-bershel (approved):

jcar87:
Thanks so much @erik-bershel!

PR #13290, opened 2025-11-11
feat: change azcopy script to download directly from github releases

PrashantRaj18198:
# Description
The `azcopy` script currently uses a static URL for downloading. There seems to be a window during new releases of azcopy where the assets are available on github releases and a new release is also available but the static URL returns 404. I don't have full context of this static URL here, how it is patched to point to the releases, but this URL doesn't seem very robust. Switching to use the github releases assets would make for a more resilient script, adding the changes for that here.

Related issue: https://github.com/actions/runner-images/issues/13279

## Check list

### DIFF
--- a/images/macos/scripts/build/install-azcopy.sh
+++ b/images/macos/scripts/build/install-azcopy.sh
@@ -6,14 +6,22 @@
 
 source ~/utils/utils.sh
 
-if is_Arm64; then
-    url="https://aka.ms/downloadazcopy-v10-mac-arm64"
+# Determine the architecture for azcopy download
+# azcopy uses "amd64" instead of "x64" in their release names
+# For reference check this specific release
+# https://github.com/Azure/azure-storage-azcopy/releases/tag/v10.30.1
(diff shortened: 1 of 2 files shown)


erik-bershel:
Oh, I see we're on the same page. 😄 I'm glad you're paying attention to the project and wanting to make it better—there's nothing more satisfying for a maintainer.

There's a problem. We don't accept contributions to macOS for technical reasons. 😞 There are two options: 
- First, you leave only the changes for Ubuntu, and team will review it separately. 
- Second, I'll simply add you as a co-author here: [link](https://github.com/actions/runner-images/compare/main...github-maccloud:runner-images:upd-azcopy-direct-release-links).

What would you prefer? 🤔

PrashantRaj18198:
@erik-bershel Let do co-author then, it would be best if we could add the changes for macos as well.

erik-bershel:
Closing in favor: https://github.com/actions/runner-images/pull/13297

PR #13355, opened 2025-11-26
ghcup: use wget as default downloader

andosius:
# Description
I had a hard time creating custom images with the default downloader for haskell versions (curl).
Changing it to wget allows me to create templates for proxmox without build failures.

curl ended up returning error code 56 during the download of the first or second archive and broke the image generation process.

## Check list

^ Atleast in my environment on Ubuntu 24.04.


GitPaulo:
Thank you for your contribution. This will be closed due to: https://github.com/actions/runner-images/issues/13353#issuecomment-3582633580

PR #13368, opened 2025-11-27
Add provisioner shell to list installed apt packages

GitPaulo:
# Description

This PR adds a new step to the Ubuntu 22.04 and 24.04 image build pipelines to output a sorted list of installed apt packages and their versions.

## Check list

### DIFF
--- a/images/ubuntu/templates/build.ubuntu-22_04.pkr.hcl
+++ b/images/ubuntu/templates/build.ubuntu-22_04.pkr.hcl
@@ -194,7 +194,7 @@ build {
 
   provisioner "shell" {
     execute_command = "sudo sh -c '{{ .Vars }} {{ .Path }}'"
-    inline          = ["dpkg-query -W -f='$${Package} $${Version}\\n' | sort"]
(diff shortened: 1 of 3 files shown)


images/ubuntu/templates/build.ubuntu-22_04.pkr.hcl
```diff
@@ -192,6 +192,11 @@ build {
     scripts          = ["${path.root}/../scripts/build/configure-snap.sh"]
   }
 
+  provisioner "shell" {
+    execute_command = "sudo sh -c '{{ .Vars }} {{ .Path }}'"
+    inline          = ["dpkg-query -W -f='$${Package} $${Version}\\n' | sort"]
```
maxim-lobanov: a couple of suggestions:
1. I don't think that we need double `$` in line 🤔 
2. Consider creating a small script like `list-apt-packages.sh`
3. Add some lines before printing packages list. Packer logs are pretty long and it can be difficult to find list of packages in output. May be something like `echo "Print the list of all installed apt packages"`

PR #13380, opened 2025-12-03
[macOS] Delete DOTNET_ROOT environment variable

erik-bershel:
# Description

- .NET is installed in the default location, meaning the latest installed version is automatically available. Anyone wishing to use a different version in a different instance must set the variable manually or use a suitable automation tool ([setup-dotnet](https://github.com/actions/setup-dotnet) for GitHub Actions, [UseDotNet@2](https://learn.microsoft.com/en-us/azure/devops/pipelines/tasks/reference/use-dotnet-v2?view=azure-pipelines) for Azure DevOps).
- Setting DOTNET_SKIP_FIRST_TIME_EXPERIENCE=1 and DOTNET_NOLOGO=1

#### Related issue:

https://github.com/actions/runner-images/issues/13365
https://github.com/actions/runner-images/issues/13470

## Check list

### DIFF
--- a/images/macos/assets/bashrc
+++ b/images/macos/assets/bashrc
@@ -22,8 +22,9 @@ export PATH="/usr/local/opt/curl/bin:$PATH"
 export PATH=$HOME/.cargo/bin:$PATH
 
 export RCT_NO_LAUNCH_PACKAGER=1
-export DOTNET_ROOT=$HOME/.dotnet
 export DOTNET_MULTILEVEL_LOOKUP=0
+export DOTNET_SKIP_FIRST_TIME_EXPERIENCE=1
+export DOTNET_NOLOGO=1
 
 export HOMEBREW_NO_AUTO_UPDATE=1
 export HOMEBREW_NO_INSTALL_CLEANUP=1


nagilson:
I think this is a great change and moves towards the correct behavior.

With that said, I believe this is a breaking change. Does the runner-images team have an A-B, or safer rollout mechanism which requires an image upgrade? 

Basically, before the bashrc file force overrode DOTNET_ROOT to the `$HOME/.dotnet`. That is indeed the default location and what is normally used, so the default experience will not be changed. Users do have to explicitly do or import something that overrides DOTNET_ROOT to a custom value, but engineers may have expected the prior behavior, so this may cause their processes to suddenly use an unexpected version of the dotnet.exe (DOTNET_ROOT).

This will certainly cause live failures in CI as many folks set DOTNET_ROOT. Again, this is the correct behavior, but we should shift to this gently.

erik-bershel:
> I think this is a great change and moves towards the correct behavior.
> 
> With that said, I believe this is a breaking change. Does the runner-images team have an A-B, or safer rollout mechanism which requires an image upgrade?
> 
> Basically, before the bashrc file force overrode DOTNET_ROOT to the `$HOME/.dotnet`. That is indeed the default location and what is normally used, so the default experience will not be changed. Users do have to explicitly do or import something that overrides DOTNET_ROOT to a custom value, but engineers may have expected the prior behavior, so this may cause their processes to suddenly use an unexpected version of the dotnet.exe (DOTNET_ROOT).
> 
> This will certainly cause live failures in CI as many folks set DOTNET_ROOT. Again, this is the correct behavior, but we should shift to this gently.

I think this change will be around for a while. It'll likely hit our images in the second half of January or somewhere around, after the announcement.

nagilson:
https://github.com/actions/runner-images/issues/13470 It looks like this is still being done - why was this PR closed, is there a new PR?

erik-bershel:
I'll reopen it later - my clean-up workflow deleted feature branch. 😄

images/macos/assets/bashrc
```diff
@@ -22,8 +22,9 @@ export PATH="/usr/local/opt/curl/bin:$PATH"
 export PATH=$HOME/.cargo/bin:$PATH
 
 export RCT_NO_LAUNCH_PACKAGER=1
-export DOTNET_ROOT=$HOME/.dotnet
 export DOTNET_MULTILEVEL_LOOKUP=0
+export DOTNET_SKIP_FIRST_TIME_EXPERIENCE=1
```
GerardSmit, line 26: Do we need to set `DOTNET_SKIP_FIRST_TIME_EXPERIENCE`?

This environment variable was removed in .NET Core 3.0.
The last supported version was .NET Core 2.2 which is EOL since December 23, 2019

Starting from .NET Core 3.0 the ENV `DOTNET_NOLOGO` should be used ([docs](https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-environment-variables#dotnet_skip_first_time_experience)) which is set the next line.

nagilson:
Bump on this - I'm sure @littlelittlecloud is interested as well.

images/macos/assets/bashrc
nagilson: DOTNET_NOLOGO is good at this point. There's no harm in the other setting though.

PR #13563, opened 2026-01-20
Remove Deprecated Workload Component.MDD.Linux.GCC.arm

patrix87:
# Description
Removes from windows Toolsets.
The workload is deprecated and no longer seems to be installable preventing both windows 2022 and 2025 from building correctly.
Deprecation Mention Source: https://learn.microsoft.com/en-us/visualstudio/install/workload-component-id-vs-community?view=visualstudio

## Check list
- [ YES ] Related issue / work item is attached
- [ N/A ] Tests are written (if applicable)
- [ N/A ] Documentation is updated (if applicable)
- [ YES ] Changes are tested and related VM images are successfully generated

### DIFF
--- a/images/windows/toolsets/toolset-2022.json
+++ b/images/windows/toolsets/toolset-2022.json
@@ -229,7 +229,6 @@
             "Microsoft.VisualStudio.Workload.Universal",
             "Microsoft.VisualStudio.Workload.VisualStudioExtension",
             "Component.MDD.Linux",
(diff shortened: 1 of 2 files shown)


patrix87:
I ran both build with and without this workload with no other changes to validate the issue and the fix, but I cannot share the logs. I'm confident this resolves the issue as it did in our pipelines.

PR #13633, opened 2026-02-05
[ubuntu] Update docker to 29.1.5, compose to 2.40.3

shamil-mubarakshin:
# Description
This PR updates `docker` and `docker compose` on ubuntu images:
- docker is updated to `29.1.5`
- docker compose is updated to `2.40.3`

## Check list

### DIFF
--- a/images/ubuntu/toolsets/toolset-2204.json
+++ b/images/ubuntu/toolsets/toolset-2204.json
@@ -222,11 +222,11 @@
             },
             {
                 "package": "docker-ce-cli",
-                "version": "28.0.4"
+                "version": "29.1.5"
             },
             {
                 "package": "docker-ce",
-                "version": "28.0.4"
+                "version": "29.1.5"
             }
         ],
         "plugins": [
@@ -237,7 +237,7 @@
             },
             {
                 "plugin": "compose",
-                "version": "2.38.2",
+                "version": "2.40.3",
                 "asset": "linux-x86_64"
             }
         ]
--- a/images/ubuntu/toolsets/toolset-2404.json
+++ b/images/ubuntu/toolsets/toolset-2404.json
@@ -199,11 +199,11 @@
             },
             {
                 "package": "docker-ce-cli",
-                "version": "28.0.4"
+                "version": "29.1.5"
             },
             {
                 "package": "docker-ce",
-                "version": "28.0.4"
+                "version": "29.1.5"
             }
         ],
         "plugins": [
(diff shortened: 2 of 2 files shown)


shepmaster:
This is a bit of a stab in the dark, but one of my jobs (using 29.1.5) now [fails with a pretty strange error](https://github.com/rust-lang/rust-playground/actions/runs/21973205981/job/63479071611):

```
image with reference shepmaster/rust-stable:latest was found but does not provide any platform
```

The same job [from the day before](https://github.com/rust-lang/rust-playground/actions/runs/21932097529/job/63394489999) (with 28.0.4) seems fine.

---

The TL;DR of the job is that I build a docker image, push it to GHCR, then pull it back, re-tag it, and push it a second time to GHCR and Docker Hub. The error message suggests that somehow I've built an image that isn't `linux/amd64`, but if I pull the same image to my laptop I can run it as `linux/amd64`.

Do you have any gut feeling if this Docker update might not play well with GHCR or something?

images/ubuntu/toolsets/toolset-2204.json
```diff
@@ -222,11 +222,11 @@
             },
             {
                 "package": "docker-ce-cli",
-                "version": "28.0.4"
+                "version": "29.1.5"
```
MadLittleMods, line 225: Heads-up, Docker Engine 29.0.0 uses `containerd` as the default image store now.

> `containerd` image store is now the default for **fresh installs**. This doesn't apply to daemons configured with `userns-remap` (see [moby#47377](https://github.com/moby/moby/issues/47377)).
>
> *-- 29.0.0 (2025-11-10), https://docs.docker.com/engine/release-notes/29/#2900**

This appears to cause behavioral changes with how local vs remote images are chosen. As experienced with our CI breaking, https://github.com/element-hq/synapse/pull/19460#discussion_r2818080879

To add onto `containerd` being the problem, previously another developer was experiencing the same problem locally and noticed `containerd` being the only difference with their setup (resolved after disabling), see https://github.com/element-hq/synapse/pull/18210#discussion_r2499646952

I'm guessing this is also what is contributing to the problems that @shepmaster is experiencing, https://github.com/actions/runner-images/pull/13633#issuecomment-3900315434

MadLittleMods: For people following along in this thread, this PR was reverted in https://github.com/actions/runner-images/pull/13708

PR #13708, opened 2026-02-20
Revert "[ubuntu] Update docker to 29.1.5, compose to 2.40.3 (#13633)"

shamil-mubarakshin:
# Description
This reverts commit c218bde724e0ae9730d9ce101bca654a4c2e521d.

## Check list

### DIFF
--- a/images/ubuntu/toolsets/toolset-2204.json
+++ b/images/ubuntu/toolsets/toolset-2204.json
@@ -222,11 +222,11 @@
             },
             {
                 "package": "docker-ce-cli",
-                "version": "29.1.5"
+                "version": "28.0.4"
             },
             {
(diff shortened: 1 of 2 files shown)


images/ubuntu/toolsets/toolset-2204.json
```diff
@@ -222,11 +222,11 @@
             },
             {
                 "package": "docker-ce-cli",
-                "version": "29.1.5"
+                "version": "28.0.4"
```
MadLittleMods, line 225: What's the context/reasoning for reverting?

It did cause problems as I pointed out in https://github.com/actions/runner-images/pull/13633#discussion_r2819281945 but I'm curious if there were other factors here.

PR #13777, opened 2026-03-09
[Ubuntu] Moved libicu installation from helper script to toolset

v-SergeiRadich:
# Description

## Issue
- `libicu` package is not listed among installed APT packages of Ubuntu 24.04 and has fixed version 70

## Solution
- Removed old approach with basically hardcoded `libicu` installation from helper script
- Added proper installation using toolset

#### Related issue:
- https://github.com/actions/runner-images/issues/13772

## Check list


v-SergeiRadich:
'libicu' is being installed as a dependency, no need to add it to toolset.

PR #13782, opened 2026-03-11
[Windows] Fix 406 when installing java tools

gavintodd:
# Description
Fix error during java tool installation on Windows images
 
>Invoke-RestMethod : The remote server returned an error: (406) Not Acceptable

Added accept header when calling to Adoptium api in `images/windows/scripts/build/Install-JavaTools.ps1`. When I was attempting to create Windows images it was failing during java install with 406 from Adoptium.

- Fixes actions/runner-images#13781

## Check list

### DIFF
--- a/images/windows/scripts/build/Install-JavaTools.ps1
+++ b/images/windows/scripts/build/Install-JavaTools.ps1
@@ -53,7 +53,7 @@ function Install-JavaJDK {
     )
 
     # Get Java version from api
-    $assetUrl = Invoke-RestMethod -Uri "https://api.adoptium.net/v3/assets/latest/${JDKVersion}/hotspot"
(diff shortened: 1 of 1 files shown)


skalpin:
I came to make a similar PR, but I don't think this is necessary anymore. Yesterday the api was throwing a 406 error on version 11 without the ?vendor=eclipse at the end e.g. https://api.adoptium.net/v3/assets/latest/11/hotspot?vendor=eclipse

Today the api is returning correctly _without_ the vendor variable:
You can try it yourself here: https://api.adoptium.net/q/swagger-ui/#/Assets/getLatestAssets

Or just plop this into a browser: https://api.adoptium.net/v3/assets/latest/11/hotspot

gavintodd:
I was getting this error on various versions in the image generation process but locally was never able to reproduce it. From my understanding 406 is specifically content negotiation and the default that powershell uses is different then curl and that is why it wasn't manifesting on Ubuntu images. I attempted 4 image generations and they failed on 11 or 17 every time. Once I added the Accept header it worked just fine.

shamil-mubarakshin (approved):

Issue #13750, opened 2026-03-02
Using BUILD_RG_NAME to not create RG


gavintodd:
### Description

My org has a permissions structure that won't work to create RGs at start of the process. I added `$env:BUILD_RG_NAME = "<rg_name>"` and then i started getting the error

"* Specify either a location to create the resource group in or an existing build_resource_group_name, but not both."

The issue i'm running into is the validate is called as

```powershell
    Write-Host "Validating packer template..."
    & $PackerBinary validate `
        "-only=$($PackerTemplate.BuildName).*" `
        "-var=client_id=fake" `
        "-var=client_secret=fake" `
        "-var=subscription_id=$($SubscriptionId)" `
        "-var=tenant_id=fake" `
        "-var=location=$($AzureLocation)" `
        "-var=image_os=$($PackerTemplate.ImageOS)" `
        "-var=managed_image_name=$($ManagedImageName)" `
        "-var=managed_image_resource_group_name=$($ResourceGroupName)" `
        "-var=install_password=$($InstallPassword)" `
        "-var=allowed_inbound_ip_addresses=$($AllowedInboundIpAddresses)" `
        "-var=azure_tags=$($TagsJson)" `
        $PackerTemplate.Path
```

But the `GenerateResourcesAndImage.ps1` has
```powershell
        [Parameter(Mandatory = $True)]
        [string] $AzureLocation,
```

How am i'm supposed to use a premade RG in the process?

### Platforms affected

- [ ] Azure DevOps
- [ ] GitHub Actions - Standard Runners
- [ ] GitHub Actions - Larger Runners

### Runner images affected

- [ ] Ubuntu 22.04
- [x] Ubuntu 24.04
- [ ] Ubuntu Slim
- [ ] macOS 14
- [ ] macOS 14 Arm64
- [ ] macOS 15
- [ ] macOS 15 Arm64
- [ ] macOS 26
- [ ] macOS 26 Arm64
- [ ] Windows Server 2022
- [x] Windows Server 2025
- [ ] Windows Server 2025 with Visual Studio 2026

### Image version and build link

Main branch all versions

### Is it regression?

Not sure, org permissions structure used to allow me to create temp RG

### Expected behavior

Leverages existing RG

### Actual behavior

Throws error either

```
Cannot process command because of one or more missing mandatory
     | parameters: AzureLocation.
```

or if passed as ""

```
Cannot bind argument to parameter 'AzureLocation' because it is an empty
     | string.
```

### Repro steps

[...]

PR #13796, opened 2026-03-12
Change AzureLocation to optional in GenerateResourcesAndImage.ps1‎

v-GeorgyPuzakov:
Description

Fix GenerateResourcesAndImage to support using an existing build resource group via BUILD_RG_NAME without requiring AzureLocation.

Previously, AzureLocation was a mandatory PowerShell parameter, which caused failures when users intentionally provided only BUILD_RG_NAME (existing RG flow).
This change makes AzureLocation optional so the script no longer fails at parameter binding stage and allows the expected Packer validation/build behavior to handle valid/invalid combinations.

This fixes: [#13750](https://github.com/actions/runner-images/issues/13750)

### DIFF
--- a/images/macos/scripts/build/install-llvm.sh
+++ b/images/macos/scripts/build/install-llvm.sh
@@ -9,5 +9,7 @@ source ~/utils/utils.sh
 llvmVersion=$(get_toolset_value '.llvm.version')
 
 brew_smart_install "llvm@${llvmVersion}"
+# After brew update to 5.1.0 we have to manually unlink llvm to avoid conflicts with pre-installed llvm on macOS
+brew unlink "llvm@${llvmVersion}"
 
 invoke_tests "LLVM"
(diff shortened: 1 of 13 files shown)


helpers/GenerateResourcesAndImage.ps1
```diff
@@ -166,7 +167,7 @@ Function GenerateResourcesAndImage {
         [ImageType] $ImageType,
         [Parameter(Mandatory = $False)]
         [string] $ManagedImageName = "Runner-Image-$($ImageType)",
-        [Parameter(Mandatory = $True)]
+        [Parameter(Mandatory = $False)]
```
shamil-mubarakshin, line 169: This line change alone should be enough to support use case, when `BUILD_RG_NAME` env variable is set. Everything else looks like an unnecessary overcomplication of this helper.
Lets test this

PR #13993, opened 2026-05-04
[Ubuntu] Add xml maven latest version parser

v-sergei-pyshnoi:
# Description
This PR adds functionality get latest patch version for Maven with provided major.minor version from toolset

#### Related issue:
https://github.com/actions/runner-images/issues/13961

## Check list


v-AndriiKhyliuk (approved):

bowtie-ltsa:
@v-AndriiKhyliuk 

at this moment in time, trying to build your project locally, maven 3.9.16 is chosen, since it is latest, but it is only available on dlcdn.apache.org. It is not available (at this moment) from archive.apache.org, and the build fails.

image

worth filing an issue?

PR #14045, opened 2026-05-13
Add GitHub Copilot CLI to hosted runner images

salmanmkc:
Install `@github/copilot` into the **agent tool cache** during image bake for Ubuntu 24.04 and 26.04. The CLI is staged at `$AGENT_TOOLSDIRECTORY/copilot-cli/<version>/x64/` with the matching `x64.complete` marker file, so a runtime consumer using `@actions/tool-cache` can find it via `tc.find("copilot-cli", ...)` and expose it on PATH with `core.addPath`.

This enables agentic workflows that use the Copilot engine to skip the runtime `npm install` step on a cache hit, reducing workflow startup time. The toolcache version is read from `https://raw.githubusercontent.com/github/gh-aw-actions/main/.github/aw/compat.json` (the agentic-workflows compatibility matrix), specifically the `max-agent` value on the catch-all row.

### Changes

- `images/ubuntu/scripts/build/install-copilot-cli.sh` — fetch the toolcache version from `compat.json`, install via `npm install -g --prefix <toolcache>/x64/`, write `x64.complete` marker. Hard-fails the bake on fetch/parse failure (matches `install-awf.sh` semantics).
- `images/ubuntu/scripts/tests/Tools.Tests.ps1` — Pester test verifying the toolcache layout (`/opt/hostedtoolcache/copilot-cli/*/x64/bin/copilot`, `x64.complete` marker present). Mirrors the existing AWF test block.
- `images/ubuntu/templates/build.ubuntu-24_04.pkr.hcl` — invoke `install-copilot-cli.sh` in the build (after `install-nodejs.sh` so `npm` is available).
- `images/ubuntu/templates/build.ubuntu-26_04.pkr.hcl` — same.

### Why toolcache instead of system PATH

A runtime consumer needs to (a) pick a specific version that satisfies a semver range and (b) ensure the binary is on PATH for the agent step. The standard pattern for both is `@actions/tool-cache`: `tc.find("copilot-cli", range)` returns the cached path, `core.addPath(<path>/bin)` exposes it. That requires the on-disk layout to be `$RUNNER_TOOL_CACHE/<tool>/<version>/<arch>/` with a `.complete` marker — which is exactly what this script produces. AWF (`install-awf.sh` in this same repo) uses the same pattern.

### Backward compatibility

This change is **no-op** for workflows that don't use `@actions/tool-cache` to find the Copilot CLI. The bake step neither shadows nor removes anything from the global PATH (the previous global `npm install -g` location is replaced because nothing else needs the binary there). Workflows that today install Copilot CLI via `npm install -g` at runtime keep doing so unchanged.


images/ubuntu/templates/build.ubuntu-24_04.pkr.hcl
```diff
@@ -111,6 +111,7 @@ provisioner "shell" {
       "${path.root}/../scripts/build/install-cmake.sh",
       "${path.root}/../scripts/build/install-codeql-bundle.sh",
       "${path.root}/../scripts/build/install-awf.sh",
+      "${path.root}/../scripts/build/install-copilot-cli.sh",
       "${path.root}/../scripts/build/install-container-tools.sh",
       "${path.root}/../scripts/build/install-dotnetcore-sdk.sh",
```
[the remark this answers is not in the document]
salmanmkc: Fixed in 921ae04 \u2014 `install-copilot-cli.sh` now runs right after `install-nodejs.sh` in both Ubuntu 24.04 and 26.04 provisioner lists. While I was in there I also pinned the install to a known-good version pulled from the agentic workflows compatibility matrix (`raw.githubusercontent.com/github/gh-aw-actions/main/.github/aw/compat.json`), so the baked tool-cache version matches what the runtime setup action will resolve to. Falls back to `latest` on any fetch failure so an outage on `raw.githubusercontent.com` does not block the image bake.

images/ubuntu/templates/build.ubuntu-26_04.pkr.hcl
```diff
@@ -110,6 +110,7 @@ provisioner "shell" {
       "${path.root}/../scripts/build/install-cmake.sh",
       "${path.root}/../scripts/build/install-codeql-bundle.sh",
       "${path.root}/../scripts/build/install-awf.sh",
+      "${path.root}/../scripts/build/install-copilot-cli.sh",
       "${path.root}/../scripts/build/install-container-tools.sh",
```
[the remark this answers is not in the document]
salmanmkc: Same fix in 921ae04 \u2014 see the 24.04 reply above for the full rationale.

images/ubuntu/scripts/build/install-copilot-cli.sh
```diff
@@ -0,0 +1,66 @@
...
+# Install into the agent tool cache so the gh-aw-actions/setup action's
+# tc.find("copilot-cli", ...) lookup succeeds on hosted runners. Layout:
+#   $AGENT_TOOLSDIRECTORY/copilot-cli/<version>/x64/{bin,lib}
+#   $AGENT_TOOLSDIRECTORY/copilot-cli/<version>/x64.complete
+# Mirrors the conventions used by install-awf.sh and install-codeql-bundle.sh.
+copilot_toolcache_path="$AGENT_TOOLSDIRECTORY/copilot-cli/$copilot_version/x64"
+mkdir -p "$copilot_toolcache_path"
+
+echo "Installing GitHub Copilot CLI v$copilot_version into $copilot_toolcache_path (pinned via compat.json)..."
+npm install -g --prefix "$copilot_toolcache_path" \
```
maxim-lobanov: if it is possible to use GitHub releases for installing copilot CLI, it would be more prefered way: https://github.com/github/copilot-cli/releases/

images/ubuntu/scripts/build/install-copilot-cli.sh
```diff
@@ -0,0 +1,66 @@
...
+    echo "Error: compat.json catch-all max-agent ('$copilot_version') is not a valid SemVer version."
+    exit 1
+fi
+
+# Install into the agent tool cache so the gh-aw-actions/setup action's
+# tc.find("copilot-cli", ...) lookup succeeds on hosted runners. Layout:
+#   $AGENT_TOOLSDIRECTORY/copilot-cli/<version>/x64/{bin,lib}
+#   $AGENT_TOOLSDIRECTORY/copilot-cli/<version>/x64.complete
+# Mirrors the conventions used by install-awf.sh and install-codeql-bundle.sh.
+copilot_toolcache_path="$AGENT_TOOLSDIRECTORY/copilot-cli/$copilot_version/x64"
```
paveliak: > x64

Could you make this script arch-agnostic so it works for `arm64` too?

salmanmkc: addressed now

images/ubuntu/scripts/build/install-copilot-cli.sh
salmanmkc: addressed thanks

maxim-lobanov (dismissed):

v-AndriiKhyliuk (dismissed):

images/ubuntu/scripts/build/install-copilot-cli.sh
```diff
@@ -0,0 +1,84 @@
...
+##         compatibility matrix.
+##  Supply chain security: Copilot CLI - checksum validation
+################################################################################
+
+# Source the helpers for use with the script
+source $HELPER_SCRIPTS/install.sh
+
+# Detect host architecture to pick the matching release asset and toolcache
+# subdirectory.
+case "$(uname -m)" in
```
shamil-mubarakshin: Please utilize `is_arm64` and `is_x64` helpers from `os.sh`, e.g. [here](https://github.com/actions/runner-images/blob/main/images/ubuntu/scripts/build/install-kubernetes-tools.sh#L10)

salmanmkc: done

v-AndriiKhyliuk (approved):

PR #14437, opened 2026-07-22
[windows] Preinstall MSYS2 on Windows 11 ARM64 images

### LINKED ISSUE #14087: [windows-11-arm] Pre-install msys2
**Type of issue**
- [ ] Error when running a workflow.
- [x] Request for additional tools or software for an existing image.
- [ ] Request for a new image, Linux or Windows.
- [ ] Other

**Describe the issue**
MSYS2 is not installed in `C:\msys64` as it is on other Windows images

**To Reproduce**
try using `msys2/setup-msys2` action with `release: false` option.

**Expected behavior**
`C:\msys64` exists and contains MSYS2

**Additional context**
I thought about using `msys2/setup-msys2` with `release: true` and `location: C:\`, but the action checks for and errors out in that configuration because
`'location' cannot be 'C:' because that contains the built-in MSYS2 installation
      in GitHub Actions environments. See option 'release', in order to use that installation:
      https://github.com/msys2/setup-msys2#release`

v-davit-ioramashvili:
# Description

Preinstall MSYS2 on both Windows 11 ARM64 images, so `C:\msys64` exists as it does on the x64 images.

**Size: 0.26 GB. Installation time: 1-8 min** (dominated by unpacking the MSYS2 base filesystem; varies with runner disk I/O - measured 72 s and 462 s on two hosted ARM runners, with identical results).

No mingw packages are bundled — `MsysPackages` stays empty, matching every other Windows toolset since Windows 2022. Anyone who needs a toolchain installs it with `pacman` (~38 s) and picks their own package set.

#### Related issue:
Closes #14087

## Changes
- Added `Install-Msys2.ps1` to both ARM64 templates; it now picks the native arm64 installer, since the x86_64 one misbehaves there (msys2/msys2-installer#96)
- `Configure-Shell.ps1` creates `msys2bash.cmd` on ARM64, and fixes the x64 `MSYSTEM` value to `MINGW64` (thanks @mmuetzel)
- Dropped the ARM64 skips in `MSYS2.Tests.ps1` and the `Test-IsX64` gates on the Shells/MSYS2 report sections, so the ARM64 readme documents MSYS2 as the Windows 2025 one does

As on the x64 images, MSYS2 is installed but not added to PATH. The generated readme carries the same note.

## Testing
Ran the real install scripts on a hosted `windows-11-arm` runner:
- installs to 0.26 GB; MSYS2 Pester tests pass
- `msys2/setup-msys2@v2` with `release: false` works against the preinstalled MSYS2 — the action installs its own packages on top, and a C file compiles
- `msys2bash.cmd` opens a working shell; the post-generation first-launch script creates the user profile
- `Get-ShellTarget` and `Get-PacmanVersion` run on ARM64 (Pacman 6.1.0)
- machine PATH unchanged — no `msys64` entry
- MSYSTEM casing checked on windows-2025: lowercase `mingw64` leaves `PATH[0]` at `/usr/local/bin`, `MINGW64` correctly prepends `/mingw64/bin`

Validation runs (public fork, all green):
- [PR validation, both architectures](https://github.com/v-davit-ioramashvili/runner-images/actions/runs/32847713880)
  - **windows-11-arm**: install, Pester tests, CLANGARM64 wrapper, post-gen first launch, software report, machine PATH, and `setup-msys2` with `release: false` reusing the preinstalled MSYS2
  - **windows-2025**: confirms x64 is unaffected - the wrapper still selects `MINGW64` and resolves `MSYSTEM_PREFIX=/mingw64` with `/mingw64/bin` first on PATH

## Check list

### DIFF
--- a/images/windows/scripts/build/Configure-Shell.ps1
+++ b/images/windows/scripts/build/Configure-Shell.ps1
@@ -5,7 +5,7 @@ New-Item -Path $shellPath -ItemType Directory | Out-Null
 if ((Test-IsX64) -or (Test-IsArm64)) {
     # add a wrapper for C:\msys64\usr\bin\bash.exe
     # arm64 images carry the CLANGARM64 prefix, x64 images carry MINGW64
-    $msystem = if (Test-IsArm64) { "CLANGARM64" } else { "mingw64" }
+    $msystem = if (Test-IsArm64) { "CLANGARM64" } else { "MINGW64" }
 @"
 @echo off
 setlocal
--- a/images/windows/scripts/build/Configure-Shell.ps1
+++ b/images/windows/scripts/build/Configure-Shell.ps1
@@ -2,16 +2,18 @@
 $shellPath = "C:\shells"
 New-Item -Path $shellPath -ItemType Directory | Out-Null
 
-if (Test-IsX64) {
+if ((Test-IsX64) -or (Test-IsArm64)) {
     # add a wrapper for C:\msys64\usr\bin\bash.exe
-@'
+    # arm64 images carry the CLANGARM64 prefix, x64 images carry MINGW64
+    $msystem = if (Test-IsArm64) { "CLANGARM64" } else { "mingw64" }
+@"
 @echo off
 setlocal
 IF NOT DEFINED MSYS2_PATH_TYPE set MSYS2_PATH_TYPE=strict
-IF NOT DEFINED MSYSTEM set MSYSTEM=mingw64
+IF NOT DEFINED MSYSTEM set MSYSTEM=$msystem
 set CHERE_INVOKING=1
 C:\msys64\usr\bin\bash.exe -leo pipefail %*
-'@ | Out-File -FilePath "$shellPath\msys2bash.cmd" -Encoding ascii
+"@ | Out-File -FilePath "$shellPath\msys2bash.cmd" -Encoding ascii
 }
 
 # gitbash <--> C:\Program Files\Git\bin\bash.exe
--- a/images/windows/scripts/build/Configure-Shell.ps1
+++ b/images/windows/scripts/build/Configure-Shell.ps1
@@ -2,10 +2,9 @@
 $shellPath = "C:\shells"
 New-Item -Path $shellPath -ItemType Directory | Out-Null
 
-if ((Test-IsX64) -or (Test-IsArm64)) {
-    # add a wrapper for C:\msys64\usr\bin\bash.exe
-    # arm64 images carry the CLANGARM64 prefix, x64 images carry MINGW64
-    $msystem = if (Test-IsArm64) { "CLANGARM64" } else { "MINGW64" }
+# add a wrapper for C:\msys64\usr\bin\bash.exe
+# arm64 images use the CLANGARM64 environment, x64 images MINGW64
+$msystem = if (Test-IsArm64) { "CLANGARM64" } else { "MINGW64" }
 @"
 @echo off
 setlocal
@@ -14,7 +13,6 @@ IF NOT DEFINED MSYSTEM set MSYSTEM=$msystem
 set CHERE_INVOKING=1
 C:\msys64\usr\bin\bash.exe -leo pipefail %*
 "@ | Out-File -FilePath "$shellPath\msys2bash.cmd" -Encoding ascii
-}
 
 # gitbash <--> C:\Program Files\Git\bin\bash.exe
 New-Item -ItemType SymbolicLink -Path "$shellPath\gitbash.exe" -Target "$env:ProgramFiles\Git\bin\bash.exe" | Out-Null
--- a/images/windows/scripts/build/Install-Msys2.ps1
+++ b/images/windows/scripts/build/Install-Msys2.ps1
@@ -14,11 +14,16 @@ function Install-Msys2 {
     # We can't use Resolve-GithubReleaseAssetUrl function here
     # because msys2-installer releases don't have a consistent versioning scheme
 
+    # The msys2 runtime is x86_64-only and runs under emulation on arm64, but the
+    # installer itself has a native arm64 build - the x86_64 one misbehaves there.
+    # See https://github.com/msys2/msys2-installer/issues/96
+    $installerPrefix = if (Test-IsArm64) { "^msys2-arm64" } else { "^msys2-x86_64" }
+
     $assets = (Invoke-RestMethod -Uri "https://api.github.com/repos/msys2/msys2-installer/releases/latest").assets
-    $downloadUri = ($assets | Where-Object { $_.name -match "^msys2-x86_64" -and $_.name.EndsWith(".exe") }).browser_download_url
+    $downloadUri = ($assets | Where-Object { $_.name -match $installerPrefix -and $_.name.EndsWith(".exe") }).browser_download_url
     $installerName = Split-Path $downloadUri -Leaf
-  
-    # Download the latest msys2 x86_64, filename includes release date
+
+    # Download the latest msys2 installer, filename includes release date
     Write-Host "Download msys2 installer $installerName"
     $installerPath = Invoke-DownloadWithRetry $downloadUri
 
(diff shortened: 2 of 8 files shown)


images/windows/toolsets/toolset-win-11-arm64.json
```diff
@@ -99,7 +99,48 @@
...
+                            "cpack",
+                            "ctest"
+                        ]
+                    },
+                    {
+                        "name": "make",
+                        "executables": [
+                            "mingw32-make"
+                        ]
+                    },
```
mmuetzel, line 133: For a pre-installed `make` it would probably be more useful to install this package instead:
https://packages.msys2.org/packages/make?variant=x86_64 (i.e., the "Cygwin" `make`)

mmuetzel: The main difference is probably that `make` uses a POSIX shell and `mingw32-make` uses `cmd.exe`.
Because most `Makefile`s are written with Linux first in mind, many of the more complex `Makefile`s assume a working shell with the corresponding intrinsics. Attempting to use `mingw32-make` with such a `Makefile` will likely fail.

Very few projects write `Makefile`s for Windows `cmd.exe`. So the applications where one would want to use `mingw32-make` are limited.

For very simple Makefiles, either one would probably work.

mmuetzel: It might be best to install the [`base-devel`](https://packages.msys2.org/packages/base-devel?variant=x86_64) group instead (which contains the "correct" version of `make`).

images/windows/scripts/build/Configure-Shell.ps1
```diff
@@ -2,16 +2,18 @@
 $shellPath = "C:\shells"
 New-Item -Path $shellPath -ItemType Directory | Out-Null
 
-if (Test-IsX64) {
+if ((Test-IsX64) -or (Test-IsArm64)) {
     # add a wrapper for C:\msys64\usr\bin\bash.exe
-@'
+    # arm64 images carry the CLANGARM64 prefix, x64 images carry MINGW64
+    $msystem = if (Test-IsArm64) { "CLANGARM64" } else { "mingw64" }
```
mmuetzel: This might have been a pre-existing issue. But the correct spelling for the value of the `MSYSTEM` environment variable is `MINGW64`.
Also, the UCRT64 environment is recommended over the MINGW64 environment by now.

```suggestion
    $msystem = if (Test-IsArm64) { "CLANGARM64" } else { "UCRT64" }
```

mmuetzel: Maybe, switching from MINGW64 to UCRT64 deserves a separate PR. (Doing that would probably require a bunch of other changes at the same time.)

I'd still recommend the following:

```suggestion
    $msystem = if (Test-IsArm64) { "CLANGARM64" } else { "MINGW64" }
```

images/windows/scripts/build/Configure-Shell.ps1
```diff
@@ -2,16 +2,18 @@
 $shellPath = "C:\shells"
 New-Item -Path $shellPath -ItemType Directory | Out-Null
 
-if (Test-IsX64) {
+if ((Test-IsX64) -or (Test-IsArm64)) {
     # add a wrapper for C:\msys64\usr\bin\bash.exe
-@'
+    # arm64 images carry the CLANGARM64 prefix, x64 images carry MINGW64
+    $msystem = if (Test-IsArm64) { "CLANGARM64" } else { "MINGW64" }
```
kmilos, line 8: Note that UCRT64 has been the default installed MSYSTEM on x64 for a while now and that MINGW64 is deprecated:

https://www.msys2.org/news/#2022-10-29-changing-the-default-environment-from-mingw64-to-ucrt64
https://www.msys2.org/news/#2026-03-15-deprecating-the-mingw64-environment

images/windows/scripts/build/Install-Msys2.ps1
```diff
@@ -103,7 +108,8 @@ function Install-MingwPackages {
 Install-Msys2
 
 # Add msys2 bin tools folders to PATH temporary
-$env:PATH = "C:\msys64\mingw64\bin;C:\msys64\usr\bin;$origPath"
+$msysExecDir = if (Test-IsArm64) { "clangarm64" } else { "mingw64" }
```
kmilos, line 111: Ditto

images/windows/scripts/build/Configure-Shell.ps1
mmuetzel: In my humble opinion, it would be nice to switch to UCRT64. But that should be done in an independent pull request.
This here is about adding MSYS2 to the Windows on ARM images. (It's not about the Intel/AMD Windows images.)

kmilos: > This here is about adding MSYS2 to the Windows on ARM images. (It's not about the Intel/AMD Windows images.)

True, but at the same time it would create an immediate difference in behaviour between arm64 and x64 images - the CLANGARM64 environment produces binaries linked against UCRT, while legacy MINGW64 still uses MSVCRT.

Not that a more striking difference already doesn't exists (no MSYS2 😉), so a two-step approach is fine by me as well.

v-davit-ioramashvili: The UCRT64 switch is prepared as a follow-up — I'll open it once this merges. Verified on windows-2025 that the wrapper then resolves MSYSTEM_PREFIX=/ucrt64 with /ucrt64/bin first on PATH.

images/windows/toolsets/toolset-win-11-arm64.json
v-davit-ioramashvili: Thanks — I gave this a try on a `windows-11-arm` runner. The package itself is fine: `pacman -S make` installs cleanly at 1.62 MiB and gives GNU Make 4.4.1 in `/usr/bin`.

The blocker is elsewhere. Populating `MsysPackages.msys2` makes `Install-Msys2Packages` run, and that function unconditionally does `pacman -R p7zip` and throws if it fails:

```
--- Install msys2 packages
Total Installed Size:  1.62 MiB
installing make...
--- Remove p7zip/7z package due to conflicts
error: target not found: p7zip
```

p7zip isn't part of the msys2 base install any more, so the build dies right after the package installs. That array has been empty on every toolset since Windows 2019 was removed in January, so this went unnoticed.

So adding `make` needs that removal made conditional first. I'd rather fix it as its own change than fold it into this PR.

mmuetzel: Oof. I didn't expect that this would cause so much trouble.
Alternatively, you could probably just drop installing any `make` package (the `mingw32-make` is pretty useless imho). Then users could install it with `pacman -S make` (or with the [msys2/setup-msys2](https://github.com/msys2/setup-msys2) action).

v-SergeiRadich (changes requested):
Please attach link to test runs.

images/windows/scripts/build/Configure-Shell.ps1
```diff
@@ -2,16 +2,18 @@
 $shellPath = "C:\shells"
 New-Item -Path $shellPath -ItemType Directory | Out-Null
 
-if (Test-IsX64) {
+if ((Test-IsX64) -or (Test-IsArm64)) {
```
v-SergeiRadich: Please refine this code - this condition is going to be true every time, because we have only ARM and x64 images.

v-SergeiRadich (changes requested):

v-SergeiRadich (approved):

images/windows/toolsets/toolset-win-11-arm64.json
Ashe14: > Oof. I didn't expect that this would cause so much trouble. Alternatively, you could probably just drop installing any `make` package (the `mingw32-make` is pretty useless imho). Then users could install it with `pacman -S make` (or with the [msys2/setup-msys2](https://github.com/msys2/setup-msys2) action).

It fixed it only install MSYS2.

v-AndriiKhyliuk (approved):

Peter0x44:
@skeeto

images/windows/toolsets/toolset-win-11-arm64.json
MortimerSC: Fixes #14567

Issue #14478, opened 2026-07-30
[UBUNTU 24.04] 20260726.254.1 sudo timedatectl set-timezone fails


yorkmoy:
### Description

We are seeing sudo timedatectl set-timezone fail with Access denied message in the 20260726.254.1 runner image that has started to rollout.

### Platforms affected

- [ ] Azure DevOps
- [x] GitHub Actions - Standard Runners
- [ ] GitHub Actions - Larger Runners

### Runner images affected

- [ ] Ubuntu 22.04
- [ ] Ubuntu 22.04 Arm64
- [x] Ubuntu 24.04
- [ ] Ubuntu 24.04 Arm64
- [ ] Ubuntu 26.04
- [ ] Ubuntu 26.04 Arm64
- [ ] Ubuntu Slim
- [ ] macOS 14
- [ ] macOS 14 Arm64
- [ ] macOS 15
- [ ] macOS 15 Arm64
- [ ] macOS 26
- [ ] macOS 26 Arm64
- [ ] Windows Server 2022
- [ ] Windows Server 2025
- [ ] Windows Server 2025 with Visual Studio 2026
- [ ] Windows Desktop 11
- [ ] Windows Desktop 11 with Visual Studio 2026

### Image version and build link

Image: ubuntu-24.04
Version: 20260726.254.1
https://github.com/actions/runner-images/releases/tag/ubuntu24%2F20260726.254

### Is it regression?

  Yes. The command is working in Version: 20260720.247.2   Image Release: https://github.com/actions/runner-images/releases/tag/ubuntu24%2F20260720.247

### Expected behavior

Run sudo timedatectl set-timezone America/New_York
  sudo timedatectl set-timezone America/New_York
  shell: /usr/bin/bash -e {0}
  env:
    TEST_BRANCH: 
    SPEC: 
    NODE_VERSION: 24
    AWS_ACCESS_KEY_ID: ***
    AWS_SECRET_ACCESS_KEY: ***
    AWS_REGION: us-east-1
    TOTAL_INSTANCES: 5

### Actual behavior

  sudo timedatectl set-timezone America/New_York
  shell: /usr/bin/bash -e {0}
  env:
    TEST_BRANCH: 
    SPEC: 
    NODE_VERSION: 24
    AWS_ACCESS_KEY_ID: ***
    AWS_SECRET_ACCESS_KEY: ***
    AWS_REGION: us-east-1
    TOTAL_INSTANCES: 5
[...]

Issue #14477, opened 2026-07-30
[Ubuntu] Owner of `/etc` and `/usr` changed from `root` to `runner` in 20260726


ajaust:
### Description

We observed that in the most recent runner image, the directories `/usr`, `/etc` and some of their subdirectories are owned by `runner` instead of `root`. 

I think that the ownership changes due to [this commit](https://github.com/actions/runner-images/commit/c299d4dbfdca73bbd022988d059a947234b70d23) and more explicitly [these lines in `install-container-tools.sh`](https://github.com/actions/runner-images/blob/4055b52178bcc904dc9d059b0c01d247188aa967/images/ubuntu/scripts/build/install-container-tools.sh#L38-L39) that unpack an archive directly to `/`.

### Platforms affected

- [ ] Azure DevOps
- [x] GitHub Actions - Standard Runners
- [ ] GitHub Actions - Larger Runners

### Runner images affected

- [x] Ubuntu 22.04
- [ ] Ubuntu 22.04 Arm64
- [x] Ubuntu 24.04
- [ ] Ubuntu 24.04 Arm64
- [ ] Ubuntu 26.04
- [ ] Ubuntu 26.04 Arm64
- [ ] Ubuntu Slim
- [ ] macOS 14
- [ ] macOS 14 Arm64
- [ ] macOS 15
- [ ] macOS 15 Arm64
- [ ] macOS 26
- [ ] macOS 26 Arm64
- [ ] Windows Server 2022
- [ ] Windows Server 2025
- [ ] Windows Server 2025 with Visual Studio 2026
- [ ] Windows Desktop 11
- [ ] Windows Desktop 11 with Visual Studio 2026

### Image version and build link

- Ubuntu 22.04: 20260726.241.1
- Ubuntu 24.04: 20260726.254.1

We did not test the whether the ARM runners are also affected.

Below are links to workflow runs that hit the new runner version and therefore `/etc` and `/usr` are owned by `runner`.

#### `/usr`

- Ubuntu 22.04: https://github.com/ajaust/ubuntu-runner-usr-owner/actions/runs/30533812613/job/90842146609
- Ubuntu 24.04: https://github.com/ajaust/ubuntu-runner-usr-owner/actions/runs/30532992734/job/90839475069

#### `/etc`

- Ubuntu 22.04: https://github.com/ajaust/ubuntu-runner-usr-owner/actions/runs/30541949529/job/90868708617
[...]

Issue #14473, opened 2026-07-29
[Ubuntu 24.04] podman broken on 20260726.254: uses /usr/bin/crun 1.14.1 instead of the crun 1.28 in /usr/local/bin


shyney7:
### Description

Every `podman run` and `podman build` fails on the ubuntu-24.04 image with:

```
Error: OCI runtime error: crun: unknown version specified
```

The image has podman 5.8.4 and crun 1.28, but podman is not using that crun. It resolves the runtime to `/usr/bin/crun`, which is still the Ubuntu archive build at 1.14.1:

```
$ podman --version
podman version 5.8.4
$ podman info --format '{{.Host.OCIRuntime.Path}}'
/usr/bin/crun
$ /usr/bin/crun --version
crun version 1.14.1
$ /usr/local/bin/crun --version
crun version 1.28
$ command -v crun
/usr/local/bin/crun
```

podman 5.8.4 writes `"ociVersion":"1.2.1"` into config.json. crun 1.14.1 rejects any version string that does not contain "1.0" (src/libcrun/container.c at the 1.14.1 tag):

```c
if (def->ociVersion && strstr (def->ociVersion, "1.0") == NULL)
    return crun_make_error (err, 0, "unknown version specified");
```

so nothing starts. crun dropped that check in 1.16.

One thing that makes this annoying to debug: `crun --version` reports 1.28, because /usr/local/bin comes first on PATH. It looks like a current crun is installed while podman is actually running the old one.

Installing crun 1.28 over /usr/bin/crun fixes it. I confirmed that on a runner that had the broken image.

### Platforms affected

- [x] GitHub Actions - Standard Runners

### Runner images affected

- [x] Ubuntu 24.04

### Image version and build link

Version: 20260726.254.1

https://github.com/actions/runner-images/releases/tag/ubuntu24%2F20260726.254

### Is it regression?

Yes. Image 20260720.247.2 works. It has podman 4.9.3, which writes ociVersion 1.0.x, and crun 1.14.1 accepts that. The podman 4.9.3 to 5.8.4 bump is the only container change listed in the release notes for 20260726.254.

Because the rollout is staged, jobs currently pass or fail depending on which image they get. The same commit can pass and fail minutes apart, which is how this first showed up for me.

ubuntu-26.04 is not affected. It has podman 5.7.0 with crun 1.21 at /usr/bin/crun.

### Expected behavior

podman runs containers on the ubuntu-24.04 image.

### Actual behavior

[...]

PR #14412, opened 2026-07-17
[Ubuntu] Update `podman` (5.8.4) to the Ubuntu 22.04 and 24.04 images.

v-GeorgyPuzakov:
# Description

**Improvement**

Bring a current `podman` (5.8.4) to the Ubuntu 22.04 and 24.04 images.

Both images install `podman` from `apt`, which only offers outdated versions — **3.4.4** on 22.04 and **4.9.3** on 24.04 — whose Docker-compatible API is too old for the current Docker daemon. This breaks `podman push docker-daemon:`:

```
client version 1.41 is too old. Minimum supported API version is 1.44
```

(reported in actions/runner-images#13691). Ubuntu 26.04 already ships `podman` 5.x via `apt` and is unaffected. Ubuntu provides no maintained newer-`podman` apt repository (the Kubic/OBS `devel:kubic:libcontainers` project was discontinued), so a newer version cannot be sourced through `apt` on 22.04/24.04.

**Change** — `images/ubuntu/scripts/build/install-container-tools.sh`:

- **22.04 & 24.04:** install a pinned, checksum-verified static `podman` bundle ([`mgoltzsche/podman-static` v5.8.4](https://github.com/mgoltzsche/podman-static)), which is self-contained (podman + crun/runc + conmon + netavark + aardvark-dns + pasta + fuse-overlayfs + catatonit). `buildah`/`skopeo` continue to come from `apt` (neither depends on `podman`); `uidmap` is added for rootless support.
- **26.04:** unchanged — keeps `apt` `podman` 5.x.
- **Supply chain:** the bundle version is pinned and both `amd64`/`arm64` archives are verified against a pinned SHA-256 before extraction.
- **AppArmor (24.04+):** Ubuntu ≥ 23.10 enables `kernel.apparmor_restrict_unprivileged_userns=1`. The distro `podman` package ships an `/etc/apparmor.d/podman` profile that grants the `userns` permission, but the static binary in `/usr/local/bin` has none, so rootless `podman` fails with `failed to reexec: Permission denied`. An equivalent profile covering `/usr/{bin,local/bin}/podman` is installed when the restriction is enabled.
- **Networking:** the `netavark` `firewall_driver = "iptables"` workaround (previously 26.04-only, actions/runner-images#14230) now applies to all Ubuntu versions, since `netavark` is podman 5's default backend. The obsolete 22.04-only `containernetworking-plugins` (CNI) step is removed.

#### Related issue:
- Tracking: github/hosted-runners-images#549
[...]

PR #14480, opened 2026-07-30
[Ubuntu] Fix podman crun selection and /usr, /etc ownership broken by the podman bundle install

v-dunjadenic:
# Description
This is a bug fix. It covers two regressions that come from the same change  (the podman static bundle install added in #14412).

**Related issues:** [#14473](https://github.com/actions/runner-images/issues/14473), [#14477](https://github.com/actions/runner-images/issues/14477), [#14478](https://github.com/actions/runner-images/issues/14478)

## 1. podman uses the wrong crun (#14473)

The image ends up with two cruns: the current one from the bundle in `/usr/local/bin`, and the old distro one in `/usr/bin` that `buildah` depends on. Podman does not use `PATH` to find its runtime, it checks a fixed list of paths where `/usr/bin` comes first. Because of this it picks the old one (1.14.1 on 24.04, 0.17 on 22.04). That crun rejects the config podman 5.x writes, so every `podman run` and `podman build` fails:

```
Error: OCI runtime error: crun: unknown version specified
```

**Fix:** write a `containers.conf.d` drop-in telling podman which crun to use. 

## 2. /usr and /etc become owned by `runner` (#14477)

The bundle archive is built on a GitHub Actions runner, so every entry inside is recorded as owned by uid 1001 (`runner`), including the `usr` and `etc` directory entries themselves. The script unpacks it as root straight into `/`, and tar restores that stored ownership, handing `/usr` and `/etc` to an unprivileged user.

**Fix:** extract with `--no-same-owner --no-overwrite-dir`, so files land as `root` and existing system directories are left untouched.

Ubuntu 26.04 is unaffected by both: it installs podman from apt with crun 1.21 and never extracts the bundle. Note that 22.04 is affected as well, not only 24.04 as issue #14473 suggests. No breaking changes.

## 3. `sudo timedatectl set-timezone` fails (#14478)

Same cause as #2, so fix 2 covers it and no extra change is needed. Confirmed on a rolled-back (healthy) image by changing only the ownership and back.
- [Evidence run](https://github.com/v-dunjadenic/runner-images/actions/runs/30638756543)

```
/usr, /etc owned by root    -> set-timezone OK
/usr, /etc owned by runner  -> Failed to set time zone: Access denied
/usr, /etc owned by root    -> set-timezone OK
```
## Validation

- [Full run](https://github.com/v-dunjadenic/runner-images/actions/runs/30632890655)
- [Source test workflow with explanation](https://github.com/v-dunjadenic/runner-images/actions/runs/30632890655/workflow)

The workflow runs the real build script from this branch on hosted runners, in two modes per issue:

- `before-fix`: undoes the fix, then confirms the bug happens (green = reproduced)
- `after-fix`: runs the script as committed, then confirms everything works

Covered: `ubuntu-22.04`, `ubuntu-22.04-arm`, `ubuntu-24.04`, `ubuntu-24.04-arm`, `ubuntu-26.04` (control). The image tests, including the two new ones, run as part of the script in every leg.

> The affected images were rolled back, but the code on `main` still contains both bugs, so the next image build would reintroduce them.

## Check list

### DIFF
--- a/images/ubuntu/scripts/build/install-container-tools.sh
+++ b/images/ubuntu/scripts/build/install-container-tools.sh
@@ -39,6 +39,16 @@ install_podman_static() {
         "podman-linux-${arch}/usr" "podman-linux-${arch}/etc"
     rm -f "$archive_path"
 
+    # podman resolves its OCI runtime from a built-in list of absolute paths in which
+    # /usr/bin/crun precedes /usr/local/bin/crun, and the bundle does not pin the path
+    # either, so podman picks the outdated crun that buildah pulls in from apt (1.14.1
+    # on 24.04) instead of the one shipped with the bundle. That crun rejects the
+    # ociVersion written by podman 5.x with "unknown version specified".
+    # https://github.com/actions/runner-images/issues/14473
+    mkdir -p /etc/containers/containers.conf.d
+    printf '[engine]\nruntime = "crun"\n\n[engine.runtimes]\ncrun = ["/usr/local/bin/crun"]\n' \
+        | tee /etc/containers/containers.conf.d/99-fix-runtime.conf
+
     # Ubuntu >= 23.10 restricts unprivileged user namespaces via AppArmor
     # (kernel.apparmor_restrict_unprivileged_userns=1). The distro podman package
     # ships an /etc/apparmor.d/podman profile that grants the `userns` permission,
--- a/images/ubuntu/scripts/build/install-container-tools.sh
+++ b/images/ubuntu/scripts/build/install-container-tools.sh
@@ -43,11 +43,12 @@ install_podman_static() {
     # /usr/bin/crun precedes /usr/local/bin/crun, and the bundle does not pin the path
     # either, so podman picks the outdated crun that buildah pulls in from apt (1.14.1
     # on 24.04) instead of the one shipped with the bundle. That crun rejects the
-    # ociVersion written by podman 5.x with "unknown version specified".
+    # ociVersion written by podman 5.x with "unknown version specified". Only the path
+    # is corrected, and the drop-in is named to load first so user drop-ins still win.
     # https://github.com/actions/runner-images/issues/14473
     mkdir -p /etc/containers/containers.conf.d
-    printf '[engine]\nruntime = "crun"\n\n[engine.runtimes]\ncrun = ["/usr/local/bin/crun"]\n' \
-        | tee /etc/containers/containers.conf.d/99-fix-runtime.conf
+    printf '[engine.runtimes]\ncrun = ["/usr/local/bin/crun"]\n' \
+        | tee /etc/containers/containers.conf.d/00-fix-runtime.conf
 
     # Ubuntu >= 23.10 restricts unprivileged user namespaces via AppArmor
     # (kernel.apparmor_restrict_unprivileged_userns=1). The distro podman package
--- a/images/ubuntu/scripts/build/install-container-tools.sh
+++ b/images/ubuntu/scripts/build/install-container-tools.sh
@@ -34,8 +34,14 @@ install_podman_static() {
     archive_path=$(download_with_retry "$archive_url")
     use_checksum_comparison "$archive_path" "$checksum"
 
-    # The bundle ships ./usr and ./etc under a single top-level directory
-    tar -xzf "$archive_path" -C / --strip-components=1 \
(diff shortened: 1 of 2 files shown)


images/ubuntu/scripts/build/install-container-tools.sh
```diff
@@ -39,6 +39,16 @@ install_podman_static() {
...
 
+    # podman resolves its OCI runtime from a built-in list of absolute paths in which
+    # /usr/bin/crun precedes /usr/local/bin/crun, and the bundle does not pin the path
+    # either, so podman picks the outdated crun that buildah pulls in from apt (1.14.1
+    # on 24.04) instead of the one shipped with the bundle. That crun rejects the
+    # ociVersion written by podman 5.x with "unknown version specified".
+    # https://github.com/actions/runner-images/issues/14473
+    mkdir -p /etc/containers/containers.conf.d
+    printf '[engine]\nruntime = "crun"\n\n[engine.runtimes]\ncrun = ["/usr/local/bin/crun"]\n' \
+        | tee /etc/containers/containers.conf.d/99-fix-runtime.conf
```
adelton: I'm not sure the `99-fix-runtime.conf` is the best file name for this default and distro-wide configuration. Reading https://github.com/containers/common/blob/main/docs/containers.conf.5.md it seems later files overwrite configuration in earlier ones.

So if someone currently uses `/etc/containers/containers.conf.d/70-set-my-runtime.conf` to specifically set the runtime for their project, the `99-fix-runtime.conf` will mess up that expecation.

Please consider adding this change as `/etc/containers/containers.conf.d/00-fix-runtime.conf` or in a similar "defaults are specified first" fashion.

images/ubuntu/scripts/build/install-container-tools.sh
```diff
@@ -39,6 +39,16 @@ install_podman_static() {
...
     rm -f "$archive_path"
 
+    # podman resolves its OCI runtime from a built-in list of absolute paths in which
+    # /usr/bin/crun precedes /usr/local/bin/crun, and the bundle does not pin the path
+    # either, so podman picks the outdated crun that buildah pulls in from apt (1.14.1
+    # on 24.04) instead of the one shipped with the bundle. That crun rejects the
+    # ociVersion written by podman 5.x with "unknown version specified".
+    # https://github.com/actions/runner-images/issues/14473
+    mkdir -p /etc/containers/containers.conf.d
+    printf '[engine]\nruntime = "crun"\n\n[engine.runtimes]\ncrun = ["/usr/local/bin/crun"]\n' \
```
adelton: Is it necessary to explicitly set the `runtime = "crun"`? Documentation https://github.com/containers/common/blob/main/docs/containers.conf.5.md says
> Default runtime will be searched for on the system using the priority: "crun", "runc", "runj", "kata", "runsc", "ocijail"

so `crun` is early in the list.

But if someone currently configures different runtime in `/etc/containers/containers.conf`, this change will overwrite their setting.

Ideally the runner image should differ from the default podman configuration as little as possible.

v-AndriiKhyliuk (dismissed):

adelton:
Hello @v-AndriiKhyliuk, I left some concerns about the actual change. Have you seen them and could you comment on them before rushing to approve the change?

v-dunjadenic:
Hi @adelton, could you please point us to what you're referring to ? Sorry if we missed it.

adelton:
I started a review here. At https://github.com/actions/runner-images/pull/14480/changes/BASE..d4f96a91a0db69ec656f22b5c0d046aae6872c63 there are comments for two parts of the patch.

v-dunjadenic:
@adelton It looks like we're not seeing any of your comments, here's a screenshot of what I see at the link you sent:
image
We also checked with other team members, and none of them see a review. Maybe post the comments here in case there's some issues with your review ? Sorry about this again.

adelton:
Ah, this seems to be an instance of https://github.com/orgs/community/discussions/10369. Submitted now.

v-dunjadenic:
You have valid points on both, I included the suggestions in the latest commit. 

Thanks for the review, feel free to take a look at the changes and provide any additional feedback: https://github.com/actions/runner-images/pull/14480/changes/6eb6f24d7544c3dd35517ad7c6dc2dababb3ea6c

v-GeorgyPuzakov (approved):

PR #14488, opened 2026-08-01
Add aarch64-pc-windows-gnullvm target for Rust

jeroen:
Open source ecosystems based on msys2 use clang-arm64 toolchains , which require the `aarch64-pc-windows-gnullvm` target in rust. We also [need this for the R project](https://contributor.r-project.org/windows-arm64/).

This is basically the native target on arm64 for most oss projects, it is a bit surprising it is not installed.

### DIFF
--- a/images/windows/scripts/build/Install-Rust.ps1
+++ b/images/windows/scripts/build/Install-Rust.ps1
@@ -36,6 +36,7 @@ $env:Path += ";$env:CARGO_HOME\bin"
 
 if (Test-IsArm64) {
     rustup target add aarch64-pc-windows-msvc
+    rustup target add aarch64-pc-windows-gnullvm
 } else {
     # Add i686 target for building 32-bit binaries
     rustup target add i686-pc-windows-msvc


v-davit-ioramashvili:
Thanks for getting to this first, @jeroen 👍 — this is exactly the right fix. I'd opened #14493 for the same target before I spotted yours, so I'm closing mine in favour of this one.

One thing worth adding while we're here: a small assertion in `Rust.Tests.ps1` so we notice if the target ever silently drops out of a future image (nothing currently checks installed targets). It touches a file that isn't in your diff, so it can't go in as a one-click suggestion — but here's the snippet if you'd like to drop it in. Equally happy to send it as a small follow-up PR after this merges if you'd rather keep this one minimal.

In `images/windows/scripts/tests/Rust.Tests.ps1`, after the `$rustEnvNotExists` block:

```powershell
$rustTargetNames = if (Test-IsArm64) {
    @("aarch64-pc-windows-msvc", "aarch64-pc-windows-gnullvm")
} else {
    @("i686-pc-windows-msvc", "x86_64-pc-windows-gnu")
}
$rustTargets = $rustTargetNames | ForEach-Object { @{ Target = $_ } }
```

and alongside the other `It` blocks:

```powershell
It "<Target> rustup target is installed" -TestCases $rustTargets {
    (rustup target list --installed) | Should -Contain $Target
}
```

Either way — thanks again! 🙏

v-davit-ioramashvili:
Thanks again for spotting and fixing this, @jeroen — you were first to it, and it's appreciated. We've consolidated the same change plus a Rust.Tests.ps1 assertion (to catch the target ever dropping out of a future image) into [#14493](https://github.com/actions/runner-images/pull/14493), which is moving ahead on our side, so I'm closing this one to avoid two PRs for the same fix. Credited you there. Really appreciate the contribution 🙏

PR #14517, opened 2026-08-07
[Ubuntu] Remove podman bundle's non-setuid fusermount3 that breaks unprivileged FUSE mounts

### LINKED ISSUE #14516: Ubuntu 24.04 image 20260804.265: stray non-setuid, world-writable /usr/local/bin/fusermount3 shadows the setuid helper, breaking unprivileged FUSE mounts
### Description

Since image `ubuntu24/20260804.265`, mounting any FUSE filesystem the normal unprivileged way fails: the setuid-root `fusermount3` helper's `mount(2)` call returns `EPERM`. The identical call succeeds when `fusermount3` is invoked by root, and non-FUSE root mounts (e.g. tmpfs) work fine. This breaks FUSE-based CI generally (libfuse tools such as sshfs/rclone, Rust `fuser`, AppImage, fuse-overlayfs) whenever the job runs as the `runner` user.

Ruled out on a failing VM:

- `/usr/bin/fusermount3` is setuid (`-rwsr-xr-x root:root`), `/dev/fuse` present (`crw-rw-rw-`), stock `/etc/fuse.conf`
- No AppArmor profile for fusermount3 exists and `dmesg` shows zero denials; `Seccomp: 0`, `NoNewPrivs: 0`, full `CapBnd`
- Distro packages are unchanged vs the working image (`fuse3` 3.14.0-5build1, `apparmor` 4.0.1really4.0.1-0ubuntu0.24.04.7)

The only relevant delta is the kernel: `6.17.0-1020-azure` (works) → `6.17.0-1021-azure` (fails). strace of the user-invoked helper (root strace with `-u` so setuid semantics are preserved):

```
mount("/dev/fuse", ".", "fuse", MS_NOSUID|MS_NODEV, "fd=3,rootmode=40000,user_id=1001"...) = -1 EPERM (Operation not permitted)
```

The same mount performed by real root returns 0. Behaviorally, the new kernel appears to reject fuse mounts when the caller's *real* uid is unprivileged, even with euid 0 and full capabilities.

Both image versions are currently serving `ubuntu-latest`, so the failure appears intermittent run-to-run.

### Platforms affected

- [x] GitHub Actions - Standard Runners

### Runner images affected

- [x] Ubuntu 24.04

Broken: `20260804.265` / `20260804.265.1` ([release](https://github.com/actions/runner-images/releases/tag/ubuntu24%2F20260804.265)). Working: `20260720.247.2`. (Affected runs are in a private repository; self-contained repro below.)

### Is it regression?

Yes. Works on `ubuntu24/20260720.247.2` (kernel `6.17.0-1020-azure`).

### Expected behavior

Unprivileged processes can mount FUSE filesystems through the setuid `fusermount3` helper, as on all previous images.

### Actual behavior

`fusermount3: mount failed: Operation not permitted` (exit 1); the kernel returns `EPERM` from `mount(2)` despite euid 0 and full effective capabilities.

### Repro steps

```yaml
runs-on: ubuntu-latest
steps:
  - run: |
      mkdir -p /tmp/mnt
      python3 - <<'EOF'
      import os, socket, subprocess
      s1, s2 = socket.socketpair()
      env = dict(os.environ) | {"_FUSE_COMMFD": str(s2.fileno())}
      p = subprocess.run(["fusermount3", "-o", "rw,nosuid,nodev", "/tmp/mnt"],
                         pass_fds=[s2.fileno()], env=env, capture_output=True, text=True)
      print(p.returncode, p.stderr)
      EOF
```

This replicates the exact mount handshake libfuse performs, without needing a FUSE daemon. It prints `1 fusermount3: mount failed: Operation not permitted` on `20260804.265` and `0` on `20260720.247.2`. Any real FUSE tool (`sshfs`, `rclone mount`) reproduces it equally. Prefixing the helper invocation with `sudo` makes it succeed on both images.

v-GeorgyPuzakov:
# Description

**Bug fix**

Unprivileged FUSE mounts (`sshfs`, `rclone`, `fuse-overlayfs`, AppImage, …) fail with `EPERM` on Ubuntu 22.04/24.04 when run as the `runner` user:

```
fusermount3: mount failed: Operation not permitted
```

**Cause:** the static `podman` bundle added in #14412 ships `/usr/local/bin/fusermount3` (mode `0755`, **not setuid**). Since `/usr/local/bin` precedes `/usr/bin` in `PATH`, it shadows the distro's setuid-root `/usr/bin/fusermount3`, so the helper runs without privileges. Ubuntu 26.04 is unaffected (no bundle).

**Fix** — `images/ubuntu/scripts/build/install-container-tools.sh`: remove the bundle's `/usr/local/bin/fusermount3` so the setuid distro helper is used (`fuse-overlayfs` still resolves it via `PATH`). Added a `Containers` regression test asserting `fusermount3` resolves to `/usr/bin/fusermount3`. Complements #14480, which did not touch `fusermount3`.

#### Related issue:
- Fixes https://github.com/actions/runner-images/issues/14516

## Check list
- [x] Tests are written (if applicable) — `Containers` regression test for the `fusermount3` path
- [x] Documentation is updated (if applicable) — n/a

### Build results

| Image | Status | Link |
|---|---|---|
| Ubuntu 22.04 X64 | IN PROGRESS | https://github.com/actions/runner-images-ci/actions/runs/31171442196 |
| Ubuntu 24.04 X64 | IN PROGRESS | https://github.com/actions/runner-images-ci/actions/runs/31171444735 |

### DIFF
--- a/images/ubuntu/scripts/build/install-container-tools.sh
+++ b/images/ubuntu/scripts/build/install-container-tools.sh
@@ -56,6 +56,11 @@ install_podman_static() {
     printf '[engine.runtimes]\ncrun = ["/usr/local/bin/crun"]\n' \
         | tee /etc/containers/containers.conf.d/00-fix-runtime.conf
 
+    # The bundle's non-setuid fusermount3 shadows the distro's setuid
+    # /usr/bin/fusermount3 (PATH order) and breaks unprivileged FUSE mounts.
+    # https://github.com/actions/runner-images/issues/14516
+    rm -f /usr/local/bin/fusermount3
+
     # Ubuntu >= 23.10 restricts unprivileged user namespaces via AppArmor
     # (kernel.apparmor_restrict_unprivileged_userns=1). The distro podman package
     # ships an /etc/apparmor.d/podman profile that grants the `userns` permission,
--- a/images/ubuntu/scripts/tests/Tools.Tests.ps1
+++ b/images/ubuntu/scripts/tests/Tools.Tests.ps1
@@ -346,6 +346,11 @@ Describe "Containers" {
         $(stat -c "%U" $Directory) | Should -Be "root"
     }
 
+    # https://github.com/actions/runner-images/issues/14516
+    It "fusermount3 resolves to the setuid distro helper, not the podman bundle shadow" -Skip:(Test-IsUbuntu26) {
+        (Get-Command fusermount3).Source | Should -Be "/usr/bin/fusermount3"
+    }
+
 }
 
 Describe "nvm" {


v-AndriiKhyliuk (approved):

v-SergeiRadich (approved):

hlme7788145-a11y (commented):
a.com

PR #14643, opened 2026-08-27
[Ubuntu] Stop apt-get update stalling on a degraded mirror

### LINKED ISSUE #14594: [Ubuntu] apt update stalls 20+ min retrying azure.archive.ubuntu.com before falling back
### Description

`sudo apt update` intermittently stalls for **20+ minutes** on `ubuntu-latest` (Ubuntu 24.04)
because apt keeps retrying `azure.archive.ubuntu.com`, which returns `Ign:` for every index, before
falling back to `archive.ubuntu.com`.

The step never fails — it just hangs. Without an explicit `timeout-minutes`, the job runs toward the
6-hour default limit.

### Evidence

Log excerpt from an affected step (step duration: **22m40s**, entirely inside `apt update`, cancelled
manually — it never reached the following commands):

```
Ign:3 http://azure.archive.ubuntu.com/ubuntu noble-updates InRelease
Ign:4 http://azure.archive.ubuntu.com/ubuntu noble-backports InRelease
Ign:5 http://azure.archive.ubuntu.com/ubuntu noble-security InRelease
Ign:2 http://azure.archive.ubuntu.com/ubuntu noble InRelease
... (26 `Ign:` lines total against azure.archive.ubuntu.com)
Hit:2 https://archive.ubuntu.com/ubuntu noble InRelease
Get:3 https://archive.ubuntu.com/ubuntu noble-updates InRelease [126 kB]
Get:4 https://archive.ubuntu.com/ubuntu noble-backports InRelease [126 kB]
Get:5 https://archive.ubuntu.com/ubuntu noble-security InRelease [126 kB]
##[error]The operation was canceled.
```

26 `Ign:` lines against the Azure mirror versus 13 `Hit:`/`Get:` on the fallback.

### Frequency

Measured across the last 14 runs of one job in a single repository:

| duration | runs |
| --- | --- |
| 3–4 min (normal) | 11 |
| 32 min | 1 |
| 80 min | 1 |
| 22 min (cancelled) | 1 |

Roughly **20% of runs** were affected over a ~24-hour window. Two jobs in the *same* workflow run and
the *same* commit behaved differently — one completed the identical step in 1m56s while the other
stalled — which points at intermittent mirror reachability rather than anything in the workflow.

### Expected behaviour

`apt update` should fail over to the working mirror quickly, or fail fast, rather than stalling for
tens of minutes.

### Related (both closed)

- #12599 — Apt install started failing since yesterday
- #5656 — `apt-get update` failed on the latest ubuntu-22.04 image

### Workaround, in case it helps others

Bounding apt's own retry/timeout behaviour makes the fallback happen in seconds, and treating the
step as non-fatal avoids blocking the job:

```yaml
- name: Install dependencies
  timeout-minutes: 10
  run: |
    APT_OPTS="-o Acquire::Retries=1 -o Acquire::http::Timeout=10 -o Acquire::https::Timeout=10"
    timeout 90 sudo apt-get $APT_OPTS update -qq || echo "::warning::apt-get update timed out"
    timeout 150 sudo apt-get $APT_OPTS install -y --no-install-recommends <packages> \
      || echo "::warning::apt-get install timed out"
```

Adding `timeout-minutes` to the job is what turns a silent multi-hour hang into a fast, diagnosable
failure.

v-dunjadenic:
# Description
- Replace `APT::Acquire::Retries "10"` with `Acquire::Retries "1"`.
- Set `Acquire::http::Timeout` / `Acquire::https::Timeout` to `15`.
- Assert the effective values in `Apt.Tests.ps1` and `ubuntu-slim/test.sh`.

### Reasoning:
When a mirror degrades, `apt-get update` stalls for 20-80 minutes instead of failing over. Three things caused that.

apt reads `Acquire::Retries`, not `APT::Acquire::Retries`. the latter appears nowhere in apt's source, so these images have always run on apt's default of 3. Correcting the key alone would have made things worse: retries are spent on the same*URI before apt moves to the next mirror in `/etc/apt/apt-mirrors.txt`, so a value of 10 pushes time-to-failover from ~2 minutes to ~8.5 minutes per index item. 

`Acquire::http::Timeout` was never set, but apt's built-in default is already 30s (unchanged across every apt in 22.04/24.04/26.04), so setting it to 30 explicitly is a no-op. 15s is safe because it is a connect/inactivity timeout, not a transfer budget. It does not abort slow-but-progressing downloads.

Time-to-failover also scales with the number of index items, because the mirror method re-sorts from priority:1 for every item and keeps no failure state. Reducing that item count is split into a separate PR so it can be validated on its own.

Mirror order is unchanged. The Azure mirror runs in the same datacentres as the runners, so it is fast and its traffic stays inside the network. The two fallbacks behind it, `archive.ubuntu.com` and `security.ubuntu.com`, are two names for the same servers. Putting them first would send every runner's apt traffic to servers Canonical owns and we don't, and would leave one thing to fail instead of two. 

Tests assert `apt-config dump` output rather than file contents, so a wrong key name or a later override fails the build instead of silently reverting to apt's defaults.

#### Related issue:
https://github.com/actions/runner-images/issues/14594

#### Past PRs: 
- https://github.com/actions/runner-images/pull/14596
- https://github.com/actions/runner-images/pull/14584

## Check list

### DIFF
--- a/images/ubuntu-slim/scripts/build/configure-apt.sh
+++ b/images/ubuntu-slim/scripts/build/configure-apt.sh
@@ -14,8 +14,15 @@ source $HELPER_SCRIPTS/os.sh
 # systemctl disable apt-daily-upgrade.timer
 # systemctl disable apt-daily-upgrade.service
 
-# Enable retry logic for apt up to 10 times
-echo "APT::Acquire::Retries \"10\";" > /etc/apt/apt.conf.d/80-retries
+# Bound apt's acquire behaviour so a stalled mirror fails over in seconds instead of minutes.
+# apt reads Acquire::Retries (default 3), not APT::Acquire::Retries, and spends every retry on the
+# same URI before trying the next mirror in /etc/apt/apt-mirrors.txt, so a high count delays failover.
+# https://github.com/actions/runner-images/issues/14594
+cat <<EOF > /etc/apt/apt.conf.d/80-retries
+Acquire::Retries "1";
+Acquire::http::Timeout "15";
+Acquire::https::Timeout "15";
+EOF
 
 # Configure apt to always assume Y
 echo "APT::Get::Assume-Yes \"true\";" > /etc/apt/apt.conf.d/90assumeyes
--- a/images/ubuntu-slim/scripts/build/configure-apt.sh
+++ b/images/ubuntu-slim/scripts/build/configure-apt.sh
@@ -52,6 +52,9 @@ else
     cat /etc/apt/sources.list.d/ubuntu.sources
 fi
 
+echo 'APT mirrors'
+cat /etc/apt/apt-mirrors.txt
+
 apt-get update
 
 echo "ubuntu ALL=(ALL) NOPASSWD:ALL" >> /etc/sudoers
--- a/images/ubuntu-slim/scripts/build/configure-apt.sh
+++ b/images/ubuntu-slim/scripts/build/configure-apt.sh
@@ -24,6 +24,8 @@ Acquire::http::Timeout "15";
 Acquire::https::Timeout "15";
 EOF
 
+echo 'Acquire::IndexTargets::deb::DEP-11::DefaultEnabled "false";' > /etc/apt/apt.conf.d/90-index-targets
+
 # Configure apt to always assume Y
 echo "APT::Get::Assume-Yes \"true\";" > /etc/apt/apt.conf.d/90assumeyes
 
--- a/images/ubuntu-slim/scripts/build/configure-apt.sh
+++ b/images/ubuntu-slim/scripts/build/configure-apt.sh
@@ -14,7 +14,7 @@ source $HELPER_SCRIPTS/os.sh
 # systemctl disable apt-daily-upgrade.timer
 # systemctl disable apt-daily-upgrade.service
 
-# Bound apt's acquire behaviour so a stalled mirror fails over in seconds instead of minutes.
+# Bound apt's acquire behavior so a stalled mirror fails over in seconds instead of minutes.
 # apt reads Acquire::Retries (default 3), not APT::Acquire::Retries, and spends every retry on the
 # same URI before trying the next mirror in /etc/apt/apt-mirrors.txt, so a high count delays failover.
 # https://github.com/actions/runner-images/issues/14594
--- a/images/ubuntu-slim/scripts/build/configure-apt.sh
+++ b/images/ubuntu-slim/scripts/build/configure-apt.sh
@@ -24,8 +24,6 @@ Acquire::http::Timeout "15";
 Acquire::https::Timeout "15";
 EOF
 
-echo 'Acquire::IndexTargets::deb::DEP-11::DefaultEnabled "false";' > /etc/apt/apt.conf.d/90-index-targets
-
 # Configure apt to always assume Y
 echo "APT::Get::Assume-Yes \"true\";" > /etc/apt/apt.conf.d/90assumeyes
 
--- a/images/ubuntu-slim/test.sh
+++ b/images/ubuntu-slim/test.sh
@@ -99,3 +99,10 @@ run_test "docker buildx is installed" docker buildx version
 
 # Quick check: ensure the imagedata JSON file was created during image build
 run_test "imagedata JSON file exists" test -f /imagegeneration/imagedata.json
+
+# Assert the effective apt acquire bounds, since a setting written under a key apt does not read is
+# silently ignored: https://github.com/actions/runner-images/issues/14594
+run_test "apt acquire retries are bounded" bash -c 'apt-config dump Acquire::Retries | grep -Fxq "Acquire::Retries \"1\";"'
+run_test "apt http timeout is bounded" bash -c 'apt-config dump Acquire::http::Timeout | grep -Fxq "Acquire::http::Timeout \"15\";"'
+run_test "apt https timeout is bounded" bash -c 'apt-config dump Acquire::https::Timeout | grep -Fxq "Acquire::https::Timeout \"15\";"'
+run_test "apt sources resolve through the mirror list" bash -c 'grep -Rq "mirror+file:/etc/apt/apt-mirrors.txt" /etc/apt/'
--- a/images/ubuntu-slim/test.sh
+++ b/images/ubuntu-slim/test.sh
@@ -105,4 +105,5 @@ run_test "imagedata JSON file exists" test -f /imagegeneration/imagedata.json
 run_test "apt acquire retries are bounded" bash -c 'apt-config dump Acquire::Retries | grep -Fxq "Acquire::Retries \"1\";"'
 run_test "apt http timeout is bounded" bash -c 'apt-config dump Acquire::http::Timeout | grep -Fxq "Acquire::http::Timeout \"15\";"'
 run_test "apt https timeout is bounded" bash -c 'apt-config dump Acquire::https::Timeout | grep -Fxq "Acquire::https::Timeout \"15\";"'
+run_test "apt DEP-11 index target is disabled" bash -c 'apt-config dump Acquire::IndexTargets::deb::DEP-11::DefaultEnabled | grep -Fxq "Acquire::IndexTargets::deb::DEP-11::DefaultEnabled \"false\";"'
 run_test "apt sources resolve through the mirror list" bash -c 'grep -Rq "mirror+file:/etc/apt/apt-mirrors.txt" /etc/apt/'
--- a/images/ubuntu-slim/test.sh
+++ b/images/ubuntu-slim/test.sh
@@ -105,5 +105,4 @@ run_test "imagedata JSON file exists" test -f /imagegeneration/imagedata.json
 run_test "apt acquire retries are bounded" bash -c 'apt-config dump Acquire::Retries | grep -Fxq "Acquire::Retries \"1\";"'
 run_test "apt http timeout is bounded" bash -c 'apt-config dump Acquire::http::Timeout | grep -Fxq "Acquire::http::Timeout \"15\";"'
 run_test "apt https timeout is bounded" bash -c 'apt-config dump Acquire::https::Timeout | grep -Fxq "Acquire::https::Timeout \"15\";"'
-run_test "apt DEP-11 index target is disabled" bash -c 'apt-config dump Acquire::IndexTargets::deb::DEP-11::DefaultEnabled | grep -Fxq "Acquire::IndexTargets::deb::DEP-11::DefaultEnabled \"false\";"'
 run_test "apt sources resolve through the mirror list" bash -c 'grep -Rq "mirror+file:/etc/apt/apt-mirrors.txt" /etc/apt/'
--- a/images/ubuntu/scripts/build/configure-apt.sh
+++ b/images/ubuntu/scripts/build/configure-apt.sh
@@ -17,10 +17,11 @@ systemctl disable apt-daily-upgrade.service
 # Bound apt's acquire behavior so a stalled mirror fails over in seconds instead of minutes.
 # apt reads Acquire::Retries (default 3), not APT::Acquire::Retries, and spends every retry on the
 # same URI before trying the next mirror in /etc/apt/apt-mirrors.txt, so a high count delays failover.
-# arm64 sources come from ports.ubuntu.com, which the mirror list does not cover, leaving retries as
-# the only failover apt has, so the count stays at apt's default.
+# 22.04 and 24.04 arm64 are served by ports.ubuntu.com, which the mirror list does not cover, leaving
+# retries as the only failover apt has there, so the count stays at apt's default. 26.04 merged arm64
+# into the main archive, so it uses the mirror list like x64 does.
 # https://github.com/actions/runner-images/issues/14594
-if is_arm64; then
+if is_ubuntu22_arm64 || is_ubuntu24_arm64; then
     apt_retries=3
 else
     apt_retries=1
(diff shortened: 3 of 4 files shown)


images/ubuntu-slim/scripts/build/configure-apt.sh
```diff
@@ -14,8 +14,15 @@ source $HELPER_SCRIPTS/os.sh
...
-echo "APT::Acquire::Retries \"10\";" > /etc/apt/apt.conf.d/80-retries
+# Bound apt's acquire behaviour so a stalled mirror fails over in seconds instead of minutes.
+# apt reads Acquire::Retries (default 3), not APT::Acquire::Retries, and spends every retry on the
+# same URI before trying the next mirror in /etc/apt/apt-mirrors.txt, so a high count delays failover.
+# https://github.com/actions/runner-images/issues/14594
+cat <<EOF > /etc/apt/apt.conf.d/80-retries
+Acquire::Retries "1";
+Acquire::http::Timeout "15";
+Acquire::https::Timeout "15";
+EOF
```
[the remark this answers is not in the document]
v-dunjadenic: Valid, fixed.

images/ubuntu-slim/test.sh
```diff
@@ -99,3 +99,10 @@ run_test "docker buildx is installed" docker buildx version
 
 # Quick check: ensure the imagedata JSON file was created during image build
 run_test "imagedata JSON file exists" test -f /imagegeneration/imagedata.json
+
+# Assert the effective apt acquire bounds, since a setting written under a key apt does not read is
+# silently ignored: https://github.com/actions/runner-images/issues/14594
+run_test "apt acquire retries are bounded" bash -c 'apt-config dump Acquire::Retries | grep -Fxq "Acquire::Retries \"1\";"'
+run_test "apt http timeout is bounded" bash -c 'apt-config dump Acquire::http::Timeout | grep -Fxq "Acquire::http::Timeout \"15\";"'
+run_test "apt https timeout is bounded" bash -c 'apt-config dump Acquire::https::Timeout | grep -Fxq "Acquire::https::Timeout \"15\";"'
+run_test "apt sources resolve through the mirror list" bash -c 'grep -Rq "mirror+file:/etc/apt/apt-mirrors.txt" /etc/apt/'
```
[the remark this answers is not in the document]
v-dunjadenic: `apt-config dump` prints the resolved config tree, not the contents of the files. `Configuration::Set()` is `Lookup(Name, true)` followed by `Itm->Value = Value`, so a later drop-in overwrites the node instead of adding a second entry. There is only ever one value per key. An override would show up as a single line with the new value, and the grep would fail. Keeping `grep -Fxq` on purpose.

v-dunjadenic:
This has been placed in draft since I'm waiting for the CI and want to explore a couple of other things regarding potential improvements first.

v-AndriiKhyliuk (dismissed):

images/ubuntu/scripts/build/configure-apt.sh
```diff
@@ -14,8 +14,28 @@ systemctl stop apt-daily-upgrade.timer
 systemctl disable apt-daily-upgrade.timer
 systemctl disable apt-daily-upgrade.service
 
-# Enable retry logic for apt up to 10 times
-echo "APT::Acquire::Retries \"10\";" > /etc/apt/apt.conf.d/80-retries
+# Bound apt's acquire behaviour so a stalled mirror fails over in seconds instead of minutes.
+# apt reads Acquire::Retries (default 3), not APT::Acquire::Retries, and spends every retry on the
+# same URI before trying the next mirror in /etc/apt/apt-mirrors.txt, so a high count delays failover.
+# arm64 sources come from ports.ubuntu.com, which the mirror list does not cover, so there retries are
+# the only failover apt has and the count stays at apt's default.
```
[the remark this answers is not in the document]
v-dunjadenic: Valid, addressed

images/ubuntu-slim/test.sh
```diff
@@ -99,3 +99,11 @@ run_test "docker buildx is installed" docker buildx version
...
 # Quick check: ensure the imagedata JSON file was created during image build
 run_test "imagedata JSON file exists" test -f /imagegeneration/imagedata.json
+
+# Assert the effective apt acquire bounds, since a setting written under a key apt does not read is
+# silently ignored: https://github.com/actions/runner-images/issues/14594
+run_test "apt acquire retries are bounded" bash -c 'apt-config dump Acquire::Retries | grep -Fxq "Acquire::Retries \"1\";"'
+run_test "apt http timeout is bounded" bash -c 'apt-config dump Acquire::http::Timeout | grep -Fxq "Acquire::http::Timeout \"15\";"'
+run_test "apt https timeout is bounded" bash -c 'apt-config dump Acquire::https::Timeout | grep -Fxq "Acquire::https::Timeout \"15\";"'
+run_test "apt DEP-11 index target is disabled" bash -c 'apt-config dump Acquire::IndexTargets::deb::DEP-11::DefaultEnabled | grep -Fxq "Acquire::IndexTargets::deb::DEP-11::DefaultEnabled \"false\";"'
+run_test "apt sources resolve through the mirror list" bash -c 'grep -Rq "mirror+file:/etc/apt/apt-mirrors.txt" /etc/apt/'
```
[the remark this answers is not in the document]
v-dunjadenic: `ubuntu-slim` is x64 only by construction. Every download in the build is hardcoded to x86_64.

images/ubuntu-slim/scripts/build/configure-apt.sh
```diff
@@ -14,8 +14,17 @@ source $HELPER_SCRIPTS/os.sh
...
```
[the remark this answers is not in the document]
v-dunjadenic: `ubuntu-slim` is x64 only by construction. Every download in the build is hardcoded to x86_64.

images/ubuntu/scripts/build/configure-apt.sh
```diff
@@ -14,8 +14,29 @@ systemctl stop apt-daily-upgrade.timer
...
+cat <<EOF > /etc/apt/apt.conf.d/80-retries
+Acquire::Retries "$apt_retries";
+Acquire::http::Timeout "15";
+Acquire::https::Timeout "15";
+EOF
+
+# DEP-11/AppStream is desktop software-catalog metadata with no CI use. Skipping it drops 15 of the 51
+# index items and 7.6 MB from every apt-get update, which is also 15 fewer chances to hit a sick mirror.
+# Sorts after appstream's own /etc/apt/apt.conf.d/50appstream, so it wins.
+echo 'Acquire::IndexTargets::deb::DEP-11::DefaultEnabled "false";' > /etc/apt/apt.conf.d/90-index-targets
```
drielenr: I really like this, I wasn't aware of these index items that don't get to be used by CI. 
But since this a separate behavior change, maybe we could ship in separate pr? Then we can validate in canary that `apt-get update` and `apt-get install` succeeds, and we isolate the if any problem happens

images/ubuntu/scripts/build/configure-apt.sh
```diff
@@ -46,6 +67,9 @@ else
     cat /etc/apt/sources.list.d/ubuntu.sources
 fi
 
+echo 'APT mirrors'
+cat /etc/apt/apt-mirrors.txt
```
drielenr, line 68: Does this get stored somewhere we can retrieve later? Not sure how valuable it's to print this file, we can also come to the repo to refer to it, right?

drielenr: but also it doesn't hurt keeping it, so I'm ok with it

images/ubuntu/scripts/build/configure-apt.sh
```diff
@@ -14,8 +14,29 @@ systemctl stop apt-daily-upgrade.timer
...
-# Enable retry logic for apt up to 10 times
-echo "APT::Acquire::Retries \"10\";" > /etc/apt/apt.conf.d/80-retries
+# Bound apt's acquire behavior so a stalled mirror fails over in seconds instead of minutes.
+# apt reads Acquire::Retries (default 3), not APT::Acquire::Retries, and spends every retry on the
+# same URI before trying the next mirror in /etc/apt/apt-mirrors.txt, so a high count delays failover.
+# 22.04 and 24.04 arm64 are served by ports.ubuntu.com, which the mirror list does not cover, leaving
+# retries as the only failover apt has there, so the count stays at apt's default. 26.04 merged arm64
+# into the main archive, so it uses the mirror list like x64 does.
+# https://github.com/actions/runner-images/issues/14594
+if is_ubuntu22_arm64 || is_ubuntu24_arm64; then
```
drielenr, line 24: since arm64 will not have the failover mirrors, maybe we can set the timeout to be a little bit more generous? I was thinking we can keep the 30s default timeout for them here

images/ubuntu/scripts/build/configure-apt.sh
```diff
@@ -14,8 +14,29 @@ systemctl stop apt-daily-upgrade.timer
...
+    apt_retries=3
```
drielenr, line 25: why don't we have those for slim? we don't have arm64 there?

images/ubuntu/scripts/tests/Apt.Tests.ps1
```diff
@@ -24,3 +24,45 @@ Describe "Apt" {
...
+        $aptSources | Should -Match ([regex]::Escape("ports.ubuntu.com/ubuntu-ports"))
+        $aptSources | Should -Not -Match ([regex]::Escape("mirror+file:/etc/apt/apt-mirrors.txt"))
+    }
+
+    It "Mirror list entries have unique priorities" {
+        $priorities = Get-Content "/etc/apt/apt-mirrors.txt" | ForEach-Object { ($_ -split "priority:")[1] }
+        $priorities.Count | Should -BeGreaterThan 1
+        ($priorities | Select-Object -Unique).Count | Should -BeExactly $priorities.Count
+    }
+}
```
drielenr, line 68: Those tests are mostly configuration tests, not properly behavior tests. We should have some integration tests to confirm we are really retrying 1 time, that failover works, and confirm the timeout in slow-progress 
(e.g. have a local server send data slowly but at intervals shorter than 15 seconds. Assert that the transfer completes successfully. This validates the claim that the timeout is based on inactivity rather than total transfer duration)

drielenr: I was thinking we could maybe have that as part of the pr check. But I'm also ok shipping as is, and create a separate work item to track it and we can address it later

images/ubuntu/scripts/build/configure-apt.sh
Jinx150: > why don't we have those for slim? we don't have arm64 there?

Ubuntu slim is only x64 

IMG_0635

images/ubuntu/scripts/build/configure-apt.sh
Jinx150: > since arm64 will not have the failover mirrors, maybe we can set the timeout to be a little bit more generous? I was thinking we can keep the 30s default timeout for them here

See on reasoning Acquire::http::Timeout was never set, but apt's built-in default is already 30s (unchanged across every apt in 22.04/24.04/26.04), so setting it to 30 explicitly is a no-op. 15s is safe because it is a connect/inactivity timeout, not a transfer budget. It does not abort slow-but-progressing downloads.

drielenr: I think the 30s timeout is better for arm64 since it has no fallback mirror. With 15s, if the path to `ports.ubuntu.com` is congested such that responses consistently land past 15s, all four attempts fail deterministically and the job dies with no fallback. With 30s, the same request may succeed. 
And the 15s timeout does not help with slow-but-progressing downloads because it resets whenever data arrives, 15s and 30s would be identical, both can run to the 6h ceiling.

v-dunjadenic: Agreed, let's keep 30s for 22.04 and 24.04 arm64.

The timeout covers more than gaps mid-transfer. The same value is passed into the TCP connect, and it also governs the wait for the first response byte. In both of those there is no data flowing to reset it. On a lossy path SYN retransmits back off at roughly 1, 2, 4 and 8 seconds, so a handshake that would succeed at 30s can consistently miss 15s. With no failover on those images that becomes a hard failure rather than a fallback, which is exactly the case you described.

It also costs us nothing for this incident, since those two images never used `azure.archive.ubuntu.com` in the first place.

I'll make the timeout conditional alongside the retry count and update the expected values in the tests.

images/ubuntu/scripts/build/configure-apt.sh
v-dunjadenic: I think we should keep it. The repo shows what we intended to configure, the build log shows what actually landed. Those actually happened to be different here, since on 22.04 and 24.04 arm64 the rewrite silently does not match, so the file is written but never used. That is a case reading the repo would not reveal.

images/ubuntu/scripts/build/configure-apt.sh
v-dunjadenic: Thanks ! Splitting it off then, I'll link the PR here as well.

images/ubuntu/scripts/build/configure-apt.sh
v-dunjadenic: Yeah, that's correct.

images/ubuntu/scripts/tests/Apt.Tests.ps1
v-dunjadenic: Makes sense to have it, I'll create a separate item to track this

images/ubuntu/scripts/build/configure-apt.sh
v-dunjadenic: The DEP-11 change is now split off here: https://github.com/actions/runner-images/pull/14719

drielenr (approved):

